PR #9, opened 2025-03-14
adding report-to-wandb

Goekdeniz-Guelmez:


Goekdeniz-Guelmez:
@awni should be mergable now.

ivanfioravanti:
It would be great having this integrated OOB 🙏

Goekdeniz-Guelmez:
@awni should be mergeable now!

angeloskath (changes requested):
Hey @Goekdeniz-Guelmez thanks for adding that!

I think it would be better to move this out of `lora.py`. I would do the following changes
- Add a file `mlx_lm/tuner/callbacks.py`
- Add `WandBCallback` there
    - Use `try: import wandb; except: wandb = None` pattern instead of `if`
    - Initialize `wandb` in the callback constructor
    - Accept the project name, config and optional wrapped callback in the callback constructor
- Change the `--report-to-wandb` to `--wandb` or `--wandb-project` and allow setting the project or the default `None` to disable the callback

Goekdeniz-Guelmez:
Hey @angeloskath your suggestions are really helpful! I did what you said and it works as expected. Also the run config and data files are save in the `--adapter-path` path.

Goekdeniz-Guelmez:
You can now call it like `--wandb my-project-run`

Goekdeniz-Guelmez:
Hey @awni, this should be mergable now! If you need me to fix or change something, just ping me.

mlx_lm/LORA.md
```diff
@@ -387,6 +387,10 @@ tokens-per-second, using the MLX Example
 [`wikisql`](https://github.com/ml-explore/mlx-examples/tree/main/lora/data)
 data set.
 
+## Logging
+
+You can log training metrics to Weights & Biases by adding the `--report-to-wandb` flag. This requires installing wandb manually with `pip install wandb`. When enabled, all training and validation metrics will be logged to your wandb account.
```
awni, line 392: I think the flag / documentation here is wrong right? It should be `--wandb` which is the experiment name and it's disabled by default.

mlx_lm/examples/lora_config.yaml
```diff
@@ -37,6 +37,9 @@ val_batches: 25
 # Adam learning rate.
 learning_rate: 1e-5
 
+# Whether to report the logs to WandB
+report_to_wand: true
```
awni, line 41: Same here `wandb: null` as default?

mlx_lm/examples/lora_config.yaml
```diff
@@ -37,6 +37,9 @@ val_batches: 25
 # Adam learning rate.
 learning_rate: 1e-5
 
+# Whether to report the logs to WandB
```
awni, line 40: The wanbd project name (no wanb logging by default)

awni:
This looks nice! Just a couple minor comments. Could you address then we can merge?

mlx_lm/LORA.md
Goekdeniz-Guelmez: Great catch, thanks!

mlx_lm/examples/lora_config.yaml
Goekdeniz-Guelmez: Yes

Goekdeniz-Guelmez:
Thanks! They have been addressed!

awni (approved):
Awesome!

awni (approved):

awni:
@angeloskath you have to approve this in order to land it.

angeloskath (approved):
👍

PR #11, opened 2025-03-14
Fix cohere2

### LINKED ISSUE #3: Command-A error
Image

Why is this error coming up?

awni:
Close #3

### DIFF
--- a/mlx_lm/models/cohere2.py
+++ b/mlx_lm/models/cohere2.py
@@ -88,10 +88,17 @@ class Attention(nn.Module):
             if mask.shape[-1] != key_len:
                 mask = mask[..., -key_len:]
 
+        # TODO: maybe remove cast once fused mask is supported since attention
+        # may be in higher precision
+        sdpa_type = mx.float32 if queries.dtype == mx.float16 else queries.dtype
         output = scaled_dot_product_attention(
-            queries, keys, values, cache=cache, scale=self.scale, mask=mask
-        )
-
+            queries.astype(sdpa_type),
+            keys,
+            values,
+            cache=cache,
+            scale=self.scale,
+            mask=mask,
+        ).astype(queries.dtype)
         output = output.transpose(0, 2, 1, 3).reshape(B, L, -1)
         return self.o_proj(output)
 
@@ -126,9 +133,11 @@ class TransformerBlock(nn.Module):
         mask: Optional[mx.array] = None,
         cache: Optional[Tuple[mx.array, mx.array]] = None,
     ) -> mx.array:
+
         h = self.input_layernorm(x)
         attn_h = self.self_attn(h, mask, cache)
         ff_h = self.mlp(h)
+
         return attn_h + ff_h + x
 
 


vlbosch:
I can confirm this fixes the issue with regular 4 bits and mixed precision quantizations. However, I am still running into an issue when I don't quantize the embeddings-layer. Then after a 'correct' first token, it goes back to "< PAD >" only. 

I don't know whether that's an issue attributable to the models architecture or to MLX. I did notice the embeddings-layer is also used as lm_head. Or maybe the model is just sensitive to quantization and cannot handle the big difference between non-quantized embeddings and the rest quantized to 8- and 4-bits?

awni:
> However, I am still running into an issue when I don't quantize the embeddings-layer.

You'll have to say more about how you did this so that we can reproduce it.

angeloskath (approved):
👍

vlbosch:
> > However, I am still running into an issue when I don't quantize the embeddings-layer.
> 
> You'll have to say more about how you did this so that we can reproduce it.

The issue is that when I quantize the embeddings-layer, the model runs just fine. Both the regular quantizations and mixed quantizations. But I mostly make custom converts and leave out the lm_head and embeddings-layer of quantization. For example when I run this code, the model only outputs the first (correct) token and then goes back to " < PAD >". I also tried providing the dtype=bfloat16 argument, but that didn't make any difference.

```python
import argparse
from typing import Union

from mlx_lm.utils import convert
from mlx.nn import Module


def mixed_4_8(path: str, module: Module, config: dict) -> Union[bool, dict]:
    """A mixed quantization with similar choices to llama.cpp's Q4_K_M.
    Ref: https://github.com/ggerganov/llama.cpp/blob/917786f43d0f29b7c77a0c56767c0fa4df68b1c5/src/llama.cpp#L5265
    """

    if not hasattr(module, "to_quantized"):
        return False

    if 'lm_head' in path or 'embeddings' in path or 'embed_tokens' in path:
        return False

    index = int(path.split(".")[2]) if len(path.split(".")) > 2 else 0

    num_layers = config["num_hidden_layers"]
    use_more_bits = (
        index < num_layers // 8 or
        index >= 7 * num_layers // 8 or
        (index - num_layers // 8) % 3 == 2
    )
    if "v_proj" in path and use_more_bits:
        return {"group_size": 32, "bits": 8}
    if "down_proj" in path and use_more_bits:
        return {"group_size": 32, "bits": 8}
    if "lm_head" in path:
        return {"group_size": 32, "bits": 8}

    return {"group_size": 32, "bits": 4}


if __name__ == "__main__":
    parser = argparse.ArgumentParser(
        description="Convert Hugging Face model to MLX format"
    )
    parser.add_argument("--hf-path", type=str, help="Path to the Hugging Face model.")
    parser.add_argument(
        "--mlx-path", type=str, default="mlx_model", help="Path to save the MLX model."
    )

    args = parser.parse_args()
    convert(args.hf_path, args.mlx_path, quantize=True, quant_predicate=mixed_4_8)
```

vlbosch:
To add, it seems that regardless of the quantization used, Command A loses coherency at around 5000 tokens. At the tipping point, it start to repeat itself first then proceeding to straight gibberish.

vlbosch:
> To add, it seems that regardless of the quantization used, Command A loses coherency at around 5000 tokens. At the tipping point, it start to repeat itself first then proceeding to straight gibberish.

@awni Unfortunately with the latest releases of MLX-LM and MLX (even compiled from source) I still face this issue. What can I do to help further triage this?

awni:
This should be fixed post https://github.com/ml-explore/mlx/pull/2029

vlbosch:
> This should be fixed post [ml-explore/mlx#2029](https://github.com/ml-explore/mlx/pull/2029)

@awni Thanks for your quick response. I tried specifically with MLX after that commit and verified that the fix is indeed present, but the issue with Command A remains. Also tried to reconvert the model after the fix, just in case that would make a difference. But at least 4 bits with group-size 32 doesn't work despite of https://github.com/ml-explore/mlx/pull/2029

awni:
Ok, please file an issue with the command to reproduce the behavior you are seeing.

Issue #15, opened 2025-03-16
Cannot serve at IPv6 address


chaihahaha:
When running mlx-lm with IPV6 server address ::, this error shows up:

![Image](https://github.com/user-attachments/assets/e6046858-5ad6-420e-8baa-abe6b647cec3)

PR #16, opened 2025-03-16
Add support for IPv6 server

chaihahaha:
This PR will fix https://github.com/ml-explore/mlx-lm/issues/15
IPv6 server will help self-hosting users in countries where IPv4 addresses are insufficient but IPv6 addresses are abundant.


angeloskath (approved):
Nice, thanks!

angeloskath:
To be fair I don't think that we need IPv6 here but it is nice to support it with this small diff. It feels right that it works when passing `--host ::1` instead of erroring out.

PR #20, opened 2025-03-17
Update mixed_quant_predicate_builder calls to include high_bits parameter

sealad886:
Explicitly pass the high_bits parameter in the mixed_2_6 and mixed_3_6 quant predicates instead of using default value.

Fixes [ml-explore/mlx-examples#17](https://github.com/ml-explore/mlx-lm/issues/17)

Migrated from [mlx-examples](https://github.com/ml-explore/mlx-examples/pull/1348) repo.


awni (approved):
Looks great, thanks!

PR #29, opened 2025-03-18
Use the prompt cache for completions in server

awni:
Previously we only used the prompt cache for multi-turn conversation. This also uses the cache for multiple queries on the same long prompt.


barronalex (approved):
Thanks for this -- looks good!

PR #32, opened 2025-03-18
dequantize dsv3

awni:
Since we can load fp8, allow dequantization in mlx-lm directly to simplify creating new quants of dsv3.


barronalex (approved):
Nice!

PR #37, opened 2025-03-19
Refactor utils

awni:
This is an attempt to refactor the `utils.py` file from being a single monolith for everything.

What we lose is `python -m mlx_lm.convert` is not quite valid anymore since we import `convert` in `__init__` and Python doesn't like that:

```
<frozen runpy>:128: RuntimeWarning: 'mlx_lm.convert' found in sys.modules after import of package 'mlx_lm', but prior to execution of 'mlx_lm.convert'; this may result in unpredictable behaviour
```

So instead one will have to use:

- `mlx_lm.convert` (preferred method)
- `python -m mlx_lm convert` (notice the missing period)

If we are ok with this approach, I will do the same for `mlx_lm.generate`.


angeloskath:
Hmm. I don't mind but I generally am quite relaxed with these things. I am using both but replacing a dot with a space shouldn't be a problem.

Having said that, we can always just rename it to something else ie `_convert` and import it in `__init__.py` and `convert.py` . Personally I like better what you are doing now but this could be an option.

angeloskath (approved):

PR #39, opened 2025-03-19
Fix gemma3 and cohere2 sliding window

### LINKED ISSUE #35: Gemma 3 and Command A become unusable after 5000 tokens
With the latest mlx-lm v0.22.1 I tried running Gemma 3 and Command A. Both seem to completely lose coherency at around ±5000 tokens. Gemma 3 at Q8 and Command A at Q4 and a mixed Q4-8 quant. I also convert with group size 32, if that might be related.

Also tried other models like Qwen 2.5 32B with the latest mlx-lm, but they stay coherent.

awni:
As title, closes #35

### DIFF
--- a/mlx_lm/models/cohere2.py
+++ b/mlx_lm/models/cohere2.py
@@ -179,6 +179,7 @@ class CohereModel(nn.Module):
                 == self.args.sliding_window_pattern - 1
             )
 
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
👍

PR #40, opened 2025-03-20
Updates for causal mask

awni:
Pre/Post for 4-bit Mistral 7B on M2 Utra

Prompt: 33898 tokens, 676.212 tokens-per-sec
Peak memory: 12.257 GB

Prompt: 33898 tokens, 869.279 tokens-per-sec
Peak memory: 9.111 GB


angeloskath (approved):
👍

PR #44, opened 2025-03-21
Adding support for Sliding Window Attention

Goekdeniz-Guelmez:


Goekdeniz-Guelmez:
So, the forward pass works, with sliding window, I'll try to do it with a qwen2.5 model when Im home.

Goekdeniz-Guelmez:
@awni should be mergeable!!

awni:
@Goekdeniz-Guelmez can you say more about what this is for? IIUC it's supposed to be an efficient implementation of sliding window attention? If so would be useful to see some benchmarks to understand what this is buying us.

Goekdeniz-Guelmez:
@awni I beleihe MLX doesnt have Sliding window natively, and hence Gemma3 and Qwen3 have that, I thought I would just add that support. But yes, it’s an efficient implementation of sliding window attention to handle long-context inputs more efficiently. I’ll later put together some benchmarks comparing it to standard attention (latency, memory, maybe throughput).

PR #53, opened 2025-03-25
Set wired limit and compile step

awni:
- Benchmark was done on top of https://github.com/ml-explore/mlx/pull/2006
- We get a nice speed boost from both wiring memory and compiling the step function
- Compiling the step function also reduce RAM substantially

Training a 1B with ~2048 sequence length and batch size 4:

Pre:
```
Iter 10: Train loss 1.415, Learning Rate 1.000e-05, It/sec 0.225, Tokens/sec 1819.193, Trained Tokens 80880, Peak mem 47.302 GB
```

Post:

```
Iter 10: Train loss 1.382, Learning Rate 1.000e-05, It/sec 0.284, Tokens/sec 2295.674, Trained Tokens 80880, Peak mem 40.508 GB
```

### DIFF
--- a/mlx_lm/tuner/trainer.py
+++ b/mlx_lm/tuner/trainer.py
@@ -4,6 +4,7 @@ import glob
 import shutil
 import time
 from dataclasses import dataclass, field
+from functools import partial
 from pathlib import Path
 from typing import List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
🚀

PR #55, opened 2025-03-27
fix missing self references in ChatDataset class

Goekdeniz-Guelmez:
Instance Attribute Scope Fixing, so added missing `self` in `offset = len(tokenizer.apply_chat_template(...))`, and in `def itemlen(idx: int)`.


awni (approved):
Thanks for the fix!

PR #63, opened 2025-03-31
Update logits to new mlx + version bump

awni:
- Use the model dtype for logits and loss to save RAM (since logsumexp accumulates in high precision this should be ok). Only cast the losses (which are much smaller) prior to the reduction.
- Use `mx.fast.rms_norm` in plamo2


barronalex (approved):
Looks good!

PR #65, opened 2025-04-02
Remove alpha parameter from LoRA defaults

nfergu:
This PR removes the alpha parameter from the default LoRA config. I don't think it's used, per [this comment](https://github.com/ml-explore/mlx-examples/issues/982#issuecomment-2353883500):

> We don't ever explicitly compute α. The scale parameter is used directly W + scale * a @ b.T. If you wanted to work out the implied α you could multiply the scale by the lora rank you are using e.g. scale * rank.

So it's confusing to me to have it in the defaults. Particularly since the default (16) doesn't seem to match the default rank (8) and scale (10.0).

But maybe I've missed something. Feel free to close this PR if so.

### DIFF
--- a/mlx_lm/lora.py
+++ b/mlx_lm/lora.py
@@ -67,7 +67,7 @@ CONFIG_DEFAULTS = {
     "config": None,
     "grad_checkpoint": False,
     "lr_schedule": None,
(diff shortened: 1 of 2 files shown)


awni (approved):
Thanks!

PR #66, opened 2025-04-03
Count accept rate of speculative decoding

wangkuiyi:
This PR is only for reproducing the problem that speculative does not accelerate the inference of model Qwen 2.5 32B 3bit.  It is **not** indent to merge.

```shell
python tests/benchmark_speculative_decoding.py  | tee /tmp/logg
```

Copying numbers from the output file `/tmp/logg` into a spreadsheet gave us the following:

Screenshot 2025-04-02 at 11 19 39 PM

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -514,10 +514,12 @@ def speculative_generate_step(
     # Set these so the finally block doesn't raise
     num_draft = 0
     n = 0
+    speculation = 0  # count the calls to the draft model
     try:
         while True:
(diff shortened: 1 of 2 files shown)


wangkuiyi:
Added an experiment using the 4-bit version of the main model. We can see the performance improvement from speculative decoding:

Screenshot 2025-04-03 at 8 39 14 AM

wangkuiyi:
I’m closing this PR as it has fulfilled its purpose of reporting the observations.

Since it doesn’t offer a clean or elegant solution for printing speculative decoding performance data, I don’t believe it’s worth merging.

PR #67, opened 2025-04-03
fix / improve dsv3

awni:
3 days later...

Several things in this PR

- The big fix is not using the bias shifted scores to weight the expert hidden states but using the original scores instead
- Some precision improvements to DSV3
  - When dequanting from the original do it in fp32 since the scales are in fp32 and then down cast (rather than down cast the scales and then multiply). This matches how the DSV3 dequant kernel works and should give results closer to the original.
  - Use the right epsilon for some RMSNorms
  - Keep the expert score bias in fp32. Casting them to bf16 is quite destructive. And since this computation is done in fp32 anyway there is no perf issue. Adds a mechanism to let a model indicate which keys should not be cast.
  - Fix the pipeline parallel dsv3

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -102,7 +102,13 @@ def convert(
 
     weights = dict(tree_flatten(model.parameters()))
     dtype = getattr(mx, dtype)
-    weights = {k: v.astype(dtype) for k, v in weights.items()}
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
Thank you for fixing these! I am even more excited to use the models now!!!

PR #68, opened 2025-04-03
Add support for temperature

chaytanc:
Modify generate to add support for temperature argument


awni:
It already supports that via a sampler. See e.g. https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/generate.py#L809

PR #70, opened 2025-04-04
Distributed Eval

barronalex:
Add distributed support to mlx_lm.evaluate.

On one machine:
```
{
    "alias": "winogrande",
    "acc,none": 0.6156274664561957,
    "acc_stderr,none": 0.01367156760083619
}
```

On two machines:
```
{
    "alias": "winogrande",
    "acc,none": 0.6156274664561957,
    "acc_stderr,none": 0.01367156760083619
}
```

Also cleans up the code according to the comments from the original [PR](https://github.com/ml-explore/mlx-examples/pull/1174) and fixes a few bugs:

- Add a causal mask with sequence length padding
- Fix `--no-apply-chat-template`
- Default args.limit to None. This matches `lm_eval` and before you were forced to set a limit.

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -55,6 +55,10 @@ def _pad_inputs(inputs):
 
 @register_model("mlxlm")
 class MLXLM(LM):
+
+    tokenizer_name = lm_eval.models.huggingface.HFLM.tokenizer_name
+    apply_chat_template = lm_eval.models.huggingface.HFLM.apply_chat_template
+
     def __init__(
         self,
         path_or_hf_repo: str,
@@ -236,9 +240,6 @@ class MLXLM(LM):
         inv_sort = np.argsort(sorted_indices)
         results = [results[inv_sort[i]] for i in range(len(inv_sort))]
         return results
-
-    tokenizer_name = lm_eval.models.huggingface.HFLM.tokenizer_name
-    apply_chat_template = lm_eval.models.huggingface.HFLM.apply_chat_template
 
     def loglikelihood_rolling(self, requests) -> list[float]:
         """Compute full log-likelihood of a string, with no truncation, for perplexity computation
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -394,8 +394,9 @@ def main():
     task_names = "_".join(args.tasks)
     ver = version("lm_eval")
     filename = f"eval_{model_name}_{task_names}_{args.num_shots:02d}_v_{ver}.json"
-    output_path = output_dir / filename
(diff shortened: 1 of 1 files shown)


mlx_lm/evaluate.py
```diff
@@ -223,16 +199,54 @@ def loglikelihood(self, requests) -> list[tuple[float, bool]]:
...
+        scores = np.array(scores[:num_results])
+        is_greedy = np.array(is_greedy[:num_results])
+
+        results = [(score, ig) for score, ig in zip(scores, is_greedy)]
+        inv_sort = np.argsort(sorted_indices)
+        results = [results[inv_sort[i]] for i in range(len(inv_sort))]
+        return results
 
     tokenizer_name = lm_eval.models.huggingface.HFLM.tokenizer_name
-    apply_chat_template = lm_eval.models.huggingface.HFLM.apply_chat_template
```
angeloskath, line 235: Out of curiosity is the tokenizer name above used by lm harness? Should it be moved to the top or set in `__init__` as an instance member rather than as a class member?

angeloskath (approved):
Looks great and works awesomely :-) 

So far I tested with Qwen and Gemma and both need the `--no-apply-chat-template`. Which models do need the chat template?

I think we should guard the result writing with `if mx.distributed().init().rank() == 0` or sth along these lines.

awni:
> So far I tested with Qwen and Gemma and both need the --no-apply-chat-template. Which models do need the chat template?

Even if they are chat models? And what do you mean by "need"? Like the results are much worse if you use the template? If so, that seems strange 🤔

angeloskath:
Hm I just realized that the behavior changed from before. No the results aren't worse it breaks with `AttributeError: 'list' object has no attribute 'tokenizer'`.

Maybe I need to update some package but `lm_eval.models.huggingface.HFLM.apply_chat_template` is the call that errors out in line 246.

@barronalex any ideas?

barronalex:
Good catch here! I didn't mean to change that part of the chat template logic but it must have come in during a rebase. I've reverted it.

mlx_lm/evaluate.py
barronalex: tokenizer_name is a method we're borrowing from the hugging face [lm eval backend](https://github.com/EleutherAI/lm-evaluation-harness/blob/11ac352d5f670fa14bbce00e423cff6ff63ff048/lm_eval/api/model.py#L190) so I think it's OK to have it as an instance member here.
I have moved them to the top though as I agree that makes it a bit easier to read.

angeloskath: Gotcha makes sense.

PR #71, opened 2025-04-04
Move from `python -m mlx_lm.generate` to `python -m mlx_lm generate`

angeloskath:
We had done the move and added the deprecation notice but the commands weren't available in `mlx_lm/__main__.py`. Added all of them and more deprecation notices.


awni (approved):
Thanks!!

PR #62, opened 2025-03-31
Adds support for speculative decoding to mlx_lm.server

Art9681:
This pull request introduces speculative decoding to the `mlx_lm` server to improve text generation performance. The main changes include adding support for a draft model, modifying the prompt cache, and updating the server to handle speculative decoding. Additionally, new tests are added to ensure the functionality of the draft model.

### Speculative Decoding Support:
* `mlx_lm/SERVER.md`: Added documentation for enabling speculative decoding using a draft model.
* `mlx_lm/server.py`: Introduced a draft model and associated parameters, updated the prompt cache to handle the draft model, and modified the completion handling to use speculative decoding.

### Testing:
* `tests/test_server.py`: Added new tests for the server with the draft model to ensure speculative decoding works correctly, including tests for handling completions, chat completions, streaming, and prompt cache functionality.

### Test Results:
```
$ python -m unittest tests/test_server.py                                 
Fetching 9 files: 100%|███████████████████████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 63019.59it/s]
127.0.0.1 - - [31/Mar/2025 11:27:46] "POST /v1/chat/completions HTTP/1.1" 200 -
.127.0.0.1 - - [31/Mar/2025 11:27:46] "POST /v1/chat/completions HTTP/1.1" 200 -
.127.0.0.1 - - [31/Mar/2025 11:27:46] "POST /v1/completions HTTP/1.1" 200 -
.127.0.0.1 - - [31/Mar/2025 11:27:46] "GET /v1/models HTTP/1.1" 200 -
Fetching 9 files: 100%|███████████████████████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 10169.38it/s]
Fetching 9 files: 100%|███████████████████████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 14080.10it/s]
127.0.0.1 - - [31/Mar/2025 11:27:48] "POST /v1/chat/completions HTTP/1.1" 200 -
.127.0.0.1 - - [31/Mar/2025 11:27:48] "POST /v1/completions HTTP/1.1" 200 -
.127.0.0.1 - - [31/Mar/2025 11:27:48] "POST /v1/chat/completions HTTP/1.1" 200 -
127.0.0.1 - - [31/Mar/2025 11:27:48] "POST /v1/chat/completions HTTP/1.1" 200 -
.127.0.0.1 - - [31/Mar/2025 11:27:48] "POST /v1/chat/completions HTTP/1.1" 200 -
[...]

PR #85, opened 2025-04-08
Server draft

awni:
Alternate of https://github.com/ml-explore/mlx-lm/pull/62 with some modifications.

### DIFF
--- a/mlx_lm/SERVER.md
+++ b/mlx_lm/SERVER.md
@@ -41,6 +41,19 @@ curl localhost:8080/v1/chat/completions \
    }'
 ```
 
+## Speculative Decoding
+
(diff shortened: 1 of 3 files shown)


Art9681:
Great work. Thank you!

shrikrishnaholla:
Looking forward to this MR. I run a local llm server for my dev team on mlx and I'm looking forward to the speedup I'm expecting this to bring!

Raize:
I'm also looking forward to seeing this merged.

angeloskath (approved):
This looks great! Sorry for taking too long to review.

I think in another PR we should probably refactor it a bit so that the model provider is the only one holding the model (and tokenizer) at any given time similar to holding the draft model for instance. I don't think it is as simple as setting it as a member variable after load (this would probably work but it would not be different) which is why I am not proposing something specific.

PR #86, opened 2025-04-08
Support nemotron-nas

cavit99:
Adding support for nemotron-nas such as
> nvidia/Llama-3_1-Nemotron-Ultra-253B-v1
> nvidia/Llama-3_3-Nemotron-Super-49B-v1

[Paper](https://arxiv.org/abs/2411.19146)


awni:
Thanks for the addition! Very nice! I made a few minor edits, removed some unneeded stuff and also the code was pretty heavy on comments compared to the style of the rest of mlx-lm so I removed a few extraneous ones here and there. Just FYI for future reference.

awni (approved):

PR #87, opened 2025-04-10
Remove vision_config from llama4 and from model conversion

neilmehta24:
Remove vision_config from llama4 since it's unused, so it shouldn't be a requirement for loading a model

Also, remove vision_config from the model conversion script so that it's easier for MLX users to see if the model was converted with vision enabled


awni (approved):
Thanks @neilmehta24 !

PR #88, opened 2025-04-10
Support AWQ Quantization

barronalex:
Adds a command to use [AWQ](https://arxiv.org/abs/2306.00978) to quantize a model.

AWQ uses a calibration dataset to scale and clip `nn.Linear` layers, resulting in more accurate quantization. This is currently hard-coded to wiki text but should be easy to customize in the script.

Currently we support a subset of the model types in mlx-lm: Llama, Qwen and Gemma3. I prioritized the most popular according to the HF [mlx-community](https://huggingface.co/mlx-community?sort_models=downloads#models).

# Performance

Since the scales and clips are fused into the previous layer, the resulting model runs at exactly the same tokens per second as the normally quantized model.

# Quality
Quality benchmarks are run on 8 zero shot `lm_eval` tasks using `mlx_lm.evaluate`:
```
winogrande boolq arc_challenge arc_easy hellaswag openbookqa piqa social_iqa
```

`n-grid=20` `num-samples=256`

Scores are a percentage of average bf16 performance across the 8 tasks.
For all quants, embedding/lm_head layers are quantized with 4 bits, 32 group size.

## Qwen 2.5 7B

| Quant | % of bf16 |
| --- | --- |
| bf16 |  100.0 |
| 3 bit AWQ | 96.9 |
| 3 bit | 92.4 |

## Llama 3.2 1B
| Quant | % of bf16 |
| --- | --- |
| bf16 |  100.0 |
| 4 bit AWQ | 99.0 |
| 4 bit | 97.7 |
| 3 bit AWQ | 94.2 |
| 3 bit | 87.4 |

## Gemma 3 1B
| Quant | % of bf16 |
| --- | --- |
| bf16 |  100.0 |
| 4 bit AWQ | 98.4 |
| 4 bit | 96.1 |

# Quantization Time
AWQ is much more expensive at quantization time than our affine quants. A few examples:

M1 Max, 1B model, `--n-grid 10 --num-samples 32`:  15 minutes

M2 Ultra, 7B model, `--n-grid 20 --num-samples 256`: 3 hours

The script supports distributed execution which linearly speeds up the run time.

In practice I have seen more samples result in better task performance (as shown in the original paper), but I haven't done a comprehensive ablation.

# MOE
There's experimental MOE support in the script. I ran a few tests with MOE models and didn't see much improvement in the downstream task performance. It's possible there a bug or maybe MOE models require many more samples since all the experts need to be taken into account by a single scale.

### DIFF
--- a/mlx_lm/awq.py
+++ b/mlx_lm/awq.py
@@ -1,0 +1,439 @@
+# Learned quantization using AWQ:
+
+# References:
+# AWQ
+# https://arxiv.org/abs/2306.00978
+# https://github.com/mit-han-lab/llm-awq
+
+import argparse
+import glob
+import shutil
+from pathlib import Path
+from typing import Callable
+
+import mlx.core as mx
+import mlx.nn as nn
+import numpy as np
+from datasets import Dataset, load_dataset
+from mlx.utils import tree_flatten, tree_map_with_path
+from mlx_lm.models.base import create_attention_mask
+from mlx_lm.tokenizer_utils import TokenizerWrapper
+from mlx_lm.utils import fetch_from_hub, get_model_path, save_config, save_weights
+from tqdm import tqdm
+
+
+def mse(x, y):
+    return ((x - y).astype(mx.float32)) ** 2
+
+
+def run_layer(layer: nn.Module, x: mx.array, batch_size: int = 32, **kwargs):
+    y = []
+    for i in range(0, x.shape[0], batch_size):
+        y.append(layer(x[i : i + batch_size], **kwargs))
+        mx.eval(y)
+    y = mx.concatenate(y, axis=0)
+    return y
+
+
+def dist_split(x: mx.array, group: mx.distributed.Group):
+    B = x.shape[0]
+    N = group.size()
+    assert B % N == 0
+    r = group.rank()
+    local_B = (B + N - 1) // N
+    return x[r * local_B : (r + 1) * local_B]
+
+
+def search_best_scale(
+    layers: list[nn.Module],
+    x: mx.array,
+    quantize_func: Callable,
+    block: nn.Module | None = None,
+    layer_kwargs: dict | None = None,
+    n_grid: int = 20,
+):
+    group = mx.distributed.init() if mx.distributed.is_available() else None
+    layer_kwargs = layer_kwargs or {}
+
+    block = block or layers[0]
+    out = block(x, **layer_kwargs)
+
+    x_max = x.abs().mean(axis=(0, 1))
+
+    best_error = float("inf")
+    best_scales = None
+
+    weights = tree_flatten(block.parameters())
+
+    for ratio in tqdm(range(n_grid)):
+        ratio = ratio * 1 / n_grid
+        scales = mx.maximum(x_max**ratio, 1e-4).reshape(-1)
+        scales = scales / (scales.max() * scales.min()).sqrt()
+        for layer in layers:
+            layer.weight = quantize_func(layer.weight * scales) / scales
+
+        out_q = run_layer(block, x, **layer_kwargs)
+        loss = mse(out, out_q).sum()
+        if group is not None:
+            loss = mx.distributed.all_sum(loss, stream=mx.cpu) / group.size()
+        loss /= out.size
+        mx.eval(loss)
+        is_best = loss < best_error
+        if is_best:
+            best_error = loss
+            best_scales = scales
+
+        # reload the original weights
+        block.load_weights(weights)
+
+    best_scales = best_scales.reshape(-1)
+    mx.eval(best_scales)
+    return best_scales
+
+
+def apply_scale(prev_op, layers, scales):
+    # Apply the scales to the layers
+    if isinstance(prev_op, nn.Linear):
+        assert len(layers) == 1
+        prev_op.weight = prev_op.weight / scales[:, mx.newaxis]
+        if hasattr(prev_op, "bias"):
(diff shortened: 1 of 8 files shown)


mlx_lm/awq.py
```diff
@@ -0,0 +1,541 @@
...
+    model: nn.Module,
+    tokenizer: TokenizerWrapper,
+    config,
+    model_path: Path,
+    mlx_path: str,
+):
+    weights = dict(tree_flatten(model.parameters()))
+
+    mlx_path = Path(mlx_path)
+    save_weights(mlx_path, weights, donate_weights=True)
```
awni, line 463: I think for `donate_weights` to be useful here, we need to `del model`. But it also probably isn't super relevant either way since we presumably need the entire model in memory

mlx_lm/awq.py
```diff
@@ -0,0 +1,541 @@
...
+
+    # Select random chunks
+    starts = mx.random.randint(
+        0, len(tokens) - sequence_length - 1, shape=(num_samples, 1)
+    )
+    data = tokens[starts + mx.arange(sequence_length)]
+    if tokenizer.bos_token_id:
+        data = mx.concatenate(
+            [mx.full((*data.shape[:1], 1), tokenizer.bos_token_id), data], axis=-1
+        )
```
awni, line 449: It's a bit odd to use the bos token here since the segments are just randomly pulled out from a long stretch of text. I'm not sure if you tried without it.. but my guess is it's probably not useful and maybe even slightly worse to use it. I might just use `sequence_length+1` segments instead.

mlx_lm/awq.py
```diff
@@ -0,0 +1,541 @@
...
+        class Catcher(nn.Module):
+            def __call__(self, x: mx.array, *args, **kwargs):
+                # Store the input features on the original modules.
+                if hasattr(module, "input_feat"):
+                    module.input_feat = mx.concatenate([module.input_feat, x], axis=0)
+                else:
+                    module.input_feat = x
+
+                # Also store the MOE indices if applicabale
+                module.indices = None
```
awni, line 363: Does it make sense to always set this to `None` here? I might be misunderstanding.. but it seems like you won't be able to aggregate the indices across multiple calls if they are always set to `None`

mlx_lm/awq.py
```diff
@@ -0,0 +1,540 @@
...
+def clip_block(
+    block: nn.Module, quantize_func: Callable, group_size: int, n_grid: int = 20
+):
+
+    def apply_clip(path, module):
+        if (
+            isinstance(module, nn.Linear)
+            and "q_proj" not in path
+            and "k_proj" not in path
+        ):
```
awni, line 316: Why not a learned clip for those as well? Is that an empirical thing that it doesn't help much /hurts to clip them?

mlx_lm/awq.py
```diff
@@ -0,0 +1,535 @@
...
+                    block.mlp.gate_proj,
+                    block.mlp.up_proj,
+                ],
+            ),
+            LayerConfig(
+                prev_op=block.mlp.up_proj,
+                layers=[
+                    block.mlp.down_proj,
+                ],
+            ),
```
awni, line 221: It's a little confusing to me that we find an optimal scale for the `gate_proj` and `up_proj` but then we fuse the scale found for the `down_proj` into the `up_proj` which could revert the improvement for the `up_proj`. Maybe we should search and apply the `down_proj` before we search for the `gate_proj` and `up_proj`.

mlx_lm/awq.py
awni: Found this comment: https://github.com/mit-han-lab/llm-awq/blob/main/awq/quantize/auto_clip.py#L74-L76

Not sure exactly what it means.. but it seems this is the same behavior of as the original.

mlx_lm/awq.py
awni: Fixed.

awni:
I decoupled the config from the algorithm a bit so that it's easier to add support for new models. I left the configs in the main file. If this grows we may consider splitting out the model configs into a separate location.

awni:
Added an `mlx_lm.upload` to make it easy to upload models in a saved directory. So now we can do an awq and then easily upload to HF like.

awni:
I added a short readme. Otherwise, I think this is good to go! I'll merge once the tests clear.

awni (approved):

ivanfioravanti:
I added default options in README because I was planning some tests and gave for granted 4bit was default, but it's not.
Just to clarify this better without opening the script.

Why --grid-size 10 as default, when in paper is 20, just because it's faster? Thanks @barronalex

barronalex:
grid-size=10 is faster and I didn't see large quality differences but it's definitely worth experimenting.

ahmetaltunel:
It makes sense that "datasets" (required by awq.py) and "lm-eval" (evaluate.py) are optional extras (as extras_require [test], [evaluate]) to keep the mlx-lm installation lightweight.

However, users might not realize they need to install the extras separately and could run into errors.

To help with this, could you consider adding a brief note in the relevant readme (e.g., AWQ.md  +evaluation section) to mention installing the required dependencies?

Potentially like:
awq/test extras: pip install mlx-lm[test]
evaluate extras: pip install mlx-lm[evaluate]
or having a separate extras_require for awq:  pip install mlx-lm[awq]

awni:
I added a `pip install mlx-lm[awq]` and a comment about it in the readme. Thanks for the input.

PR #97, opened 2025-04-16
Feature/resumable lora with metadata

gingofthesouth:
Summary
-------

This pull request introduces resumable LoRA fine-tuning for mlx-lm, motivated by real-world issues encountered during training. When training was interrupted due to Out-Of-Memory (OOM) errors on macOS Metal, there was previously no way to resume progress. Training would always restart from iteration 1, overwriting adapter files leaving us back at square one.

This change adds:

*   The ability to resume training from the most recent adapter checkpoint (--resume-adapter-file)
*   Metadata saved in adapter .safetensors files to track the training iteration
*   Optimizer state saved and restored so training resumes accurately. 
*   Test coverage to validate resume behavior, metadata integrity, and optimizer state

Changes
-------

*   train\_model() and train() updated to track iteration count
*   trainer.py saves the current training step  and optimizer state in adapter metadata
*   On resume, the --resume-adapter-file checkpoint is loaded and the iteration count and optimizer state are restored
*   Filenames of the form 000XXXX\_adapters.safetensors continue correctly from the last step
    
*   Added test script test\_lora\_resume.py that tests saving and resuming from iteration and optimizer metadata
        

Why this is needed
------------------

Without resume functionality, training interruptions due to memory limits or system crashes result in wasted compute time and no way to continue training. This update provides robust continuation support while maintaining compatibility with existing workflows.

Test Plan
---------

*   Manual testing with --resume-adapter-file on interrupted training
*   Automated test test\_lora\_resume.py validates end-to-end resume behavior

### DIFF
--- a/mlx_lm/lora.py
+++ b/mlx_lm/lora.py
@@ -215,9 +215,34 @@ def train_model(
         raise ValueError(f"Received unknown fine-tune-type {args.fine_tune_type}")
 
     # Resume from weights if provided
+    start_iteration = 1
     if args.resume_adapter_file is not None:
-        print(f"Loading fine-tuned weights from {args.resume_adapter_file}")
+        adapter_file = Path(args.resume_adapter_file)
+        if adapter_file.is_dir():
+            safetensor_files = sorted(
+                adapter_file.glob("*_adapters.safetensors"),
+                key=lambda f: int(f.name.split("_")[0]),
+                reverse=True,
+            )
+            if not safetensor_files:
+                raise ValueError("No adapter files found to resume from.")
+            latest = safetensor_files[0]
+            print(f"Auto-resuming from latest adapter file: {latest}")
+            args.resume_adapter_file = str(latest)
+        else:
+            print(f"Resuming from: {args.resume_adapter_file}")
         model.load_weights(args.resume_adapter_file, strict=False)
+
+        # Log resume state and extract iteration
+        from safetensors.numpy import safe_open
+
+        with safe_open(args.resume_adapter_file, framework="numpy") as f:
+            metadata = dict(f.metadata())
+            print("✅ Resuming from checkpoint:")
+            print("  Metadata:", metadata)
+            if "iteration" in metadata:
+                start_iteration = int(metadata["iteration"]) + 1
+                print(f"  Continuing from iteration {start_iteration}")
 
     print_trainable_parameters(model)
 
@@ -264,6 +289,7 @@ def train_model(
         train_dataset=train_set,
         val_dataset=valid_set,
         training_callback=training_callback,
+        start_step=start_iteration,
     )
 
 
--- a/mlx_lm/lora.py
+++ b/mlx_lm/lora.py
@@ -15,7 +15,13 @@ import yaml
 
 from .tokenizer_utils import TokenizerWrapper
 from .tuner.datasets import load_dataset
-from .tuner.trainer import TrainingArgs, TrainingCallback, evaluate, train
+from .tuner.trainer import (
+    TrainingArgs,
+    TrainingCallback,
+    evaluate,
+    load_checkpoint,
+    train,
+)
 from .tuner.utils import (
     build_schedule,
     linear_to_lora_layers,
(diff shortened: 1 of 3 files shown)


chimezie:
Is restoring the step number alone sufficient for resuming the optimizer and learning schedule state, as described [here](https://github.com/ml-explore/mlx-examples/issues/1237)?

gingofthesouth:
> Is restoring the step number alone sufficient for resuming the optimizer and learning schedule state, as described [here](https://github.com/ml-explore/mlx-examples/issues/1237)?

Currently, restoring the step number is sufficient to resume adapter weight training from the correct iteration (and this pr accounts for that when resuming from adapter checkpoints with no metadata). However, it does not restore optimizer internals (e.g., momentum buffers, learning rate schedule progress). This patch adds metadata as a foundation for fully resumable training. Future work could store the optimizer state and LR schedule status using this metadata to improve training continuity and reproducibility.

gingofthesouth:
Who do I have to pay to get a code review around here? 😅

awni:
@gingofthesouth thanks for the contribution and sorry for the delay. Just to set your expectations it might take us a little while (~week) to get to this, but we'll try and take a look soon.

gingofthesouth:
> @gingofthesouth thanks for the contribution and sorry for the delay. Just to set your expectations it might take us a little while (~week) to get to this, but we'll try and take a look soon.

No worries at all, we’re all juggling a lot, so I completely understand. I mostly just wanted to get the ball rolling. As I dive deeper into the library and learn more about LLM training and tuning, I’m finding more areas where I could potentially contribute — like optimiser behaviour and restoring state when resuming training.
I also noticed there’s no automated MR pipeline in place (unit tests, linting, static analysis, or even an agent-assisted first-pass review), so could potentially help there if that was of any value to the project.

it's a fun journey.

mlx_lm/lora.py
```diff
@@ -215,10 +215,35 @@ def train_model(
...
+            print(f"Auto-resuming from latest adapter file: {latest}")
+            args.resume_adapter_file = str(latest)
+        else:
+            print(f"Resuming from: {args.resume_adapter_file}")
         model.load_weights(args.resume_adapter_file, strict=False)
 
+        # Log resume state and extract iteration
+        from safetensors.numpy import safe_open
+
+        with safe_open(args.resume_adapter_file, framework="numpy") as f:
```
angeloskath, line 239: You can load the metadata in mlx by doing `_, metadata = mx.load(args.resume_adapter_file, return_metadata=True)`.

mlx_lm/lora.py
```diff
@@ -215,10 +215,35 @@ def train_model(
         raise ValueError(f"Received unknown fine-tune-type {args.fine_tune_type}")
 
     # Resume from weights if provided
+    start_iteration = 1
```
angeloskath, line 218: I would refactor the following code into a separate function in `trainer.py`, see my comment about saving, it would be a its counterpart.

mlx_lm/tuner/trainer.py
```diff
@@ -339,11 +340,22 @@ def step(batch):
         # Save adapter weights
         if it % args.steps_per_save == 0 and rank == 0:
             adapter_weights = dict(tree_flatten(model.trainable_parameters()))
-            mx.save_safetensors(str(args.adapter_file), adapter_weights)
```
angeloskath, line 342: I would refactor the following into a `save_checkpoint` function. In order for resume to be really useful we need to save the optimizer state as well so together with the adapter weights, I would save the `optimizer.state` as well.

This also makes any lr schedule work as expected as the optimizer will be using the same step number etc.

Then you would need a `load_checkpoint` function which I would write quite close to the one above that takes a file, a model and an optimizer and returns the start iteration.

tests/test_lora_resume.py
angeloskath, line 1: We are not using `pytest` but `unittest` see `tests/test_finetune.py`.

angeloskath (changes requested):
This is I think something useful but it needs a bit of work. I left comments throughout.

mlx_lm/lora.py
gingofthesouth: Thank you for pointing this out. I've moved to this method for retrieving the metadata.

mlx_lm/tuner/trainer.py
gingofthesouth: Good idea. I've moved the code to trainer.py and refactored it to include the optimizer code too.
I had a few struggles but think I got there in the end. I am going to be honest, I'm finding coding in python is very foreign to me. I keep thinking to build utility classes, extensions, protocols and dependancy injection patterns. Thus I find myself a little lost about where to put code.

Any further feedback will be warmly received as I am learning the language ecosystem as I go.

mlx_lm/lora.py
gingofthesouth: Done.

tests/test_lora_resume.py
gingofthesouth: I viewed the other tests and cleaned this up. Hopefully the new test code aligns better with the existing testing patterns and test libraries.

gingofthesouth: Hi @angeloskath do you have any feedback on the changes I have made relating to your review?

zcbenz:

PR #100, opened 2025-04-17
Inform the gather mm regarding the sorted indices

angeloskath:
As title. It requires https://github.com/ml-explore/mlx/pull/2078 . I am not sure I like the `should_sort` name and crude condition so let me know what you think. I could change it to something like the following

```
indices.size / num_experts >= 8
```
which makes more sense imho.

### DIFF
--- a/mlx_lm/models/switch_layers.py
+++ b/mlx_lm/models/switch_layers.py
@@ -71,7 +71,7 @@ class QuantizedSwitchLinear(nn.Module):
     def num_experts(self):
         return self.weight.shape[0]
 
-    def __call__(self, x, indices):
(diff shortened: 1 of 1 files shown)


awni:
What's the logic behind the current condition? It does seem like the condition in your comment makes more sense / at least is easier to understand.

angeloskath:
Yeah there isn't much exploration behind the current condition. The current one in my mind is `if we_are_in_prompt_processing:`.

angeloskath:
The more I think about it. The previous condition was also important for routing to the non-batch gather_qmm so I will make a hybrid one and push it.

awni (approved):
🚀 !

PR #102, opened 2025-04-18
fix gemma3 in fp16

awni:
Thanks @Blaizzy for pointing out the fix for fp16.


angeloskath (approved):
🚀

Blaizzy:
My pleasure ❤️💪🏾

This should solve the problem!

No need to re-upload the model weights in bf16

PR #107, opened 2025-04-19
Patch tokenizer utils to match MLX-VLM

Blaizzy:
This PR will allow mlx-vlm to use mlx-lm's tokenizer_utils directly instead of duplicating.

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -1,4 +1,5 @@
 import json
+from json import JSONDecodeError
 from functools import partial
 from typing import List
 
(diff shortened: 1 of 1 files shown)


mlx_lm/tokenizer_utils.py
```diff
@@ -341,7 +342,7 @@ def _is_bpe_decoder(decoder):
     return isinstance(decoder, dict) and decoder.get("type", None) == "ByteLevel"
 
 
-def load_tokenizer(model_path, tokenizer_config_extra={}, eos_token_ids=None):
+def load_tokenizer(model_path, tokenizer_config_extra={}, return_tokenizer=True, eos_token_ids=None):
```
awni, line 345: What's the purpose of that flag?

Blaizzy: MLX-VLM uses the processor so I only need the detokenizer class and not the tokenizerwrapper class.

awni: What about the eos token ids? Do you manage those separately? They are part of the `TokenizerWrapper` and pretty important for the stopping condition.

Blaizzy: Yes, I manage them separately at the processor level.

Precisely, I've created a stopping criteria class. It does the same, but also allows users to pass custom tokens or a callable.

PRs:
- https://github.com/Blaizzy/mlx-vlm/pull/310
- https://github.com/Blaizzy/mlx-vlm/pull/307

awni:
Can you run the format hooks? O/w good to go here.

Blaizzy:
Awesome, Done ✅

awni (approved):
Thanks @Blaizzy glad you are moving towards using more mlx-lm!! That will simplify stuff a lot.

Blaizzy:
My pleasure! ❤️💪🏾

Indeed it will simplify a lot and it's a great way for us to collaborate.

Issue #103, opened 2025-04-19
Bug during inference with large prompt context (16K+) in MOE


ivanfioravanti:
Tested with latest mlx-lm 0.23 and mlx 0.25.
I tried different temperature and max-tokens, on M3 Ultra and M4 Max.

`cat 16k.txt | mlx_lm.generate --model mlx-community/Llama-4-Scout-17B-16E-Instruct-4bit --max-tokens 1000 --prompt -`
Output generation starts well, then it gets stuck and print garbage (same token till end of inference)

![Image](https://github.com/user-attachments/assets/1aaf6bb7-5979-40f8-8116-5fb4b423e827)

Attached the file used for testing, first chapter of public domain Pride & Prejudice. It works well till 8K
[16k.txt](https://github.com/user-attachments/files/19819663/16k.txt)

Issue is not present with non -MOE models: 
` cat 16k.txt | mlx_lm.generate --model mlx-community/gemma-3-27b-it-qat-4bit --max-tokens 1000 --prompt -`

awni:
This was broken in mlx==0.24.1 so I don't think it's related to the recent MoE update and in fact may not be related to MoE at all. It's more likely an issue with Llama4. Will investigate.

PR #108, opened 2025-04-19
fix llama4 cache

awni:
Closes https://github.com/ml-explore/mlx-lm/issues/103

@Blaizzy another good thing for you to use in mlx-vlm from mlx-lm are the cache routines. E.g. you'll get the bug fix here if you use them.

### DIFF
--- a/mlx_lm/cache_prompt.py
+++ b/mlx_lm/cache_prompt.py
@@ -148,7 +148,7 @@ def main():
         pass
 
     print()
-    print(f"Peak memory: {mx.metal.get_peak_memory() / 1e9:.3f} GB")
+    print(f"Peak memory: {mx.get_peak_memory() / 1e9:.3f} GB")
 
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
🙏

Blaizzy:
Thanks @awni!

Will do ASAP. Cache routines are 95% complete since v0.1.22.

Just need to migrate ChunkedCache today.

If I remember correctly I copied it last week because MLX-LM release with llama 4 wasn't out yet so it would break my release.

awni:
> If I remember correctly I copied it last week because MLX-LM release with llama 4 wasn't out yet so it would break my release.

Makes sense. You can always let us know if you need a urgent release for mlx-lm.. it's easy to do a patch release.

Blaizzy:
Thanks, that's great to know ❤️

I will surely do that

PR #110, opened 2025-04-20
Adding support for XTC (eXclude Top Choice) 

Dirky14:
XTC (eXclude Top Choice) is a widely used logit processor in the storytelling community in LLM, by removing the most likely tokens from consideration, thus allowing more creativity into the generations. This PR adds XTC support by adapting the text-generation-webui's implementation of the sampler creator p-e-w  to mlx_lm. 

Because it's implemented as a logit processor, it's for now only usable in mlx_lm.server, as mlx_lm.generate doesn't implement them (unless I missed something). I didn't really know if it was intended or not, so I didn't push the generate modifications I did in local.

Since it's my first real open-source contribution, it's certainly not perfect, so I'm open to any improvement suggestions !

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -60,6 +60,8 @@ def make_logits_processors(
     logit_bias: Optional[Dict[int, float]] = None,
     repetition_penalty: Optional[float] = None,
     repetition_context_size: Optional[int] = 20,
+    xtc_probability: Optional[float] = 0.0,
+    xtc_threshold: Optional[float] = 0.0,
 ):
     """
     Make logits processors for use with ``generate_step``.
@@ -70,6 +72,9 @@ def make_logits_processors(
         repetition_context_size (int, optional): The number of tokens to
           consider for repetition penalty. Default: ``20``.
         logit_bias (dictionary, optional): Additive logit bias.
+        xtc_probability (float, optional): The probability of applying XTC
+          sampling.
+        xtc_threshold (float, optional): The threshold for XTC sampling.
 
     Returns:
         List[Callable[[mx.array, mx.array], mx.array]]:
@@ -92,6 +97,8 @@ def make_logits_processors(
         logits_processors.append(
             make_repetition_penalty(repetition_penalty, repetition_context_size)
         )
+    if xtc_probability > 0.0:
+        logits_processors.append(make_xtc(xtc_probability, xtc_threshold))
     return logits_processors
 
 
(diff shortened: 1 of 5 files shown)


awni (changes requested):
This looks like a nice implementation and we can definitely add it. A couple things here:
- Please change it from a logits_processor to a sampler. The distinction is a bit subtle.. but I think it makes more sense as a sampler since it doesn't require the past history of tokens
- Please add some tests [here](https://github.com/ml-explore/mlx-lm/blob/main/tests/test_sample_utils.py)

Dirky14:
Just moved my code to sampler. I agree it make more sense like that.
I did tests on edge case (probability = 0 and 1, since otherwise it would be complicated to test, with 2 examples for testing the behavior of "discarding all but the last token above the threshold"). Tell me if I have to change anything, and thanks for your review !

mlx_lm/sample_utils.py
```diff
@@ -219,6 +232,55 @@ def apply_top_p(logits: mx.array, top_p: float) -> mx.array:
...
+    # removed by XTC sampling
+    special_tokens_mask = mx.zeros(indices_to_remove.shape[1], getattr(mx, "bool_"))
+    special_tokens_mask[special_tokens_ids] = True
+    indices_to_remove = mx.logical_and(
+        indices_to_remove, mx.logical_not(special_tokens_mask)
+    )
+
+    logits_edited = mx.where(indices_to_remove, -float("inf"), logits)
+
+    return mx.where(mx.random.uniform(0, 1) >= xtc_probability, logits, logits_edited)
```
angeloskath, line 281: I think a much simpler way to get the `indices_to_remove_mask` is 

```python
probs = mx.softmax(logits, axis=-1)
indices_to_remove = probs > (probs + probs < xtc_threshold).min()
```

so I think the whole implementation can be

```python
def apply_xtc(logits, xtc_probability, xtc_threshold, special_tokens_ids):
    # we can avoid the softmax if logits is already normalized
    mask = probs > (probs + probs < xtc_threshold).min()
    mask[special_tokens_ids] = False
    return mx.where(
        mx.random.uniform(0, 1) > xtc_probability,
        logits,
        mx.where(mask, -mx.inf, logits),
    )
```

this deviates from the original in the following way, when the probability closest to the threshold is the same for many tokens then the original would only keep 1 vs this will keep all that have that exact probability value. It should be both rare and useful I think.

Dirky14: Thanks for the review ! A few notes after checking the discussion of the implementation in llamacpp : 

- It seems like there is a difference between the original implementation and the llamacpp one. The llamacpp (and, in fact, the original proposal of the author, since it was changed later in the PR) ones doesn't remove the EOS/newline tokens from the mask. Apparently, some tests were done and nobody saw a degradation of the quality of the generations between the two implementations. However, it could impact the generation length, as EOS tokens would have a probability near 1, and would be slashed by XTC if we don't remove them from the mask, so I think it would be preferable to keep special token ids exclusion.

- I agree that your implementation is clearly simpler. However, I'm not sure if a sampler should act differently between inference systems (even if there is an exception here for the EOS tokens exclusion). If you think it's an acceptable difference, I will change the tests and the code accordingly for the sake of the simplicity.

angeloskath: I think it is worth it given the huge gap in readability and low probability of it affecting the experience, meaning it is unlikely for the large probability tokens to have exactly the same probability.

Dirky14: I applied your suggestion then, just edited a bit the code (I tested your implementation, and it switched every probs to -inf, and the mask[special_tokens_ids] = False was producing gibberish output on some of my prompt tests). Don't hesitate to tell me if you have any other suggestions/modifications to make. Thanks again for the review !

angeloskath: Hey thanks for doing the additions, no need for the added complexity I just missed some parentheses around the comparison.

```python
mask = probs > (probs + (probs < xtc_threshold)).min()
```

doing this with where (as you have it) is also fine

```python
```

For the rest I like the if check but no need for merging masks or anything just add the ellipsis
```python
if special_tokens_ids:
    mask[..., special_tokens_ids] = False
```

angeloskath (approved):

angeloskath:
@awni I think it requires your approval since you requested changes. So feel free to approve and merge whenever you want.

awni (approved):
LGTM!

PR #114, opened 2025-04-22
Try and scan only mlx compatible LLMs

awni:
For clients which use the models endpoint of mlx-lm.server it helps to try and only find LLMs to avoid cluttering the list with useless/incompatible repo ids.

### DIFF
--- a/mlx_lm/_version.py
+++ b/mlx_lm/_version.py
@@ -1,3 +1,3 @@
 # Copyright © 2023-2024 Apple Inc.
 
-__version__ = "0.23.0"
+__version__ = "0.23.1"
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -680,10 +680,18 @@ class APIHandler(BaseHTTPRequestHandler):
         self._set_completion_headers(200)
         self.end_headers()
 
+        files = ["config.json", "model.safetensors.index.json", "tokenizer_config.json"]
+
+        def probably_mlx_lm(repo):
+            if repo.repo_type != "model":
+                return False
+            file_names = {f.file_path.name for f in repo.refs["main"].files}
+            return all(f in file_names for f in files)
+
         # Scan the cache directory for downloaded mlx models
         hf_cache_info = scan_cache_dir()
         downloaded_models = [
-            repo for repo in hf_cache_info.repos if "mlx" in repo.repo_id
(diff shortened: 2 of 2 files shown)


angeloskath (approved):
👍

BenceSzalai:
I've found this from #130.

This change eliminated valid models from the response of the `/models` endpoint.

I've tested and some older models, which have a `weights.00.safetensors` file instead of `model.safetensors.index.json` , and they work perfectly fine, so they should?/could be listed, but they are no longer after this change.

I guess the check should be more like, to not render some models unavailable for some clients:

```py
        files = ["config.json", "tokenizer_config.json"]

        def probably_mlx_lm(repo):
            if repo.repo_type != "model":
                return False
            file_names = {f.file_path.name for f in repo.refs["main"].files}
            return all(f in file_names for f in files) and ("model.safetensors.index.json" in file_names or "weights.00.safetensors" in file_names)
```

Let me know if I should make a PR / please forgive my ignorance if I miss some reason for the other format models should be no longer listed. (In which case probably they should be marked on hugging-face as deprecated format or sg as it can be quite confusing, also makes selecting models a bit more tedious, as before downloading one should check the format of the files, whereas [mlx-community](https://huggingface.co/mlx-community) kind of suggest that the models there should work with MLX.)

EDIT: Sorry, just seen your comment under #130 just now, after posting this. So I guess listing a model that does not work is still better than not listing a model, that otherwise would work if were listed.

awni:
> So I guess listing a model that does not work is still better than not listing a model, that otherwise would work if were listed.

In this case it's only very old models (like more than a year old) that won't be listed which is why I'm not that worried about it. On the one hand it's a trivial change to make like what you posted above. On the other hand.. I kind of want to encourage people to use newer models (as in more recently converted) since our quantization has improved and we can standardize the format.

awni:
> whereas mlx-community kind of suggest that the models there should work with MLX.

Note: not every model in mlx-community works with mlx-lm and a model doesn't have to be in mlx-community to work with mlx-lm. There are all sorts of other models in the mlx-community like VLMs, speech and audio models, etc. That's why the most accurate solution is to check the config.json model type.

BenceSzalai:
How about adding a warning to console if a model found with "weights.00.safetensors" saying sg. like:

"The format of this model is outdated. Please use a newer model or make a fresh conversion from the original model to utilize the benefits of the latest format."

I can make a PR for it if you have no objections.

PR #115, opened 2025-04-23
Update AWQ docs to include default values

ivanfioravanti:
Default options included to clarify that 3bit quantization is deafult.


awni:
I think maybe it's worth updating the default to 4bit as it is more expected and still works a lot better than 3 bit.

ivanfioravanti:
I totally agree. 4bit would be better.

awni:
Ok I changed the defaults herre. Thanks for the PR @ivanfioravanti !

awni (approved):

PR #116, opened 2025-04-23
Add Baichuan M1 Model

cmcmaster1:
As the name says, added the Baichuan M1 architecture (see https://huggingface.co/baichuan-inc/Baichuan-M1-14B-Instruct).


cmcmaster1:
There are some issues I've found. I'll reopen when I've fixed them.

cmcmaster1:
Fixed. It wasn't working for multiturn. Works now.

awni:
Thanks for the addition. I pushed a couple changes to your branch.

I think the main thing we need to sort out is the handling of the "convolution" in the attention layer. There are two problems I see there:
1. Everything is loops  when it should be vectorized as much as possible
2. The state should not be held by the model but by the cache

The second will either require a new cache type or some way of storing an arbitrary payload on the base cache.

I'm happy to help with either or both of these. Let me know how you'd like to proceed.

awni:
Ok I worked on 1 and 2 a bit. Simplified the conv implementation and got rid of the loops. Added a cache list class so we can have the a conv cache and kv cache for the same block.

awni (approved):
Thanks for the addition! Will merge once the tests clear.

cmcmaster1:
Thanks! You did all of that while I was sleeping. I did refactor to a vectorized version of the convolution, but given W=2 it didn't really make any difference and it looked ugly. But yours is much cleaner than mine. Thanks for fixing my code and merging so quickly! I've tested it out and it's genuinely the best medical model for its size.

ccckblaze:
how about baichuan M2? @cmcmaster1

PR #117, opened 2025-04-23
During model conversion, default dtype to config.json torch_dtype

neilmehta24:
When converting models, if the `dtype` argument is unspecified, load the weights as specified in the config.json `torch_dtype`. If `torch_dtype` key is not found or the value is unrecognized, fallback to the model weights' dtype. The conversion CLI will still respect the user's `--dtype` argument if they wish to override this behavior


awni (approved):
Looks awesome, thanks!!

PR #120, opened 2025-04-24
Add top-k sampling switch to the generate CLI

neilmehta24:
Add a `--top-k` parameter to the `mlx_lm generate` CLI tool. Also, update the default `make_sampler` top-k parameter to `0`, to be consistent across the code base.


awni:
Nice contribution! Could you rebase this on main @neilmehta24 and then I can run the tests? (Just landed another sampler `xtc` which is causing the conflict).

PR #122, opened 2025-04-25
Fix evaluate.py oversight

N8python:
main() is never being called!


awni:
You should be able to run it like:
`mlx_lm.evaluate` or `python -m mlx_lm evaluate`

We deprecated running sub packages directly since it has issues making them importable.

N8python:
Ah! Sorry - quite the mistake on my part.

awni:
No indeed, we should document this command a bit better.

PR #127, opened 2025-04-26
Update baichuan_m1.py

cmcmaster1:
The sanitize method was failing with quantized models. I've simplified it to be consistent with implementations in other models. It works now.

### DIFF
--- a/mlx_lm/models/baichuan_m1.py
+++ b/mlx_lm/models/baichuan_m1.py
@@ -204,10 +204,17 @@ class Model(nn.Module):
             caches.append(CacheList(conv_cache, kv_cache))
         return caches
 
-    def sanitize(self, weights):
-        weights = dict(weights)
-        if "lm_head.weight" not in weights:
-            weights["lm_head.weight"] = weights["model.embed_tokens.weight"]
+    def sanitize(self, weights: dict) -> dict:
+        is_quantized = "lm_head.scales" in weights
+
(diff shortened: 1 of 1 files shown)


mlx_lm/models/baichuan_m1.py
```diff
@@ -204,14 +204,10 @@ def make_cache(self) -> List[Any]:
...
-            weights.pop("lm_head.weight", None)
-        else:
-            # Pre-normalize the lm_head
-            w = weights["lm_head.weight"]
-            w = w / (mx.linalg.norm(w, axis=-1, keepdims=True) + 1e-7)
-            weights["lm_head.weight"] = w
+    def sanitize(self, weights):
+        weights = dict(weights)
+        if "lm_head.weight" not in weights:
+            weights["lm_head.weight"] = weights["model.embed_tokens.weight"]
```
awni, line 210: Good catch. Though I think we should keep the pre-normalization for non-quantized LM heads otherwise that will be broken.

So concretely.. check if `lm_head.scales` is in `weights`. If it is then it's quantized already and we don't need normalizatoin. If it's not then do the normalization.

cmcmaster1: Done.

awni (approved):

awni (approved):
Thanks for the fixes!!

PR #128, opened 2025-04-26
Update AWQ evaluate and upload

ivanfioravanti:
Added seed to evaluate command and fixed upload parameters.

### DIFF
--- a/mlx_lm/AWQ.md
+++ b/mlx_lm/AWQ.md
@@ -43,7 +43,7 @@ downstream tasks using `mlx_lm.evaluate`. For example:
 ```bash
 mlx_lm.evaluate \
     --model mlx_model \
-    --tasks winogrande boolq arc_challenge arc_easy hellaswag openbookqa piqa social_iqa --seed 42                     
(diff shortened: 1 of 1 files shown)


mlx_lm/AWQ.md
```diff
@@ -43,15 +43,15 @@ downstream tasks using `mlx_lm.evaluate`. For example:
 ```bash
 mlx_lm.evaluate \
     --model mlx_model \
-    --tasks winogrande boolq arc_challenge arc_easy hellaswag openbookqa piqa social_iqa                      
+    --tasks winogrande boolq arc_challenge arc_easy hellaswag openbookqa piqa social_iqa --seed 42
```
awni, line 46: Just curious, what is the seed there for? We already have it at a default and for most people I don't think they would need to set it.

ivanfioravanti: You are totally right! I missed the 123 default value!

awni (approved):
Thanks for the fix @ivanfioravanti  !! 🙏

PR #129, opened 2025-04-27
Distilled weight quantization

awni:
Relies on https://github.com/ml-explore/mlx/pull/2129

```
mlx_lm.dwq --model hf/repo
```

This learns the quantization `scales` and `biases` parameters using gradient descent and a KL loss on the unquantized model outputs. Hence "distilled" weight quantization (DWQ).

One perk of this over AWQ is it works out of the box with any model in mlx-lm and any quantized layer (embedding, swtich, linear, etc). So much more general and the code is much simpler.

Some results on Llama 3.2 1B (not all done yet)

Model | Average accuracy
------ | -------
fp16 | 47.95 
Q4 | 46.21
Q4 DWQ, 256 samples |  47.27
Q4 DWQ, 512 samples |  47.57
Q3 | 39.87
Q3 DWQ, 256 samples | 44.65
Q3 DWQ, 512 samples | 44.99
Q3 DWQ, 1024 samples | 45.14

Eval done with:

```
mlx_lm.evaluate --model mlx_model  --tasks winogrande boolq arc_challenge arc_easy hellaswag openbookqa piqa social_iqa
```

### DIFF
--- a/mlx_lm/dwq.py
+++ b/mlx_lm/dwq.py
@@ -4,15 +4,19 @@ import argparse
 import copy
 import glob
 import shutil
+import time
+import types
 from pathlib import Path
 
 import mlx.core as mx
 import mlx.nn as nn
 import mlx.optimizers as optimizers
-from datasets import load_dataset
+import numpy as np
 from mlx.utils import tree_flatten, tree_map
 
 from mlx_lm.tokenizer_utils import TokenizerWrapper
+from mlx_lm.tuner.datasets import load_dataset
+from mlx_lm.tuner.trainer import iterate_batches
 from mlx_lm.tuner.utils import print_trainable_parameters
 from mlx_lm.utils import (
     create_model_card,
@@ -24,23 +28,15 @@ from mlx_lm.utils import (
 )
 
 
-def dist_split(x: mx.array, group: mx.distributed.Group):
-    B = x.shape[0]
-    N = group.size()
-    assert B % N == 0
-    r = group.rank()
-    local_B = (B + N - 1) // N
-    return x[r * local_B : (r + 1) * local_B]
-
-
 def dwq_quantize(
     model,
     q_model,
     opt,
-    inputs: mx.array,
+    data,
     group_size: int = 64,
     bits: int = 3,
     batch_size: int = 2,
+    max_seq_length: int = 2048,
     dtype: mx.Dtype = mx.bfloat16,
 ):
     group = mx.distributed.init()
@@ -51,16 +47,22 @@ def dwq_quantize(
     def log_norm(x):
         return x - mx.logsumexp(x, axis=-1, keepdims=True)
 
-    def loss_fn(params, x, targets):
+    def loss_fn(params, x, targets, lengths):
         q_model.update(tree_map(lambda x: x.astype(dtype), params))
         logits = q_model(x).astype(mx.float32)
-        return nn.losses.kl_div_loss(log_norm(logits), targets, reduction="mean")
+        losses = nn.losses.kl_div_loss(log_norm(logits), targets, reduction="none")
+        mask = mx.arange(1, targets.shape[1] + 1) <= lengths[:, 1:]
+        ntoks = mask.sum()
+        loss = (mask * losses).sum() / ntoks
+        return loss, ntoks
 
     # TODO distributed support
-    def step(inputs, targets, params):
-        loss, grads = mx.value_and_grad(loss_fn)(params, inputs, targets)
+    def step(inputs, targets, lengths, params):
+        (loss, ntoks), grads = mx.value_and_grad(loss_fn)(
+            params, inputs, targets, lengths
+        )
         params = opt.apply_gradients(grads, params)
-        return loss, params
+        return loss, ntoks, params
 
     # Accumulate learned weights in higher precision
     params = tree_map(
@@ -68,36 +70,22 @@ def dwq_quantize(
         q_model.trainable_parameters(),
     )
 
-    losses = []
-    it = 1
-    for i in range(0, inputs.shape[0], batch_size):
-        batch = inputs[i : i + batch_size]
+    avg_loss = None
+    tokens = 0
+    tic = time.time()
+    for it, (batch, lengths) in enumerate(
+        iterate_batches(data, batch_size, max_seq_length)
+    ):
         targets = log_norm(model(batch).astype(mx.float32))
         mx.eval(targets)
-        loss, params = step(batch, targets, params)
+        loss, ntoks, params = step(batch, targets, lengths, params)
         mx.eval(loss, params)
-        losses.append(loss.item())
-        print(f"Iter: {it}, Loss: {loss.item():.3f}")
-        if (it + 1) % 10 == 0:
-            loss_avg = sum(losses) / len(losses)
-            print(f"Average Loss {loss_avg:.3f}")
-            losses = []
-        it += 1
+        loss = loss.item()
+        tokens += ntoks.item()
+        toks_per_sec = tokens / (time.time() - tic)
+        avg_loss = 0.9 * (avg_loss or loss) + 0.1 * loss
+        print(f"{it=}, {loss=:.3f}, {avg_loss=:.4f}, {tokens=}, {toks_per_sec=:.3f}")
     q_model.update(tree_map(lambda x: x.astype(dtype), params))
-
-
-def load_wikitext(
-    tokenizer, num_samples: int = 32, sequence_length: int = 2048, split: str = "train"
-) -> mx.array:
-    dataset = load_dataset("Salesforce/wikitext", "wikitext-2-raw-v1", split=split)
-    texts = "\n\n".join(dataset["text"])
-    tokens = tokenizer.encode(texts, return_tensors="mlx")[0]
-
-    # Select random chunks
-    starts = mx.random.randint(
-        0, len(tokens) - sequence_length - 1, shape=(num_samples, 1)
-    )
-    return tokens[starts + mx.arange(sequence_length)]
 
 
 def save_model(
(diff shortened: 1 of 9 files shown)


awni:
Qwen 1.7B results are actually quite nice. The DWQ recovers the full bf16 perf on the eval. In fact.. the DWQ scores slightly higher than BF16.. which must be noise because we are using the bf16 to distill to.

Precision | Accuracy
---- | ----
BF16 | 45.90
Q4 | 44.94
Q4 AWQ | 45.28
Q4 DWQ | 46.46

awni:
Some updates:
- Added support for distributed
- Added a temperature parameter to sharpen the scores prior to the loss. I didn't fully x-validate that yet but 2.0 seems to work better than not setting it
- Use Adam instead of SGD
- Train all other floating point params as well (basically the layer norm params), works better.. and why not.
- Changed switch layers to use a class activation instead of the function since the nb::function isn't copyable.

awni:
Also updated the documentation to include dwq. I think this is ready to review @angeloskath and land. Though I'm sure there is more to refine.

ayushnangia:
@awni what are your student and teacher model?

mlx_lm/models/switch_layers.py
```diff
@@ -184,7 +184,7 @@ def __init__(
         input_dims: int,
         hidden_dims: int,
         num_experts: int,
-        activation=nn.gelu_approx,
+        activation=nn.GELU(approx="precise"),
```
angeloskath, line 187: For better printing?

mlx_lm/dwq.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+        grads = nn.average_gradients(grads)
+        params = opt.apply_gradients(grads, params)
+        return loss, ntoks, params
+
+    # Accumulate learned weights in higher precision
+    params = tree_map(
+        lambda x: x.astype(mx.float32),
+        q_model.trainable_parameters(),
+    )
```
angeloskath, line 79: Interesting, did you actually notice a difference? A subtle point here is that this results in the optimizer states being fp32 as well which is what you wanted I am sure but it is hidden behind the subtlety that we use the params to initialize the states.

mlx_lm/dwq.py
```diff
@@ -0,0 +1,209 @@
...
+        type=float,
+        default=0.5,
+        help="Temperature scaling for the loss.",
+    )
+    args = parser.parse_args()
+
+    group = mx.distributed.init()
+
+    num_samples = args.num_samples
+    if group is not None and num_samples % group.size() > 0:
```
angeloskath, line 177: Group will never be none, it may be singleton but it will always be a valid object.

mlx_lm/dwq.py
```diff
@@ -0,0 +1,209 @@
...
+    tokenizer.save_pretrained(mlx_path)
+
+    save_config(config, config_path=mlx_path / "config.json")
+    create_model_card(mlx_path, hf_path)
+
+
+def load_data(tokenizer, num_samples: int):
+    args = types.SimpleNamespace(
+        hf_dataset={
+            "path": "allenai/tulu-3-sft-mixture",
```
angeloskath, line 130: Probably should be a path we pass into the script ...

angeloskath (approved):
This is perfect! Left random comments mostly due to curiosity.

mlx_lm/dwq.py
awni: > A subtle point here is that this results in the optimizer states being fp32 as well which is what you wanted

Yea.. I think it's pretty important for those to be in fp32 (unfortunately because it's a bit slower).

> did you actually notice a difference? 

Yes. I just re-ran it with the latest with and without fp32 for the trainable parameters. FWIW I've almost always had to do this to get good bf16 training (MLX or otherwise). In our LoRA example the lora params are fp32 so we don't have this problem.

Here's a figure comparing the difference.

Screenshot 2025-05-03 at 8 52 55 AM

awni:
> @awni what are your student and teacher model?

@ayushnangia the student is the quantized model, the teacher is the non-quantized model (e.g. fp16 of bf16)

Anemll:
```bash
mlx_lm.evaluate --model mlx_model  --tasks winogrande boolq arc_challenge arc_easy hellaswag openbookqa piqa social_iqa
```
is missing  --no-apply-chat-template,  at least for LLAMA it will produce incorrect / lower than expected results.
I'm not sure if it effects comparison for quantization, but definitely creates an issue when trying to compare different backends.



BoolQ:
--no-apply-chat-template 
{
    "alias": "boolq",
    "acc,none": 0.7097859327217125,
    "acc_stderr,none": 0.007938079855173703
}

without:
{
    "alias": "boolq",
    "acc,none": 0.5532110091743119,
    "acc_stderr,none": 0.008695392261996192
}

PS. perhaps mlx_lm.evaluate should have some default settings for different datasets?

awni:
Can you say more about that @Anemll ?

I'm not an expert on evals.. but my current understanding is that for chat / instruct tuned models, one should use the chat template? And for base models, one shouldn't. (This is how `mlx_lm.evaluate` functions now).

dessatel:
I’m not an evaluation expert either, so please double‑check this with others.
My recommendation is based on empirical results and prior reports.

Why --no‑apply-chat-template
These multiple‑choice benchmarks (BoolQ, ARC, PIQA, Social IQA, WinoGrande, HellaSwag) were designed for plain language‑model scoring, not for dialogue agents. Nearly all expect the model to output a single token—True, False, or a letter like A.

When you wrap the prompt in a chat template, extra role tags and new‑lines appear before that first graded token. The harness then scores the wrong token, which drags accuracy down. 

Evidence
	•	GitHub discussion of template‑related drops: https://github.com/EleutherAI/lm-evaluation-harness/issues/1841
	•	Paper on first‑token mismatch for chat‑tuned models: https://arxiv.org/abs/2402.14499
	•      My BoolQ test

IMO, for single‑token MCQs, either (1) remove the template or (2) switch to grading the full text output instead of the first token.

To be safe, It might be worthwhile to run each benchmark  with and without the system prompt on an original (baseline) model—just to see how much the template alone moves the needle. 
Some of the “gains” you observed in quantized Social Benchmarks  might be attributable to template effects .

Anemll:
I’m not an evaluation expert either, so please double‑check this with others.
My recommendation is based on empirical results and prior reports.

Why --no‑apply-chat-template
These multiple‑choice benchmarks (BoolQ, ARC, PIQA, Social IQA, WinoGrande, HellaSwag) were designed for plain language‑model scoring, not for dialogue agents. Nearly all expect the model to output a single token—True, False, or a letter like A.

When you wrap the prompt in a chat template, extra role tags and new‑lines appear before that first graded token. The harness then scores the wrong token, which drags accuracy down. 

Evidence
	•	GitHub discussion of template‑related drops: https://github.com/EleutherAI/lm-evaluation-harness/issues/1841
	•	Paper on first‑token mismatch for chat‑tuned models: https://arxiv.org/abs/2402.14499
	•      My BoolQ test

IMO, for single‑token MCQs, either (1) remove the template or (2) switch to grading the full text output instead of the first token.

To be safe, It might be worthwhile to run each benchmark  with and without the system prompt on an original (baseline) model—just to see how much the template alone moves the needle. 
Some of the “gains” you observed in quantized Social Benchmarks  might be attributable to template effects .

Anemll:
Just to add,  you might have seen "the instruct" model template on the Open‑LLM Leaderboard docs:

“For instruction models add --apply_chat_template and --fewshot_as_multiturn.”
https://huggingface.co/docs/leaderboards/open_llm_leaderboard/about#reproducibility

The leaderboard uses its own fork of lm‑eval‑harness. For multi‑choices like MMLU‑PRO, BBH, and GPQA, that fork extracts the answer letter from the full text output, so leaving the chat template on does not hurt accuracy as much.

awni:
Indeed I do get much better results with the chat template disabled for Qwen3 4B on MMLU. The instruct model is a bit worse than the base model with the chat template disabled. 76 vs 71 (that's on a 2200 example subset). The tech report only reported base model perf on MMLU, and it seems like those numbers ball park make sense.

Anemll:
social_iqa is interesting.
Is DWQ using templates during tuning?

Comparison of Accuracy (acc,none): No Template vs. Template Applied

Llama 3.2 1B  ( instruction tuned)
| Task | No Template (%) | Template Applied (%) | Change (%) |
| :------------ | :-------------- | :------------------- | :--------- |
| social_iqa | 43.1% | 43.9% | +0.8% |
| piqa | 74.6% | 73.7% | -0.9% |
| hellaswag | 45.6% | 43.7% | -1.9% |
| openbookqa | 26.6% | 24.0% | -2.6% |
| arc_challenge | 34.6% | 31.7% | -2.9% |
| winogrande | 61.9% | 57.1% | -4.8% |
| arc_easy | 69.0% | 57.6% | -11.4% |
| boolq | 71.0% | 56.1% | -14.9% |

awni:
> Is DWQ using templates during tuning?

By default yes. You can specify the dataset and depending on the format it may or may not use the template.

I've only DWQ-trained instruct models so far. I did try a non-chat-template calibration dataset (the same one we use for AWQ), it didn't seem to make much difference for the evals I looked at at the time (MMLU with chat-template).

awni:
Thanks for the results btw. It seems like we need to be more careful when using the chat template. For most of the loglikelihood tasks (e.g. where you are scoring responses to a prompt) it may make sense to default it to be off.

Anemll:
I agree, adding a custom regex script may be necessary, especially with ‘think’ mode triggered by prompts across multiple models. 
I plan to try Hugging Face forks to see if they perform better: https://github.com/huggingface/lm-evaluation-harness/

Anemll:
Meant to say ‘think’ mode, but markup butchered my post.

PR #131, opened 2025-04-28
Add ability to define default parameters for server in a settings file

BenceSzalai:
When using `mlx_lm.server` to play with models locally, it can become a huge time waster to research and figure out how the parameters can be set for each tool that connects to the API. Some tools have nice UI to add arbitrary body parameters, some tools only have some common parameters on the UI, and need setting these in config files, and some tools does not allow adding custom body parameters at all, in which case one has to write and use a utility proxy to sit between the client and the server to modify the request. This is a huge time waster and also makes it hard to maintain all these configurations over various tools.

Here the key idea is that while some parameters may worth changing per client or even per request, some should be configurable as defaults for all requests. One such thing for me is proper `draft_model` for each of my installed models, which can easily 2x the generation speed. This should be a one time set and forget thing, regardless of what client will consume the server after.

For this reason, I've implemented a config file to sit in the user's home folder where the parameters can be preset for all requests.

In the config file I put everything under the main key of `server` thinking that maybe in the future other modules could utilize the same file as well. Also, under server I put everything under `parameters`, so other settings can be added later for the `server` module too.

One thing I am quite indecisive about is how the file should be read. First I made it only read the file once, and reuse the same value until server restart, but it makes fiddling with the file really tedious, to always restart the server, and since the settings are only loaded once per request I think it is better to load it every time. But this also means if other modules start to use the same `get_settings()` call, they must make sure the result is handled properly to avoid repeated reading of the settings file e.g. inside a loop or so.

Let me know what do you think!

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -4,11 +4,12 @@ import http
 import json
 import threading
 import unittest
+from unittest.mock import patch
 
 import requests
 
 from mlx_lm.server import APIHandler
-from mlx_lm.utils import load
+from mlx_lm.utils import get_settings, load
 
 
 class DummyModelProvider:
@@ -54,6 +55,9 @@ class TestServer(unittest.TestCase):
         cls.server_thread = threading.Thread(target=cls.httpd.serve_forever)
         cls.server_thread.daemon = True
         cls.server_thread.start()
+        # Prevent user settings from being loaded and interfere with test runs
+        cls.get_settings_patcher = patch("mlx_lm.server.get_settings", return_value={})
+        cls.mock_get_settings = cls.get_settings_patcher.start()
 
     @classmethod
     def tearDownClass(cls):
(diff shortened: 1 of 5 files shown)


BenceSzalai:
I've just realised I was sloppy and did not use a feature branch. Let me know if I should resubmit it, since I'm not sure this way it can be squashed for merge. Not like I'd assume it'll be necessarily merged, I understand that adding a global settings file is a large change, and you may propose other way, like providing a settings file using a CLI param or idk.

tests/test_server.py
```diff
@@ -54,6 +55,9 @@ def setUpClass(cls):
         cls.server_thread = threading.Thread(target=cls.httpd.serve_forever)
         cls.server_thread.daemon = True
         cls.server_thread.start()
+        # Prevent user settings from being loaded and interfere with test runs
+        cls.get_settings_patcher = patch("mlx_lm.server.get_settings", return_value={})
+        cls.mock_get_settings = cls.get_settings_patcher.start()
```
BenceSzalai, line 60: I'm not sure if this is good approach. It defo highlights an issue introduced by a global settings. Maybe a more proper mocking should be used instead, but honestly, I could not get it working in a more global way. (Python is not my native language.) My concern with this is that this patterns should become repeated in all tests for any modules which use the settings file.

BenceSzalai:
It is a mess, that I started this from main, will close and fix my repo.

Issue #109, opened 2025-04-20
feat: Support custom local model directory in `mlx_lm.server`


ahmetaltunel:
**Problem:**
Currently, the `mlx_lm.server` only discovers models located within the standard Hugging Face cache directory via `huggingface_hub.scan_cache_dir()`. This means:
1.  The `GET /v1/models` endpoint does not list models that users might have downloaded, fine-tuned, or stored in other local directories (e.g., `~/.models/`, shared network drives).
2.  Even if a user knows the name/ID of a local model, when making a request (e.g., `POST /v1/chat/completions`), the server's `ModelProvider` might still attempt to resolve and download this ID from Hugging Face Hub if it's not found in the standard cache, rather than loading it directly from the known local path.

**Use Case:**
Users often maintain their own model inventories outside the default HF cache, including:
*   Models downloaded manually or via other tools.
*   Fine-tuned model versions.
*   Quantized models stored in specific project folders.
*   Models stored on shared infrastructure for team access.

It would be beneficial for the server to be aware of and prioritize these local model locations.

**Proposed Solution:**
Introduce a mechanism, potentially a CLI flag like `--model-dir <path>`, that allows users to specify one or more custom directories for the server to scan.

When this flag is used:
1.  **Discovery:** The `GET /v1/models` endpoint should merge the results from `scan_cache_dir()` with models found within the specified `--model-dir`(s). It should recursively scan the provided path(s) for valid model directories (containing `config.json`, weights, etc.) and include them in the list, perhaps using their relative path from `--model-dir` as their ID.
2.  **Loading:** The `ModelProvider.load()` logic should be updated. When resolving a model ID provided in an API request, it should first check if a corresponding subdirectory exists within the specified `--model-dir`(s). If found, it should load the model directly from that local path, bypassing any attempts to download from Hugging Face Hub for that ID. If not found locally, it can fall back to the current behavior (checking HF cache / attempting download).

**Benefits:**
*   **Visibility:** Provides a complete inventory of available models (cache + local) via the API.
*   **Offline/Private Use:** Enables seamless use of locally stored models without requiring internet access.
*   **Simplified Workflow:** Users can refer to local models by their simple directory name/ID instead of needing full paths in API requests.
*   **Performance:** Avoids unnecessary network lookups for models already present locally.

**Additional Considerations:**
*   Handling potential ID conflicts between the cache and local directories.

This feature would significantly improve the usability and flexibility of `mlx_lm.server` for users managing local model collections.

[...]

PR #134, opened 2025-04-29
Support custom local model directory in mlx_lm.server

rohinish404:
This pr introduces a **--model-dir** cli flag as mentioned in #109. This allows users to specify one or more local directories that the server should scan for MLX-compatible models.

1. **Model Discovery**: The GET /v1/models endpoint now scans the directories provided via --model-dir (recursively) in addition to the Hugging Face cache. Models found locally are included in the returned list, using their relative path from the specified --model-dir as their id.
2. **Model Loading**: The ModelProvider logic is updated to prioritize resolving model identifiers requested in API calls (like POST /v1/chat/completions). It first checks if the identifier corresponds to a valid model subdirectory within any of the specified --model-dir paths. If found, it loads the model directly from that local path. Only if the model is not found locally does it fall back to the previous behavior of treating the identifier as a Hugging Face Hub ID (checking the cache and attempting download if necessary).

## How to Use:

Example commands that can be run - 

```
mlx_lm.server --model mlx-community/gemma-3-12b-it-qat-6bit --model-dir ~/models
```
This will try to find the model with the id **gemma-3-12b-it-qat-6bit** in **~/models/mlx-community**

```
mlx_lm.server --model quantized/Qwen2.5-7B-Instruct-1M-4bit--model-dir ~/models
```
This will try to find the model with the id **Qwen2.5-7B-Instruct-1M-4bit** in **~/models/quantized**

Similarly, multiple directories can be added in after the **--model-dir** flag.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -142,6 +142,15 @@ def process_message_content(messages):
                 raise ValueError("Only 'text' content type is supported.")
             message["content"] = "".join(text_fragments)
 
+def is_valid_model_dir(dir_path: Path) -> bool:
+            """Checks if a directory contains minimal files indicating a potential model."""
+            has_config = (dir_path / "config.json").is_file()
+            has_tokenizer_config = (dir_path / "tokenizer_config.json").is_file()
+            has_weights = any(dir_path.glob("*.safetensors")) or \
+                          (dir_path / "weights.npz").is_file() or \
+                          (dir_path / "model.safetensors.index.json").is_file()
+            return has_config and has_tokenizer_config and has_weights
+
 
 @dataclass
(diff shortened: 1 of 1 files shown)


arty-hlr:
This shouldn't be necessary, why do `mlx_lm.chat` and `mlx_lm.generate` work with local models, but `mlx_lm.server` doesnt? See https://github.com/ml-explore/mlx-lm/issues/109#issuecomment-2869641496

awni:
Thanks for the contribution. From looking at the diff.. the complexity is a bit too high for the feature, so I'm going to close this. If more people are looking for something like this or if you can find a way to support it with a smaller in a simpler way, we can revisit.

PR #138, opened 2025-04-30
Optimize prompt cache handling and add tests (Fixes #137)

### LINKED ISSUE #137: Server: Smarter prompt cache invalidation
Currently the server will reset the prompt cache if there is any difference between the cached tokens and the prompt tokens.

On my M1 MacBook Air with a system prompt of 1500 tokens this means that responses take > 10 seconds compared to < 1 second if the cache is reset.

This is an issue for two reasons:

1. The client may elect not to include a recent message in the next prompt.
2. The token IDs in the generated response can be different to the token IDs from the prompt due to tokenizer discrepancies.

I have encountered both of these scenarios.

The issue is here:

https://github.com/ml-explore/mlx-lm/blob/c48de5082c44c4fa73383420421682f42b77113f/mlx_lm/server.py#L495-L505

Any difference in the tokens will cause the entire cache to be invalidated. 

It would be better to find the common prefix and trim to that point.

jolonf:
This fixes the overly eager prompt cache resetting where any difference in tokens causes the entire prompt cache to be re-made, which causes long delays. 

The fix finds the common prefix between the cache and the prompt and trims the cache if the common prefix is shorter than the existing cache.

Scenarios that benefit from this is when the client decides not to include a recent message in the prompt or if there are tokenizer discrepancies between the generator and the prompt. These issues commonly only affect recent (later) messages in the prompt and can more efficiently be trimmed rather than creating a new prompt cache.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -30,7 +30,7 @@ from ._version import __version__
 from .generate import stream_generate
 from .models.cache import can_trim_prompt_cache, make_prompt_cache, trim_prompt_cache
 from .sample_utils import make_logits_processors, make_sampler
-from .utils import load
+from .utils import common_prefix_len, load
 
 
 def get_system_fingerprint():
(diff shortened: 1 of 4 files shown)


BenceSzalai:
Hi! This sounds like an awesome improvement.

May I wonder what kind of "tokenizer discrepancies" would cause cache invalidation? Like do you mean that the tokenizer would re-encode the prompt into a different vector than the output vector it was de-tokenized from before? I can imagine it could happen for example if there are tokens for "A" and "B", used in the output during generation, but also there is a token for "AB", so when the prompt is re-tokenized it could use the token for "AB" end up being a different vector. Or what does it look like?

jolonf:
One specific case I have, I am getting the LLM to invoke functions, in this particular case it is `mlx-community/Llama-3.2-Instruct-3B-4-bit`, the LLM generates the response:

`ACTION: LIST_DIR(path='.')`

When I send this back to the server the last 3 token IDs are different:

Cached Prompt: `[1151, 55128, 128009]` 
Prompt: `[52886, 873, 128009]` 

Even though they decode to the same string: `='.')<|eot_id|`

| Cached tkid | Prompt tkid | Cached str | Prompt str
--------------|-------------|------------|----------
1151                |  52886         |  `='`               | `='.`
55128            | 873               | `.')`                | `')`
128009         | 128009        | `<\|eot_id\|>`  | `<\|eot_id\|>`

altaic:
Interesting. In that case, perhaps it'd also be worthwhile to decode mismatched tokens, compare their strings, and skip invalidation if the strings are equal.

mlx_lm/utils.py
```diff
@@ -504,3 +504,29 @@ def save_config(
     # write the updated config to the config_path (if provided)
     with open(config_path, "w") as fid:
         json.dump(config, fid, indent=4)
+
+
+def common_prefix_len(list1, list2):
```
awni, line 509: Could you replace the calls to https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/evaluate.py#L29-L35 with calls to this function instead?

mlx_lm/server.py
```diff
@@ -493,42 +493,96 @@ def generate_response(
...
         ):
+            logging.debug(
+                f"*** Resetting cache (model change or no common prefix). com_prefix_len: {com_prefix_len} ***"
+            )
             self.prompt_cache.model_key = self.model_provider.model_key
             self.prompt_cache.cache = make_prompt_cache(self.model_provider.model)
             if self.model_provider.draft_model is not None:
                 self.prompt_cache.cache += make_prompt_cache(
                     self.model_provider.draft_model
                 )
```
awni, line 533: I think it's time to refactor this into a little helper to make it easier to change stuff in the future. Maybe make a function like `reset_prompt_cache`. It could even an inner function inside this one.

awni:
This is a really nice contribution! Could you address the comments? Then I'll run the tests and we can merge it.

awni:
> When I send this back to the server the last 3 token IDs are different:

That's an interesting case. I would probably use the tokens returned by the server rather than re-encode the text. In general there is no guarantee that `generated_tokens == encode(decode(generated_tokens))`

jolonf:
> > When I send this back to the server the last 3 token IDs are different:
> 
> That's an interesting case. I would probably use the tokens returned by the server rather than re-encode the text. In general there is no guarantee that `generated_tokens == encode(decode(generated_tokens))`

But the client doesn't have access to the tokens, only the decoded string? And the client can also willing change the message history so the server can't rely on its own cache as the reference.

I've made the changes you have suggested, will push them shortly.

awni:
> But the client doesn't have access to the tokens, only the decoded string? 

In some cases they should be available https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/server.py#L655

> I've made the changes you have suggested, will push them shortly.

Thanks!

awni (approved):
Looks great! Will merge when the tests clear.

BenceSzalai:
> Interesting. In that case, perhaps it'd also be worthwhile to decode mismatched tokens, compare their strings, and skip invalidation if the strings are equal.

So would this be another feature to add on top? Is there any reason not to try to do this?

It could go something like determine the common prefix length, if it is shorter than the whole input, start to decode the tokens starting with the first mismatch, and if they match after decoding extend the common prefix to cover those too. After that we could replace them and just restart with calculate common prefix again with the corrected input sequence as a naive algo. Rinse and repeat until a real divergence is found or the whole input is validated. Indeed that's suboptimal, because it recompares the same tokens over and over again for every divergence.

A more optimal one could be to count the number of reusable tokens in the cache, and when the cache and input diverges, but they decode to the same string, increment the count with the appropriate number, and when they start to match again call the calculate common prefix length on the remaining sequence again, this way all tokens are only counted at max once.

I'm afraid thought this would mean that technically the cache and the input is not a match in terms of tokens. So a possible caveat could be that this means that we are feeding the model different token sequence for a given input compared to what it was during training, which can raise concern provided that the input is otherwise tokenized in a deterministic way, which I think it is. Maybe this is a showstopper for this approach? Because without this addition, the cache would kind of self-correct, and would only be truncated the first time, but on all subsequent requests it'd now be a proper match, because after the first response from the model, it'll be re-tokenized anyway. So maybe it is better to pay the price of cache truncation, but make sure the inputs are processed as they were intended for the model. Escpecially that this would only ever truncate last response amount of tokens from the cache, so the largest portion, covering the whole conversation up to that point is retained anyway.

I'm curious of the opinions, 'cause I really only have very little experience with LLMs, so please forgive my ignorance if i'm talking nonsense.

jolonf:
You're welcome to try implementing it 😀

I could imagine some complexities around the token boundaries. The characters might match but not at the start of token, so the algorithm could be finicky and would need to come up with a lot test cases.

It also depends how much of an issue this is in practice. In my use case I'm dealing with longer histories, larger system prompt, and relatively shorter messages. So if the last message is invalid, that is a proportionately smaller amount of the enttire prompt, so discarding the entire recent message from the cache isn't going to have a large performance impact.

However if the use case was only a few but very large messages, then your optimization might be worth looking at.

BenceSzalai:
> So if the last message is invalid, that is a proportionately smaller amount of the entire prompt, so discarding the entire recent message from the cache isn't going to have a large performance impact.

Yes, I've just figured that myself just now, but I tend to edit my own comments rather than spamming the topic with my flow of thoughts as additional comments, so i've been added it there while you responded.

Truly it'd not be a huge benefit and could degrade model performance because of not feeding the model with the input tokenised the same way as it got it during training. So it's a pass from my side. :)

Thanks for thinking with me!

PR #142, opened 2025-05-01
:sparkles: feat: Add support for Xiaomi MiMo model

johnmai-dev:
Huggingface: https://huggingface.co/XiaomiMiMo
Github: https://github.com/XiaomiMiMo/MiMo @XiaomiMiMo

### DIFF
--- a/mlx_lm/models/mimo.py
+++ b/mlx_lm/models/mimo.py
@@ -1,0 +1,258 @@
+from dataclasses import dataclass
+from typing import Any, Dict, Optional, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .rope_utils import initialize_rope
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str
+    hidden_size: int
+    num_hidden_layers: int
+    intermediate_size: int
+    num_attention_heads: int
+    rms_norm_eps: float
+    vocab_size: int
+    num_key_value_heads: int
+    max_position_embeddings: int = 32768
+    rope_theta: float = 10000.0
+    rope_traditional: bool = False
+    rope_scaling: Optional[Dict[str, Union[float, str]]] = None
+    tie_word_embeddings: bool = False
+    num_nextn_predict_layers: int = 2
+
+
+class Attention(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+
+        dim = args.hidden_size
+        self.n_heads = n_heads = args.num_attention_heads
+        assert args.num_key_value_heads is not None
+        self.n_kv_heads = n_kv_heads = args.num_key_value_heads
+
+        head_dim = args.hidden_size // n_heads
+        self.scale = head_dim ** -0.5
+
+        self.q_proj = nn.Linear(dim, n_heads * head_dim, bias=True)
+        self.k_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=True)
+        self.v_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=True)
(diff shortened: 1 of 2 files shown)


johnmai-dev:
PyCharm 2025-05-02 02 21 27

mlx_lm/models/mimo.py
```diff
@@ -0,0 +1,263 @@
...
+                mtp_logits = [
+                    self.model.embed_tokens.as_linear(mtp_out)
+                    for mtp_out in mtp_outputs
+                ]
+            else:
+                out = self.lm_head(out)
+                mtp_logits = [self.lm_head(mtp_out) for mtp_out in mtp_outputs]
+
+            return out, mtp_logits
+        else:
```
awni, line 244: I don't think we should support the `use_mtp` for now. It doesn't fit so well with the rest of mlx-lm. I would jsut delete that branch / remove the option.

mlx_lm/models/mimo.py
```diff
@@ -0,0 +1,263 @@
...
+            h = layer(h, mask, c)
+
+        h = self.norm(h)
+
+        if use_mtp and input_embeds is not None:
+            mtp_outputs = []
+            for mtp_layer in self.mtp_layers:
+                mtp_out = mtp_layer(input_embeds, h, mask)
+                mtp_outputs.append(mtp_out)
+            return h, mtp_outputs
```
awni, line 208: Remove `use_mtp`

mlx_lm/models/mimo.py
```diff
@@ -0,0 +1,263 @@
...
+        assert self.vocab_size > 0
+        self.embed_tokens = nn.Embedding(args.vocab_size, args.hidden_size)
+        self.layers = [
+            TransformerBlock(args=args) for _ in range(args.num_hidden_layers)
+        ]
+        self.norm = nn.RMSNorm(args.hidden_size, eps=args.rms_norm_eps)
+
+        self.mtp_layers = [
+            MiMoMTPLayers(args=args) for _ in range(args.num_nextn_predict_layers)
+        ]
```
awni, line 180: We could sanitize those in the `sanitize` function.

mlx_lm/models/mimo.py
```diff
@@ -0,0 +1,263 @@
...
+        input_embeds: mx.array,
+        hidden_states: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Any] = None,
+    ) -> mx.array:
+        input_embeds = self.token_layernorm(input_embeds)
+        previous_hidden_states = self.hidden_layernorm(hidden_states)
+
+        concatenated = mx.concatenate([previous_hidden_states, input_embeds], axis=-1)
+        hidden_states = self.input_proj(concatenated)
```
awni, line 145: What's the purpose of the `input_embeds` parameter?

awni:
Looks awesome, thanks for adding that! It looks very nicely done! I think it makes sense to get rid of all the MTP stuff from the model file. We can always add it back when/if we intend to support it. Otherwise this is good to go!

mlx_lm/models/mimo.py
johnmai-dev: Source: https://huggingface.co/XiaomiMiMo/MiMo-7B-SFT/blob/main/modeling_mimo.py#L36

johnmai-dev:
> Looks awesome, thanks for adding that! It looks very nicely done! I think it makes sense to get rid of all the MTP stuff from the model file. We can always add it back when/if we intend to support it. Otherwise this is good to go!

Done.🍻

awni (approved):
Looks great!

mlx_lm/models/mimo.py
enduringstack: May I ask why not support mtp?

awni: The outer code needs to support using mtp as a speculative generation method and currently it doesn't so there's not much point to having it in the module itself.

Issue #140, opened 2025-05-01
Use a model's default generation config if it exists


awni:
E.g. Phi-4 reasoning works much better using the default config. We should read it if it exits:

https://huggingface.co/microsoft/Phi-4-reasoning/blob/main/generation_config.json

xz-keg:
t=0 doesn't work for most reasoning models. They likely fall into repetition. 

I make a pull request to change the default sampler from max to a more common setting of topp=0.95, t=0.6. #147

reneleonhardt:
generation_config.json is included since 0.24.1, is there some progress to actually use it for reasoning models like Phi-4?

reneleonhardt:
Is #486 enough to add generation_config.json to config.json?

TomLucidor:
What is the current situation with this? What if `generation_config.json` did not match to `config.json`? I found this issue with Nemotron-3-Nano as a one-off but not sure if this is more common than expeceted https://github.com/cubist38/mlx-openai-server/issues/122

PR #143, opened 2025-05-02
Refactor save and include generation config

awni:
- Refactor saving a model into `utils.save` so that we don't have multiple places where all the logic is duplicated
- Add `generation_config.json` to the files that we copy from the source. This way models converted with MLX keep the generation config so we can use it moving forward (see #140)


angeloskath (approved):
Nice!

PR #147, opened 2025-05-04
add a default sampler in generate.py

xz-keg:
The current effective default sampler when calling "generate'' is the t=0 max sampler which does not bring good performance, especially for reasoning LLMs. 
  
  Add a default sampler (t=0.6, topp=0.95, which is the most common setting for reasoning models ) to replace it.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -35,8 +35,8 @@ from .utils import load
 
 DEFAULT_PROMPT = "hello"
 DEFAULT_MAX_TOKENS = 100
-DEFAULT_TEMP = 0.0
-DEFAULT_TOP_P = 1.0
+DEFAULT_TEMP = 0.6
+DEFAULT_TOP_P = 0.95
 DEFAULT_MIN_P = 0.0
 DEFAULT_TOP_K = 0
 DEFAULT_XTC_PROBABILITY = 0.0
@@ -349,8 +349,11 @@ def generate_step(
         kv_group_size=kv_group_size,
         kv_bits=kv_bits,
     )
-
-    sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
+    default_sampler=make_sampler(
+        temp=DEFAULT_TEMP,
+        top_p=DEFAULT_TOP_P
+    )
+    sampler = sampler or default_sampler
 
     def _step(y):
         with mx.stream(generation_stream):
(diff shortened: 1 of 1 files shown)


awni:
Thanks for the contribution, but let's keep the defaults as they are. You are free to pass in whatever sampler you like. I'd prefer to push for a more general solution that makes an effort to use the generation config (https://github.com/ml-explore/mlx-lm/issues/140) since that way we can have the best setting for each model.

xz-keg:
> Thanks for the contribution, but let's keep the defaults as they are. You are free to pass in whatever sampler you like. I'd prefer to push for a more general solution that makes an effort to use the generation config (#140) since that way we can have the best setting for each model.

You actually can't. Make_sampler is not included in __init__.py so you cannot create a sampler yourself outside the package. And the default max sampler is unusable in most situations.

 Which means users need to manually change the code inside package after each update, which is very annoying.

awni:
> You actually can't. Make_sampler is not included in init.py so you cannot create a sampler yourself outside the package. And the default max sampler is unusable in most situations.

You should be able to. Here's an example:

```python
from mlx_lm import generate, load

checkpoint = "mistralai/Mistral-7B-Instruct-v0.3"
model, tokenizer = load(path_or_hf_repo=checkpoint)

prompt = "Why is the sky blue?"
conversation = [{"role": "user", "content": prompt}]

    conversation=conversation, add_generation_prompt=True
)
sampler = make_sampler(temp=0.8)

response = generate(
    model=model,
    tokenizer=tokenizer,
    prompt=prompt,
    max_tokens=128,
    verbose=True,
    sampler=sampler,
)
```

xz-keg:
> > You actually can't. Make_sampler is not included in init.py so you cannot create a sampler yourself outside the package. And the default max sampler is unusable in most situations.
> 
> You should be able to. Here's an example:
> 
> ```python
> from mlx_lm import generate, load
> from mlx_lm.sample_utils import make_sampler
> 
> checkpoint = "mistralai/Mistral-7B-Instruct-v0.3"
> model, tokenizer = load(path_or_hf_repo=checkpoint)
> 
> prompt = "Why is the sky blue?"
> conversation = [{"role": "user", "content": prompt}]
> 
> prompt = tokenizer.apply_chat_template(
>     conversation=conversation, add_generation_prompt=True
> )
> sampler = make_sampler(temp=0.8)
> 
> response = generate(
>     model=model,
>     tokenizer=tokenizer,
>     prompt=prompt,
>     max_tokens=128,
>     verbose=True,
>     sampler=sampler,
> )
> ```

okay I know this undocumented method works. You should put it in documentation.

PR #148, opened 2025-05-04
Allow loading already quantized model

awni:
Allow loading an already quantized model as the student rather than always quantizing the teacher. This let's us:
- More easily used mixed quant configs
- Use a cascade of e.g. AWQ -> DWQ


angeloskath (approved):
👍

PR #152, opened 2025-05-04
Added flag --limit-to-loaded-model to cli and limit request to list…

fu:
… only loaded model. This will provide possibility to launch multiple servers on same host and each instance can stick to its loaded model instead of lazy loading other available models


awni:
I'm not following the purpose of this PR. You can already run multiple instances of mlx_lm.server with different ports and use different models simultaneously. Or you can swap the model you are using in the server by specifying the model field. Unless you specify a new model, it won't change.. so the default behavior is to use the loaded model.

I'm going to close this for now. If you comment with some use case that I might be missing that this solves.. we can re-open / revisit.

PR #153, opened 2025-05-04
Quantization aware training

awni:
This adds quantization aware training via distillation to the base model. I'm not yet sure it's worth adding. Just making the PR for reference.

I think it closely much matches the implementation in the Gemma3 tech report / QAT paper. See e.g. section 2.3 of [Gemma3 tech report](https://storage.googleapis.com/deepmind-media/gemma/Gemma3Report.pdf).

I want to run evals against AWQ / DWQ / etc to see how they all compare. Also if we do add it.. there is a lot of common code with DWQ so it would be good to refactor a bit.

Also the straight through estimators are probably pretty inefficient so maybe those should be done with a custom function instead.

### DIFF
--- a/mlx_lm/dwq.py
+++ b/mlx_lm/dwq.py
@@ -45,9 +45,10 @@ def dwq_quantize(
     data,
     batch_size: int = 2,
     max_seq_length: int = 2048,
-    temperature: float = 0.5,
-    dtype: mx.Dtype = mx.bfloat16,
+    temperature: float = 1.0,
+    activation_layer_step: float = 0.25,
     activation_loss_weight: float = 1e-1,
+    dtype: mx.Dtype = mx.bfloat16,
 ):
     group = mx.distributed.init()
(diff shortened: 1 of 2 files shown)


awni:
Some benchmark results on Gemma3 4B.

- Benchmark is on ~2200 samples from MMLU zero shot
- All quants are group size 64
- bf16 / q8 / q6 are very similar
- q4 Google QAT is the Google QAT trained model. It's really meant for Q4_0 group size 32 so it makes some sense that it doesn't help MLX group size 64 Q4_1
- DWQ and MLX QAT are both on the same dataset with 2048 samples. They give similar results.
- AWQ helps but is generally worse
- Training the QAT and DWQ longer helped a bit (2048 > 1024 by ~1% or so)
- Q4 AWQ + DWQ basically matches q8 / bf16.

Screenshot 2025-05-06 at 12 46 37 PM

PR #154, opened 2025-05-05
Add support for chunked request bodies (llama-swap compatibility)

reznej:
## Motivation
I use mlx-lm in combination with [Open WebUI](https://github.com/open-webui/open-webui). One issue I've encountered is that once a model is loaded into memory, it stays resident indefinitely—even after it's no longer in use. Since I also rely on this machine for other productive tasks, I wanted to enable automatic memory release after inactivity, similar to the keep_alive behavior in [Ollama](https://github.com/ollama/ollama).

During my research, I came across [Llama-Swap](https://github.com/mostlygeek/llama-swap), a lightweight proxy to manage different models. One of its features is a `ttl` config parameter, which unloads the model after a specified idle timeout—freeing up system resources automatically.

## Problem
However, while integrating llama-swap into my setup, I discovered that mlx-lm (as of v0.24.0) is not compatible with transfer-Encoding: chunked, which seemingly llama-swap uses for proxying HTTP requests. This causes inference requests to fail with a 502 due to missing content-Length.

## Solution
This PR introduces a minimal addition to the `do_POST` method in `server.py`, enabling mlx-lm to decode chunked HTTP request bodies. It preserves existing behavior for requests with a content-length header. Therefore, this change should have no effect on current usage patterns but enables broader compatibility with toolchains that rely on chunked streaming.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -293,8 +293,22 @@ class APIHandler(BaseHTTPRequestHandler):
             return
 
         # Fetch and parse request body
-        content_length = int(self.headers["Content-Length"])
-        raw_body = self.rfile.read(content_length)
+        content_length = self.headers.get("Content-Length")
+        if content_length:
+            raw_body = self.rfile.read(int(content_length))
+        else:
+            chunks = []
+            while True:
+                line = self.rfile.readline()
+                if not line:
+                    break
+                size = int(line.strip(), 16)
+                if size == 0:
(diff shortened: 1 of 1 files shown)


yihongang:
I think llama-swap [wants to use content-length](https://github.com/mostlygeek/llama-swap/pull/13), but it's just broken at the moment. There's an open PR to fix it, but we may take some time to agree on approach.

zcbenz:

Issue #149, opened 2025-05-04
[AWQ] Align MLX-LM with original implementation


casper-hansen:
The original implementation uses the configuration as seen below. I would recommend putting this command in the documentation as this will likely improve accuracy compared to 32 samples and a grid size of 10. A sequence length of 2048 is also _nice_, but it may be more demanding on memory requirements.

```
mlx_lm.awq --model HF_MODEL_PATH --bits 4 --num_samples 128 --n_grid 20 --sequence_length 512
```

The original implementation also uses for asymmetric quantization, which should improve accuracy as well. Is this planned to be supported in the future?

awni:
I'm  ok with changing the defaults to match the original settings. Is that what you use in AutoAWQ?

> The original implementation also uses for asymmetric quantization, which should improve accuracy as well. Is this planned to be supported in the future?

All of MLX quantization is asymmetric.

casper-hansen:
> I'm ok with changing the defaults to match the original settings. Is that what you use in AutoAWQ?

Yes, those defaults are used in AutoAWQ and in the original research repository.

> All of MLX quantization is asymmetric.

Good to hear this! It’s not seemingly apparent from the quantization code, but nonetheless this is good.

awni:
> Good to hear this! It’s not seemingly apparent from the quantization code, but nonetheless this is good.

Instead of calling those parameters `zeros` we call them `biases`. So every quantized weight has `scales` and `biases` parameters associated with them. You'll notice the `mx.quantize` returns three tensors as output (quantized weights, scales and biases).

awni:
@casper-hansen we updated the defaults to match AutoAWQ. It seems to work well, thanks for the tip!

I'm wondering if you could share some experience on the data side.
- How important is the calibration data in your experience? We use wikitext but I'm playing around with an [alternative](https://github.com/ml-explore/mlx-lm/pull/159).
- Is it important to use e.g. prompt templates correctly and generally have the data formatting match the real-world model use as much as possible?

The calibration may or may not be that sensitive to the data.. any insights there would be great to hear!

casper-hansen:
> * How important is the calibration data in your experience? We use wikitext but I'm playing around with an [alternative](https://github.com/ml-explore/mlx-lm/pull/159).

[...]

PR #157, opened 2025-05-06
improvements / defaults for awq

awni:
- Use the suggested defaults as in #149
- Matches the standard AWQ clipping implementation (use a single clip value)
- Adds a fallback to the original weights for a given block if the loss is nor reduced


angeloskath (approved):
Looks good!

It isn't immediately obvious to me why one-sided clipping is better but it definitely is a more gradual "corruption" of the matrix ...

awni:
Nor to me. But the loss reduction for the one sided is generally a bit better than the two-sided 🤷‍♂️ .

PR #158, opened 2025-05-06
Add Evaluation Progress

rgeorgi:
Adding a tqdm progress bar for evaluate to make it more transparent what’s happening when --test is called, rather than an empty screen.


awni (approved):
LGTM, thanks!

PR #159, opened 2025-05-06
Better calibration data for AWQ

awni:
Uses a more diverse dataset for AWQ calibration.

On Gemma3 4B this leads to a nice improvement in downstream results (48.4 MMLU vs 50.96)


angeloskath (approved):
👍

PR #161, opened 2025-05-07
DWQ extra activation penalty

awni:
Depends on #148 

Adds a penalty between the hidden states of the student and the teacher:
- The penalty uses L1 (L2 was too sensitive to large values).
- Defaults to use 4 layers equally spaced
- Loss is `kl_loss + activation_weight * activation_loss`
- Updates some default values which work better

New results on 2200 MMLU questions look pretty good for Gemma3 and Qwen3 4b models:

Screenshot 2025-05-07 at 7 45 10 AM

### DIFF
--- a/mlx_lm/LEARNED_QUANTS.md
+++ b/mlx_lm/LEARNED_QUANTS.md
@@ -40,6 +40,33 @@ For a full list of options run:
 mlx_lm.dwq --help
 ```
 
+#### Tips
+
+- DWQ works best distilling to lower precision, anywhere from 2-bit to 4-bit
+  models.
+- Distilling 16-bit precision to 8-bit and even 6-bit often doesn't work well.
+  The loss starts out so low that it's difficult to reduce further.
+- Decreasing the quantization group size (e.g. `--group-size 32`) doubles the
+  number of tunable parameters and can work much better.
+- If the loss is oscillating and not going down consistently, try reducing the
+  learning rate. If it is decreasing but slowly, try increasing the learning
+  rate.
+- As a rule of thumb, lower precision can benefit from a higher learning rate
+  since the loss starts out higher. Conversely, higher precision needs a lower
+  learning rate.
+
+
+#### Memory Use
+
+A few options to reduce memory use for DWQ:
+
+- Distill from an 8-bit model instead of a 16-bit model. The 8-bit
+  models are usually as good as 16-bit precision models.
+- Use a shorter maximum sequence length. The default is 2048. Using
+  `--max-seq-length 512` reduces the memory and still gets good results.
+- Use a smaller batch size, e.g. `--batch-size 1`
+
+
 ### AWQ 
 
 Use `mlx_lm.awq` to run AWQ on a given model. For example:
--- a/mlx_lm/dwq.py
+++ b/mlx_lm/dwq.py
@@ -45,9 +45,10 @@ def dwq_quantize(
     data,
     batch_size: int = 2,
     max_seq_length: int = 2048,
-    temperature: float = 0.5,
-    dtype: mx.Dtype = mx.bfloat16,
+    temperature: float = 1.0,
+    activation_layer_step: float = 0.25,
     activation_loss_weight: float = 1e-1,
+    dtype: mx.Dtype = mx.bfloat16,
 ):
     group = mx.distributed.init()
     world_size = group.size()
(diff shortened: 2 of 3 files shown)


angeloskath (approved):
🚀

vgoklani:
The fact that your Q4 DWQ is better than the bf16 is very suspicious.

awni:
@vgoklani indeed. The Qwen3 models are consistently slightly better post DWQ which is a bit unsettling. Current hypothesis is that this has to do overtraining to non-thinking mode which matches the conditions of the eval as well. 

As next steps on that:
- it's worth checking if adding some thinking samples to the calibration data will fix that
- benchmarking a task which does use thinking (running this already)

awni:
Did some perplexity benchmarks thanks to @N8python for putting the benchmark together. The dataset has thinking traces in it. See [here](https://huggingface.co/datasets/ServiceNow-AI/R1-Distill-SFT).

On Qwen3 4B:

Precision | PPL
--- | ----
BF16 | 1.95
Q4 | 2.05
DWQ Q4 | 1.90

awni:
Also did a DWQ with 6 bit distilled to the bf16 model (h/t @n8python). Results on MMLU. The DWQ Q6 is still the best, but the variations are pretty small.

Precision | Acc
---- | -----
BF16 | 54.17
Q6 | 53.95
DWQ Q6 | 54.96

Generally it seems like the process of quantizing + DWQ distilling generally leads to a better model even than the baseline on the benchmarks we've looked at. It's not so intuitive..

In theory I suppose it's possible quantization + distillation as a combined action acts as a regularizer leading to overall better generalization. It's also possible we're getting lucky on the benchmarks and the model is worse in areas we haven't tested.

awni:
MMLU Pro computer science is a generative benchmark (so can test it with thinking enabled). Here's the results with thinking enabled on Qwen3 4B:

Precision | Accuracy
----- | -----
BF16 | 73.17
Q4 | 68.29
DWQ Q4 | 69.76


So while we do get a nice improvement with DWQ it looks like we may not get as big an improvement in "thinking" mode as non thinking mode. Which could be an artifact of the calibration data.

awni:
I made a calibration dataset specifically for Qwen3 using some rollouts which have thinking traces generated from the 30B MoE. In case anyone wants to play around with it `--data-path mlx-community/qwen3_dwq_calibration_1332`.

CC @mzbac @n8python @mark-lord

mzbac:
I have been experimenting a bit with the 4-bit DWQ using mlx-community/qwen3_dwq_calibration_1332. It doesn't seem to improve much on the thinking mode output. will try some real-world use cases for non-thinking outputs later.

awni:
I ran a few evals with different variations of the calibration dataset. So 50/50 if 50% from the original and 50% rollouts with thinking traces. 100/0 is 100% from the original (the current default). The MMLU is a non-thinking eval. The MMLU Pro CS is a thinking (generative) eval.

TLDR it does seem using a higher ratio of thinking traces helps with the thinking eval.. but it also hurts with the non-thinking eval. The MMLU Pro CS eval is not huge (~400 questions) so there should be some variance there. But the trend is fairly consistent.

For now though I would probably still use the default 100/0 dataset as it seems more helpful overall. Ideally we can find a calibration dataset/ method that helps a bit more on thinking without hurting so much on non-thinking. Also this is really quite specific to Qwen3...

Model | MMLU  | MMLU Pro CS
----- | ----  | ----
bf16       | 54.17 | 73.17
Q4         | 45.61 | 68.29
DWQ 100/0  | 55.35 | 69.76
DWQ 88/12  | 51.27 | 68.05
DWQ 75/25  | 44.91 | 70.49
DWQ 50/50  | 42.19 | 71.22

mzbac:
> I ran a few evals with different variations of the calibration dataset. So 50/50 if 50% from the original and 50% rollouts with thinking traces. 100/0 is 100% from the original (the current default). The MMLU is a non-thinking eval. The MMLU Pro CS is a thinking (generative) eval.
> 
> TLDR it does seem using a higher ratio of thinking traces helps with the thinking eval.. but it also hurts with the non-thinking eval. The MMLU Pro CS eval is not huge (~400 questions) so there should be some variance there. But the trend is fairly consistent.
> 
> For now though I would probably still use the default 100/0 dataset as it seems more helpful overall. Ideally we can find a calibration dataset/ method that helps a bit more on thinking without hurting so much on non-thinking. Also this is really quite specific to Qwen3...
> 
> Model	MMLU	MMLU Pro CS
> bf16	54.17	73.17
> Q4	45.61	68.29
> DWQ 100/0	55.35	69.76
> DWQ 88/12	51.27	68.05
> DWQ 75/25	44.91	70.49
> DWQ 50/50	42.19	71.22

Is there a specific setting for using the thinking traces in DWQ? I looked at the code, and it seems that DWQ doesn't use the reasoning content from the dataset. Maybe that's the why I converted DWQ quant with qwen3_dwq_calibration_1332, doesn't seem to be much improvement.

awni:
You shouldn't need a specific setting. The chat template handles the `reasoning_content` in any of the `assistant` messages by putting it between thinking tags.

priontific:
Super interesting results... out of interest, did you try doing 0% original and 100% rollouts with thinking traces?

awni:
I didn't try that. It would be interesting to compare as well but I didn't generate enough rollouts for that experiment (would need 1024 to have a comparable number of samples).

PR #162, opened 2025-05-08
add default params to better run quen3

rrziehe:
these new server.py args can be used to configure better default values for:

- num_draft_tokens
- temp
- top_p
- top_k
- min_p
- max_tokens

another arg is added to disable thinking for qwen3 models

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -307,12 +307,18 @@ class APIHandler(BaseHTTPRequestHandler):
         self.stream_options = self.body.get("stream_options", None)
         self.requested_model = self.body.get("model", "default_model")
         self.requested_draft_model = self.body.get("draft_model", "default_model")
-        self.num_draft_tokens = self.body.get("num_draft_tokens", self.model_provider.cli_args.num_draft_tokens)
+        self.num_draft_tokens = self.body.get(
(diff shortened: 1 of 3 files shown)


mlx_lm/server.py
```diff
@@ -759,7 +765,9 @@ def handle_chat_completions(self) -> List[int]:
                 messages,
                 body.get("tools", None),
                 add_generation_prompt=True,
-                enable_thinking = False if self.model_provider.cli_args.disable_thinking else None,
+                enable_thinking=(
+                    False if self.model_provider.cli_args.disable_thinking else None
+                ),
```
awni, line 770: I don't think we should hard code this field name since it's not standard across tokenizers. Instead it would be good to allow to pass a json dict of extra tokenizer template parameters which could be model specific. See e.g. https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/evaluate.py#L367-L373

awni:
This is a nice addition! Could you address the comment I left, then we can merge it?

mlx_lm/server.py
rrziehe: done, thanks

awni (approved):

awni (approved):

PR #164, opened 2025-05-09
Add search strategy

nfergu:
This is a proposal for adding a "search strategy" to token generation. This enables different strategies for selecting generated tokens. For example, see this (not tested) [ prototype of beam search](https://github.com/nfergu/mlx-lm/pull/2) that is based on this.

I have left the interface of `generate_step` unchanged by defaulting to a "linear search" strategy, which does the same thing as before. In addition, there is a new function called `generate_with_search` which accepts a `SearchStrategy` implementation, which allows other strategies to be used. I haven't plumbed `generate_with_search` into the main `stream_generate` but perhaps it should be? 

There's perhaps some tidy up to be done, but I thought I'd get this out for feedback before I did too much more on it.

Let me know what you think. Happy to make changes. One alternative to this PR would be that search strategies would duplicate the code that is currently in `generate_with_search` within the `SearchStrategy` implementation itself. Or we  refactor the code that is currently in `generate_with_search` into a helper function that search strategies would call. If you think either of those are nicer approaches I'll close this PR.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -293,20 +293,27 @@ class SearchStrategy(ABC):
     @abstractmethod
     def generate(
         self,
-        y: mx.array,
+        prompt: mx.array,
         prompt_cache: List[Any],
         quantize_cache_fn: Callable[[Any], None],
         total_prompt_tokens: int,
+        prompt_progress_callback: Callable[[int, int], None],
     ) -> Generator[Tuple[int, mx.array], None, None]:
         """
         Generates a sequence of tokens from the model.
 
         Args:
-            y (mx.array): The next token sequence.
-            prompt_cache (List[Any]): The prompt cache.
-            quantize_cache_fn (Callable[[Any], None]): A function to quantize the cache.
-            total_prompt_tokens (int): The total number of tokens in the prompt.
-
+            prompt (mx.array): The prompt to use for generation. Note: this may not be the
+                complete prompt if some of it has been cached in the `prompt_cache`.
+            prompt_cache (List[Any]): The prompt cache. This is normally updated when the model
+                is called to generate new tokens.
+            quantize_cache_fn (Callable[[Any], None]): A function to quantize the cache. This
+                should be called when the cache has been updated (by the model).
+            total_prompt_tokens (int): The total number of tokens in the prompt. This should
+                be passed to the `prompt_progress_callback` when the prompt has been processed.
+            prompt_progress_callback (Callable[int, int]): A call-back which takes the
+               prompt tokens processed so far and the total number of prompt tokens.
+               This should be called when the prompt has been processed.
         Yields:
             Tuple[int, mx.array]: A token and a vector of log probabilities.
         """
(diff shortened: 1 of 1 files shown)


mlx_lm/generate.py
```diff
@@ -322,15 +352,16 @@ def generate_step(
...
            when ``kv_bits`` is non-None. Default: ``0``.
-        prompt_prorgress_callback (Callable[int, int]): A call-back which takes the
+        prompt_progress_callback (Callable[int, int]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
+        search_strategy (SearchStrategy): The search strategy to use. This defaults to
+            `LinearSearch`.
 
     Yields:
-        Tuple[mx.array, mx.array]: One token and a vector of log probabilities.
+        Tuple[int, mx.array]: One token and a vector of log probabilities.
```
nfergu, line 361: Not related to the main change, but I think the typing was wrong here. I'm pretty sure this method generates int, as it calls `.item()` on the token array, but I might be missing something.

mlx_lm/generate.py
```diff
@@ -284,21 +285,50 @@ def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_
...
+        pass
+
+
 def generate_step(
     prompt: mx.array,
     model: nn.Module,
     *,
     max_tokens: int = 256,
-    sampler: Optional[Callable[mx.array, mx.array]] = None,
+    sampler: Optional[Callable[[mx.array], mx.array]] = None,
```
nfergu, line 321: Not related to the main change, but I think the typing was wrong here. The sampler has a single `mx.array` argument and returns an `mx.array`.

mlx_lm/generate.py
```diff
@@ -284,21 +285,50 @@ def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_
...
     logits_processors: Optional[List[Callable[[mx.array, mx.array], mx.array]]] = None,
     max_kv_size: Optional[int] = None,
     prompt_cache: Optional[Any] = None,
     prefill_step_size: int = 2048,
     kv_bits: Optional[int] = None,
     kv_group_size: int = 64,
     quantized_kv_start: int = 0,
-    prompt_progress_callback: Optional[Callable[int, int]] = None,
-) -> Generator[Tuple[mx.array, mx.array], None, None]:
+    prompt_progress_callback: Optional[Callable[[int, int], None]] = None,
```
nfergu, line 329: Not related to the main change, but I think the typing was wrong here. The callback has two `int` arguments, and doesn't return anything.

mlx_lm/generate.py
```diff
@@ -284,21 +285,56 @@ def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_
...
+    A strategy for obtaining a sequence of tokens from the model.
+    """
+
+    @abstractmethod
+    def generate(
+        self,
+        prompt: mx.array,
+        prompt_cache: List[Any],
+        quantize_cache_fn: Callable[[Any], None],
+        total_prompt_tokens: int,
```
nfergu, line 299: I think it's a bit awkward that `generate` needs to take `total_prompt_tokens` and `prompt_progress_callback`, but I could immediately see a nice way around this.

zcbenz:

PR #165, opened 2025-05-09
Fixed calculation of n_dims for odd bit quants

cblomert:
Fixed a rounding error when creating LoRA-adapters for odd-bit-quantized models (e.g. 6 bit variants and all others that are not divisors of 32)


awni (approved):
Good catch, thanks!!

altaic:
I'm surprised that `imul` is numerically different to `mul` for ints. Am I missing something (other than coffee)?

awni:
dims * (32//bits) vs (dims * 32) // bits

PR #167, opened 2025-05-11
Performance regression when reusing trainer with gradient checkpointing

cblomert:
When repeatedly calling train() with activated gradient checkpointing (e.g. for HPO) the __call__ method would recursively be mapped to a checkpointed fn, which led to a performance decrease on each iterative run. 
This can be fixed by restoring the original __call__ method after training finishes.

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -113,6 +113,7 @@ def convert(
 
     print("[INFO] Loading")
     model_path = get_model_path(hf_path, revision=revision)
+    print("[INFO] Model path:", model_path)
     model, config, tokenizer = fetch_from_hub(model_path, lazy=True)
 
     if isinstance(quant_predicate, str):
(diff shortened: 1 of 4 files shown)


awni:
I think a with statement could work nicely for this. And maybe just around the call to `step` (it should be very cheap):

```python
with grad_checkpoint(model) if args.grad_checkpoint else nullcontext(): 
    step()
```
  
Then you can change the `grad_checkpoint` to be a context manager:

```python
def grad_checkpoint(model):                                                    
    """
    Context manager for gradient checkpointing.         
    """ 
    layer = model.layers[0]
    fn = type(layer).__call__
        
    def checkpointed_fn(model, *args, **kwargs):                               
        def inner_fn(params, *args, **kwargs):                                 
            model.update(params)                                               
            return fn(model, *args, **kwargs)                                  

        return mx.checkpoint(inner_fn)(model.trainable_parameters(), *args, **kwargs)   
    type(layer).__call__ = checkpointed_fn 
    try:
      yield None
    finally:                                                        
      type(layer).__call__ = fn                
```

awni:
@cblomert are you interested in coming back to this?

zcbenz:

PR #168, opened 2025-05-11
Beam search

nfergu:
Beam search prototype. Not tested yet


nfergu:
Closed. Meant to target fork.

PR #169, opened 2025-05-11
Add ability to define default parameters for server in a settings file

BenceSzalai:
Same as #131 was, but I closed that because I accidentally opened that from my main,which is obviously is a mess.

When using `mlx_lm.server` to play with models locally, it can become a huge time waster to research and figure out how the parameters can be set for each tool that connects to the API. Some tools have nice UI to add arbitrary body parameters, some tools only have some common parameters on the UI, and need setting these in config files, and some tools does not allow adding custom body parameters at all, in which case one has to write and use a utility proxy to sit between the client and the server to modify the request. This is a huge time waster and also makes it hard to maintain all these configurations over various tools.

Here the key idea is that while some parameters may worth changing per client or even per request, some should be configurable as defaults for all requests. One such thing for me is proper `draft_model` for each of my installed models, which can easily 2x the generation speed. This should be a one time set and forget thing, regardless of what client will consume the server after.

For this reason, I've implemented a config file to sit in the user's home folder where the parameters can be preset for all requests.

In the config file I put everything under the main key of `server` thinking that maybe in the future other modules could utilize the same file as well. Also, under server I put everything under `parameters`, so other settings can be added later for the `server` module too.

One thing I am quite indecisive about is how the file should be read. First I made it only read the file once, and reuse the same value until server restart, but it makes fiddling with the file really tedious, to always restart the server, and since the settings are only loaded once per request I think it is better to load it every time. But this also means if other modules start to use the same `get_settings()` call, they must make sure the result is handled properly to avoid repeated reading of the settings file e.g. inside a loop or so.

Let me know what do you think!

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -4,11 +4,12 @@ import http
 import json
 import threading
 import unittest
+from unittest.mock import patch
 
 import requests
 
 from mlx_lm.server import APIHandler
-from mlx_lm.utils import load
+from mlx_lm.utils import get_settings, load
 
 
 class DummyModelProvider:
@@ -54,6 +55,9 @@ class TestServer(unittest.TestCase):
         cls.server_thread = threading.Thread(target=cls.httpd.serve_forever)
         cls.server_thread.daemon = True
         cls.server_thread.start()
+        # Prevent user settings from being loaded and interfere with test runs
+        cls.get_settings_patcher = patch("mlx_lm.server.get_settings", return_value={})
+        cls.mock_get_settings = cls.get_settings_patcher.start()
 
     @classmethod
     def tearDownClass(cls):
--- a/mlx_lm/SERVER.md
+++ b/mlx_lm/SERVER.md
@@ -137,7 +137,8 @@ example above `"max_tokens": 256` would be applied to even the
 `mlx-community/Qwen2.5-Coder-32B-Instruct-8bit` model if it would not have
 `"max_tokens": 16384` written under its own key.
 
-You have to restart the server for any changes in this file to take effect!
+Any changes in this file take effect for any subsequent requests without a need to
+restart the server.
 
 
 ### Response Fields
--- a/mlx_lm/SERVER.md
+++ b/mlx_lm/SERVER.md
@@ -92,6 +92,53 @@ curl localhost:8080/v1/chat/completions \
 - `num_draft_tokens`: (Optional) The number of draft tokens the draft model
   should predict at once. Defaults to `3`.
 
+#### Custom defaults
+You can set custom defaults for the server by `~/.mlx_lm.settings.json` in
+your home directory. In this file you can specify a different default value
+for all the request fields listed above.
+
+```json
+{
+  "server": {
+    "parameters": {
+      "default": {
+        "max_tokens": 256
+      }
+    }
+  }
+}
+```
+This will set the default `max_tokens` to `256` for all requests, which does not
+contain `max_tokens` in the request body.
+
+The default values can be overridden on a per-model basis as well. For example to
+force speculative decoding for the `mlx-community/Qwen2.5-Coder-32B-Instruct-8bit`,
+this can be added:
+
+```json
+{
+  "server": {
+    "parameters": {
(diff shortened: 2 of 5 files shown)


tests/test_server.py
```diff
@@ -54,6 +55,9 @@ def setUpClass(cls):
         cls.server_thread = threading.Thread(target=cls.httpd.serve_forever)
         cls.server_thread.daemon = True
         cls.server_thread.start()
+        # Prevent user settings from being loaded and interfere with test runs
+        cls.get_settings_patcher = patch("mlx_lm.server.get_settings", return_value={})
+        cls.mock_get_settings = cls.get_settings_patcher.start()
```
BenceSzalai, line 60: I'm not sure if this is good approach. It defo highlights an issue introduced by a global settings. Maybe a more proper mocking should be used instead, but honestly, I could not get it working in a more global way. (Python is not my native language.) My concern with this is that this patterns should become repeated in all tests for any modules which use the settings file.

awni:
Thanks for the contribution. I don't think this machinery is yet necessary in mlx-lm.server. It would serve your use cases just as well to read your own settings file and populate the request before sending it to the server?

I would suggest you do that for now. If we see others finding a need for something like this we can reconsider having a global settings file.

BenceSzalai:
Thank you for looking into it!

> It would serve your use cases just as well to read your own settings file and populate the request before sending it to the server?

That's right indeed when polling the server from custom code, but when the server is used from other tools, frontends etc it is harder to make sure all these settings are in place if even all of these can be configured in those.

It is a question of where the service provided by the server should be defined? In the server itself or in the client consuming the api. From my perspective allowing the client to control these parameters is all good, also it is how the openai api works, which should be the baseline to follow. On the other hand for any setting not explicitly stated by the consumer there should be a default value. So it comes down if those defaults are configurable by whoever runs the server or are these defaults part of the sw / repo itself.

So basically we can say there are 3 theoretical levels to configure the model before inference:
- client - consumer
- server configuration - server administrator
- server software - server developer (owner of this repo)

My suggestion is about adding the missing mid-level, the level of the server administrator.

Amyway, indeed I have to and I do accept your decision. I hope you may reconsider it at some point. Meanwhile I'll probably just monkey patch my instance or make a small proxy to sit in between my clients and the server.

awni:
Thanks for the constructive explanation. 

> My suggestion is about adding the missing mid-level, the level of the server administrator.

What do you think about using command line arguments for that instead? Like for max tokens you do `mlx_lm.server --max-tokens 512` when you launch the server. And that's what it will use unless otherwise specified?

I can see one downside is that you can't change it without restarting the server.. but that seems minor to me.

BenceSzalai:
For me another downside seems to be that these parameters are best set per model. Using CLI implies that any server instance would always serve only one given model, but in my use-cases - which may or may not be indicative of others - I poll the running server from multiple tools and for multiple different models. Context length is indeed a parameter that can be different per model, which may need to be reflected or may affect the desired `max_tokens` value, but what I use most the time the config file for (in my patched version) is to preset `draft_model` and `num_draft_tokens` for all my main ones to achieve faster inference.

I am happy to discuss the details of this, but I don't want to take your time away from other more important things.

So I'm wondering what part of using a config file you see issue with? Because I am myself a bit sceptical about what I did, that is saying a config file should be at a predefined path kind of introducing a global state into the software. That could cause issues if someone forgets what they have set there or when sever process runs with different user account etc.

But maybe a CLI option would be more appropriate to point to a config file. That would also solve the issue with tests and not setting a `--config` (or so) option could ensure the server runs in it's pure setup, so basically any troubleshooting could just require remove the config argument as step 0.

But if you have other concerns, like increase of code complexity or so, obviously this would not address those.

awni:
Yea I think you summed up my concerns well and I agree a CLI argument for the path to the config is much more palatable.

I think the other thing is I'd like to understand this problem a bit better and the suite of options we have to solve it to make sure we solve it as optimally as possible. So understanding which config parameters need to live in a config file vs from the CLI vs part of the model config itself is useful.

PR #166, opened 2025-05-09
mistral3

awni:
Closes #136

Issue #136, opened 2025-04-29
ERROR:root:Model type mistral3 not supported.


dennisnoto-etana:
trying to run the mlx convert py for mistralai/Mistral-Small-3.1-24B-Instruct-2503 and get 

ERROR:root:Model type mistral3 not supported.

On versions 
mlx                               0.25.1
mlx-lm                          0.24.0

!python3 -m mlx_lm convert \
        --hf-path {huggingFaceModel} \
        --mlx-path {macModel} \
        --dtype bfloat16

model directory does not have files for mistral models.  Been using MLX for gemma3 and llama with no issues. 

Full error

Fetching 23 files: 100%|█████████████████████| 23/23 [00:00<00:00, 20051.75it/s]
ERROR:root:Model type mistral3 not supported.
Traceback (most recent call last):
  File "/Users/dennisnoto/Documents/NotoData-Dev/JupyterNotebooks/Anaconda-2025-3/anaconda3/lib/python3.12/site-packages/mlx_lm/utils.py", line 71, in _get_classes
    arch = importlib.import_module(f"mlx_lm.models.{model_type}")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dennisnoto/Documents/NotoData-Dev/JupyterNotebooks/Anaconda-2025-3/anaconda3/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1324, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'mlx_lm.models.mistral3'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/dennisnoto/Documents/NotoData-Dev/JupyterNotebooks/Anaconda-2025-3/anaconda3/lib/python3.12/site-packages/mlx_lm/__main__.py", line 27, in <module>
    submodule.main()
[...]

PR #172, opened 2025-05-12
Allow to use mistral3 model type for lora

prldev:
This complete request  [#166](https://github.com/ml-explore/mlx-lm/issues/136) to be able to use Mistral3 during lora.
Without this change, mlx_vlm.lora was aborted with message "Lora does not support mistral3"


awni (approved):

awni:
Thanks!

PR #173, opened 2025-05-13
Don't hide exception with wrong message when HF model isn't found

awni:
As title.. 

The default HF exception when the repository isn't found is fine.. I'm not sure why we catch and re-raise a custom exception.

Also in cases where you run out of disk space or the download fails for a different reason we were hiding that with the wrong model not found message.


angeloskath (approved):
👍

Issue #175, opened 2025-05-13
wandb integration issues


ivanfioravanti:
Hi,
I have seen some small issues in the wandb integration:
- steps reported = real steps / 10
- validation loss is report on step + 1, instead of the correct one. i.e. 20,41,63,84, etc.

/cc super @Goekdeniz-Guelmez

Goekdeniz-Guelmez:
Fix comming this evening :D Thanks!!!!!

ivanfioravanti:
Thanks! @Goekdeniz-Guelmez this integration is great! No more need for custom implementations.

awni:
Thanks for flagging this issue! Should be resolved in #177.

PR #177, opened 2025-05-13
Wandb fix

Goekdeniz-Guelmez:
Addresses this Issue #175

### DIFF
--- a/mlx_lm/tuner/trainer.py
+++ b/mlx_lm/tuner/trainer.py
@@ -261,7 +261,7 @@ def train(
 
             if training_callback is not None:
                 val_info = {
-                    "iteration": it,
+                    "iteration": it - 1,
                     "val_loss": val_loss,
                     "val_time": val_time,
                 }
@@ -299,7 +299,7 @@ def train(
 
             if training_callback is not None:
                 train_info = {
-                    "iteration": it,
+                    "iteration": it // args.steps_per_report,
                     "train_loss": train_loss,
                     "learning_rate": learning_rate,
                     "iterations_per_second": it_sec,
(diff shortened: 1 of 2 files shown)


mlx_lm/tuner/trainer.py
```diff
@@ -299,7 +299,7 @@ def step(batch):
 
             if training_callback is not None:
                 train_info = {
-                    "iteration": it,
+                    "iteration": it // args.steps_per_report,
```
awni, line 302: I don't think it makes sense to change this. The iteration should just be the iteration we are currently on right?

Goekdeniz-Guelmez: Oh yea my bad, that would make only sense if we were explicitly trying to report in units of reporting steps but that’s not meaningful to the training process or metrics analysis.

awni (approved):

mlx_lm/tuner/trainer.py
```diff
@@ -261,7 +261,7 @@ def step(batch):
 
             if training_callback is not None:
                 val_info = {
-                    "iteration": it,
+                    "iteration": it - 1,
```
awni, line 264: Also I don't think this fixes the issue from @ivanfioravanti. I think what we need to do is pass the `step` to `wandb.log`.  E.g. `wandb.log(train_info, step=info["iteration"])`

Otherwise each call to `log` creates a new step and validation and train will be out of sync.

Goekdeniz-Guelmez: Do you mean something like:

```python
def on_train_loss_report(self, train_info: dict):
        wandb.log(train_info, step=train_info.get("iteration"))
        if self.wrapped_callback:
            self.wrapped_callback.on_train_loss_report(train_info)
```

awni (approved):
Thanks!

Goekdeniz-Guelmez:
Always a pleasure!

PR #179, opened 2025-05-14
Add input_embeddings input to generate_step, Gemma 3, Qwen 2

mattjcly:
### Overview

Enables `input_embeddings` to be used instead of `prompt`. This allows for prompt enrichment such as injecting image embeddings for models like Gemma3 and Qwen2/2.5VL.

I originally was only going to pipe this for Gemma3, but since `mlx-community/Qwen1.5-0.5B-Chat-4bit` (Qwen2 arch) is the current text model used for testing, I added `input_embeddings` to the Qwen2 model as well. This should eventually prove useful to be able to run Qwen2/2.5VL with image embeddings as input.

### Testing
- test_generate.py test for `stream_generate` with input embeddings
- test_generate.py test for `stream_generate` with input embeddings, with `prefill_step_size` < number of embeddings
```
matt@Matts-MacBook-Pro [10:35:14] [~/Workspace/forks/mlx-lm] [input-embeddings-gemma3]
-> % python -m unittest tests/test_generate.py   
Fetching 9 files: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 134337.14it/s]
Fetching 9 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 21423.80it/s]
.
----------------------------------------------------------------------
Ran 6 tests in 0.793s
```
- POC confirmation of the ability to compute Gemma3 image embeddings outside of `stream_generate`, then send them in with correct understanding of the image by the text model.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -31,7 +31,7 @@ from .models.cache import (
 )
 from .sample_utils import make_sampler
 from .tokenizer_utils import TokenizerWrapper
-from .utils import load
+from .utils import does_model_support_input_embeddings, load
 
 DEFAULT_PROMPT = "hello"
 DEFAULT_MAX_TOKENS = 100
@@ -298,6 +298,7 @@ def generate_step(
     kv_group_size: int = 64,
     quantized_kv_start: int = 0,
     prompt_progress_callback: Optional[Callable[int, int]] = None,
+    input_embeddings: Optional[mx.array] = None,
 ) -> Generator[Tuple[mx.array, mx.array], None, None]:
     """
     A generator producing token ids based on the given prompt from the model.
@@ -322,14 +323,17 @@ def generate_step(
         kv_group_size (int): Group size for KV cache quantization. Default: ``64``.
         quantized_kv_start (int): Step to begin using a quantized KV cache.
            when ``kv_bits`` is non-None. Default: ``0``.
-        prompt_prorgress_callback (Callable[int, int]): A call-back which takes the
+        prompt_progress_callback (Callable[int, int]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
+       input_embeddings (mx.array, optional): Input embeddings to use in place of
+           prompt tokens, if provided. Default: ``None``.
 
     Yields:
         Tuple[mx.array, mx.array]: One token and a vector of log probabilities.
     """
+    if input_embeddings is not None and not does_model_support_input_embeddings(model):
+        raise ValueError("Model does not support input embeddings.")
 
-    y = prompt
     tokens = None
 
     # Create the KV cache for generation
@@ -352,15 +356,27 @@ def generate_step(
 
     sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
 
-    def _step(y):
+    def _step(token_ids=None, embeddings=None):
+        nonlocal tokens
+
         with mx.stream(generation_stream):
-            logits = model(y[None], cache=prompt_cache)
+            if embeddings is not None:
+                logits = model(
+                    None, cache=prompt_cache, input_embeddings=embeddings[None]
+                )
+            elif token_ids is not None:
+                logits = model(token_ids[None], cache=prompt_cache)
+            else:
+                raise ValueError(
+                    "Either token_ids or embeddings must be provided to _step."
+                )
+
             logits = logits[:, -1, :]
 
-            if logits_processors:
-                nonlocal tokens
-                tokens = mx.concat([tokens, y]) if tokens is not None else y
-
+            if logits_processors and token_ids is not None:
+                tokens = (
+                    mx.concat([tokens, token_ids]) if tokens is not None else token_ids
+                )
                 for processor in logits_processors:
                     logits = processor(tokens, logits)
 
@@ -370,11 +386,21 @@ def generate_step(
             y = sampler(logprobs)
             return y, logprobs.squeeze(0)
 
+    using_embeddings = input_embeddings is not None
+
+    y = input_embeddings if using_embeddings else prompt
     with mx.stream(generation_stream):
-        total_prompt_tokens = y.size
+        total_prompt_tokens = y.shape[0]
         prompt_processed_tokens = 0
-        while y.size > prefill_step_size:
-            model(y[:prefill_step_size][None], cache=prompt_cache)
+        while y.shape[0] > prefill_step_size:
+            if using_embeddings:
+                model(
+                    None,
+                    cache=prompt_cache,
+                    input_embeddings=y[:prefill_step_size][None],
+                )
+            else:
+                model(y[:prefill_step_size][None], cache=prompt_cache)
             quantize_cache_fn(prompt_cache)
             mx.eval([c.state for c in prompt_cache])
             prompt_progress_callback(prompt_processed_tokens, total_prompt_tokens)
@@ -382,13 +408,13 @@ def generate_step(
             y = y[prefill_step_size:]
             mx.clear_cache()
 
-        y, logprobs = _step(y)
+        y, logprobs = _step(embeddings=y) if using_embeddings else _step(token_ids=y)
 
     mx.async_eval(y, logprobs)
     n = 0
     while True:
         if n != max_tokens:
-            next_y, next_logprobs = _step(y)
+            next_y, next_logprobs = _step(token_ids=y)
             mx.async_eval(next_y, next_logprobs)
         if n == 0:
             mx.eval(y)
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -326,7 +326,7 @@ def generate_step(
         prompt_progress_callback (Callable[int, int]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
         input_embeddings (mx.array, optional): Input embeddings to use in place of
-           prompt tokens, if provided. Default: ``None``.
+           prompt tokens. Default: ``None``.
 
     Yields:
         Tuple[mx.array, mx.array]: One token and a vector of log probabilities.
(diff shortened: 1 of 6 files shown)


mlx_lm/generate.py
```diff
@@ -322,14 +323,17 @@ def generate_step(
         kv_group_size (int): Group size for KV cache quantization. Default: ``64``.
         quantized_kv_start (int): Step to begin using a quantized KV cache.
            when ``kv_bits`` is non-None. Default: ``0``.
-        prompt_prorgress_callback (Callable[int, int]): A call-back which takes the
+        prompt_progress_callback (Callable[int, int]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
+       input_embeddings (mx.array, optional): Input embeddings to use in place of
```
awni, line 328: (indentation is off)
```suggestion
        input_embeddings (mx.array, optional): Input embeddings to use in place of
```

mattjcly: Oops! Good catch. I guess black/pre-commit doesn't do docstrings

mlx_lm/models/qwen2.py
```diff
@@ -137,8 +137,12 @@ def __call__(
         inputs: mx.array,
         mask: mx.array = None,
         cache=None,
+        input_embeddings: Optional[mx.array] = None,
     ):
-        h = self.embed_tokens(inputs)
+        if input_embeddings is not None:
+            h = input_embeddings
+        else:
+            h = self.embed_tokens(inputs)
```
awni, line 145:
```suggestion
        h = input_embeddings or self.embed_tokens(inputs)
```

mlx_lm/models/gemma3_text.py
```diff
@@ -175,9 +175,12 @@ def __call__(
...
         mask: mx.array = None,
         cache=None,
+        input_embeddings: Optional[mx.array] = None,
     ):
-
-        h = self.embed_tokens(inputs)
+        if input_embeddings is not None:
+            h = input_embeddings
+        else:
+            h = self.embed_tokens(inputs)
```
awni, line 183: nit:

```suggestion
        h = input_embeddings or self.embed_tokens(inputs)
```

mlx_lm/models/qwen2.py
mattjcly: When I try this, I get:
```
  File "/Users/matt/Workspace/forks/mlx-lm/mlx_lm/models/qwen2.py", line 142, in __call__
    h = input_embeddings or self.embed_tokens(inputs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: [convert] Only length-1 arrays can be converted to Python scalars.
```
Full log:
```
/Users/matt/Workspace/forks/mlx-lm/.venv/bin/python /Users/matt/Applications/PyCharm.app/Contents/plugins/python-ce/helpers/pycharm/_jb_unittest_runner.py --target test_generate.TestGenerate.test_stream_generate_input_embeddings_prefill 
Testing started at 10:41 AM ...
Launching unittests with arguments python -m unittest test_generate.TestGenerate.test_stream_generate_input_embeddings_prefill in /Users/matt/Workspace/forks/mlx-lm/tests

Fetching 9 files: 100%|██████████| 9/9 [00:00<00:00, 48708.05it/s]

Error
  File "/Users/matt/Workspace/forks/mlx-lm/tests/test_generate.py", line 136, in test_stream_generate_input_embeddings_prefill
    for generation_result in stream_generate(
  File "/Users/matt/Workspace/forks/mlx-lm/mlx_lm/generate.py", line 662, in stream_generate
  File "/Users/matt/Workspace/forks/mlx-lm/mlx_lm/generate.py", line 651, in <genexpr>
    token_generator = (
                      ^
  File "/Users/matt/Workspace/forks/mlx-lm/mlx_lm/generate.py", line 397, in generate_step
    model(
  File "/Users/matt/Workspace/forks/mlx-lm/mlx_lm/models/qwen2.py", line 172, in __call__
    out = self.model(inputs, mask, cache, input_embeddings)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matt/Workspace/forks/mlx-lm/mlx_lm/models/qwen2.py", line 142, in __call__
    h = input_embeddings or self.embed_tokens(inputs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: [convert] Only length-1 arrays can be converted to Python scalars.



Ran 1 test in 0.396s

FAILED (errors=1)

Process finished with exit code 1
```

mattjcly: Very open to a way to get this to work though. I am using python 3.11.10:
```
matt@Matts-MacBook-Pro [10:42:26] [~/Workspace/forks/mlx-lm] [input-embeddings-gemma3 *]
-> % /Users/matt/Workspace/forks/mlx-lm/.venv/bin/python --version
Python 3.11.10
```

mlx_lm/models/gemma3_text.py
mattjcly: Same response as above :(

mattjcly: Maybe I'm missing something?

mlx_lm/generate.py
```diff
@@ -322,14 +323,17 @@ def generate_step(
...
+           prompt tokens, if provided. Default: ``None``.
 
     Yields:
         Tuple[mx.array, mx.array]: One token and a vector of log probabilities.
     """
+    if input_embeddings is not None and not does_model_support_input_embeddings(model):
+        raise ValueError("Model does not support input embeddings.")
```
awni, line 335: When you pass `input_embeddings` are you also passing the original prompt / what do you pass for the prompt in that case?

mattjcly: Good q: I was debating whether or not the original prompt should be made optional

I suppose it could either be made optional, or the expectation is to just pass an empty array if you don't want to do anything with logits processors. Optional seems like the more clear option to me from an "api describes behavior" perspective, but then that would change the call signatures of of `stream_generate`/`generate_step`

mlx_lm/generate.py
```diff
@@ -352,15 +356,27 @@ def generate_step(
...
 
-            if logits_processors:
-                nonlocal tokens
-                tokens = mx.concat([tokens, y]) if tokens is not None else y
+            logits = logits[:, -1, :]
 
+            if logits_processors and token_ids is not None:
+                tokens = (
+                    mx.concat([tokens, token_ids]) if tokens is not None else token_ids
+                )
```
awni, line 379: I wonder if it makes sense to refactor this a bit. Like maybe accept a token history into this function and do the book-keeping outside this function.

It's already a bit broken .. we don't properly include the token history since we do the prompt processing in steps outside this function.

It doesn't have to be in this PR .. just thinking that we'll want to fix / better specify this behavior, especially when using `input_embeddings`

mlx_lm/models/gemma3_text.py
awni: No that's my bad. Disregard.

mlx_lm/generate.py
awni: I think the downside with that is you will get different behavior when using input_embeddings vs using the tokens.

Another option is to always require the original token ids (and maybe do a check to see if their length matches the `input_embeddings`. What do you think about that? Is it awkward in your calling code?

mlx_lm/generate.py
```diff
@@ -352,15 +356,27 @@ def generate_step(
...
+            if embeddings is not None:
+                logits = model(
+                    None, cache=prompt_cache, input_embeddings=embeddings[None]
+                )
+            elif token_ids is not None:
+                logits = model(token_ids[None], cache=prompt_cache)
+            else:
+                raise ValueError(
+                    "Either token_ids or embeddings must be provided to _step."
+                )
```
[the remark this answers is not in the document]
mattjcly: Sounds good to me! Stylistically wasn't sure what aligned best with mlx-lm, I think this makes sense.

Not related to encapsulation, but rather the semantics of the implementation: sometimes using dims as an indirect indicator if tokens vs. embeddings can confuse me on first glance, especially since modifications like `y[None]` happen sometimes and add a dim. Not sure what you think about that, but totally fine with moving forward with your suggestion as-is

mlx_lm/generate.py
mattjcly: Unfortunately I think that would be awkward, since the original token in the case of an image may be something like "<image>" and then be expanded to, say, 500+ embeddings worth of tokens.

Maybe for now we just do "one or the other"? And can either enforce that be requiring an empty array for the prompt tokens (doesn't change function signatures), or can switch prompt tokens to an optional. What do you think? I'd be inclined to start with just requiring an empty array for the prompt tokens, and can refactor to optional in a follow-up? but good with either

mlx_lm/generate.py
mattjcly: I would agree - on the surface I think I like the idea of accepting a token history into the function

mattjcly: Would be inclined to make that a separate pr like you said, since there are potential further api-change implications

mlx_lm/generate.py
```diff
@@ -352,15 +361,22 @@ def generate_step(
 
     sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
 
+    def _model_call(y):
+        if y.ndim == 3:
+            return model(None, cache=prompt_cache, input_embeddings=y)
+        else:
+            return model(y, cache=prompt_cache)
```
awni, line 368: I think the gain in the rest of the code simplicity outweighs the cost of using `ndim` here. I agree it's not ideal though as it's a little brittle... thanks for making that change!

awni (approved):
Looks awesome! Will merge after the tests clear!

PR #180, opened 2025-05-16
fix upload to hub usage

IANTHEREAL:
Refer the implementation, fix the incorrect usage in fuse and merge

```
def upload_to_hub(path: str, upload_repo: str):
    """
    Uploads the model to Hugging Face hub.

    Args:
        path (str): Local path to the model.
        upload_repo (str): Name of the HF repo to upload to.
    """
```


awni (approved):
Good catch, thanks for the fix!

PR #186, opened 2025-05-18
fix server.py null tool content (#185)

### LINKED ISSUE #185: mlx-lm.server gets exception with "content":null in tool call messages
While testing tool calling from LangChain, I found that when I send the conversation messages back to the LLM (running on mlx-lm.server backend), there's an exception caused by the "content: null" JSON that comes over in the tool call message.  I extracted this into a curl call:

`curl -H 'Host: localhost:8080' -H 'Accept: application/json' -H 'Content-Type: application/json' -X POST http://localhost:8080/v1/chat/completions -d '{"messages":[{"content":"what'"'"'s 2+3?","role":"user"},{"content":null,"role":"assistant","tool_calls":[{"type":"function","id":"79f24cae-0234-4f2a-83a5-bc8ab91f2c61","function":{"name":"add","arguments":"{\"a\": 2, \"b\": 3}"}}]},{"content":"5","role":"tool","tool_call_id":"79f24cae-0234-4f2a-83a5-bc8ab91f2c61"}],"model":"mlx-community/Qwen3-30B-A3B-4bit","max_completion_tokens":40960,"stream":true,"temperature":0.6,"tools":[{"type":"function","function":{"name":"add","description":"Adds a and b.","parameters":{"properties":{"a":{"type":"integer"},"b":{"type":"integer"}},"required":["a","b"],"type":"object"}}}],"top_p":0.95}'`

The message that causes trouble is the second one (prettified):

```
{
  "content": null,
  "role": "assistant",
  "tool_calls": [
    {
      "type": "function",
      "id": "79f24cae-0234-4f2a-83a5-bc8ab91f2c61",
      "function": {
        "name": "add",
        "arguments": "{\"a\": 2, \"b\": 3}"
      }
    }
  ]
}
```

The portion of the stack trace belonging to `server.py` and below is:

The portion of the stack trace in the `server.py` and below is:

```
File "/Users/jpangburn/Documents/vsc_workspaces/mlx-lm/mlx_lm/server.py", line 767, in handle_chat_completions
    prompt = self.tokenizer.apply_chat_template(
        messages,
    ...<2 lines>...
        **self.model_provider.cli_args.chat_template_args,
    )
  File "/Users/jpangburn/python_envs/langchainnew/lib/python3.13/site-packages/transformers/tokenization_utils_base.py", line 1695, in apply_chat_template
    rendered_chat = compiled_template.render(
        messages=chat,
    ...<3 lines>...
        **template_kwargs,
    )
  File "/Users/jpangburn/python_envs/langchainnew/lib/python3.13/site-packages/jinja2/environment.py", line 1295, in render
    self.environment.handle_exception()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/jpangburn/python_envs/langchainnew/lib/python3.13/site-packages/jinja2/environment.py", line 942, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "<template>", line 34, in top-level template code
TypeError: argument of type 'NoneType' is not iterable
```

Ultimately what happens is the template expects content to have a string value, not None.  I also tried this with a Gemma3 model to see if it was just Qwen3 "mlx-community/gemma-3-4b-it-qat-4bit", but it got a similar error (didn't complain about iterable, but didn't like None).

I'm not sure if this is the correct way to solve this, but in the server.py I replaced the line `self.body = json.loads(raw_body.decode())` with `self.body = json.loads(raw_body.decode(), object_hook=lambda d: {**d, "content": d.get("content") or ""})` and now it doesn't have the problem.

I'll fork and create a branch with this fix.  If this is the correct fix, let me know and I'll create a pull request.

jpangburn:
Avoid exception with null content in tool messages.

Closes #185

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -295,7 +295,10 @@ class APIHandler(BaseHTTPRequestHandler):
         # Fetch and parse request body
         content_length = int(self.headers["Content-Length"])
         raw_body = self.rfile.read(content_length)
-        self.body = json.loads(raw_body.decode())
+        self.body = json.loads(
+            raw_body.decode(),
+            object_hook=lambda d: {**d, "content": d.get("content") or ""},
+        )
         indent = "\t"  # Backslashes can't be inside of f-strings
         logging.debug(f"Incoming Request Body: {json.dumps(self.body, indent=indent)}")
         assert isinstance(
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -130,6 +130,38 @@ class TestServer(unittest.TestCase):
         self.assertIn("id", response_body)
         self.assertIn("choices", response_body)
 
+    def test_handle_chat_completions_with_null_tool_content(self):
+        url = f"http://localhost:{self.port}/v1/chat/completions"
+        chat_post_data = {
+            "model": "chat_model",
+            "max_tokens": 10,
+            "temperature": 0.7,
+            "top_p": 0.85,
+            "repetition_penalty": 1.2,
+            "messages": [
+                {"role": "user", "content": "what is 2+3?"},
+                {
+                    "role": "assistant",
+                    "content": None,
+                    "tool_calls": [
+                        {
+                            "type": "function",
+                            "id": "123",
+                            "function": {
+                                "name": "add",
+                                "arguments": '{"a": 2, "b": 3}',
+                            },
+                        }
+                    ],
+                },
+                {"role": "tool", "content": "5", "tool_call_id": "123"},
+            ],
+        }
+        response = requests.post(url, json=chat_post_data)
+        response_body = response.text
+        self.assertIn("id", response_body)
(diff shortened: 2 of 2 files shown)


awni:
Thanks for the fix! I left a comment about it, could you address then we can merge?

Also fwiw it looks like Qwen3 models just updated their chat template recently and now they work with content as `None`..

jpangburn:
@awni thanks for the comments and suggestion! Good to know about Qwen3, but probably still worth integrating this fix at this point since other popular models fail with content `None`?

awni:
>  but probably still worth integrating this fix at this point since other popular models fail with content None

Agreed!

awni (approved):
Thanks!!

jpangburn:
Thank you guys for this great project!  Glad to help a tiny bit!

PR #187, opened 2025-05-19
Faster top-p and min-p sampling

awni:
```
mlx_lm.generate --model mlx-community/Qwen3-1.7B-4bit --prompt "Write a story about Einstein" -m 512 --top-p 0.95 --temp 0.6 --seed 1
```

Pre: 283.09 tok/s
Post: 291.67 tok/s

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -184,7 +184,6 @@ def apply_min_p(
     selected_logprobs = mx.where(tokens_to_remove, -float("inf"), sorted_logprobs)
 
     # Create a mapping to rearrange back to original indices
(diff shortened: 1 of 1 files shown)


mlx_lm/sample_utils.py
```diff
@@ -185,7 +185,12 @@ def apply_min_p(
 
     # Create a mapping to rearrange back to original indices
     # Use argsort of sorted_indices to get the inverse permutation
```
angeloskath, line 187: Nit: No longer true.

angeloskath (approved):
👍

mlx_lm/sample_utils.py
awni: Of course 😆

PR #188, opened 2025-05-19
Fix URL for LoRA Data in Error Message

rgeorgi:
Updating the error for incorrect LoRA training data to point to the correct page instead of 404.


awni (approved):
Thanks!!

PR #202, opened 2025-05-29
Dynamic quants

awni:
Adds and `mlx_lm.dynamic_quant` for computing better mixed precision quantizations (open to better names if anyone has suggestsions).

In my benchmarks so far this usually works better.. sometimes by a lot than the default q4_m setting (which is what you get when you do `--quant-predicate mixed_4_6` with `mlx_lm.convert`). 

Will share some PPL numbers.

Qwen3-0.6B-base

Quant   |  BPW | Perplexity
--------- | ----- | -----
bf16 | 16.0 | 9.97
Q4, 32 | 5.0 | 11.92
Q4, 64 | 4.5 | 12.58
Mixed (Q4_K_M) | 4.70 | 11.86
Dynamic | 4.70 | 11.71
Dynamic | 5.0 | 11.25
DWQ | 4.5 | 11.24

Gemma3 1B PT

Quant   |  BPW | Perplexity
--------- | ----- | -----
bf16 | 16.0 | 22.558
Q4, 32 | 4.5 | 25.91
Q4, 64 | 5.0 | 26.717
Mixed (Q4_K_M) | 5.13 | 24.68
Dynamic | 5.0 | 24.533
Dynamic | 5.13 | 24.44

Qwen 3 8B

Quant | BPW | Perplexity
---- | ---- | ----
bf16 | 16.0 | 5.942
Q4, 32 | 5.0 | 6.920	
Q4, 64 | 4.5 | 7.366
Mixed (Q4_K_M) | 4.89 | 6.385
Dynamic | 5.0 | 6.206
Dynamic | 4.89 | 6.24

### DIFF
--- a/mlx_lm/quant/dynamic_quant.py
+++ b/mlx_lm/quant/dynamic_quant.py
@@ -218,7 +218,7 @@ def main():
             return {"bits": args.high_bits, "group_size": args.high_group_size}
         return True
 
-    weights, config = quantize_model(
+    model, config = quantize_model(
         model,
         config,
         q_group_size=args.low_group_size,
@@ -233,7 +233,7 @@ def main():
     save(
         args.mlx_path,
         model_path,
-        weights,
+        model,
         tokenizer,
         config,
         hf_repo=args.model,


awni:
Ok I think this is ready for review! The results are looking pretty solid and it's relatively efficient to compute. It gives a really nice bump over homogenous quants even at the same bpw and seems to consistently work better than the MLX mixed quants (which use the same settings as LCP Q4_K_M).

@angeloskath let me know if you have any thoughts on this one!

I set the default parameters to use q4/q5 so should wait till https://github.com/ml-explore/mlx/pull/2226 lands.

angeloskath:
I am playing with it a bit, it is very cool, will stamp it in a bit :-)

I am testing to see if we gain anything by doing it in steps as in compute sensitivities, then select the most sensitive x% of weights to quantize high. Then recompute sensitivities with these quantized and so on and so forth. If set to 1 weight it will be very slow but probably better.

mzbac:
This looks awesome and seems directly compatible with mlx-lm and the swift-mlx-example's MLXLLM.

vlbosch:
This is very exciting! Is this by any chance the MLX implementation of [Super Weights in Large Language Models](https://arxiv.org/html/2411.07191v1) I noticed this paper was written "while interning at Apple" — and this implementation feels like a natural extension of that work. 😊

I was wondering: does the sensitivity analysis always require the full model weights (in bf16/fp16) to be in memory? Or would it be possible to perform this analysis in a layer-wise fashion instead? This would significantly reduce the memory requirements and allow dynamic quantization of very large models (e.g. Qwen3-235B) on systems with "limited" unified memory (like an M3 Max with 128GB), not needing a M3 Ultra with 512GB.

I understand this would likely take (much) longer to run, but it would allow many more users to take advantage of this powerful new quantization method.

Thanks again for your great work and for continuing to improve MLX quantization support!

awni:
>  Is this by any chance the MLX implementation of Super Weights in Large Language Models I noticed this paper was written "while interning at Apple" 

It's unrelated to that work. The method here is quite different. 

> I was wondering: does the sensitivity analysis always require the full model weights (in bf16/fp16) to be in memory? Or would it be possible to perform this analysis in a layer-wise fashion instead?

The targets have to come from the high precision model.. though it's probably ok for that model to be 8-bit instead of 16-bit. There's also other tricks we could play to reduce the working set size.. like precomputing the targets so you don't need to have two models in RAM at the same time.

vlbosch:
> > Is this by any chance the MLX implementation of Super Weights in Large Language Models I noticed this paper was written "while interning at Apple"
> 
> It's unrelated to that work. The method here is quite different.
> 

Thanks for the clarification! Could you briefly explain how this implementation differs from the "Super Weights" paper? Is it that their approach focuses on individual "super weights" while this dynamic quantization method operates at the layer level?

> > I was wondering: does the sensitivity analysis always require the full model weights (in bf16/fp16) to be in memory? Or would it be possible to perform this analysis in a layer-wise fashion instead?
> 
> The targets have to come from the high precision model.. though it's probably ok for that model to be 8-bit instead of 16-bit. There's also other tricks we could play to reduce the working set size.. like precomputing the targets so you don't need to have two models in RAM at the same time.

Precomputing the targets to avoid having two models (original and quantized) in memory at the same time makes sense! But I thought such a function was already incorporated in this PR, with the "--sensitivities" option?

I was thinking — could we take this a step further by also computing those targets in a layer-wise fashion, rather than requiring the full model in memory? Running a forward pass through the model, saving activations per layer and so on, to calculate the sensitivities without the complete (8- or 16-bit) model in memory. This would allow the complete sensitivity analysis on systems with limited memory, and could also be useful for people using custom 'calibration sets' (e.g., language-specific datasets instead of English-only data).

Is this something that could be integrated into the current workflow, or would it require a mayor rewrite?

awni:
The targets are the final output of the high-precision model. There isn't a layer-wise component there. It might be possible to split the quantization of the model up into sections and use layer wise targets instead.. but that would require some work to explore.

angeloskath:
Pushed the sensitivity change, feel free to merge at will.

| Model  | Quant | BPW | Perplexity |
| ------------- | ------------- | --- | --- |
| Qwen 3 0.6 Base  | bf16 | 16 | 9.97  |
| Qwen 3 0.6 Base  | dynamic | 5.01 | 11.099  |
| Gemma 3 1B PT  | bf16 | 16 | 22.558  |
| Gemma 3 1B PT  | dynamic | 4.88 | 24.305  |

awni:
That's awesome!. The Gemma 3 1B is both smaller and better than the previous 5.0 dynamic. Though I wonder why the actual bpw isn't closer to the target 🤔 (assuming you were using the default 5.0 taget?).

angeloskath:
Yeah I think the next weight to be quantized is the embedding matrix, which is huge and will push this to 5.3 or sth... I even put 5.1 as the target and got the same.

The obvious next update here is to change this to keep selecting until we can't select the best one and then see if we can select the next and so on. But I think it makes sense to see if a more complicated search across several compression levels is possible with reasonable complexity. For instance we can trade-off negative impact to positive impact to gain some bits as well, eg there may be some weight which when moved to 3 or 2 bits has smaller impact to the loss than moving a weight to 5 bits or 6 bits 🤷‍♂️

angeloskath:
As I was writing the above, the 5.2 target finished and we get

```
Original PPL: 22.558
[INFO] Quantized model with 5.181 bits per weight.
Quantized PPL: 23.788
```

PR #207, opened 2025-06-02
Add total params to metadata + cleanup

awni:
Adds `total_parameters` to `model.safetensors.index.json` so HF can correctly parse quantized model sizes.

- Cleaned up how we save model weights to simplify just passing a model
- Remove `merge` command. I'm 99% sure no-one used it and it's a maintenance burden.. but I can add it back if I'm wrong about that

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -33,7 +33,7 @@ if os.getenv("MLXLM_USE_MODELSCOPE", "False").lower() == "true":
 else:
     from huggingface_hub import snapshot_download
 
-from mlx.utils import tree_flatten, tree_reduce
+from mlx.utils import tree_flatten, tree_map, tree_reduce
 from transformers import PreTrainedTokenizer
 
 # Local imports
@@ -393,7 +393,7 @@ def save_model(
         "weight_map": {},
     }
     if donate_model:
-        del model
+        model.update(tree_map(lambda _: mx.array([]), model.parameters()))
 
     # Write the weights and make sure no references are kept other than the
     # necessary ones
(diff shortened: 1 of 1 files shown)


mlx_lm/utils.py
```diff
@@ -503,11 +507,11 @@ def save(
     tokenizer: TokenizerWrapper,
     config: Dict[str, Any],
     hf_repo: Optional[str] = None,
-    donate_weights: bool = True,
+    donate_model: bool = True,
```
angeloskath, line 510: The weights argument should probably be changed to `model` as well.

mlx_lm/utils.py
```diff
@@ -387,13 +385,20 @@ def save_weights(
...
-    index_data = {"metadata": {"total_size": total_size}, "weight_map": {}}
+    index_data = {
+        "metadata": {
+            "total_size": total_size,
+            "total_parameters": get_total_parameters(model),
+        },
+        "weight_map": {},
+    }
+    if donate_model:
+        del model
```
angeloskath: Unfortunately this has to be a bit more complicated now. Something like

```python
empty_weights = tree_map(mx.zeros_like, weights)
model.update(empty_weights)
del weights
del model
```

mlx_lm/utils.py
awni: Oops.. good catch.

mlx_lm/utils.py
awni: Why so?

awni: Nvm.. I get it.

PR #208, opened 2025-06-03
some nits in lora

awni:
- Fix fusing with dequantization to not do an extra quantize / dequantize in fuse
- Add some defaults to the config
- Use higher precision in lora fuse


angeloskath (approved):
🙏

PR #209, opened 2025-06-04
adds support for file prompt caching

prncvrm:
This PR helps users to read a txt, or any fileformat to be parsed as prompt to cache
Would be useful for larger prompts or CI pipelines to create new cached models
@awni FYI


awni:
This isn't necessary. You can just pipe the prompt in like so:

```
cat file.txt | mlx_lm.cache_prompt --prompt "-"
```

PR #210, opened 2025-06-04
reset cache

prncvrm:
- In cases of Multi-Turn, when loading the model via cache, the cache gets updated with each iteration
- As a user, i might have a need to reset back the cache to original state, before i generate another response from the already loaded memory, or cleanoff after the response has been generated
- can be called in following fashion
```
def reset_cache(prompt_cache):
    for prompt in prompt_cache:
        prompt.reset_to_original_cache()
````        

@awni let me know your thoughts on the same :-)

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -217,8 +217,13 @@ class KVCache(_BaseCache):
         self.values = None
         self.offset = 0
         self.step = 256
+        self._initial_state = None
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by awni** [comment]
You should need to do something like this.

You can just do a deep copy of the cache using `copy.deepcopy`:

```python
initial_cache = copy..deepcopy(cache)
generate(..., prompt_cache=cache)
# Reset to initial state if needed
cache = initial_cache
```

PR #217, opened 2025-06-07
Enable tool use in the server and add an example using openai client

awni:
Adds some methods to the tokenizer to determine properties of a model / tokenizer:
- If it supports tool use and the tool start and end tokens
- If it supports thinking and the think start and end tokens

It's a bit hacky right now how it figures this out, but that part is easy to fix if upstream packages provide the information in a more usable way.

Also updated mlx-lm.server to properly return tool calls in compliance with the openai spec. And added an example showing how to use that with the `openai` client.


mlx_lm/tokenizer_utils.py
```diff
@@ -267,6 +267,30 @@ def __init__(
             if eos_token_ids is not None
             else {tokenizer.eos_token_id}
         )
+        self._think_start = None
+        self._think_end = None
+        self._tool_call_start = None
+        self._tool_call_end = None
+
+        THINK_TOKENS = [("<think>", "</think>")]
+        TOOL_CALL_TOKENS = [("<tool_call>", "</tool_call>")]
```
pcuenca, line 276: Interesting! Do you have any thoughts on how we can make this better?

awni: Yea a couple thoughts:
- First off, I think ever LLM vendor should make sure tools are explicitly demarcated in the output. Some LLM (e.g. llama3) don't put the tools in-between special tokens. And then the burden is on the user to figure out if it's a tool call or not which is quite messy.
- Whether or not a model supports tool use (and thinking) should be part of the config or tokenizer config. And the tool start and tool end and think start and think end ideally would be methods on the tokenizer itself and/or part of the config.


I think the above two changes while relatively minor would make a world of difference for having a clean implementation of tool use for many different models..

Rocketknight1: Yes, for sure! I work on chat templates at Hugging Face and this is something we'd like as well; right now, the range of formats is very broad, and we do something basically like what you're doing here, we look for references to tools in the chat template to decide if the model supports tools or not. It's obviously hacky but there isn't really anything better!

awni: Yea for this kind of thing I think it has to be put in the config by model providers. But Hugging Face has the power to make that a standard!

PR #183, opened 2025-05-17
Fix server cache

awni:
Avoid trimming the entire prompt, always leave at least one token.

Closes #182

PR #221, opened 2025-06-09
change the method of Leave at least one token

trillionmonster:
change the method of Leave at least one token 

[related](https://github.com/ml-explore/mlx-lm/pull/183#issuecomment-2954424694)


trillionmonster:
This issue is really urgent — my mom can always see the last thing I said in the chat. Every time I run pip install, I have to manually go into the source code and change it again. It's so frustrating.
@awni

awni:
Can you provide a way to reproduce the issue so we can investigate it?

awni:
Closing as inactive.

PR #222, opened 2025-06-09
Remove sentencepiece dependencies

mitmul:
Temporary fix while waiting for the new version release of sentencepiece that supports Python 3.13 on macOS environment.


awni:
Hmm.. but don't we need sentencepiece for some models?

mitmul:
Oops, I accidentally opened a PR to the upstream repository. This was originally meant to be a temporary branch for internal use only. My apologies!

PR #225, opened 2025-06-09
README: Declare Conda as preferred way to install the package

pasky:
Issues with pip include no support for 3.13 (sentencepieces issue), possible issues in legacy setups with intel vs. arm64 (it's an issue of the setups but rarely met outside of packages like mlx), possibly scipy-related issues.

Conda just works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,18 +13,17 @@ Some key features include:
   with support for quantized models.
 * Distributed inference and fine-tuning with `mx.distributed`
 
-The easiest way to get started is to install the `mlx-lm` package:
-
-**With `pip`**:
-
-```sh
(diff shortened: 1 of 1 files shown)


awni:
Support for 3.13 is orthogonal to pip vs conda. The fix for that is coming from a downstream dependency.. and should be available soon (https://github.com/google/sentencepiece/pull/1084) 🤞

pasky:
(I'll just note that the linked issue is open since... Feb 6. *shrug*)

awni:
Ok, let's give the maintainers another few days. If still no movement we'll make sentencepiece an optional dependency

Most models don't need it so it should be fine.

PR #228, opened 2025-06-11
Identifying layers to convert to LoRA by pattern or indicating all linear layers

chimezie:
Adds support for matching layers to convert to LoRa by REGEX patterns:

```yaml
lora_parameters:
  key_patterns: [".+gate_proj.*", ".+down_proj.+", ".+up_proj.+", ".+(q|v|k|o)_proj.+"]
```

Or by specifying "all" in any of the keys to indicate converting *all* linear layers to LoRa:

```yaml
lora_parameters:
  keys: ["all"]
```

In either case, the keys of the converted layers are also printed for reference.


ivanfioravanti:
Nice one!!!

mlx_lm/tuner/utils.py
```diff
@@ -34,6 +35,44 @@ def build_schedule(schedule_config: Dict):
...
+
+    Returns:
+        bool: A boolean indicating whether the module should be converted to LoRA
+    """
+    if key_patterns is None:
+        key_patterns = set()
+    should_convert = (
+        (
+            all_linear_layers
+            and isinstance(module, (nn.QuantizedLinear, nn.Linear, LoRASwitchLinear))
```
angeloskath: That check should be `..., SwitchLinear, QuantizedSwitchLinear)` instead.

I would write that tuple outside of the condition and change the order a bit because this way the condition can select a layer that isn't a linear layer.

```python
convertible_layers = (nn.QuantizedLinear, ...., QuantizedSwitchLinear)
is_convertible = isinstance(module, convertible_layers) or hasattr(module, "to_lora")
should_convert = (
    all_linear_layers
    or layer_key in keys
    or any(p.match(layer_key) for p in key_patterns)
)
...
return is_convertible and should_convert
```

mlx_lm/tuner/utils.py
```diff
@@ -167,13 +207,29 @@ def to_lora(layer):
         keys = set(["attn.attention.q_proj", "attn.attention.v_proj"])
     else:
         raise ValueError(f"Lora does not support {model.model_type}")
-
+    all_linear_layers = "all" in keys
+    if all_linear_layers:
+        print("Applying LoRA to all linear layers")
```
angeloskath, line 277: I would add

```python
should_convert = partial(
    should_convert_to_lora,
    keys=keys,
    all_linear_layers=all_linear_layers,
    key_patterns=key_patterns,
)
```

which then makes the list comprehensions way more readable eg

```python
lora_layers = [(k, to_lora(m)) for k, m in l.named_modules() if should_convert(k, m)]
```

angeloskath (changes requested):
This is nice! I left a few beautification comments and one bug but after that we can merge.

mlx_lm/tuner/utils.py
chimezie: Thank you.  I have incorporated those changes to `should_convert`

mlx_lm/tuner/utils.py
chimezie: @angeloskath  Thanks, it does make it much more readable.  I've also changed this

zcbenz:

PR #232, opened 2025-06-12
Implementation of AFM in MLX

angeloskath:
This adds an implementation of the released AFM loosely ported from [tamm](https://pypi.org/project/tamm/). People can read more regarding fetching the weights at https://developer.apple.com/apple-intelligence/foundation-models-adapter/ .

I have 3 scripts that take the PT weights and map them to MLX and vice-versa which I think I 'll PR in mlx-examples perhaps with a training walkthrough later this week.

The things to consider in this PR are:
- The `ConcatenateKVCache`, I use this basically when the user hasn't provided a KV cache as a convenient way to extract the keys and values from the last layer.
- The `NewlineTokenizer`. The AFM has this quirk where `<n>` maps to `\n` when decoding and vice-versa. I could massage the HF tokenizer to do the mapping during decode but no matter how much I tried I couldn't convince it to do the mapping during encoding without subclassing. If people think this is too much clutter I can try to find another solution.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -345,7 +345,7 @@ def generate_step(
             model,
             max_kv_size=max_kv_size,
         )
-    elif len(prompt_cache) != len(model.layers):
+    elif not hasattr(model, "make_cache") and len(prompt_cache) != len(model.layers):
         raise ValueError("Wrong number of layers in the prompt cache.")
 
     prompt_progress_callback = prompt_progress_callback or (lambda *_: None)
--- a/mlx_lm/models/afm7.py
+++ b/mlx_lm/models/afm7.py
@@ -10,7 +10,7 @@ import mlx.core as mx
 import mlx.nn as nn
 
 from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
-from .cache import ConcatenateKVCache
+from .cache import ConcatenateKVCache, KVCache
 from .rope_utils import initialize_rope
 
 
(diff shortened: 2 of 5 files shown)


mlx_lm/generate.py
```diff
@@ -345,7 +345,7 @@ def generate_step(
             model,
             max_kv_size=max_kv_size,
         )
-    elif len(prompt_cache) != len(model.layers):
+    elif not hasattr(model, "make_cache") and len(prompt_cache) != len(model.layers):
         raise ValueError("Wrong number of layers in the prompt cache.")
```
awni: I'm ok to just get rid of this check tbh. I don't think it's ever done anything useful.

awni: Also there is a similar check in the `speculative_generate_step` that needs to be updated or removed.

mlx_lm/models/afm7.py
```diff
@@ -0,0 +1,397 @@
...
+        cache=None,
+    ):
+        h = self.embedding(inputs)
+
+        if mask is None:
+            mask = create_attention_mask(h, cache)
+
+        if cache is None:
+            cache = [None] * len(self.layers)
+            cache[-1] = ConcatenateKVCache()
```
awni, line 363: That's a clever solution.

mlx_lm/tuner/utils.py
```diff
@@ -54,6 +54,13 @@ def linear_to_lora_layers(
     """
 
     def to_lora(layer):
+        if hasattr(layer, "to_lora"):
+            return layer.to_lora(
```
awni: Maybe worth asserting `not use_dora` or smthn.. just to not silently do something unexpected..

mlx_lm/models/afm7.py
```diff
@@ -0,0 +1,397 @@
+# Copyright © 2023-2024 Apple Inc.
```
awni: nit
```suggestion
# Copyright © 2025 Apple Inc.
```

awni (approved):
Awesome!!

PR #240, opened 2025-06-17
More dynamic quant

angeloskath:
Not sure if we want to merge this...

It does the following:
- Calculate the approximate improvement for a set of possible quants together with the extra bits required
- Find a set of quants that result in total extra bits less than the available budget such that the improvement is maximized

The last part is solved via a slightly weird 0-1 knapsack where we can only select one quant per layer.

First the results:

| Model | Original | Quantized Before | BPW Before | Quantized now | BPW now |
| --- | --- | --- | --- | --- | --- |
| Qwen/Qwen3-0.6B | 15.023 | 16.281 | 5.011 | 16.206 | 5.000 |
| Qwen/Qwen3-1.7B | 11.349 | 11.984 | 4.905 | 11.524 | 5.000 |
| Qwen/Qwen3-4B | 9.682 | 9.892 | 5.002 | 9.659 | 5.000 |
| Qwen/Qwen3-8B | 7.830 | 7.995 | 4.971 | 7.999 | 5.000 |
| google/gemma-3-1b-it | 36.221 | 36.747 | 5.100 | 35.504* | 5.000 |
| google/gemma-3-4b-it | 34.064 | 29.843 | 4.916 | 25.459 | 5.000 |

In all cases except the gemma3 1b marked with star above I put the default parameters which searches from 4.25 bits to 6 bits per weight. In that case I put the min bpw to 4.5 cause it worked way better.

There is also the elephant in the room that the quantized model shouldn't achieve better perplexity than the original but given that we are training on it I guess it is maybe ok.

Now the reason to not merge it would be that this search is maybe a bit more prone to errors or weird results. Perhaps I should make it possible to provide the options explicitly so the search isn't as slow either (it is still way way faster than calculating gradients or perplexities).

### DIFF
--- a/mlx_lm/models/switch_layers.py
+++ b/mlx_lm/models/switch_layers.py
@@ -53,12 +53,6 @@ class QuantizedSwitchLinear(nn.Module):
         # Freeze this model's parameters
         self.freeze()
 
-    def unfreeze(self, *args, **kwargs):
-        """Wrap unfreeze so that we unfreeze any layers we might contain but
-        our parameters will remain frozen."""
-        super().unfreeze(*args, **kwargs)
-        self.freeze(recurse=False)
-
     @property
     def input_dims(self):
         return self.scales.shape[2] * self.group_size
--- a/mlx_lm/quant/dwq.py
+++ b/mlx_lm/quant/dwq.py
@@ -13,7 +13,8 @@ from mlx.utils import tree_flatten, tree_map
 from tqdm import tqdm
 
 from mlx_lm.tuner.datasets import load_dataset
-from mlx_lm.tuner.trainer import iterate_batches
+from mlx_lm.tuner.losses import kl_div_loss
+from mlx_lm.tuner.trainer import grad_checkpoint, iterate_batches
 from mlx_lm.tuner.utils import print_trainable_parameters
 from mlx_lm.utils import (
     fetch_from_hub,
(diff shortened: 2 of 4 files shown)


awni:
> There is also the elephant in the room that the quantized model shouldn't achieve better perplexity than the original but given that we are training on it I guess it is maybe ok.

I've seen these kinds of inversions before for instruction tuned models (which is why I switched  to use the base model when measuring perplexity as I think it's a more reliable metric there).  For the instruct models I think we'd be better off looking at KL to the original instead of perplexity, maybe I should add that as the metric instead?

But gosh those Gemma perplexities are especially unusual..

awni:
For the base 1B gemma model I'm getting nice results:

22.777 new at 5.0 bpw
24.305 old at 4.9 bpw

However, it seems to be critical to set the 4-bit group size to 64. At 128 it is worse than the old dynamic. Any idea about why that is? And is there any reason not to just default it to 64 (other than a little extra flexibility in the search)?

angeloskath:
Yeah I will change that so we can give specific bpw options to consider. The reason why it fails is because setting it to 128 means we are starting everything from 4.25 bpw and then they have to move to 4.5 or 5 or 5.25 or 5.5 or 6  (default parameters). So basically the gradient score underestimates the perf boost of going from 4.25 to 4.5 and overestimates going from 4.25 to 5 or more and as a result we get a worse model.

I don't know if that is hand-wavy but another way of seeing that is that the further away the starting point from the possible options the less relevant the linear approximation is. For instance if our budget was 4.5 bpw then it would be better.

PR #253, opened 2025-06-24
Fix `IndexError` in `NaiveStreamingDetokenizer`

Muhtasham:
## Description
Fixed an IndexError that occurs in the `NaiveStreamingDetokenizer` class's `text` property when trying to access the last character of an empty string. Added a length check before accessing `self._current_text[-1]`.

## Problem
When the tokenizer tries to clean up tokenization spaces, it attempts to check if the current text ends with a space without first verifying that there is any text to check. This causes an IndexError when `self._current_text` is empty.

## Solution
Added a length check `len(self._current_text) > 0` before attempting to access the last character.


Muhtasham:
cc @awni

PR #265, opened 2025-07-01
Allow converting local models

awni:
Before it was broken to do:

```bash
mlx_lm.convert --hf-path my/local/path -q
```


angeloskath (approved):
🙏

byteprobe:
made a local fix for converting local models, but your update is a more robust solution. thanks! 🙏

PR #266, opened 2025-07-01
Allow prompt and input_embeddings

will-lms:
The [mlx-engine](https://github.com/lmstudio-ai/mlx-engine) [unified architecture](https://lmstudio.ai/blog/unified-mlx-engine) uses mlx-vlm to compute token and image embeddings, then passes those to mlx-lm for generation. Currently, "unified" models only pass those embeddings and not the input token IDs.

The Gemma3n implementation requires token IDs even when passing input embeddings directly, as they are needed to compute the Per Layer Embeddings.

This PR modifies the `generate_step` function to allow for just `prompt`, just `input_embeddings`, or both as input. These are then piped through to the underlying model calls. None of this is specific to Gemma3n, so it can be re-used by other models that require the same inputs.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -329,16 +329,18 @@ def generate_step(
            when ``kv_bits`` is non-None. Default: ``0``.
         prompt_progress_callback (Callable[int, int]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
-        input_embeddings (mx.array, optional): Input embeddings to use in place of
-           or in conjunction with prompt tokens. Default: ``None``.
+        input_embeddings (mx.array, optional): Input embeddings to use in conjunction
+           with prompt tokens. Default: ``None``.
 
     Yields:
         Tuple[mx.array, mx.array]: One token and a vector of log probabilities.
     """
+    if len(prompt) == 0:
+        raise ValueError("Prompt must be non-empty.")
     if input_embeddings is not None:
         if not does_model_support_input_embeddings(model):
             raise ValueError("Model does not support input embeddings.")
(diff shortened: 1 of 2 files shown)


mlx_lm/generate.py
```diff
@@ -363,47 +363,71 @@ def generate_step(
 
     sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
 
-    def _model_call(y):
-        if y.ndim == 3:
-            return model(None, cache=prompt_cache, input_embeddings=y)
+    def _model_call(input_tokens: mx.array, input_embeddings: mx.array | None):
```
awni: The `|` is Python 3.10+. So maybe better to drop the type hints here (or use a `Optional` if you want type hints)

will-lms:
Pushed some revision with:
1. Rebase
2. Address comment by using `Optional`
3. Address offline conversation by requiring prompt be non-empty.

mlx_lm/generate.py
```diff
@@ -363,47 +365,67 @@ def generate_step(
...
-    y = input_embeddings if using_embeddings else prompt
     with mx.stream(generation_stream):
-        total_prompt_tokens = y.shape[0]
+        total_prompt_tokens = prompt.shape[0]
         prompt_processed_tokens = 0
-        while y.shape[0] > prefill_step_size:
-            _model_call(y[:prefill_step_size][None])
+        while total_prompt_tokens - prompt_processed_tokens > prefill_step_size:
+            _model_call(
+                input_tokens=(prompt[:prefill_step_size][None]),
```
awni:
```suggestion
                input_tokens=prompt[:prefill_step_size][None],
```

PR #219, opened 2025-06-08
Add bitnet1.58 with custom metal kernel

Blaizzy:

PR #268, opened 2025-07-03
Feat: add falcon-e support for bitnet models

younesbelkada:
Follow up PR to: #219 - this is an alternative approach to support llama-compatible bitnet models such as Falcon-E or Falcon3-1.58bit series. 
The trick is to check inside the config if the attribute `quantization_config` exists, and check if the bitnet class used is the old (without norm and with invert scales), or the new one (with norm and without invert weight scales)

Confirmed that both Falcon-E and Microsoft bitnet works with this PR

Screenshot 2025-07-03 at 8 28 42 AM

Screenshot 2025-07-03 at 8 29 04 AM

@awni @Blaizzy @angeloskath

### DIFF
--- a/mlx_lm/models/bitlinear_layers.py
+++ b/mlx_lm/models/bitlinear_layers.py
@@ -9,8 +9,7 @@ from mlx.nn.layers.quantized import QuantizedLinear
 def bitnet_quantize(model, modules_to_not_convert=None, invert_weight_scales: bool = False):
     quantize_layers = []
     for name, module in model.named_modules():     
-        if modules_to_not_convert is None:
-            modules_to_not_convert = []
+        modules_to_not_convert = modules_to_not_convert or []
 
         # Replace nn.Linear layers, but skip 'lm_head'
         if name not in modules_to_not_convert and isinstance(module, nn.Linear):
--- a/mlx_lm/models/bitlinear_layers.py
+++ b/mlx_lm/models/bitlinear_layers.py
@@ -2,30 +2,35 @@
 
 import mlx.core as mx
 import mlx.nn as nn
-from mlx.utils import tree_unflatten
 from mlx.nn.layers.quantized import QuantizedLinear
+from mlx.utils import tree_flatten, tree_unflatten
 
 
 def bitnet_quantize(model, quantization_config: dict):
     quantize_layers = []
-    modules_to_not_convert = quantization_config.get("modules_to_not_convert") or []
-    invert_weight_scales = quantization_config.get("linear_class", "") != "autobitlinear"
+    modules_to_not_convert = quantization_config.get("modules_to_not_convert", [])
+    invert_weight_scales = (
+        quantization_config.get("linear_class", "") != "autobitlinear"
+    )
 
-    for name, module in model.named_modules():     
+    for name, module in tree_flatten(model.leaf_modules(), is_leaf=nn.Module.is_module):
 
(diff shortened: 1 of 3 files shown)


mlx_lm/utils.py
```diff
@@ -203,6 +204,17 @@ def class_predicate(p, m):
             bits=quantization["bits"],
             class_predicate=class_predicate,
         )
+    elif (quantization_config := config.get("quantization_config", None)) is not None:
+        # Handle legacy quantization config
+        quantization_method = quantization_config.get("quant_method")
+        if quantization_method == "bitnet" and config.get("model_type") != "bitnet":
```
awni: Do we need to check the model type here? If the model type is bitnet the `bitnet_quantize` function should be a no-op. Alternatively if we want to support other precisions for the `bitnet` model maybe it's worth changing the layers to be `Linear` layers.

mlx_lm/models/bitlinear_layers.py
```diff
@@ -2,9 +2,30 @@
...
 import mlx.nn as nn
+from mlx.utils import tree_unflatten
 from mlx.nn.layers.quantized import QuantizedLinear
 
 
+def bitnet_quantize(model, modules_to_not_convert=None, invert_weight_scales: bool = False):
+    quantize_layers = []
+    for name, module in model.named_modules():     
+        if modules_to_not_convert is None:
+            modules_to_not_convert = []
```
awni: Do this above the loop. And you can do:
```suggestion
        modules_to_not_convert = modules_to_not_convert or []
```

mlx_lm/models/bitlinear_layers.py
```diff
@@ -2,9 +2,30 @@
...
+
+        # Replace nn.Linear layers, but skip 'lm_head'
```
awni: Nit: this comment is incorrect. It skips any layers in the list.

mlx_lm/utils.py
```diff
@@ -203,6 +204,17 @@ def class_predicate(p, m):
...
+            from .models.bitlinear_layers import bitnet_quantize
+
+            model = bitnet_quantize(
+                model, 
+                quantization_config.get("modules_to_not_convert", None),
+                invert_weight_scales= quantization_config.get("linear_class", "") != "autobitlinear"
```
awni: I'm tempted to suggest we just send the config into the `bitnet_quantize` function and do all this logic in there. But I don't feel too strongly about it.

awni (commented):
This is looks pretty nice! Minor comments, please address then I think we can merge it.

Blaizzy:
Awesome work @younesbelkada!

mlx_lm/utils.py
younesbelkada: That's much cleaner indeed !

mlx_lm/utils.py
younesbelkada: Indeed that worked !

> Alternatively if we want to support other precisions for the bitnet model maybe it's worth changing the layers to be Linear layers.

For this I didn't experiment if the new bitnet layers are "universal" meaning we can convert them back to bf16 - what we did for falcon-e is that we pushed the bf16 on a separate repo and people can use it as llama arch. The bf16 model is simply the same model with weight_scale being injected inside the model weights but to do that you need to operate on the pre-quantized checkpoint (not the ternary one). For the microsoft bitnet, it would correspond to these weights: https://huggingface.co/microsoft/bitnet-b1.58-2B-4T-bf16. It will need some testing to check if the inference works well

Issue #274, opened 2025-07-05
module 'lm_eval' has no attribute 'models'


ivanfioravanti:
The issue is present in latest lm-eval 0.4.9 package.

python -m mlx_lm evaluate --model mlx-community/DeepSeek-R1-0528-4bit --tasks medmcqa
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/ifioravanti/github/ifioravanti/mlx-lm/mlx_lm/__main__.py", line 27, in <module>
    submodule = importlib.import_module(f"mlx_lm.{subcommand}")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ifioravanti/.pyenv/versions/3.12.9/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/Users/ifioravanti/github/ifioravanti/mlx-lm/mlx_lm/evaluate.py", line 63, in <module>
    class MLXLM(LM):
  File "/Users/ifioravanti/github/ifioravanti/mlx-lm/mlx_lm/evaluate.py", line 65, in MLXLM
    tokenizer_name = lm_eval.models.huggingface.HFLM.tokenizer_name
                     ^^^^^^^^^^^^^^
  File "/Users/ifioravanti/github/lm-evaluation-harness/lm_eval/__init__.py", line 18, in __getattr__
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
AttributeError: module 'lm_eval' has no attribute 'models'

angeloskath:
[...]

PR #275, opened 2025-07-05
fix: update import for huggingface model in evaluate.py

ivanfioravanti:
This fix #274

PR #276, opened 2025-07-06
Fix for Issue #274

chimezie:


angeloskath:
I will close this as a duplicate of #275. But I also commented on the issue (#274) since it doesn't appear on my end.

PR #278, opened 2025-07-07
Allow generation without README

will-lms:
A recent commit ([e8c2cfce](https://github.com/ml-explore/mlx-lm/commit/e8c2cfce6a3c039136c14fdfe9a5a279a811db56)) introduced a new requirement that a `README.md` must be present to load a model. This PR modifies `get_model_path` to return `None` for hf_path if the model card is not present in the local model path.


awni (approved):
Thanks!

PR #280, opened 2025-07-08
KL loss and memory improvements for DWQ and dynamic quant

angeloskath:
This uses https://github.com/ml-explore/mlx/pull/2335 to train MoE DWQs and also adds a custom metal kernel for the KL loss which speeds it up a bit (it depends on how big of a fraction of the total cost it was) and saves quite a bit of memory especially for models with large vocabularies.

Dynamic quant for `Qwen3-0.6B-Base`
```
Before:                   43s and 17.9GB
After:                    40s and 11.4GB
After with checkpointing: 47s and 9GB
```

Dynamic quant for `Qwen3-8B`
```
Before:                   7m 37s and 91.6GB
After:                    7m 27s and 88.2GB
After with checkpointing: 8m 40s and 74.6GB
```

For DWQ we can now train properly the switch layers and with gradient checkpointing and the kl-loss we can use larger batch size so we get better utilization of the batch kernel. For example

DWQ of `Qwen3-30B-A3B`
```
Before:                     48M params and 113GB and 150 tps
After batch size 1:        954M params and 124GB and 115 tps
After batch size 8 & ckpt: 954M params and 115GB and 250 tps
```

### DIFF
--- a/mlx_lm/models/switch_layers.py
+++ b/mlx_lm/models/switch_layers.py
@@ -53,12 +53,6 @@ class QuantizedSwitchLinear(nn.Module):
         # Freeze this model's parameters
         self.freeze()
 
-    def unfreeze(self, *args, **kwargs):
-        """Wrap unfreeze so that we unfreeze any layers we might contain but
-        our parameters will remain frozen."""
-        super().unfreeze(*args, **kwargs)
-        self.freeze(recurse=False)
-
     @property
     def input_dims(self):
         return self.scales.shape[2] * self.group_size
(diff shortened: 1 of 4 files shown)


angeloskath:
To be honest the memory improvements from the KL loss are less than what I expected but since we do have it now 🤷‍♂️

awni:
For DWQ with Qwen3 0.6B it reduces it quite a lot (57GB -> 30.9GB) and it's also considerably faster (1450 -> 2000 tok/s).

PR #282, opened 2025-07-08
remove sentencepiece

### LINKED ISSUE #227: Pip Install Crashes - Version 0.25.2
Cannot install Version 0.25.2 on MAC Studio M4 64GB using pip install

Error: "ERROR: Failed building wheel for sentencepiece"

awni:
It seems that [sentencepiece](https://github.com/google/sentencepiece) has been somewhat abandoned and is not updating to support newer Python versions, e.g. 3.13. So let's remove it as a required dependency.. if they figure out maintenance we can bring it back.

Most models do not need it.. some older ones do. In which case your only option is to use Python 3.12 or less. Hugging Face will issue an error for those cases like:

```
ValueError: Cannot instantiate this tokenizer from a slow version. If it's based on sentencepiece, make sure you have sentencepiece installed.
```

which I think is clear enough. 

Closes #227


byteprobe:
this has been long overdue. thanks

PR #289, opened 2025-07-09
Allow trust_remote_code in convert.py

christian-lms:
Some recent MoE models (Hunyuan, Ernie 4.5) have quants up on mlx-community that are incompatible with the code in `main` (even after reverting to the respective initial support commits). To create updated quants, one must use `convert.py`, but these models don't yet have native support in HF Transformers and so require `trust_remote_code=True` in `AutoTokenizer.from_pretrained` to load their custom model code. There isn't currently a way to pipe this kwarg in from the surface API in `convert.py`, so this PR adds that.

cc @awni

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -21,7 +21,6 @@ from .utils import (
 def mixed_quant_predicate_builder(
     recipe: str, model: nn.Module
 ) -> Callable[[str, nn.Module, dict], Union[bool, dict]]:
-
     high_bits = 6
     group_size = 64
 
@@ -96,6 +95,7 @@ def convert(
     quant_predicate: Optional[
         Union[Callable[[str, nn.Module, dict], Union[bool, dict]], str]
     ] = None,
+    trust_remote_code: bool = False,
 ):
     # Check the save path is empty
     if isinstance(mlx_path, str):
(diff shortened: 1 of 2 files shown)


awni:
> Some recent MoE models (Hunyuan, Ernie 4.5) have quants up on mlx-community that are incompatible with the code in main (even after reverting to the respective initial support commits).

That sounds like a bug. Which models specifically don't work?

awni (approved):
Thanks this looks good!

christian-lms:
> > Some recent MoE models (Hunyuan, Ernie 4.5) have quants up on mlx-community that are incompatible with the code in main (even after reverting to the respective initial support commits).
> 
> 
> 
> That sounds like a bug. Which models specifically don't work? 

As far as I tested, mlx-community/ERNIE-4.5-21B-A3B-PT-4bit and mlx-community/Hunyuan-A13B-Instruct-4bit. Both complain during the strict check that they're getting MLP expert weight tensors they don't expect (on phone so don't have the error offhand), but they work fine newly converted. 

https://github.com/ml-explore/mlx-lm/commit/5fad2cca9f4cae9932d27c0ce91d9469ea7a1389 seems to be the culprit for Ernie since the weights in mlx-community seem to have been uploaded before this commit, can't find the equivalent for Hunyuan

awni:
> mlx-community/ERNIE-4.5-21B-A3B-PT-4bit 

Looks like it was quantized before changing to use switch layers. I re-uploaded it.

> mlx-community/Hunyuan-A13B-Instruct-4bit

That one works for me.

PR #291, opened 2025-07-10
Add LFM2 (WIP)

Blaizzy:


mlx_lm/models/lfm2.py
```diff
@@ -138,71 +116,72 @@ def __call__(
         x: mx.array,
         cache: Optional[Any] = None,
     ):
+        import pdb
+
+        pdb.set_trace()
```
neilmehta24, line 121: Was this intended to be merged?

awni: Ah that's blunder.

awni: I guess we don't test that layer in the tests :/

awni: Thanks for catching, fixed on main.

Blaizzy:
My pleasure! ❤️

PR #292, opened 2025-07-10
Fix gemma3n config load bug

neilmehta24:
The gemma3n pytorch weights were re-uploaded. The re-upload removed the `query_pre_attn_scalar` field from the config, and now conversions of that model do not load. To fix, remove the `query_pre_attn_scalar` config requirement, since it isn't being used anyway.

Ref:
https://huggingface.co/google/gemma-3n-E4B-it/commit/72639b6c011b3d1357b0777871e0cfba6cac5b04
https://huggingface.co/google/gemma-3n-E2B-it/commit/f5c5abb3bcbe5ab954b06dc6a9cb10ef07b02290


awni (approved):
Ty!

PR #308, opened 2025-07-14
Allow empty prompt with input_embeddings

will-lms:
This PR allows the following `generate_step` parameter combinations. Currently only 1 and 3 are allowed.

1. `len(prompt) > 0 and input_embeddings is None` i.e. prompt only
2. `len(prompt) == 0 and input_embeddings is not None` i.e. input embeddings only
3. `input_embeddings is not None and len(prompt) == len(input_embeddings)` i.e. sequence lengths match

This still provides a helpful check that sequence lengths match when both are provided, but allows for just `input_embeddings` to be provided when the underlying model implementation only requires one or the other.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -329,21 +329,25 @@ def generate_step(
            when ``kv_bits`` is non-None. Default: ``0``.
         prompt_progress_callback (Callable[int, int]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
-        input_embeddings (mx.array, optional): Input embeddings to use in conjunction
-           with prompt tokens. Default: ``None``.
+        input_embeddings (mx.array, optional): Input embeddings to use instead of or in
+          conjunction with prompt tokens. Default: ``None``.
 
     Yields:
         Tuple[mx.array, mx.array]: One token and a vector of log probabilities.
     """
-    if len(prompt) == 0:
-        raise ValueError("Prompt must be non-empty.")
     if input_embeddings is not None:
         if not does_model_support_input_embeddings(model):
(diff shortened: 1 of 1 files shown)


mlx_lm/generate.py
```diff
@@ -386,7 +390,7 @@ def _step(input_tokens: mx.array, input_embeddings: Optional[mx.array] = None):
 
             logits = logits[:, -1, :]
 
-            if logits_processors:
+            if logits_processors and (tokens is not None or len(input_tokens) > 0):
```
awni: I don't think we need the `tokens is not None` condition here?

mlx_lm/generate.py
```diff
@@ -403,6 +407,9 @@ def _step(input_tokens: mx.array, input_embeddings: Optional[mx.array] = None):
 
     with mx.stream(generation_stream):
         total_prompt_tokens = prompt.shape[0]
```
awni: That line is superfluous.

will-lms: Oops will remove

mlx_lm/generate.py
will-lms: Yes I think not. The `tokens is not None and len(input_tokens) <= 0` should never happen anyway given the call pattern for `_step`. Will remove.

awni (approved):
LGTM! Thanks!

yukiarimo:
Is it basically just sending an image to an LLM, without text?

awni:
The preprocessed text and image tokens are merged into the combined `input_embeddings` for VLMs.

yukiarimo:
Can you please pinpoint where exactly it happens in the code? I wanna deep dive in MLX's inner workings

PR #313, opened 2025-07-15
add v1/models/repo_id

### LINKED ISSUE #304: Add /v1/models/:id endpoint to server
This is required by tools like Copilot.

awni:
Closes #304

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -842,7 +842,9 @@ class APIHandler(BaseHTTPRequestHandler):
         """
         Respond to a GET request from a client.
         """
-        if self.path == "/v1/models":
+        if self.path.startswith("/v1/models"):
+            self.handle_models_request()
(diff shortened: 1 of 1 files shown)


mlx_lm/server.py
```diff
@@ -842,7 +842,9 @@ def do_GET(self):
         """
         Respond to a GET request from a client.
         """
-        if self.path == "/v1/models":
+        if self.path.startswith("/v1/models"):
+            self.handle_models_request()
+        elif self.path == "/v1/models":
```
angeloskath: I think you can remove that condition as it is handled in the above one.

angeloskath (approved):
Left a small nitpick otherwise great.

mlx_lm/server.py
awni: Good catch.

PR #315, opened 2025-07-16
Improve /v1/models/:id endpoint

mohd-akram:
Fix response format, improve handling of slashes in model ID and return correct status if a model is not found.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -4,6 +4,7 @@ import argparse
 import json
 import logging
 import platform
+import re
 import socket
 import time
 import uuid
@@ -22,6 +23,7 @@ from typing import (
     Tuple,
     Union,
 )
+from urllib.parse import unquote
 
 import mlx.core as mx
 from huggingface_hub import scan_cache_dir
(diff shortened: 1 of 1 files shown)


mlx_lm/server.py
```diff
@@ -865,15 +867,12 @@ def handle_models_request(self):
...
-        self._set_completion_headers(200)
-        self.end_headers()
-
         files = ["config.json", "model.safetensors.index.json", "tokenizer_config.json"]
 
         parts = self.path.split("/")
         filter_repo_id = None
         if len(parts) > 3:
-            filter_repo_id = "/".join(parts[3:])
+            filter_repo_id = unquote(parts[3])
```
awni, line 875: Doesn't that need to be re joined?

awni: For example,  I might have a model id like `org/model` that will be in `parts[3]` and `parts[4]`

mohd-akram: That can be done by encoding the slash to %2F, so you'll have /v1/models/org%2Frepo, which the unquote will then decode. In the OpenAI API, the ID doesn't contain slashes and is meant to take up a single path component so IMO the server should handle it in a compatible way. This also allows having another component after without ambiguity, which is common in APIs, eg /v1/models/org%2Frepo/details.

zcbenz:

PR #316, opened 2025-07-17
Update W&B logging crash in MLX-LM-LORA

Goekdeniz-Guelmez:
Ensures all metrics are JSON-serializable by converting MX arrays before logging to wandb.
Avoids TypeError: Object of type array is not JSON serializable.

### DIFF
--- a/mlx_lm/tuner/callbacks.py
+++ b/mlx_lm/tuner/callbacks.py
@@ -31,19 +31,20 @@ class WandBCallback(TrainingCallback):
             )
         self.wrapped_callback = wrapped_callback
         wandb.init(project=project_name, dir=log_dir, config=config)
-    
+
     def _convert_to_serializable(self, data: dict) -> dict:
-        return {
-            k: v.tolist() if hasattr(v, "tolist") else v
-            for k, v in data.items()
-        }
+        return {k: v.tolist() if hasattr(v, "tolist") else v for k, v in data.items()}
 
     def on_train_loss_report(self, train_info: dict):
-        wandb.log(self._convert_to_serializable(train_info), step=train_info.get("iteration"))
+        wandb.log(
+            self._convert_to_serializable(train_info), step=train_info.get("iteration")
+        )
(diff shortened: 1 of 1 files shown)


awni:
There shouldn't be any `mx.array` being sent to the callbacks. If there are that would be the bug to fix. Are there?

Goekdeniz-Guelmez:
Not here in mlx-lm itself, but in my mlx-lm-lora for online_dpo, grpo, etc. they return the rewards etc. as mx arrays from the loss function. Imo, converting them inside the loss function wouldn’t be clean, it breaks the separation between training logic and logging. Instead, handling serialization inside the callback is the safer and more modular solution.

mlx_lm/tuner/callbacks.py
```diff
@@ -31,13 +31,19 @@ def __init__(
             )
         self.wrapped_callback = wrapped_callback
         wandb.init(project=project_name, dir=log_dir, config=config)
+    
+    def _convert_to_serializable(self, data: dict) -> dict:
+        return {
+            k: (v.item() if hasattr(v, "item") else v.tolist() if hasattr(v, "tolist") else v)
```
awni: Part of the problem here is we don't know what arrays are scalars and what arrays need to be logged as lists. It is an error to call `a.item` if `a` is not a single value but it will still have the attribute.

If you want arrays logged as scalars I think we should require that one call item ahead of time. And then standardize using `tolist` on all `mx.array` which are input to the logger directly. Does that make sense?

Goekdeniz-Guelmez: indeed, also simplifies things

awni:
One more thing: could you run the [pre-commit hooks](https://github.com/ml-explore/mlx-lm/blob/main/CONTRIBUTING.md#contributing-to-mlx-lm) ?  Thanks!

ShaohonChen:
Excellent work👍!

Goekdeniz-Guelmez:
Perfect thanks

PR #331, opened 2025-07-26
Add early stopping

hansvdam:
when run with early stopping param, lora training when evalution loss goes up.
usage:
```
args = {}
...
args["early_stopping"] = True
...
lora.run(types.SimpleNamespace(**args), training_callback)
```

### DIFF
--- a/mlx_lm/tuner/trainer.py
+++ b/mlx_lm/tuner/trainer.py
@@ -64,7 +64,10 @@ class TrainingArgs:
         default=False,
         metadata={"help": "Use gradient checkpointing to reduce memory use."},
     )
-
+    early_stopping: bool = field(
+        default=False,
+        metadata={"help": "Whether to stop when the evaluation loss increases, and overfitting starts."},
+    )
 
 def default_loss(model, batch, lengths):
     inputs = batch[:, :-1]
@@ -233,6 +236,7 @@ def train(
     trained_tokens = 0
     train_time = 0
     # Main training loop
+    last_validation_loss = float("inf")
     for it, batch in zip(
         range(1, args.iters + 1),
         iterate_batches(
@@ -273,6 +277,13 @@ def train(
                     "val_time": val_time,
                 }
                 training_callback.on_val_loss_report(val_info)
+            if args.early_stopping and val_loss > last_validation_loss:
+                print(
+                    f"Early stopping at iteration {it}, "
+                    f"validation loss {val_loss} is higher than the previous loss {last_validation_loss}."
+                )
+                break
+            last_validation_loss = val_loss
 
             tic = time.perf_counter()
 
@@ -336,6 +347,7 @@ def train(
 
     # Save final weights
     if rank == 0:
-        adapter_weights = dict(tree_flatten(model.trainable_parameters()))
+        if not args.early_stopping:  # if early stopping is not used, save the final weights, otherwise the last checkpoint is the best model
+            adapter_weights = dict(tree_flatten(model.trainable_parameters()))
         mx.save_safetensors(str(args.adapter_file), adapter_weights)
         print(f"Saved final weights to {args.adapter_file}.")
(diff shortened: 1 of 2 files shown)


mlx_lm/tuner/trainer.py
```diff
@@ -336,6 +347,7 @@ def step(batch):
 
     # Save final weights
     if rank == 0:
-        adapter_weights = dict(tree_flatten(model.trainable_parameters()))
+        if not args.early_stopping:  # if early stopping is not used, save the final weights, otherwise the last checkpoint is the best model
```
awni, line 350: This should probably be a check like:
```suggestion
        if not args.early_stopping or val_loss < last_validation_loss:
```

So we are sure to save the last model if we never early stopped (e.g. the val loss always went down).

awni (commented):
Thanks for the contribution.. left one comment. Please address then we can merge!

axiom-of-choice:
I ran into the underlying problem independently while fine-tuning and was about to write this feature before finding this PR. Rather than open a competing one, here are measurements and two implementation notes, in case they help get this over the line. @hansvdam, happy to send a patch against your branch if that's useful.

**The framing that I think makes this obvious.** Independent of early stopping, the trainer computes validation loss, prints it, and discards it. Every save is `it % steps_per_save`, so validation loss never influences which checkpoint survives. That means `adapters.safetensors` holds the *last* model, not the *best* one, and a user only discovers this by keeping every checkpoint and comparing by hand. Even someone who wants all 3000 iterations wants the best checkpoint saved.

**Measured, from a real LoRA run** (Qwen3-1.7B, 8786 rows, batch 4, lr 5e-6 cosine + warmup):

| Iter | Val loss | Δ |
|---|---|---|
| 0 | 1.4537 | (base) |
| 200 | 0.7359 | −0.7178 |
| 400 | 0.7085 | −0.0274 |
| 600 | 0.6958 | −0.0127 |

96% of the total improvement lands in the first 200 of 3000 iterations. Iterations 600-3000 are ~6.5 hours on this machine for maybe 0.02-0.03 more. To be straight about the limits of this data: my run flattened but I stopped it before validation loss actually rose, so this demonstrates wasted compute, not yet a measurably worse final artifact. I'd rather state that than overclaim.

**Two issues in the current diff:**

1. `NameError` on short runs. The final-save block only rebinds `adapter_weights` when `not args.early_stopping`, but `adapter_weights` is otherwise bound only inside `if it % args.steps_per_save == 0`. With `--early-stopping --iters 50` and the default `--save-every 100`, no periodic save ever happens, so the final `mx.save_safetensors(..., adapter_weights)` references an unbound name.

2. The "last checkpoint is the best model" assumption doesn't hold in general. `steps_per_eval` defaults to 200 and `steps_per_save` to 100, and the two cadences are independent. Stopping at an eval boundary does not imply the most recent periodic save came from that same iteration, so `adapters.safetensors` can hold weights from an iteration that was never validated. Any `steps_per_eval` that isn't a multiple of `steps_per_save` makes this worse.

Both go away if the best weights are captured in memory at the moment validation improves, rather than inferred afterwards from save timing:

```python
if val_loss < best_val_loss:
    best_val_loss = val_loss
    best_weights = dict(tree_flatten(model.trainable_parameters()))
```

and then written at the end. That also matches @awni's suggestion above about making sure the last model is still saved when early stopping never triggers, since you can fall back to the final weights when `best_weights` is None.

**One design question worth settling explicitly**, since it decides the above: should `adapters.safetensors` itself always hold the best checkpoint, or should the best go to a separate `best_adapters.safetensors`? A separate file surprises no existing user but requires them to know to pick it up. Overwriting is friendlier but is a behavior change. @awni, do you have a preference? Tracking the best validation loss and writing it seems worth doing even without the `--early-stopping` flag at all, which would make it a strictly smaller change than this PR.

Verified against `main` at e5baded, mlx-lm 0.31.3, mlx 0.32.0.

axiom-of-choice:
Since this has been idle since the review, I opened #1665 to carry the approach forward, crediting your original work. It addresses @awni's inline comment about still saving the last model, and takes a different route to it: the best weights are captured in memory when validation improves, rather than inferred from save timing, which also avoids the two issues I described in my comment above.

Happy to close mine if you would rather pick this back up yourself, just say so.

zcbenz:

PR #336, opened 2025-07-29
Changed GLM-4 MoE support for DWQ quantization

ivanfioravanti:
Updated GLM-4 MoE model implementation to support DWQ quantization method

### DIFF
--- a/mlx_lm/models/glm4_moe.py
+++ b/mlx_lm/models/glm4_moe.py
@@ -151,7 +151,7 @@ def group_expert_select(
         scores = mx.flatten(scores, -2, -1)
 
     k = top_k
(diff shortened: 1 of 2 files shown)


ivanfioravanti:
This should solve #337

ivanfioravanti:
@awni WDYT?

awni:
Did the stop gradient resolve your issue? It doesn't look like the right place to me.. but I'll have to double check.

ivanfioravanti:
> Did the stop gradient resolve your issue? It doesn't look like the right place to me.. but I'll have to double check.

Yes, it worked, but your solution is surely better. 🙏

PR #341, opened 2025-07-31
Support chat_template.jinja file

### LINKED ISSUE #340: Support chat_template.jinja file
Due to some unknown reasons, the recent Qwen3 model used the `chat_template.jinja` file instead of the one set in the [tokenizer_config.json](https://huggingface.co/mlx-community/Qwen3-30B-A3B-Thinking-2507-4bit/blob/main/tokenizer_config.json) file, resulting in the situation where `tokenizer.chat_template` == None when the model was converted to mlx and loaded with mlx-lm.

One way to solve this problem is to update the Qwen3 model, and another way is to make mlx-lm compatible with the `chat_template.jinja` file format. I tend to favor the latter because it has the least impact on users.

Qwen3 has corresponding fix submissions as:
https://huggingface.co/Qwen/Qwen3-235B-A22B-Instruct-2507/commit/3d2db25e0e076aabc0e35029940678544eab7134

madroidmaq:
fix #340

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -265,6 +265,13 @@ def load(
         model_path, tokenizer_config, eos_token_ids=config.get("eos_token_id", None)
     )
 
+    if tokenizer.chat_template is None:
+        template_path = model_path / "chat_template.jinja"
+        if template_path.exists() and template_path.is_file():
+            with open(template_path, "r", encoding="utf-8") as f:
+                chat_template_content = f.read()
+            tokenizer.chat_template = chat_template_content
+
     return model, tokenizer
 
 


awni:
MLX already pulls in the chat template from the `.jinja` file if it's there (because the `AutoTokenizer` should load it).


```python

model, tokenizer = load("mlx-community/Qwen3-30B-A3B-Thinking-2507-4bit")

print(tokenizer.chat_template)
```

Is there a test case you observed where it wasn't loading the template when it should have?

madroidmaq:
@awni  I initially thought so as well, until I encountered a BUG. Below are the results of my test.

https://huggingface.co/mlx-community/Qwen3-30B-A3B-Instruct-2507-4bit-DWQ
image

https://huggingface.co/mlx-community/Qwen3-30B-A3B-Instruct-2507-4bit
image

The contents of the `tokenizer_config.json` in the two files are consistent, but the results are different. The cause of this problem has not been clarified yet.


`Qwen3-30B-A3B-Thinking` will not have such problems because it included the `chat_template` configuration in the `tokenizer_config.json` file when it was initially released.

awni:
@madroidmaq I'm not able to repro the issue you are seeing:

```python
from mlx_lm import load

model, tokenizer = load("mlx-community/Qwen3-30B-A3B-Instruct-2507-4bit-DWQ")

print(tokenizer.chat_template)
```

That prints a chat template for me. Maybe the repo you downloaded is corrupted? Can you try deleting and re-downloading?

madroidmaq:
That makes sense, I will redownload and try again.

madroidmaq:
After I re-downloaded, this issue disappeared, and I will close this issue.

PR #343, opened 2025-07-31
Add validation set for DWQ

awni:
This adds a persistent validation set for DWQ training so we can be more certain the model is making / made progress.

It also prints a big warning if the final validation loss is not better than the initial validation loss.

### DIFF
--- a/mlx_lm/quant/dwq.py
+++ b/mlx_lm/quant/dwq.py
@@ -236,8 +236,19 @@ def load_data(
 
 def main():
     parser = argparse.ArgumentParser()
-    parser.add_argument("--model", "-m", default="Qwen/Qwen3-4B")
-    parser.add_argument("--quantized-model", default=None)
+    parser.add_argument(
+        "--model",
+        "-m",
+        help="A model to distill from for DWQ. If `quantized-model` is not"
+        " given the student model will be this model quantized according"
+        " to `bits` and `group-size`.",
+        required=True,
+    )
+    parser.add_argument(
+        "--quantized-model",
+        default=None,
+        help="An already quantized model (the student model) to improve with DWQ.",
+    )
     parser.add_argument(
         "--mlx-path", default="mlx_model", help="Path to save the quantized model."
     )
@@ -296,6 +307,8 @@ def main():
         q_model, config, _ = fetch_from_hub(
             q_model_path, lazy=True, trust_remote_code=True
         )
+        if "quantization" not in config:
+            raise ValueError("Quantized model must already be quantized.")
     else:
         q_model = copy.deepcopy(model)
         _, config = quantize_model(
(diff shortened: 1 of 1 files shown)


awni:
CC @ivanfioravanti @mzbac @voxmenthe

This should be a good way to tell if a DWQ model is likely to be better than the original. If the final valid loss is worse than the valid loss starting out.. it's very unlikely to be better. In my experience DWQ can be somewhat sensitive to the learning rate. You may need to tune it for different model sizes and precision.

mzbac:
🚀, thanks, was looking for a way to quickly evaluate dwq, this helps a lot :)

ivanfioravanti:
Great one! Thanks 🙏

awni:
I switched the loss for DWQ to be a symmetric KL (jensen shannon divergence, JSD). Vibe wise it gives better results. For example Qwen3 30B A3B instruct generates a working space invaders that is nearly identical to the full precision when using JSD but noteably less good when using regular KL 

I'm still trying out evals / more settings to see if it generalizes. But I added it here in case anyone else wants to experiment with it. At the very least it might be a bit less sensitive to the learning rate.

awni:
Also here's a [4bit DWQ Qwen3 Coder Flash](https://huggingface.co/mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit-DWQ) I made with the JSD loss

mzbac:
I will do some vibe checks for real code tests using mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit-DWQ and will report back the results.

update:

Somehow, the DWQ on the 30B A3B is getting much worse in terms of instruction following and coding comprehensive, it can't even complete the bouncing ball task. I suspect something related to Moe quantization may not directly related to the DWQ quant. but I am still wait for one of my Mac Ultra to free up to run some benchmark tests to be sure.

ivanfioravanti:
I converted 2 models today with this new DWQ code and they are great! I'll try to compare with older DWQ with benchmarks.

mzbac:
Just finished the evaluation for the dwq for the coder. 30B A3B seems to have a noticeable performance drop in coding areas, but it may be related to the https://github.com/ml-explore/mlx-lm/pull/349

Unknown-2

awni:
Awesome thanks for running that eval!!

- Indeed I wonder how the naive 4-bit does after #349. 
- I think we need to improve DWQ for MoEs in general. I think it's not working as well as it should/could.

mzbac:
@awni  I did another run for mmlu pro for the new dwq with the higher precision  gate changes, the improvement looks promising 

v1 is gate quantised to 4bit and the v2 is the gate quantized in 8bit as per your PR changes
Unknown-3

awni:
That's awesome! Looks like higher precision for the gate makes a big difference.

PR #344, opened 2025-07-31
fix error on unsupported response type in server

emmanuel-ferdman:
## PR Summary
This small PR fixes the error on unsupported response type in server.


awni (approved):
Thx!

PR #349, opened 2025-08-04
Allow per model quant config

awni:
Quantizing the gate for the MoE in higher precision makes a pretty large difference to KL and virtually no difference to bits-per-weight and run-time.


angeloskath (approved):
Nice!

PR #354, opened 2025-08-05
Add gpt_oss model

christian-lms:
Contribution by [LM Studio](https://lmstudio.ai/) Team. Initial implementation, more details to follow. Please note numerical precision considerations vs. the reference implementation.

### DIFF
--- a/mlx_lm/models/gpt_oss.py
+++ b/mlx_lm/models/gpt_oss.py
@@ -1,18 +1,19 @@
-from dataclasses import dataclass
+# Copyright © 2025 Apple Inc.
 
-from .base import BaseModelArgs, create_attention_mask
 import math
+from dataclasses import dataclass
+from functools import partial
 from typing import Any, Optional
-from .cache import KVCache, RotatingKVCache
-from .rope_utils import initialize_rope
-from .switch_layers import _gather_sort, _scatter_unsort
-
 
 import mlx.core as mx
 import mlx.nn as nn
 
+from .base import BaseModelArgs, create_attention_mask
+from .cache import KVCache, RotatingKVCache
+from .rope_utils import initialize_rope
+from .switch_layers import SwitchGLU
 
-# TODO(christian): update for hf transformers config
+
 @dataclass
 class ModelArgs(BaseModelArgs):
     num_hidden_layers: int = 36
@@ -30,7 +31,6 @@ class ModelArgs(BaseModelArgs):
     rope_scaling: Any = None
 
 
-### BEGIN MLX PSEUDO-OPERATORS ###
 # These operators emulate particular methods in torch that don't exist in MLX natively
 def mlx_topk(a, k, axis=-1):
     """MLX equivalent of torch.topk"""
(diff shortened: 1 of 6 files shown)


mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+
+    if mask.shape[-1] != K.shape[-2]:
+        mask = mask[..., -K.shape[-2] :]
+    attn_weights = mx.where(mask, attn_weights, -mx.inf)
+
+    sinks = mx.tile(S.reshape(1, -1, 1, 1), [batch, 1, q_len, 1])
+
+    combined_logits = mx.concatenate([attn_weights, sinks], axis=-1)
+    # TODO(christian): So this is just numerical stabilization, which we may not(?) need
+    # Gain about 0.5 tok/s (out of 50) if we remove this
```
awni: Every sensible softmax should do exactly this max remove before the exponentiation :). So it's not necessary because we already do it under the hood.

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+        cache=None,
+        input_embeddings: Optional[mx.array] = None,
+    ):
+        if input_embeddings is not None:
+            x = input_embeddings
+        else:
+            x = self.embed_tokens(inputs)
+
+        # TODO(christian): This is how gemma3_text does it.
+        # We always expect cache to not be None in mlx-engine so /shrug
```
awni: Yea this is good to keep because we don't use a cache with e.g. fine-tuning.

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+
+        self.experts = SwitchMLP(
+            input_dims=config.hidden_size,
+            hidden_dims=config.intermediate_size,
+            num_experts=config.num_local_experts,
+        )
+        self.router = nn.Linear(config.hidden_size, config.num_local_experts, bias=True)
+
+        self.alpha = 1.702
+        self.limit = 7.0
```
awni: Those are unused?

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+        experts, indices = mlx_topk_sorted(g, k=self.num_experts_per_tok, axis=-1)
+        # N.B. As elsewhere, upcast is required in softmax
+        expert_weights = mx.softmax(experts, axis=-1, precise=True)
+        ### END ROUTER BLOCK ###
+
+        # Experts block
+        x = self.experts(x, indices)
+
+        x = x * mx.expand_dims(expert_weights, axis=2)
+        return x.sum(axis=1)
```
awni, line 201: Sometimes it's important to do the sum in fp32.. maybe check another implementation to see?

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+        self.eps = eps
+
+    def __call__(self, x: mx.array):
+        # TODO(christian): monitor for regression
+        # N.B. source of numerical imprecision between mlx and torch
+        return mx.fast.rms_norm(x, self.weight, self.eps)
+
+
+# ref. eager_attention_forward in tfm impl
+@mx.compile
```
awni: I wouldn't compile that.. it will slow things down.. unless you do a shapeless compile but it's probably not worth it for this.

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+class RMSNorm(nn.Module):
+    def __init__(self, num_features: int, eps: float = 1e-05):
+        super().__init__()
+        self.weight = mx.ones((num_features,))
+        self.eps = eps
+
+    def __call__(self, x: mx.array):
+        # TODO(christian): monitor for regression
+        # N.B. source of numerical imprecision between mlx and torch
+        return mx.fast.rms_norm(x, self.weight, self.eps)
```
awni: Why not just use `nn.RMSNorm`.. it is exactly this..

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+    # N.B. PyTorch sigmoid upcasts to float32 internally.
+    # mx.sigmoid doesn't do this, so we implement sigmoid(x) = 1 / 1 + exp(-x) ourselves
+    # ref https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cpu/UnaryOpsKernel.cpp
+    glu_scaled = (alpha * x_glu.astype(mx.float32)).astype(mx.bfloat16)
+    # N.B. It may look pointless to downcast to bf16 and immediately back up to fp32.
+    # It is actually not: the negation has to happen in bf16 or the tensors don't match!
+    # Only after negation can we recast back up for sigmoid
+    negative_glu = (-glu_scaled).astype(mx.float32)
+    sig = (1.0 / (1.0 + mx.exp(negative_glu))).astype(mx.bfloat16)
+    ### END MLX SIGMOID ###
```
awni: I don't think most of this casting is necessary from a quality standpoint, I'd expect it to make almost no difference.. I can see that you won't get exactly comparable numerical results in bf16.

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+        # unsort the output
+        x = _scatter_unsort(x, inv_order, indices.shape)
+
+        return x.squeeze(-2)
+
+
+### END STUFF HACKED IN FROM SWITCH_LAYERS ###
+
+
+def swiglu(x, alpha: float = 1.702, limit: float = 7.0):
```
awni: This function would benefit substantially from being compiled. I would use a shapeless compile.

```
@partial(mx.compile, shapeless=True)
```

I would also do the split outside the function and then pass `x_glu` and `x_linear` as inputs.

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+
+    return final_values, final_indices
+
+
+### END MLX PSEUDO-OPERATORS ###
+
+
+### BEGIN STUFF HACKED IN FROM SWITCH_LAYERS ###
+class QuantizedSwitchLinear(nn.Module):
+    """Taken from mlx_lm.models.switch_layers.
```
awni: Curious, why did you copy-paste those in?

mlx_lm/models/gpt_oss.py
```diff
@@ -0,0 +1,504 @@
...
+    num_attention_heads: int = 64
+    num_key_value_heads: int = 8
+    sliding_window: int = 128
+    rope_theta: int = 150000
+    rope_scaling: Any = None
+
+
+### BEGIN MLX PSEUDO-OPERATORS ###
+# These operators emulate particular methods in torch that don't exist in MLX natively
+def mlx_topk_sorted(a, k, axis=-1):
```
awni: Not clear to me why you need to sort the indices? The result is independent of the order.. it only depends on the partition you use.

mlx_lm/models/gpt_oss.py
christian-lms: Somehow it gives slightly different outputs with/without this. I assume that's a precision artifact of some sort.

mlx_lm/models/gpt_oss.py
christian-lms: Artifact of a previous iteration, yeah

mlx_lm/models/gpt_oss.py
christian-lms: Artifact of when we used to do this by hand... removing

mlx_lm/models/gpt_oss.py
christian-lms: Not sure exactly how much precision is necessary, we could also totally use `mx.sigmoid` here. Your call what to do here

mlx_lm/models/gpt_oss.py
christian-lms: Also an artifact of a previous implementation, though we notice slightly different results for whatever reason

mlx_lm/models/gpt_oss.py
christian-lms: We may want to actually do the split in weight sanitization: see the below comment

mlx_lm/models/gpt_oss.py
christian-lms: So really this should be a `SwitchGLU` instead of `SwitchMLP`, but the way the weights are set up, the up projection and gate projection are packed into one tensor. Between this and using an `alpha` scale in the activation function (unlike the `SwitchGLU` in `switch_layers.py`), we figured it would be best to hack in a `SwitchMLP` of the right shape for now and deal with it later. This requires us to modify certain dimensions (see `is_up`) to account for the packed gate/up proj tensors, so we copy/pasted them here to do that.

FWIW this is the tensor shape the Transformers implementation uses: https://github.com/huggingface/transformers/pull/39923/files

psm-2:
@awni Is fine-tuning included in this update?

niryuu:
`ValueError: Unsupported quantization method mxfp4`
Do we need additional libraries update?

mlx_lm/models/gpt_oss.py
awni: I do think we should handle that in the weight sanitization and reuse the switch layers.

mlx_lm/models/gpt_oss.py
awni: Right let's do it in the sanitization. Would you mind making that change?

mlx_lm/models/gpt_oss.py
awni: I think it is fine to keep the upcast if you compile this function. It shouldn't matter much in that case.

christian-lms:
@niryuu where are you getting this error?

awni:
Presumably from trying to load the original model in mlx-lm (which is expected to not work). You can reproduce it with e.g. `mlx_lm.convert --hf-path openai/gpt-oss-20b`. I'll try and add a dequant path for that so we can load them directly.

awni:
Ok you should be able to load the original models now. I added a dequant step if needed:

```
mlx_lm.convert --hf-path openai/gpt-oss-20b -q --q-bits 8
```

And:

```
mlx_lm.generate --model mlx_model --prompt "Write a story about Einstein" -m 512
```

On an M2 Ultra:

```
<|channel|>analysis<|message|>The user: "Write a story about Einstein". They want a story about Einstein. We need to produce a story. The user didn't specify length or style. We can choose a creative narrative. Maybe a fictional story where Einstein is a character, or a story about his life, or a whimsical story. We can incorporate his personality, his relativity, his curiosity, his interactions. We can make it a short story, maybe with a twist. The user didn't specify constraints. We can produce a story that is engaging, maybe with a moral. Let's think: We can write a story about Einstein as a child, or as a professor, or as a time traveler. Or we can write a story about a young student who meets Einstein. Or a story about Einstein's relativity in a metaphorical sense. Or a story about Einstein's love for music. Or a story about Einstein's relationship with his wife. Or a story about Einstein's involvement in the Manhattan Project. Or a story about Einstein's philosophical musings. Or a story about Einstein's time in America. Or a story about Einstein's interactions with other scientists. Or a story about Einstein's childhood in Ulm. Or a story about Einstein's later years. Or a story about Einstein's "thought experiment" of riding a beam of light. Or a story about Einstein's "relativity of simultaneity" in a narrative. Or a story about Einstein's "E=mc^2" as a metaphor for love. Or a story about Einstein's "relativity" in a social context. Or a story about Einstein's "relativity" in a comedic way. Or a story about Einstein's "relativity" in a magical realism style. Or a story about Einstein's "relativity" in a sci-fi setting. Or a story about Einstein's "relativity" in a children's story. The user didn't specify. We can choose a creative approach. Let's write a story about a young boy who meets Einstein in a dream, and learns about relativity and the importance of curiosity. Or a story about Einstein's time in Princeton, where he meets a young student. Or a story about Einstein's "thought experiment" of riding a light beam, but in a narrative form. Or a story about Einstein's "relativity" as a metaphor for empathy. Or a story about Einstein's "relativity" in a comedic way. Let's choose a story that is accessible, maybe a short story with a moral. Let's write a story about a young
==========
Prompt: 72 tokens, 224.537 tokens-per-sec
Generation: 512 tokens, 84.333 tokens-per-sec
Peak memory: 22.345 GB
```

christian-lms:
Yep, it works! Uploading quants now. Are we good to merge?

stakodiak:
Yep. LGTM

psm-2:
@awni Would it be possible to support fine-tuning?
`ValueError: Lora does not support gpt_oss_moe`

altaic:
@psm-2 looks like all the other support will land soon in #357

PR #356, opened 2025-08-06
Route the gpt_oss to fused sdpa

angeloskath:
This PR changes the GPT OSS implementation such that it routes to the fused sdpa.

The idea is quite simple but adds a tiny bit of annoying book keeping. Instead of adding sinks to the softmax and dropping it after the softmax we add an extra key and value that is 0 and add the sinks as bias towards these keys and values. That means that we have an additive float mask that is not quite triangular but has a specific value for the first index.

This implementation also keeps the last mask in `_previous_mask` so that we don't have to recreate it every time which does make a difference in speed.

Some numbers:

```
Generate 2k tokens

  main:                 61.9 tps
  sdpa (no mask cache): 66.5 tps
  sdpa:                 68.8 tps

Generate 1k tokens starting from 15k

  main:                 39.8 tps
  sdpa (no mask cache): 57.1 tps
  sdpa:                 57.8 tps

```

I want to do a simplification pass (and correctness pass) before we merge because it is a bit ugly in its current form.

### DIFF
--- a/mlx_lm/models/gpt_oss.py
+++ b/mlx_lm/models/gpt_oss.py
@@ -118,13 +118,15 @@ class AttentionBlock(nn.Module):
             mask = mx.concatenate([sinks, mask], axis=-1)
             return mask
 
-        # If we are training then simply re-create the mask every time to make
-        # sure gradients flow to the sinks.
-        if self.training:
+        # When training re-create the mask so that gradients flow to the sinks.
+        # When L is large then recreate the mask because otherwise it will take
+        # a pretty significant chunk of memory.
+        if self.training or L > 8:
             self._previous_mask = None
             return _make_mask(L, offset)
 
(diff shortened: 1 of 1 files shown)


mlx_lm/models/gpt_oss.py
```diff
@@ -377,6 +409,6 @@ def make_cache(self):
                 caches.append(KVCache())
             else:
                 caches.append(
-                    RotatingKVCache(max_size=self.args.sliding_window, keep=0)
+                    RotatingKVCache(max_size=self.args.sliding_window + 1, keep=1)
```
awni, line 414: That's great.

angeloskath:
Latest update improves prompt processing speed as well. To process ~15k tokens we went from 1100 tps to 1500 tps.

PR #359, opened 2025-08-07
Revert symmetric kl

awni:
Reverting for now as it's not doing anything.

I'll test out the `jsm_div_loss` and a proper symmetric KL and report back.


angeloskath (approved):
👍

PR #362, opened 2025-08-08
Add SSE keepalive to stop client disconnects during prompt processing

dysangel:
This PR adds a keepalive callback to send SSE comments during prompt processing, to prevent client timeouts.

When trying to process very large prompt in Cline/Kilocode etc, I was getting disconnected.

The changes here allowed me to run a Cline session that successfully waited 20 minutes for a >100k GLM Air session to resume, and then successfully went on to process over 3 million tokens over the course of the task.

I've attempted to add a test, but this change is obviously better tested with a real client using a medium/large model and context.

This is my first ever open source PR, so any advice on etiquette etc is appreciated. I wasn't sure if I should create an issue first or just directly submit the PR.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -680,6 +680,23 @@ class APIHandler(BaseHTTPRequestHandler):
         tool_text = ""
         in_tool_call = False
         segment = ""
+
+        # Create keepalive callback to send SSE comments during long prompt processing
+        def keepalive_callback(processed_tokens, total_tokens):
+            logging.info(
+                f"Prompt processing progress: {processed_tokens}/{total_tokens}"
+            )
+            if self.stream:
+                try:
+                    # Send SSE comment for keepalive - invisible to clients but keeps connection alive
+                    self.wfile.write(
+                        f": keepalive {processed_tokens}/{total_tokens}\n\n".encode()
+                    )
+                    self.wfile.flush()
+                except (BrokenPipeError, ConnectionResetError, OSError):
+                    # Client disconnected, ignore
+                    pass
+
         for gen_response in stream_generate(
             model=self.model,
             tokenizer=self.tokenizer,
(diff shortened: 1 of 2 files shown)


tests/test_keepalive.py
```diff
@@ -0,0 +1,57 @@
+# Copyright © 2025 Apple Inc.
+
+import io
+import unittest
+from unittest.mock import Mock
+
+
+class TestKeepalive(unittest.TestCase):
```
awni: Would you mind moving this test to [`test_server.py`](https://github.com/ml-explore/mlx-lm/blob/main/tests/test_server.py)?

mlx_lm/server.py
```diff
@@ -680,6 +680,23 @@ def handle_completion(
...
+
+        # Create keepalive callback to send SSE comments during long prompt processing
+        def keepalive_callback(processed_tokens, total_tokens):
+            logging.info(
+                f"Prompt processing progress: {processed_tokens}/{total_tokens}"
+            )
+            if self.stream:
+                try:
+                    # Send SSE comment for keepalive - invisible to clients but keeps connection alive
+                    self.wfile.write(
```
prabhu, line 692: Does this send this comment for every token processed? Can we add some throttling with a configurable interval or batch size?

Also, a nitpick. We aren't validating the values `processed_tokens` and `total_tokens`. Not sure if there is an injection opp here.

dysangel: This callback gets triggered every `prefill_step_size` tokens, which defaults to `2048`

Screenshot 2025-08-15 at 18 40 40

Since `processed_tokens` and `total_tokens` are existing internal values being generated on the server, rather than something we send from the client, I assume it should be safe?

We could just send a blank comment to the client though and it would still keep the connection alive.

prabhu: Thank you so much. This wasn't evident from the documentation which says 

```
prompt_progress_callback (Callable[int, int]): A call-back which takes the
```

but this [line](https://github.com/ml-explore/mlx-lm/blob/877cc38e6c73f25526cb0d2c50cef393122f58e5/mlx_lm/generate.py#L418) makes it clear about the prefill_step_size and the type of the arguments.

Thank you for the prompt clarification!

PR #376, opened 2025-08-17
use weight dtype for dequant

awni:
Relevant context: https://huggingface.co/mlx-community/DeepSeek-V3-0324-4bit/discussions/3

### DIFF
--- a/mlx_lm/models/deepseek_v3.py
+++ b/mlx_lm/models/deepseek_v3.py
@@ -483,10 +483,11 @@ class Model(nn.Module):
             weight = weight.reshape(
                 ((m + pad_bottom) // bs, bs, (n + pad_side) // bs, bs)
             )
+            scale_inv = scale_inv.astype(dtype)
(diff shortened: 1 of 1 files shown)


awni:
As far as I can tell the way it currently is (e.g up-cast weights to fp32) is more aligned with the deepseek [implementation](https://github.com/deepseek-ai/DeepSeek-V3/blob/f6e34dd26772dd4a216be94a8899276c5dca9e43/inference/kernel.py#L78) as well as the fp8 gemm they wrote (which does the multiplication by scales in fp32).

So I don't think we should land this. But I do wonder why the results would be better in the linked discussion 🤔  .. it might just be coincidence.

awni:
I ran an eval with both versions on a subset of MMLU pro.

The scores are:
Current: 76.79
Update: 76.43

So it seems the way we have it now performs slightly better.. but overall it's not that sensitive to how you do the dequantization 🤷 

Closing this for now.

PR #378, opened 2025-08-18
Add LFM2-VL model implementation

christian-lms:
Adds the [Liquid AI LFM2-VL](https://huggingface.co/collections/LiquidAI/lfm2-vl-68963bbc84a610f7638d5ffa) model family for text-only inference. Sibling of https://github.com/lmstudio-ai/mlx-engine/pull/209.


mlx_lm/utils.py
```diff
@@ -45,6 +45,7 @@
     "phi-msft": "phixtral",
     "falcon_mamba": "mamba",
     "kimi_k2": "deepseek_v3",
+    "lfm2-vl": "lfm2_vl",
```
awni: Should we just make the file name `lfm2-vl.py`?

christian-lms: Oh, sure. I thought Python didn't like dashes in file names but I guess that works (`nemotron-nas.py` seems to be fine)

christian-lms: changed

PR #384, opened 2025-08-19
Add Qwen2-VL model implementation

vincentamato:
Adds the [Qwen2-VL](https://huggingface.co/collections/Qwen/qwen2-vl-66cee7455501d7126940800d) model family for text-only inference. Sibling of [lmstudio-ai/mlx-engine#215](https://github.com/lmstudio-ai/mlx-engine/pull/215).

Because Qwen2-VL and Qwen2.5-VL have the same language model implementation, the [Qwen2.5-VL](https://huggingface.co/collections/Qwen/qwen25-vl-6795ffac22b334a837c0f9a5) model family is supported and mapped to `qwen2_vl`.

### DIFF
--- a/mlx_lm/models/rope_utils.py
+++ b/mlx_lm/models/rope_utils.py
@@ -192,90 +192,76 @@ class YarnRoPE(nn.Module):
             freqs=self._freqs,
         )
 
-class MRoPE(nn.Module):   
+class MRoPE(nn.Module):
     def __init__(
         self,
         dims: int,
         mrope_section: List[int],
-        traditional: bool = False,
         base: float = 10000,
-        scale: float = 1.0,
         max_position_embeddings: int = 2048,
     ):
         super().__init__()
         self.dims = dims
-        self.traditional = traditional
-        self.base = base
-        self.scale = scale
         self.mrope_section = mrope_section
-        
+        self.base = base
+        self.max_position_embeddings = max_position_embeddings
         self._split_indices = mx.cumsum(mx.array(mrope_section) * 2)[:-1].tolist()
-        
         self._inv_freq = 1.0 / (base ** (mx.arange(0, dims, 2, dtype=mx.float32) / dims))
-        
-        self.max_seq_len_cached = 0
         self._cos_cached = None
         self._sin_cached = None
-        self._update_cos_sin_cache(max_position_embeddings)
+        self._set_cos_sin_cache(seq_len=max_position_embeddings)
     
-    def _update_cos_sin_cache(self, seq_len: int, device=None):
-        if seq_len <= self.max_seq_len_cached:
-            return
-        
+    def _set_cos_sin_cache(self, seq_len: int):
         self.max_seq_len_cached = seq_len
-        
-        positions = mx.arange(seq_len, dtype=mx.float32)
-        freqs = mx.outer(positions, self._inv_freq)
-        
-        emb = mx.concatenate([freqs, freqs], axis=-1)
-        
+        t = mx.arange(self.max_seq_len_cached).astype(mx.float32)
+        freqs = mx.outer(t, self._inv_freq)
+        emb = mx.concatenate((freqs, freqs), axis=-1)
         self._cos_cached = mx.cos(emb)
         self._sin_cached = mx.sin(emb)
     
-    def _rotate_half(self, x: mx.array) -> mx.array:
-        half_dim = x.shape[-1] // 2
-        x1, x2 = x[..., :half_dim], x[..., half_dim:]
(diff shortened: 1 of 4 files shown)


awni:
I tried running this:

```
mlx_lm.generate --model Qwen/Qwen2.5-VL-3B-Instruct --prompt "Write a story about Einstein"
```

And it gave an error missing some keys:

```
ValueError: Received 824 parameters not in model: 
model.embed_tokens.weight,
model.layers.0.input_layernorm.weight,
model.layers.0.mlp.down_proj.weight,
model.layers.0.mlp.gate_proj.weight,
model.layers.0.mlp.up_proj.weight,
...
```

vincentamato:
@awni Let me know if the above commit resolves that issue. The problem was the HuggingFace models and the MLX models have different parameter formats, so I updated the sanitize method to handle both naming conventions and properly map parameters to the expected model structure.

mlx_lm/models/rope_utils.py
```diff
@@ -193,6 +193,78 @@ def __call__(self, x, offset=0):
...
+class MRoPE(nn.Module):
+    def __init__(
+        self,
+        dims: int,
+        mrope_section: List[int],
+        base: float = 10000,
+        max_position_embeddings: int = 2048,
+    ):
+        super().__init__()
+        self.dims = dims
```
mattjcly: Excuse my naivety if I'm just completely missing something: Where are dims actually used?

I am generally wondering about how the difference in qwen2vl and qwen2.5vl MRoPE dimensions is handled.

From https://arxiv.org/pdf/2502.13923:
"(3) We upgrade MRoPE in the temporal domain by aligning to absolute time, thereby facilitating more
sophisticated temporal sequence learning;"
but this may only apply to video inputs which we are not currently supporting anyway." (basically 4D in 2.5VL vs 3D in 2VL afaict)

Just want to verify that a proper understanding of the differences between the two models can be expressed with the joint implementation/routing to that implementation :)

vincentamato: No worries. The `dims` parameter is used when building the frequency table:
```python
self._inv_freq = 1.0 / (
    base ** (mx.arange(0, dims, 2, dtype=mx.float32) / dims)
)
```
This sets the rotary width for each attention head and controls how the embedding space is partitioned across temporal, height, and width dimensions. Both Qwen2-VL and Qwen2.5-VL adopt the same 3D T/H/W decomposition, so `dims` behaves identically in both. The only distinction is that Qwen2.5-VL changes how temporal positions are generated for video, aligning them to absolute time rather than frame index. For text and image inputs, which are the only modalities relevant here, the MRoPE implementation is unchanged. As a result, a single joint implementation works for both models.

More broadly, the underlying language model is the same across Qwen2-VL and Qwen2.5-VL.
From the [Qwen2-VL report](https://arxiv.org/pdf/2409.12191): 

> The LLM component of Qwen2-VL is initialized using the parameters from Qwen2

From the [Qwen2.5-VL report](https://arxiv.org/pdf/2502.13923):

> The Qwen2.5-VL series adopts large language models as its foundational component. The model is initialized with pre-trained weights from the Qwen2.5 LLM.

And the [Qwen2.5 report](https://arxiv.org/pdf/2412.15115) confirms that Qwen2.5 shares the same architecture as Qwen2:

> For dense models, we maintain the Transformer-based decoder architecture (Vaswani et al., 2017; Radford
et al., 2018) as Qwen2 (Yang et al., 2024a).

mattjcly: Thanks for all this. Makes sense, should've clarified more that I was looking for where `self.dims` was used but I see you removed the instance variables :)

mlx_lm/models/rope_utils.py
```diff
@@ -193,6 +193,76 @@ def __call__(self, x, offset=0):
         )
 
 
+class MRoPE(nn.Module):
```
awni: I must admit I'm pretty confused by this RoPE implementation. It seems equivalent to a regular rope for the configuration used by qwen2.5 VL and qwen2 VL. There are always 3 splits so you just split and recombine the sin/cos and then apply them as usual.

My recommendation would be to just assert that `len(mrope_section) == 3` if the rope type is mrope and then use a regular rope for now. It will be faster.. and we don't need to add this function unless it's actually doing something useful.

vincentamato: You're absolutely right. I apologize for my naivety and the completely unnecessary addition. I got so caught up trying to reimplement MRoPE from `mlx-vlm` that I didn't see the bigger picture. What do you think now?

awni (approved):
Thanks!!

PR #387, opened 2025-08-19
Make KL and JS metal kernels only if metal is available

vsabolcec:
KL and JS loss metal kernels are currently made regardless whether the metal backend is available, causing a crash on non-metal devices. This PR adds a check which makes the kernels only if metal is available.

### DIFF
--- a/mlx_lm/tuner/losses.py
+++ b/mlx_lm/tuner/losses.py
@@ -8,15 +8,9 @@ def can_run_metal():
     return mx.default_device() == mx.gpu and mx.metal.is_available()
 
 
-def metal_only_function(fn):
-    def wrapped_fn(*args, **kwargs):
-        if can_run_metal():
-            return fn(*args, **kwargs)
-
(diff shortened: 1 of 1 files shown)


mlx_lm/tuner/losses.py
```diff
@@ -333,6 +341,8 @@ def _make_kl_backward_kernel():
     )
 
 
+_make_kl_forward_kernel = metal_only_function(_make_kl_forward_kernel)
+_make_kl_backward_kernel = metal_only_function(_make_kl_backward_kernel)
 _kl_forward_kernel = _make_kl_forward_kernel()
 _kl_backward_kernel = _make_kl_backward_kernel()
```
awni: The wrapping seems unnecessarily complex:
```suggestion
if can_run_metal():
  _kl_forward_kernel = _make_kl_forward_kernel()
  _kl_backward_kernel = _make_kl_backward_kernel()
```

PR #388, opened 2025-08-21
fix sampling with small top k

awni:
As pointed out by @neilmehta24 there is a bug for small `top_k` when using `top_p`, e.g.:

```
mlx_lm.generate --model mlx-community/Qwen3-4B-Thinking-2507-4bit --prompt "Tell me a story about Einstein" -m 1000 --top-k 1 --top-p 0.8 --temp 0.01
```

The bug is that `top_k` doesn't preserve the distribution.. and top_p assumes the cumulative sum is 1. If it's not 1 we can get a case where all the tokens are masked in top_p after top_k. 

The fix is simply to do top_k last.

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -48,8 +48,6 @@ def make_sampler(
 
     # Create sampler chain
     sampling_methods = []
-    if top_k > 0:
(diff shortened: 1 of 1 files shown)


yagil:
Thanks to @ggerganov for pointing out the bug :)

PR #390, opened 2025-08-21
fix window attention mask

awni:
The window attention mask had one too many values as pointed out by @angeloskath

### DIFF
--- a/mlx_lm/models/gpt_oss.py
+++ b/mlx_lm/models/gpt_oss.py
@@ -166,7 +166,7 @@ class AttentionBlock(nn.Module):
 
         # We are in inference so cache the mask and try to reuse it
         if self._previous_mask is None:
-            self._previous_mask = _make_mask(L, window_size + 1)
+            self._previous_mask = _make_mask(L, window_size)
 
         return self._previous_mask[..., : min(L + offset, window_size + 1)]
 


angeloskath:
I am fine with merging this but I think that the sliding window attention is in theory defined as before with a window of 128 attending to 129 elements. So the error is in the rotating KVCache which keeps only window elements around.

So we have 2 options.

1. Apply this change but make sure that the rotating KVCache is consistent during prompt processing. (I think it keeps window+1 during prompt and window during generation)
2. Change the rotating KVCache to keep window + 1 elements.

awni:
If the sliding window size is 128.. does the current query attend the current key + 127 previous keys or 128 previous keys. This all assumes the former (itself + 127 previous keys, for a total of 128). But sounds like you are saying it should be the latter?

awni:
For gpt-oss it seems like the answer is the former (the total should be 128 not 129): https://github.com/openai/gpt-oss/blob/main/gpt_oss/torch/model.py#L161-L165

awni:
Seems like it is the same for [gemma3](https://github.com/google/gemma_pytorch/blob/main/gemma/gemma3_model.py#L203-L206)

awni:
So for gemma3 in mlx-lm I think this change fixes it. During prompt processing the mask always contains at most `sliding_window` true elements for each query (and the rotating KV cache is extended  to hold the full context for all queries during prompt processing). During generation the rotating KV cache is kept to `sliding_window` elements and we don't use a mask.

For `gpt-oss` it might not be correct because of the way we handle the extra keep token... have to double check that.

angeloskath:
I was saying indeed it would be the latter. I checked whichever implementation back then and saw that it was 128 + 1. Either way I think making it 127 + 1 is always the better choice as we don't have weird sizes and worst case we drop 1 token.

> For gpt-oss it might not be correct because of the way we handle the extra keep token

I think it should be fine otherwise it wouldn't work without the sinks.

We probably still have to check that the rotating kv cache in non-generation mode works as expected.

awni:
The gpt-oss mask looks a bit off for generation. After processing a prompt of 73 tokens with a window size of 128 there shouldn't be any masked positions yet, but it seems to be masking the first spot after the sink slot.

```
array([[2.51562, -inf, 0, ..., 0, 0, 0]], dtype=bfloat16)
```

PR #397, opened 2025-08-24
Add `mlx_lm.perplexity`

N8python:
Adds a utility, `mlx_lm.perplexity`, for computing perplexity over the same dataset used in `quant/dynamic_quant.py`. Adds standard errors for perplexity comparisons. Properly implements a [gist](https://gist.github.com/N8python/fe048698e17fc430416d28d4689603dc) that people reported was useful.

```
usage: mlx_lm.perplexity [-h] --model MODEL [--batch-size BATCH_SIZE] [--sequence-length SEQUENCE_LENGTH] [--num-samples NUM_SAMPLES] [--seed SEED]

Evaluate perplexity of MLX models

options:
  -h, --help            show this help message and exit
  --model MODEL         Path to model or Hugging Face model ID
  --batch-size BATCH_SIZE
                        Batch size for evaluation
  --sequence-length SEQUENCE_LENGTH
                        Sequence length for evaluation
  --num-samples NUM_SAMPLES
                        Number of samples to use (-1 for all available)
  --seed SEED           Random seed for data sampling
 ```

### DIFF
--- a/mlx_lm/__main__.py
+++ b/mlx_lm/__main__.py
@@ -16,6 +16,7 @@ if __name__ == "__main__":
         "fuse",
         "generate",
         "lora",
+        "perplexity",
         "server",
         "manage",
         "upload",
--- a/mlx_lm/perplexity.py
+++ b/mlx_lm/perplexity.py
@@ -1,7 +1,6 @@
-#!/usr/bin/env python3
 """
 Evaluate perplexity (PPL) of MLX models.
-Based on the eval_ppl function from dynamic.py.
+Based on the eval_ppl function from quant/dynamic_quant.py.
 """
 
 import argparse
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by awni** [comment]
This is great, thanks! The only thing I'm thinking about is whether we should have this run by default on the calibration data or instead support Hugging Face datasets (like we do for training DWQ). The latter would be more flexible since it could support multiple datasets.

**Comment by awni** [comment]
I changed it so you can specify the dataset as an HF dataset (in one of several expected formats). To get the old behavior, use:

```
mlx_lm.perplexity --model google/gemma-3-270m-it --data-path mlx-community/mlx_lm_calibration_v5
```

awni (approved):
Thanks for adding that!!

N8python:
Alright - is there anything else to do before merge?

PR #399, opened 2025-08-25
Server now attempts to load models from local cache first

aperaham:
### Summary

This PR improves the model loading behavior by enabling the server to first attempt loading models from the local Hugging Face cache before falling back to downloading them from the Hugging Face Hub. This change reduces unnecessary network requests and improves performance when models are already cached locally.

### Changes

1. **Enhanced `get_model_path` function**:
   - The function now checks if the model path exists locally.
   - If the path does not exist, it proceeds to download the model from Hugging Face.

2. **Updated `load` function**:
   - Added a new parameter `attempt_with_local_files` to control whether the server should attempt to load models from the local cache first.
   - When `attempt_with_local_files=True`, the function first tries to load the model using `local_files_only=True`.
   - If that fails with a `LocalEntryNotFoundError`, it falls back to downloading the model from Hugging Face.
   - When `attempt_with_local_files=False` (default), the behavior remains unchanged — always attempts to download from Hugging Face.

### Behavior

- **Before**: The server always made a `GET` request to Hugging Face when a model was loaded, even if it already existed locally.
- **After**: If `attempt_with_local_files=True`, the server will first check the local cache (`.cache/huggingface`) and attempt to load the model from there. If that fails, it falls back to downloading from Hugging Face.

### Example Usage

```python
# Load model from local cache if available, fallback to HF if not
model, tokenizer = load(\"meta-llama/Llama-3.2-1B\", attempt_with_local_files=True)
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,7 +80,7 @@ To determine the model layer names, we suggest either:
 To add LoRA support edit
 [`mlx_lm/tuner/utils.py`](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tuner/utils.py#L27-L60)
 
-Finally, add a test for the new modle type to the [model
+Finally, add a test for the new model type to the [model
 tests](https://github.com/ml-explore/mlx-lm/blob/main/tests/test_models.py).
 
 You can run the tests with:
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -202,11 +202,15 @@ class ModelProvider:
                     adapter_path if adapter_path else self.cli_args.adapter_path
                 ),  # if the user doesn't change the model but adds an adapter path
                 tokenizer_config=tokenizer_config,
+                attempt_with_local_files=True,
             )
         else:
             self._validate_model_path(model_path)
             model, tokenizer = load(
-                model_path, adapter_path=adapter_path, tokenizer_config=tokenizer_config
+                model_path,
+                adapter_path=adapter_path,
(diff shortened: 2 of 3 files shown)


awni:
The problem with this change is it makes downloading a bit more brittle (see example below). If there is a way to fix the below problem then we could consider adding something like this. Otherwise I would not make it the default in `mlx_lm.server`.

```python

snapshot_download(
    "google/gemma-3-270m-it",
    local_files_only=False,
    allow_patterns=[
        "*.json",
    ],
)

# Doesn't raise!
snapshot_download(
    "google/gemma-3-270m-it",
    local_files_only=True,
    allow_patterns=[
        "*.json",
        "model*.safetensors",
    ],
)```

aperaham:
@awni Thank you for your review. I appreciate you highlighting the issue with `snapshot_download` and `local_files_only`.  

I understand that this change can lead to unexpected behavior if the cache contains missing/corrupted files, and I'll update the code to include more comprehensive checks for required model files before falling back to downloading from Hugging Face. 

I'll aim to have this updated within the next couple of days.
Thanks again.

aperaham:
@awni I think it would be best to close this for now to reduce clutter. 

While I believe the approach attempted here does demonstrate the desired behavior, as pointed out, there are a few important edge cases that need more consideration. I'm currently working on a more robust, cache aware, model/asset loading mechanism which I think could be beneficial. Progress is slower than planned due to outside constraints, but I hope to present something soon. 

Thanks again for reviewing.

PR #412, opened 2025-08-29
Adding granite moe

Goekdeniz-Guelmez:
(base)  ~/Desktop/mlx-lm/ [adding-granite-moe*] python -m mlx_lm.generate --model ibm-granite/granite-3.1-3b-a800m-instruct  --prompt "write a simple python function" --max-tokens 4096
Fetching 11 files: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 11/11 [00:00<00:00, 23114.90it/s]
==========
Here's a simple Python function that calculates the factorial of a number using recursion:

```python
def factorial(n):
    # Base case: 0! = 1
    if n == 0:
        return 1
    else:
        # Recursive case: n! = n * (n-1)!
        return n * factorial(n-1)
```

You can call this function with an integer argument to get its factorial. For example, `factorial(5)` will return `120`.

Please note that this function uses recursion, which means it calls itself within its own definition. Be cautious while using recursion with large inputs, as it can lead to a stack overflow error due to the limited call stack size. For practical purposes, consider using an iterative approach or memoization to optimize the function for larger inputs.
==========
Prompt: 64 tokens, 94.628 tokens-per-sec
Generation: 184 tokens, 7.181 tokens-per-sec
Peak memory: 7.256 GB


Goekdeniz-Guelmez:
@awni @angeloskath should be mergable :D

PR #414, opened 2025-08-29
server allow specifying seed

n8sh1:


tests/test_server.py
```diff
@@ -80,15 +80,20 @@ def test_handle_completions(self):
...
 
-        response_body = response.text
+        response_body = json.loads(response.text)
 
         self.assertIn("id", response_body)
         self.assertIn("choices", response_body)
+        self.assertEqual(
+            response_body["choices"][0]["text"],
+            ", there was a beautiful princess named Lila.\n",
+        )
```
n8sh1: Alternatively this could be tested by making two requests with the same seed and asserting they produce the same text.

awni: That's a safer test imo.. since it will be less sensitive to numerical changes under the hood  (e.g. when switching back-ends). Do you mind switching it?

n8sh1: Sure

awni (approved):
LGTM, thanks!

PR #415, opened 2025-08-30
Fix mlx_lm.perplexity seed w/ np.random.seed to ensure determinism across runs

N8python:
As per: https://x.com/kernelpool/status/1961640338214588625

Numpy random seed was not set, multiple runs were not deterministic.


awni (approved):
Thanks!

PR #430, opened 2025-09-05
Changes needed to facilitate batching

awni:
The big change here is how caching and mask making is done. It is intended to be more friendly to batching (which will have more irregular masking). But it also fixes a few bugs we have now.

1. Removes `mask` as an input to models. This isn't really a well defined input and mostly just a bug. For example sending in a mask to `gpt_oss` would cause the wrong computation to happen (since it probably doesn't include sinks). It's too ambiguous to try and apply the input mask as well as any model-specific masks (e.g. should the be && together and if so, what if one is offset in a irregular way?). Mask making is now purely a property of either the model or the cache.
2. Fix bug with some sliding window attention models where we would not use a sliding window mask when no cache is provided (e.g. fine-tuning, dwq, etc)
3. Changes the API of `create_attention_mask` to be slightly less implicit. It takes a single cache object now rather than a list of caches. 
4. Add a `make_mask` method to some cache classes. For some cache classes making a correct mask is quite coupled to the internal state of the cache. 

Other changes are mostly more tests for models and testing for the above.

Subsumes a lot from #126 (which is closed).


Goekdeniz-Guelmez:
uhhh this would be awesome, It would make synthetic dataset creation in mlx-lm-lora a lot faster and better.

PR #436, opened 2025-09-08
remove manual conv class in mamba1

Goekdeniz-Guelmez:
removed the custom DepthWiseConv1d class and used the build in conv1d with the groups param, makes it easier and has a small speed improvement.

### DIFF
--- a/mlx_lm/models/mamba.py
+++ b/mlx_lm/models/mamba.py
@@ -131,8 +131,7 @@ class MambaBlock(nn.Module):
         else:
             x_full = mx.pad(x, [(0, 0), (K - 1, 0), (0, 0)])
 
-        conv_full = self.conv1d(x_full)
-        conv_out = conv_full[:, -T:, :]
+        conv_out = self.conv1d(x_full)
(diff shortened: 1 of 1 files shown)


mlx_lm/models/mamba.py
```diff
@@ -149,9 +125,17 @@ def _process_sequence(self, x, conv_cache, state_cache):
...
-        conv_out, new_conv_cache = self.conv1d(x, conv_cache)
-        x = nn.silu(conv_out)
+        K = self.conv_kernel_size
+        if conv_cache is not None:
+            x_full = mx.concatenate([conv_cache, x], axis=1)
+        else:
+            x_full = mx.pad(x, [(0, 0), (K - 1, 0), (0, 0)])
 
+        conv_full = self.conv1d(x_full)
+        conv_out = conv_full[:, -T:, :]
```
awni: I don't think you need this slice. The convolution output will always be size `T` if the input is size `T + K - 1`.

Goekdeniz-Guelmez: Removed it :D

awni (approved):
Looks good, thanks!

PR #443, opened 2025-09-09
Batch generation

awni:
There is a lot going on here.. the basics for batch generation are mostly here but it needs some tuning / refinement for perf.

The main thing to review here the APIs that are introduced. There are a few note-able ones:

### Benchmarks:

```
mlx_lm.evaluate --model Qwen/Qwen3-4B-Instruct-2507 --task mmlu_pro --limit 20 --max-tokens 2048
```
Edit: results are close (larger eval in progress):
Single:     "exact_match,custom-extract": 0.6392857142857142,
Batch:     "exact_match,custom-extract": 0.6535714285714286,

Batch is about 5x faster:
Single: 53:06.68 total
Batch: 11:47.11 total

### User facing API:

A high level API for generating completions for a bunch of prompts:

```python
texts, stats = batch_generate(model, tokenizer, prompts)
```

Q: maybe `batch_generate` should return a dictionary or a data class to make it easier to update the return type?

A lower level API when more flexibility is needed (streaming, logprobs, raw tokens, etc). Basically how `batch_generate` is implemented:

```python
    gen = BatchGenerator(model, stop_tokens=tokenizer.eos_token_ids)
    # `insert` can be called anytime (not just at the beginning). 
    # `max_tokens` can be per prompt or a single integer
    # uids are the unique identifier for each prompt (so we can assemble the responses)
    uids = gen.insert(prompts, max_tokens)
    results = {uid: [] for uid in uids}

    # as long as there is an unfinished prompt `gen.next()` will not be empty
    while responses := gen.next():
        # each response in the list has metadata along with the token including: uid, logprobs, finish_reason
        for r in responses:
             results[r.uid].append(r.token)

     texts = [tokenizer.decode(results[uid]) for uid in uids]

    # Get decoding stats (like toks/sec)
    gen_tps = gen.stats().generation_tps
```

Q: responses is a list of dataclasses. It could also be a dictionary keyed by the uid.. that may simplify things in some cases.

### Internal 

Probably the most note-able change internally is the addition of the `BatchKVCache` which manages the offsets per example and the mask making. Only `BatchKVCache` works with batch generation and for other cache types they will need a corresponding batch version.

The BatchKVCache has an API / calculus to facilitate the continuous batching. All future batch caches would need to implement these to work with the batch generation / continuous batching.

```python
# Extend a cache in place
cache.extend(other)

# In place filter the cache to contain just the given ids
cache.filter(idx)
```

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -23,8 +23,7 @@ from lm_eval.api.registry import register_model
 from lm_eval.models import huggingface
 from tqdm import tqdm
 
-from .generate import stream_generate
-from .models.base import create_causal_mask
+from .generate import batch_generate
 from .models.cache import make_prompt_cache
 from .utils import common_prefix_len, load
 
@@ -69,7 +68,7 @@ class MLXLM(LM):
     def __init__(
         self,
         path_or_hf_repo: str,
-        max_tokens: Optional[int] = None,
+        max_tokens: int,
         use_chat_template: Optional[bool] = None,
         trust_remote_code: bool = False,
     ) -> None:
@@ -78,7 +77,7 @@ class MLXLM(LM):
         self._model, self.tokenizer = load(
             path_or_hf_repo, tokenizer_config=tokenizer_config
         )
-        self._max_tokens = max_tokens or self.tokenizer.model_max_length
+        self._max_tokens = max_tokens
         self._batch_size = 8
         self.use_chat_template = use_chat_template
         if use_chat_template is None:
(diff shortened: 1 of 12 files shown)


awni:
This is ready for an initial review. Though will require an update to MLX before it can be landed.

The main TODOs (which will probably be follow up PRs):
- Add batching support for other model types (sliding window, mamba, conv)
- Tune perf

ivanfioravanti:
This is just AMAZING!!!!!

N8python:
When temp=0 do the generations match into the hundreds of tokens?

N8python:
And also - seriously amazing work. This will enable so much!!!

awni:
I’m pretty sure they will not especially for bf16 and fp16. And it’s not reasonable to expect them to. Tiny numerical errors will accumulate and result in different tokens after a while. 

This boils down to the fact that a matrix matrix op is not bit wise identical to the concatenation of matrix vector ops which boils down to non associative fp math.

N8python:
I meant diverge no more than two single-threaded temp=0 rollouts would. But the reason I was wondering is cause in MMLU_Pro they both got the exact same score:

Results match:
Single: "exact_match,custom-extract": 0.6535714285714286,
Batch: "exact_match,custom-extract": 0.6535714285714286,

awni:
Too temp=0 single thread roll outs should not diverge at all. So yea it should diverge more than that.

Regarding the exact match of the scores.. it’s interesting.. I’m not sure if it’s suspicious or not. Maybe we should run it again to be sure.

N8python:
What temp did you run the bench on?

awni:
Default which should be 0.

N8python:
Wow, that's incredibly consistent. Guess mild per-token drift really doesn't matter for MMLU Pro. Promising.

ivanfioravanti:
I have so many batch scenarios that will benefit from this! Thanks @awni 🙏

awni:
I'm starting to think I copied the results incorrectly.. I reran the eval in single and batch mode and I got different results this time:

Single: `"exact_match,custom-extract": 0.6392857142857142,`
Batch: `"exact_match,custom-extract": 0.653571428571428`

The batch is the same but the single is different.

Command:
```
mlx_lm.evaluate --model Qwen/Qwen3-4B-Instruct-2507 --task mmlu_pro --limit 20 --max-tokens 2048
```

I'll run a larger subset to see if they two numbers converge to something close

mlx_lm/evaluate.py
```diff
@@ -348,7 +350,8 @@ def main():
     parser.add_argument(
         "--max-tokens",
         type=int,
-        help="Maximum nunber of tokens to generate. Defaults to the model's max context length.",
+        help="Maximum nunber of tokens to generate.",
```
ivanfioravanti: @awni small typo here: nunber instead of number.

awni: Thanks!

awni:
Using a larger subset (limit 50) the two methods get a lot closer:

Single: "exact_match,custom-extract": 0.6557142857142857,
Batch: "exact_match,custom-extract": 0.6514285714285715

So I think the variance is reasonable.. and the results are fine to within numerical precision. Running the full MMLU (which unfortunately takes a long time for non-batch) should not show statistically significant difference.

operator911:
OMG OMG this is going to be amazing....

awni:
I'm going to merge this. Keep in mind it only supports models with global attention for now. Next step is to add other models types (e.g. Mamba-2 style, sliding window attention, etc). @Goekdeniz-Guelmez I will incorporate your comments in a follow on PR.

Also the API is still subject to change as we understand the requirements here better.

PR #446, opened 2025-09-10
Update bitnet, nemotron h to use build in reel from MLX

Goekdeniz-Guelmez:


Goekdeniz-Guelmez:
will be set to ready for review after th new versions of mlx is released

PR #451, opened 2025-09-12
Faster ssm

awni:
Adds a custom SSM kernel + some optimizations to SSM.

Since these are more common.. I think it's worth having a common + fast implementation that hybrid SSM models can use.

TODO:

- Use ssm step for other models where possible
- Possible speed up prompt processing using scan formulation instead of recurrent formulation.

### DIFF
--- a/mlx_lm/models/ssm.py
+++ b/mlx_lm/models/ssm.py
@@ -1,0 +1,108 @@
+import mlx.core as mx
+import mlx.nn as nn
+
+
+@mx.compile
+def compute_dt(dt, dt_bias, time_step_limit):
+    dt = nn.softplus(dt + dt_bias)
+    return mx.clip(dt, time_step_limit[0], time_step_limit[1])
+
+
+def ssm_step_ops(
+    hidden_states: mx.array,
+    A_log: mx.array,
+    B: mx.array,
+    C: mx.array,
+    D: mx.array,
+    dt: mx.array,
+    dt_bias: mx.array,
+    state: mx.array,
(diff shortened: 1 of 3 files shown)


mlx_lm/models/ssm.py
```diff
@@ -0,0 +1,108 @@
...
+    for t in range(seq_len):
+        state = dA[:, t, :, None, None] * state + dB_h[:, t]
+        hs.append(state)
+    Dh = D[:, None] * hidden_states
+    y = (mx.stack(hs, axis=1) @ C[..., None]).squeeze(-1) + Dh
+    return y, state
+
+
+def make_ssm_kernel():
+    source = f"""
```
angeloskath: Maybe remove `f`?

mlx_lm/models/ssm.py
```diff
@@ -0,0 +1,108 @@
...
+        input_names=["X", "A_log", "B", "C", "D", "dt", "state_in"],
+        output_names=["out", "state_out"],
+        source=source,
+    )
+
+
+_ssm_kernel = make_ssm_kernel()
+
+
+def ssm_step(
```
angeloskath, line 91: Maybe fallback if `not mx.metal.is_available()` ?

awni:
Benchmarks:

Pre:
```
Averages: prompt_tps=439.259, generation_tps=108.173, peak_memory=4.349
```

Post
```
Averages: prompt_tps=713.992, generation_tps=131.765, peak_memory=5.770
```

The faster prompt processing uses more memory

PR #474, opened 2025-09-17
Fix llama4 text and make trainable

Goekdeniz-Guelmez:


awni:
I'm curious why did you change this? Is it for a specific model that has those features?

Goekdeniz-Guelmez:
since these are in the config, I suspect the future models could have them.

awni:
I think the reason they are in the config is because `llama4` uses the config as `llama4_text`

For the chunk attention and attention tuning stuff I think it's better to assert that it's not on for this model. If it gets turned on in the future for a llama4_text model we can add it then. Otherwise better to keep it simple. Do you mind making that change?

Goekdeniz-Guelmez:
sure! one sec.

Goekdeniz-Guelmez:
OK reverted back, but added type definitions, and used scaled_dot_prod. from base.py. can be merged now!

PR #479, opened 2025-09-18
[Batch_generate] return tokens and logprobs

charbull:
On device, returning tokens and logprobs would allow to save one additional call to get the logits.
Added two boolean that defaults to false to return the logits and the tokens


awni:
For now - the best way to get this info is to use the `BatchGenerator` directly rather than wiring everything through the high-level API which we would like to keep simple. 

We can think about having a streaming high-level API similar to `stream_generate` which returns richer responses. But it would basically be a trivial wrapper around `BatchGenerator` so might as well use that for now.

PR #486, opened 2025-09-19
feat: load generation_config.json

reneleonhardt:
### Feature
- Add generation_config.json to config.json if it exists

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -89,7 +89,7 @@ def can_trim_prompt_cache(cache: List[Any]) -> bool:
     return all(c.is_trimmable() for c in cache)
 
 
-def trim_prompt_cache(cache: List[Any], num_tokens: int) -> List[Any]:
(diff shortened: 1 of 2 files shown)


awni:
I think the right thing to do is to read the generation config and use it to set the sampling parameters. Not sure what is the purpose of this PR?

reneleonhardt:
Why did you close it without waiting for my reply?
If you know where to start you can create a pull request.

awni:
@reneleonhardt what I would do is something like:
- Read the generation config (if present) when loading the rest of the model
- Use the generation config to make a sampler which gets passed in to `generate` or `stream_generate`

The reason I closed this is because it doesn't look like a step in that direction. But feel free to comment if you disagree or if you intend to change it. It's easy to re-open it.

Otherwise I will send a PR with something like the above.

reneleonhardt:
I don't know why maintainers keep saying that it's easy to re-open a pull request after they just closed it.
It's impossible, a contributor has no permission to do so.

awni:
It's easy for *me* to reopen 😅

PR #500, opened 2025-09-24
Remove act loss and add temp in DWQ

awni:


awni:
I'm going to move this out of draft. In all cases it works as well or better not to use the activation loss (and it's faster).

I was hoping to improve even more.. but I think we should go ahead and merge it.

PR #505, opened 2025-09-26
Add Code World Model support

dnakov:
Support for facebook/cwm


awni:
I think it makes more sense to add sliding window attention to the `llama.py` file rather than making a separate model for cwm. Wdyt? It looks like the same model otherwise.

dnakov:
yeah makes sense, ill update

dnakov:
done, moved to llama.py now based on the sliding_attention layer type.

awni (approved):
Ty!

PR #326, opened 2025-07-23
Adding support to set --gradient-accumulation-steps in training

Goekdeniz-Guelmez:

PR #511, opened 2025-09-29
Added gradient accumulation to training loop

dotvignesh:
This adds `--gradient-accumulation-steps` to the trainer to match a core feature in Hugging Face Transformers. 

**Details:**
* New CLI flag: `--gradient-accumulation-steps` (default `1`, no behavior change).
* Accumulates micro-batch gradients and applies one optimizer update per window.

**Why:**
* Feature parity with HF Transformers.
* Enables larger effective batches on limited memory.

### DIFF
--- a/mlx_lm/LORA.md
+++ b/mlx_lm/LORA.md
@@ -82,6 +82,10 @@ You can specify the output location with `--adapter-path`.
 You can resume fine-tuning with an existing adapter with
 `--resume-adapter-file <path_to_adapters.safetensors>`.
 
+If you need an effective batch size larger than what fits in memory, set
+`--gradient-accumulation-steps <N>` to accumulate gradients across `N`
+micro-batches before each optimizer update.
+
 #### Logging
 
 You can log training metrics to Weights & Biases using `--report-to wandb`, or
--- a/mlx_lm/examples/lora_config.yaml
+++ b/mlx_lm/examples/lora_config.yaml
@@ -47,6 +47,9 @@ steps_per_report: 10
 # Number of training steps between validations.
 steps_per_eval: 200
 
+# Number of micro-steps to accumulate before each optimizer update.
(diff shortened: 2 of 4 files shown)


angeloskath:
This is simpler than #326 but it is still not enough to merge. For example the `step` that used to include the optimizer is no longer compiled and `average_gradients` is called for every micro-step while it need only be called once before the update of the parameters.

Something like the following would be much better.

```python
@partial(mx.compile, inputs=state, outputs=state)
def step(batch, prev_grad, update):

    if prev_grad is not None:
        grad = tree_map(lambda x, y: x + y, grad, prev_grad)

    if update:
        grad = average_gradients(grad)
        if grad_accum_steps > 1:
            grad = tree_map(lambda x: x / grad_accum_steps, grad)
        optimizer.update(model, grad)
        grad = None

    return lvalue, toks, grad
```

dotvignesh:
@angeloskath makes sense! Made few updates to incorporate the suggested changes.

dotvignesh:
> I have been awating an implementation of gradient accumulation into mlx-lm. Was there any problem for merging of this commit to main branch

Not sure - it was supposed to be merged along with PR #2634
Maybe @angeloskath / @awni held it up for some other reason.

angeloskath:
Sorry @dotvignesh I was waiting for #2634 to go into a release but I will merge it and make sure we do a patch release soon, at least before we do another release for mlx-lm. To use it you should build MLX from source though (and install mlx-lm from main as well of course until the next release.)

dotvignesh:
> Sorry @dotvignesh I was waiting for #2634 to go into a release but I will merge it and make sure we do a patch release soon, at least before we do another release for mlx-lm. To use it you should build MLX from source though (and install mlx-lm from main as well of course until the next release.)

No worries at all - the apology wasn’t needed :)
Thanks for the update!

PR #512, opened 2025-09-29
[WIP] deepseek v32

awni:


awni:
This is functional... but unfortunately quite slow. You pay a penalty for the topk gather for shorter context. And from some micro benchmarks, I think it would take fairly long context for this to be more efficient than the naive attention.

We can maybe improve on the current implementation with a `gather_mm` for the `q @ k[idx].T` matmul. But I don't think we have an op for the `scores @ v[idx]` to do it in one go.

Here's the micro benchmark for reference:
```python
import copy
import mlx.core as mx
from mlx_lm import load
from mlx_lm.models.cache import CacheList, KCache, KVCache

import time

model, _ = load("mlx_model", lazy=True)

attn = model.layers[0].self_attn
#attn.indexer.index_topk = float("inf")

cache = CacheList(KVCache(), KCache())

d = 7168
l = 32768
x = mx.random.normal(shape=(1, l, d)).astype(mx.bfloat16)

attn(x, cache=cache)
mx.eval(cache.state)

q = mx.random.normal(shape=(1, 1, d)).astype(mx.bfloat16)

def fn(x, cache):
    c = copy.deepcopy(cache)
    for _ in range(10):
        x = attn(x, cache=c)
    return x

# warmup
for _ in range(10):
    mx.eval(fn(q, cache))

tic = time.time()
for _ in range(10):
    mx.eval(fn(q, cache))
toc = time.time()
print(toc - tic)
```

mlx_lm/models/cache.py
```diff
@@ -659,6 +659,55 @@ def extend(self, other):
             c.extend(other)
 
 
+class KCache(_BaseCache):
```
awni: Maybe it makes sense to make `KVCache` general to an arbitrary number of arrays?

kernelpool:
I've been doing some testing on this and the current implementation quickly breaks down if you ask the model to summarize a large function (e.g. 4k tokens long, embedded in the prompt), resulting in incoherent and/or repetitive output.

After looking through other implementations, including [deepseek-ai/DeepSeek-V3.2-Exp](https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/tree/main/inference), I was able to eventually fix the issue (missing sum over heads). But you probably want to review these first, so I've just provided the diff below.

```
diff --git a/mlx_lm/models/deepseek_v32.py b/mlx_lm/models/deepseek_v32.py
index f811253..1ef421c 100644
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -101,10 +101,23 @@ class Indexer(nn.Module):
             k = cache.update_and_fetch(k)
         if k.shape[2] <= self.index_topk:
             return None
+
+        # Compute Q @ K^T
+        scores = q @ k.swapaxes(-1, -2)
+
+        # Apply ReLU activation (critical for correct sparse attention)
+        scores = mx.maximum(scores, 0)
+
+        # Compute per-head weights and apply after matmul
         weights = self.weights_proj(x) * (self.n_heads**-0.5)
         weights = (weights * self.softmax_scale).swapaxes(-1, -2)[..., None]
-        q_scaled = q * weights
-        scores = (q * weights) @ k.swapaxes(-1, -2)
+        scores = scores * weights
+
+        # Sum over heads (matches reference kernel.py:244)
+        # Output shape: o: T.Tensor[(b, m, n), FP32] - no heads dimension
+        scores = scores.sum(axis=1)  # (b, s, k_seq)
+
+        # Apply mask and select top-k indices
         if mask is not None:
             scores = mx.where(mask, scores, -float("inf"))
         return mx.argpartition(scores, kth=-self.index_topk, axis=-1)[
@@ -188,6 +201,8 @@ class DeepseekV32Attention(nn.Module):
 
         if self.q_lora_rank is None:
             q = self.q_proj(x)
+            # Note: DeepSeek v3.2 always uses LoRA, but handle this path defensively
+            qr = x  # Fallback for indexer (shouldn't be needed in practice)
         else:
             qr = self.q_a_layernorm(self.q_a_proj(x))
             q = self.q_b_proj(qr)
@@ -217,26 +232,47 @@ class DeepseekV32Attention(nn.Module):
             keys = mx.concatenate([k_nope, k_pe], axis=-1)
 
         queries = mx.concatenate([q_nope, q_pe], axis=-1)
-        topk_indices = self.indexer(x, qr, mask, cache=cache[1])
+        # Indexer needs 3D mask (B, L, L) or None
+        # create_attention_mask returns: None, "causal", or 2D (L, L)
+        # But handle 4D (B, num_heads, L, L) for tests/edge cases
+        if mask is None or isinstance(mask, str):
+            indexer_mask = None
+        elif mask.ndim == 2:
+            # 2D (L, L) -> 3D (B, L, L)
+            indexer_mask = mx.broadcast_to(mask, (B, *mask.shape))
+        elif mask.ndim == 4:
+            # 4D (B, num_heads, L, L) -> 3D (B, L, L)
+            indexer_mask = mask[:, 0, :, :]
+        else:
+            # Already 3D (B, L, L)
+            indexer_mask = mask
+
+        topk_indices = self.indexer(x, qr, indexer_mask, cache=cache[1])
         if topk_indices is not None:
-            repeats = self.num_heads // self.config.index_n_heads
-            if L == 1:
-                topk_indices = mx.repeat(topk_indices, repeats, axis=1).squeeze(-2)[
-                    ..., None
-                ]
-                keys = mx.take_along_axis(keys, topk_indices, axis=-2)
-                values = mx.take_along_axis(values, topk_indices, axis=-2)
-            else:
-                topk_mask = mx.zeros(
-                    (B, self.config.index_n_heads, *mask.shape[-2:]), mx.bool_
-                )
-                topk_mask = mx.put_along_axis(
-                    topk_mask, topk_indices, mx.array(True), axis=-1
-                )
-                mask = mask & topk_mask
-                mask = mx.repeat(mask, repeats, axis=1)
+            # topk_indices shape: (B, L, topk) - no heads dimension after sum
+            # Reference: scatter topk_indices into -inf mask, then add to existing mask
+            # Create sparse mask (works for both decode and prefill)
+            k_seq = keys.shape[2]
+            # Create boolean sparse mask: True at topk positions
+            sparse_mask = mx.zeros((B, L, k_seq), dtype=mx.bool_)
+            ones = mx.ones(topk_indices.shape, dtype=mx.bool_)
+            sparse_mask = mx.put_along_axis(sparse_mask, topk_indices, ones, axis=-1)
+
+            # Combine with causal mask using AND
+            if mask is not None and not isinstance(mask, str):
+                # Convert to (B, L, k_seq)
+                if mask.ndim == 2:
+                    causal_mask = mx.broadcast_to(mask, sparse_mask.shape)
+                elif mask.ndim == 4:
+                    causal_mask = mask[:, 0, :, :]
+                else:
+                    causal_mask = mask
+                sparse_mask = sparse_mask & causal_mask
+
+            # Broadcast to (B, num_heads, L, k_seq)
+            mask = sparse_mask[:, None, :, :]
-            queries, keys, values, cache=cache, scale=self.scale, mask=mask
+            queries, keys, values, cache=cache[0], scale=self.scale, mask=mask
         )
         output = output.transpose(0, 2, 1, 3).reshape(B, L, -1)
         return self.o_proj(output)

```

awni:
@kernelpool thanks for the fix. That's awesome! Do you mind sending a PR? I have some questions/comments but they are better made in-line.

awni:
I think we can merge this now

mlx_lm/models/deepseek_v32.py
```diff
@@ -0,0 +1,523 @@
...
+            self.q_lora_rank, self.n_heads * self.head_dim, bias=False
+        )
+        self.wk = nn.Linear(self.dim, self.head_dim, bias=False)
+        self.k_norm = nn.LayerNorm(self.head_dim)
+        self.weights_proj = nn.Linear(self.dim, self.n_heads, bias=False)
+        self.softmax_scale = self.head_dim**-0.5
+        self.rope = initialize_rope(
+            dims=args.qk_rope_head_dim,
+            base=args.rope_theta,
+            traditional=True,
```
kernelpool:
```suggestion
            traditional=False,
```
rope in indexer is not interleaved  (see https://x.com/deepseek_ai/status/1990616182161068094 and https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/commit/8631a813356d39b09a5c1ee1cde8ed6015559eb3)

awni: Great catch!

mlx_lm/models/deepseek_v32.py
```diff
@@ -0,0 +1,523 @@
...
+        k_pe = self.rope(k_pe, offset=offset)
+        k = mx.concatenate([k_pe, k_nope], axis=-1)
+        if cache is not None:
+            k, _ = cache.update_and_fetch(k, mx.zeros([b, 1, s, 0]))
+        if k.shape[2] <= self.index_topk:
+            return None
+        scores = q @ k.swapaxes(-1, -2)
+        scores = mx.maximum(scores, 0)
+        weights = self.weights_proj(x) * (self.n_heads**-0.5)
+        weights = (weights * self.softmax_scale).swapaxes(-1, -2)[..., None]
```
angeloskath:
```suggestion
        weights = self.weights_proj(x) * (self.n_heads**-0.5 * self.softmax_scale)
        weights = weights.swapaxes(-1, -2)[..., None]
```

Nitpick

mlx_lm/models/deepseek_v32.py
```diff
@@ -0,0 +1,523 @@
...
+            )
+        else:
+            cache = [None] * 2
+            q_pe = self.rope(q_pe)
+            k_pe = self.rope(k_pe)
+            k_pe = mx.repeat(k_pe, self.num_heads, axis=1)
+            keys = mx.concatenate([k_nope, k_pe], axis=-1)
+
+        queries = mx.concatenate([q_nope, q_pe], axis=-1)
+        topk_indices = self.indexer(x, qr, mask, cache=cache[1])
```
angeloskath, line 214: If `q_lora_rank is None` then `qr` is not defined so this line will fail. Check here or in the indexer maybe?

PR #518, opened 2025-10-01
feat: Refactor granitemoehybrid to support dense and non-hybrid variants

gabe-l-hart:
## Description

This PR extends the `granitemoehybrid.py` model implementation to support "degenerate" cases of the architecture where a dense block is used in place of an MoE and/or all layers are attention layers (non-hybrid). There will be some models in the granite 4 family that require these cases.

## Notes

In order to get _any_ of these models to run locally on this branch, I also had to remove [this line](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/base.py#L136). The version of `mlx` I have (installed with `pip install -e` from the tip of `main` / `bbf14239`) does not seem to support the `sinks` argument, though I do see it implemented [here](https://github.com/ml-explore/mlx/blob/bbf14239538c18a9f8a3d61f52ab0f29589410f0/python/src/fast.cpp#L237), so it's possible that something didn't get recompiled correctly. Either way, the addition of this flag forces the lower-bound on `mlx.core` up to a version that includes that in the function signature. This could be worked around to be backwards-compatible with either a runtime version check or a `try/except` that falls back to calling without `sinks`. Alternately, the signature with `sinks` could be called only if `sinks is not None`.

## Details

Written with Claude Code. Initial prompt:

---

I need to modify the model support implemented in `mlx_lm/models/granitemoehybrid.py` in two ways:

* Support optionally using a dense block in place of MoE. The dense block should look like `mlx_lm/models/granite.py` instead of `mlx_lm/models/granitemoe.py`.

* Support the case where there are no `mamba` layers (ie non-hybrid). This should devolve to exactly `granite.py` or `granitemoe.py` depending on whether the block after attention is dense or MoE.

You can test this using the following two models:

* Dense w/ hybrid: /Users/ghart/models/dmf_models/granite-4.0-h-micro-r250918a
* Dense w/ non-hybrid: /Users/ghart/models/dmf_models/granite-4.0-micro-r250918a

---

Signed-off-by: Gabe Goodhart <ghart@us.ibm.com>

### DIFF
--- a/mlx_lm/_version.py
+++ b/mlx_lm/_version.py
@@ -1,3 +1,3 @@
 # Copyright © 2023-2025 Apple Inc.
 
-__version__ = "0.28.1"
+__version__ = "0.28.2"
--- a/mlx_lm/models/granitemoehybrid.py
+++ b/mlx_lm/models/granitemoehybrid.py
@@ -417,9 +417,7 @@ class GraniteMoeHybridModel(nn.Module):
             else None
         )
         self.ssm_idx = (
-            args.layer_types.index("mamba")
-            if "mamba" in args.layer_types
-            else None
+            args.layer_types.index("mamba") if "mamba" in args.layer_types else None
         )
 
     def __call__(
@@ -492,7 +490,10 @@ class Model(nn.Module):
                 weights[k] = v.moveaxis(2, 1)
 
         # Handle MoE weight transformation to SwitchGLU format (only for MoE models)
-        if self.args.use_moe and "model.layers.0.block_sparse_moe.input_linear.weight" in weights:
+        if (
+            self.args.use_moe
+            and "model.layers.0.block_sparse_moe.input_linear.weight" in weights
+        ):
             for l in range(self.args.num_hidden_layers):
                 prefix = f"model.layers.{l}.block_sparse_moe"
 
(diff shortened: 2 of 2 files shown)


awni:
Thanks for the contribution. Are there any models we can test this against or they are not yet available?

> so it's possible that something didn't get recompiled correctly.

Most likely that. 

The latest mlx-lm requires a recent version of mlx. I recommend upgrading both. Usually we pin a minimum MLX version for each mlx-lm version. Below that version it may not be compatible.

gabe-l-hart:
> Most likely that.

That's what I figured, so didn't include the change here. I'm not sure why my environment didn't pick up the right newer version with `pip install -U`. I'll try a fresh install tomorrow. 

> are they not yet available?

They are still private, but they'll be out soon. I've tested both configurations (hybrid-dense and non-hybrid-dense) and both work as expected.

awni (approved):
Looks good, thanks!

awni:
Do you mind running the pre-commit format hooks?

gabe-l-hart:
Dang, I missed pre-commit. Will run shortly

gabe-l-hart:
The models are out now!

* Non-hybrid / dense: https://huggingface.co/ibm-granite/granite-4.0-micro
* Hybrid / dense: https://huggingface.co/ibm-granite/granite-4.0-h-micro

PR #523, opened 2025-10-02
Add epoch argument in trainer

Goekdeniz-Guelmez:
default is iters but when given epochs, it will overnight the iters.


Goekdeniz-Guelmez:
@awni should be mergable, added the function into the utils for mlx-lm-lora and other as well.

angeloskath (commented):
Hi @Goekdeniz-Guelmez , I don't think we need to add this option. Defining iterations is fine. Has someone expressed interest for this feature before?

Goekdeniz-Guelmez:
Hey @angeloskath, not really, only me tbh, it was more of a personal convenience. I’m fine with closing this since there isn’t clear demand yet. If more people bring it up in the future, happy to revisit.

chimezie:
@Goekdeniz-Guelmez and co. I actually have had a need for this from day one, given the convenience of not having to do the arithmetic of specifying the number of iterations and the fact that specifying epochs allows you to do this in a manner agnostic to the training data size.

Goekdeniz-Guelmez:
If more people reply here, then we should consider this, thats why I opened it again.

zcbenz:

PR #531, opened 2025-10-06
Fix sparse token selection in deepseek v3.2

kernelpool:
The following fixes/changes are included in this PR

- Added missing ReLU activation (`mx.maximum(scores, 0)`) in indexer after computing attention scores, as implemented by reference [kernel.py:241](https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/main/inference/kernel.py#L241) to filter out negative attention scores before token selection.

- Changed `scaled_dot_product_attention(..., cache=cache)` to `cache=cache[0]` to pass `KVCache` object instead of `CacheList` wrapper

- Added `scores.sum(axis=1)` in indexer to aggregate per-head scores into unified token selection (matching [kernel.py:244](https://github.com/deepseek-ai/DeepSeek-V3.2-Exp/blob/main/inference/kernel.py#L244)), changing output shape from `(B, heads, L, topk)` to `(B, L, topk)` and fixing repetitive generation caused by each head selecting different tokens.

- Rewrote attention layer's sparse token handling to create boolean masks instead of per-head direct indexing, since indexer now returns unified indices for all heads rather than per-head indices.

- Unified decode/prefill sparse masking paths to handle chunked prefill correctly by always using `k_seq = keys.shape[2]` instead of assuming square masks.

### DIFF
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -101,10 +101,12 @@ class Indexer(nn.Module):
             k = cache.update_and_fetch(k)
         if k.shape[2] <= self.index_topk:
             return None
+        scores = q @ k.swapaxes(-1, -2)
+        scores = mx.maximum(scores, 0)
         weights = self.weights_proj(x) * (self.n_heads**-0.5)
         weights = (weights * self.softmax_scale).swapaxes(-1, -2)[..., None]
-        q_scaled = q * weights
-        scores = (q * weights) @ k.swapaxes(-1, -2)
+        scores = scores * weights
+        scores = scores.sum(axis=1)
         if mask is not None:
             scores = mx.where(mask, scores, -float("inf"))
         return mx.argpartition(scores, kth=-self.index_topk, axis=-1)[
@@ -219,24 +221,15 @@ class DeepseekV32Attention(nn.Module):
         queries = mx.concatenate([q_nope, q_pe], axis=-1)
         topk_indices = self.indexer(x, qr, mask, cache=cache[1])
         if topk_indices is not None:
-            repeats = self.num_heads // self.config.index_n_heads
-            if L == 1:
-                topk_indices = mx.repeat(topk_indices, repeats, axis=1).squeeze(-2)[
-                    ..., None
-                ]
-                keys = mx.take_along_axis(keys, topk_indices, axis=-2)
-                values = mx.take_along_axis(values, topk_indices, axis=-2)
-            else:
-                topk_mask = mx.zeros(
-                    (B, self.config.index_n_heads, *mask.shape[-2:]), mx.bool_
-                )
-                topk_mask = mx.put_along_axis(
-                    topk_mask, topk_indices, mx.array(True), axis=-1
-                )
-                mask = mask & topk_mask
-                mask = mx.repeat(mask, repeats, axis=1)
+            k_seq = keys.shape[2]
+            sparse_mask = mx.zeros((B, L, k_seq), dtype=mx.bool_)
+            ones = mx.ones(topk_indices.shape, dtype=mx.bool_)
+            sparse_mask = mx.put_along_axis(sparse_mask, topk_indices, ones, axis=-1)
+            if mask is not None:
+                sparse_mask = sparse_mask & mask
+            mask = sparse_mask[:, None, :, :]
         output = scaled_dot_product_attention(
-            queries, keys, values, cache=cache, scale=self.scale, mask=mask
+            queries, keys, values, cache=cache[0], scale=self.scale, mask=mask
         )
         output = output.transpose(0, 2, 1, 3).reshape(B, L, -1)
         return self.o_proj(output)
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -223,11 +223,11 @@ class DeepseekV32Attention(nn.Module):
         if topk_indices is not None:
             k_seq = keys.shape[2]
             sparse_mask = mx.zeros((B, L, k_seq), dtype=mx.bool_)
-            ones = mx.ones(topk_indices.shape, dtype=mx.bool_)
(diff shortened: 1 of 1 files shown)


kernelpool:
Added some supporting benchmarks and evaluations using **DeepSeek-V3.2-Exp-3bit**

# Benchmarks

`mlx_lm.benchmark --model /Volumes/WD_EXTRA/models/catalyst/DeepSeek-V3.2-Exp-3bit -p <n> -g 128`

  Prompt Processing Speed (tokens/sec)

  | Context Length | Without PR | With PR | Improvement |
  |----------------|------------|---------|-------------|
  | 512 tokens     | 173.030    | 172.763 | -0.15%      |
  | 1024 tokens    | 199.939    | 199.922 | -0.01%      |
  | 2048 tokens    | 209.701    | 209.912 | +0.10%      |
  | 4096 tokens    | 157.607    | 187.479 | +18.9% ✅    |

  Generation Speed (tokens/sec)

  | Context Length | Without PR | With PR | Improvement |
  |----------------|------------|---------|-------------|
  | 512 tokens     | 21.157     | 21.162  | +0.02%      |
  | 1024 tokens    | 19.675     | 19.565  | -0.56%      |
  | 2048 tokens    | 7.668      | 15.005  | +95.7% ✅    |
  | 4096 tokens    | 7.628      | 12.131  | +59.0% ✅    |

  Peak Memory Usage (GB)

  | Context Length | Without PR | With PR | Improvement |
  |----------------|------------|---------|-------------|
  | 512 tokens     | 298.082    | 298.093 | 0.00%       |
  | 1024 tokens    | 301.330    | 301.330 | 0.00%       |
  | 2048 tokens    | 308.636    | 308.636 | 0.00%       |
  | 4096 tokens    | 327.799    | 318.847 | -8.95 GB ✅  |
  
# Perplexity
  
  `python ppl.py --model /Volumes/WD_EXTRA/models/catalyst/DeepSeek-V3.2-Exp-3bit --sequence-length <n>`

  | Seq Length | Without PR | With PR | Delta    | Time (s) | Peak Mem (GB) | Tokens/sec | Samples |
  |------------|------------|---------|----------|----------|---------------|------------|---------|
  | 512        | 4.477      | 4.477   | 0.000    | 436      | 279           | 250        | 213     |
  | 1024       | 3.847      | 3.847   | 0.000    | 436      | 279           | 500        | 213     |
  | 2048       | 3.589      | 3.589   | 0.000    | 437      | 279           | 1000       | 213     |
  | 4096       | ⚠️ 725.967 | 3.872   | -722.095 | 439      | 279           | 1991       | 213     |

awni:
First of all thanks for the thorough benchmarks and generally high quality work / attention to detail. This is super useful!

I'm very surprised by the speedup you are observing in token generation speed from this. I'd like to understand that better.

mlx_lm/models/deepseek_v32.py
```diff
@@ -219,24 +221,15 @@ def __call__(
...
-                )
-                mask = mask & topk_mask
-                mask = mx.repeat(mask, repeats, axis=1)
+            k_seq = keys.shape[2]
+            sparse_mask = mx.zeros((B, L, k_seq), dtype=mx.bool_)
+            ones = mx.ones(topk_indices.shape, dtype=mx.bool_)
+            sparse_mask = mx.put_along_axis(sparse_mask, topk_indices, ones, axis=-1)
+            if mask is not None:
+                sparse_mask = sparse_mask & mask
+            mask = sparse_mask[:, None, :, :]
```
awni: The `&` here should be done after expanding the head dimension inot the `sparse_mask`. Because in some cases we can get a 4D mask as input (when using batch generation for example).

mlx_lm/models/deepseek_v32.py
```diff
@@ -219,24 +221,15 @@ def __call__(
...
         output = scaled_dot_product_attention(
-            queries, keys, values, cache=cache, scale=self.scale, mask=mask
+            queries, keys, values, cache=cache[0], scale=self.scale, mask=mask
```
awni, line 232: Good catch!

mlx_lm/models/deepseek_v32.py
```diff
@@ -219,24 +221,15 @@ def __call__(
...
-                )
-                topk_mask = mx.put_along_axis(
-                    topk_mask, topk_indices, mx.array(True), axis=-1
-                )
-                mask = mask & topk_mask
-                mask = mx.repeat(mask, repeats, axis=1)
+            k_seq = keys.shape[2]
+            sparse_mask = mx.zeros((B, L, k_seq), dtype=mx.bool_)
+            ones = mx.ones(topk_indices.shape, dtype=mx.bool_)
+            sparse_mask = mx.put_along_axis(sparse_mask, topk_indices, ones, axis=-1)
```
awni: It's not necessary to make a `ones` array. You can just put `mx.array(True)` and let broadcasting handle it:
```suggestion
            sparse_mask = mx.put_along_axis(sparse_mask, topk_indices, mx.array(True), axis=-1)
```

awni:
> I'm very surprised by the speedup you are observing in token generation speed from this. I'd like to understand that better.

I think removing the take on the keys and values and using the boolean mask is what sped this up.

Very nice though. Would you mind addressing the comments and then we can merge it?

mlx_lm/models/deepseek_v32.py
kernelpool: Oh right. I mocked up some quick tests and made sure batching was working (we need to implement `BatchKCache` etc, so can deal with that later)

kernelpool:
Thanks for the feedback! I've added the proposed changes.

mlx_lm/models/deepseek_v32.py
awni: Good point re the `BatchKCache`, I did not consider that. It might be worth fixing our KV caching strategy to allow variable keys / values so we don't have this problem.

PR #536, opened 2025-10-07
DWQ for very large models

awni:
- Adds an option to DWQ to precompute the targets
- Precomputed targets keep only the topk scores and indices to make disk space reasonable (evals on this to come, TLDR no change so far)
- Make pipeline paralell a mixin to make it easier to add to models
- Make `pipeline_load` an actual API in mlx-lm


angeloskath (approved):
Looks awesome!

mlx_lm/quant/dwq.py
```diff
@@ -18,19 +19,62 @@
...
+                enumerate(iterate_batches(data, batch_size, max_seq_length, seed=seed)),
+                total=len(data) // batch_size,
+                desc=f"Computing targets for {split}",
+                disable=rank != 0,
+            )
+        ):
+            batch = batch[:, :-1]
+            logits = model(batch)
+            # Hack to make the last op pre-eval on the CPU to avoid even timeout
+            logits = mx.stop_gradient(logits, stream=mx.cpu)
```
awni, line 56: This is an unfortunate hack. Basically if you have the last op on the GPU then the shared event that is on the GPU can timeout waiting for another rank to finish. A possible fix would be to have an atomic event for Metal the same way we do with the Fence.

awni:
Generating targets with GLM4.6 (300B seems to work ok on 2 M3 Ultras). I'll see if I can do a full DWQ with it:

Command to generate targets:

```
mlx.launch --env MLX_METAL_FAST_SYNCH=1 --hostfile hosts.json mlx_lm.dwq --model mlx-community/GLM-4.6-bf16 --batch-size 1 --max-seq-length 512 --num-samples 2048 --target-dir dwq_targets --targets-only --pipeline 
```

awni:
Also interesting that we're getting a consistent uplift from the top-k loss. I tried a smaller value of k (256) it was a bit worse, though maybe in the noise. The 30B-A3B model is still running.

chimezie:
@awni , roughly what range of parameter size is for? Over 14B? 
I'm just wondering if this was meant to benefit models of all sizes, given the size of the models you gave as examples

awni:
The topk improvement is for all models. The pipeline parallel is only needed for very large models that don’t fit in RAM of one machine. In general one should use a single machine (or data parallel) if the model fits in RAM.

ivanfioravanti:
Amazing! This will push DWQ usage even more!

awni:
I'm going to merge this because it has some nice QoL improvements for pipeline parallel. The DWQ changes work but I think still need some tuning for best possible performance. But for regular (non distributed DWQ) it is the same as it was.

PR #537, opened 2025-10-07
Add lfm2 moe

Blaizzy:
Screenshot 2025-10-07 at 11 59 20 PM

### DIFF
--- a/mlx_lm/models/lfm2_moe.py
+++ b/mlx_lm/models/lfm2_moe.py
@@ -39,8 +39,6 @@ class ModelArgs(BaseModelArgs):
     layer_types: Optional[List[str]] = None
 
     def __post_init__(self):
-        if self.num_key_value_heads is None:
-            self.num_key_value_heads = self.num_attention_heads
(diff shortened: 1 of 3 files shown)


Blaizzy:
My pleasure!

mlx_lm/models/lfm2_moe.py
```diff
@@ -192,8 +191,8 @@ def __call__(
         self,
         x: mx.array,
     ):
-        gates = self.gate(x)
-        gates = mx.softmax(gates, axis=-1, precise=True)
+        gates = self.gate(x).astype(mx.float32)
```
Blaizzy, line 194: Why do we upcast it to float32 here?

awni: This score computation in MoEs is usually quite sensitive to precision, especially with a bias which is kept in fp32 (and it is also not that expensive).

Blaizzy: That makes sense.

Thanks for clarifying!

PR #547, opened 2025-10-10
Add Qwen3-VL language model implementation

vincentamato:
Adds the [Qwen3-VL](https://huggingface.co/collections/Qwen/qwen3-vl-68d2a7c1b8a8afce4ebd2dbe) model family for text-only inference. Sibling PR of [lmstudio-ai/mlx-engine#230](https://github.com/lmstudio-ai/mlx-engine/pull/230).

### DIFF
--- a/mlx_lm/models/qwen3_vl_moe.py
+++ b/mlx_lm/models/qwen3_vl_moe.py
@@ -69,5 +69,9 @@ class Model(nn.Module):
         return weights
 
     @property
+    def quant_predicate(self):
+        return self.language_model.quant_predicate
+
+    @property
     def layers(self):
         return self.language_model.model.layers
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -509,40 +509,6 @@ class TestModels(unittest.TestCase):
             model, args.model_type, args.vocab_size, args.num_hidden_layers
         )
 
-    def test_qwen3_vl_moe(self):
-        from mlx_lm.models import qwen3_vl_moe
-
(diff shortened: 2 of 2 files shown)


tests/test_models.py
```diff
@@ -509,6 +509,40 @@ def test_qwen3_moe(self):
...
+            },
+        )
+        model = qwen3_vl_moe.Model(args)
+        self.model_test_runner(
+            model,
+            args.model_type,
+            args.text_config["vocab_size"],
+            args.text_config["num_hidden_layers"],
+        )
+
```
awni: We don't need both this and the test below, they are redundant. The new style is just to add the config to the list below so probably better to remove this one.

mlx_lm/models/qwen3_vl_moe.py
```diff
@@ -0,0 +1,73 @@
...
+                ].swapaxes(-2, -1)
+                weights[f"{prefix}.switch_mlp.down_proj.weight"] = weights.pop(
+                    f"{prefix}.experts.down_proj"
+                ).swapaxes(-2, -1)
+
+        return weights
+
+    @property
+    def layers(self):
+        return self.language_model.model.layers
```
awni, line 77: Should probably add a `quant_predicate` here as well and just call the LM-only implementation. Otherwise the gate may get lower precision which is not so good.

vincentamato:
@awni I went ahead and addressed those comments in my latest commit. How does it look?

PR #549, opened 2025-10-10
Support data parallel eval for generation tasks

awni:
As title.

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -307,7 +307,7 @@ class MLXLM(LM):
         group = mx.distributed.init()
 
         # split data accross ranks
-        pad_to = (len(requests) + group.size() - 1) // group.size()
+        total_requests = len(requests)
         requests = requests[group.rank() :: group.size()]
 
         logging.info("Generating continuation for %d sequences." % len(requests))
@@ -342,8 +342,8 @@ class MLXLM(LM):
         # Gather the completions
         if group.size() > 1:
             with mx.stream(mx.cpu):
-                n = len(completions)
-                pad = pad_to - n
+                pad_to = (total_requests + group.size() - 1) // group.size()
+                pad = pad_to - len(completions)
                 completions = [list(c.encode("utf-8")) for c in completions]
                 max_len = mx.array(max(len(c) for c in completions))
                 max_len = mx.distributed.all_max(max_len).item()
@@ -365,8 +365,8 @@ class MLXLM(LM):
                     .flatten(0, 1)
                     .tolist()
                 )
-                completions = completions[:n]
-                lengths = lengths[:n]
+                completions = completions[:total_requests]
+                lengths = lengths[:total_requests]
                 completions = [
                     bytearray(c[:l]).decode() for c, l in zip(completions, lengths)
                 ]


mlx_lm/evaluate.py
```diff
@@ -333,6 +338,17 @@ def generate_until(self, requests) -> list[str]:
...
+
+        # Gather the completions
+        if group.size() > 1:
+            max_len = mx.array(max(len(c) for c in completions))
+            max_len = mx.distributed.all_max(max_len, stream=mx.cpu).item()
+            lengths = mx.array([len(c) for c in completions])
+            completions = mx.array([c + [0] * (max_len - len(c)) for c in completions])
+            completions = mx.distributed.all_gather(completions, stream=mx.cpu).tolist()
+            lengths = mx.distributed.all_gather(lengths, stream=mx.cpu).tolist()
+            completions = [c[:l] for c, l in zip(completions, lengths)]
```
angeloskath: I don't know if I am missing something but that seems incorrect to me. Feel free to beautify the following:

```python
import time
import mlx.core as mx


group = mx.distributed.init()
rank = group.rank()
size = group.size()

requests = [(0, 7, 2), (1, 5, 2), (4, 12, 2), (5, 9, 2), (2, 8, 2)]
total_requests = len(requests)
completions = [list(range(*args)) for args in requests[rank :: size]]


N = mx.distributed.all_max(len(completions)).item()
# alternatively it is N = (total_requests + size - 1) // size
max_len = mx.distributed.all_max(max(len(c) for c in completions)).item()
completions += [[0]*max_len]*(N - len(completions))
lengths = mx.array([len(c) for c in completions])
completions = mx.array([c + [0] * (max_len - len(c)) for c in completions])
completions = mx.distributed.all_gather(completions[None]).transpose(1, 0, 2).flatten(0, 1).tolist()
lengths = mx.distributed.all_gather(lengths[None]).transpose(1, 0).flatten(0, 1).tolist()
completions = [c[:l] for c, l in zip(completions[:total_requests], lengths[:total_requests])]

if rank == 0:
    print(requests)
    for c in completions:
        print(c)

```

The main issue is that the order of requests and completions is not the same. Moreover there will be a comms error at best and a deadlock at worst given that the number of sequences can be different if they are not evenly divisible.

Regarding putting `stream=mx.cpu` I don't think it is needed, given that there is only one node in the graph and it will be placed to the comms stream it will be CPU on the Mac, GPU for NCCL etc This would only have an effect if we have comms backends with both cpu and gpu support in the future.

awni: > The main issue is that the order of requests and completions is not the same.

~Not following, what do you mean?~ Got it.

> Moreover there will be a comms error at best and a deadlock at worst given that the number of sequences can be different if they are not evenly divisible.

Good point.

mlx_lm/evaluate.py
```diff
@@ -304,11 +304,16 @@ def generate_until(self, requests) -> list[str]:
             continuation: str
                 The generated continuation.
         """
+        group = mx.distributed.init()
+
+        # split data accross ranks
+        pad_to = (len(requests) + group.size() - 1) // group.size()
```
angeloskath:
```suggestion
        total_requests = len(requests)
        pad_to = (total_requests + group.size() - 1) // group.size()
```

mlx_lm/evaluate.py
```diff
@@ -333,6 +338,39 @@ def generate_until(self, requests) -> list[str]:
...
+                    .tolist()
+                )
+                lengths = (
+                    mx.distributed.all_gather(lengths[None])
+                    .swapaxes(0, 1)
+                    .flatten(0, 1)
+                    .tolist()
+                )
+                completions = completions[:n]
+                lengths = lengths[:n]
```
angeloskath:
```suggestion
                completions = completions[:total_requests]
                lengths = lengths[:total_requests]
```

awni:
Thanks for holding my hand on this one :). 

It seems to be working as expected. On 4 nodes I get 65.57 on MMLU Pro with Qwen3 4B which closely matches the one node result on the same subset: 65.14.

PR #554, opened 2025-10-15
Add support for nanochat

dnakov:
[mlx model](https://huggingface.co/dnakov/nanochat-d20-mlx)

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1932,6 +1932,13 @@ class TestModels(unittest.TestCase):
                 "max_position_embeddings": 1000,
                 "vocab_size": 1000,
             },
+            {
+                "model_type": "nanochat",
+                "hidden_size": 1280,
+                "num_hidden_layers": 20,
+                "vocab_size": 32,
+                "intermediate_size": 128,
+            },
         ]
         for config in test_configs:
             model_type = config["model_type"]


mlx_lm/models/nanochat.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Reshape to (B, L, H, D) - do NOT transpose yet
+        queries = queries.reshape(B, L, self.num_heads, self.head_dim)
+        keys = keys.reshape(B, L, self.num_kv_heads, self.head_dim)
+        values = values.reshape(B, L, self.num_kv_heads, self.head_dim)
+
+        # Apply RoPE BEFORE transpose (PyTorch applies RoPE on (B, T, H, D) format)
+        offset = cache.offset if cache is not None else 0
+        queries = apply_rotary_emb(queries, cos, sin, offset=offset)
+        keys = apply_rotary_emb(keys, cos, sin, offset=offset)
```
awni: I think that can just be `mx.fast.rope` and all the rest of the rope machinery can be removed.

dnakov: it's a different rope, blocked layout. I tried the Traditional true and false but the outputs are very different.

dnakov: actually i think i got it, Traditional=True and reorder the even and odd ones. Updated

dnakov: up to 321 tps. any more ideas? :)

mlx_lm/models/nanochat.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # Handle KV cache after transpose
+        if cache is not None:
+            keys, values = cache.update_and_fetch(keys, values)
+
+        # Repeat KV heads if using MQA/GQA
+        if self.num_kv_heads != self.num_heads:
+            n_rep = self.num_heads // self.num_kv_heads
+            keys = mx.repeat(keys, n_rep, axis=1)
+            values = mx.repeat(values, n_rep, axis=1)
```
awni: No need for that mlx SDPA will handle GQA

mlx_lm/models/nanochat.py
```diff
@@ -0,0 +1,196 @@
...
+class MLP(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.c_fc = nn.Linear(args.hidden_size, args.intermediate_size, bias=False)
+        self.c_proj = nn.Linear(args.intermediate_size, args.hidden_size, bias=False)
+
+    def __call__(self, x: mx.array) -> mx.array:
+        # Critical: nanochat uses ReLU^2, not GELU!
+        x = self.c_fc(x)
+        x = mx.square(nn.relu(x))
```
awni: `nn.relu2` instead. Will be compiled and should be slightly faster.

mlx_lm/models/nanochat.py
```diff
@@ -0,0 +1,189 @@
...
+    Returns:
+        Tensor with RoPE applied, same shape as input
+    """
+    # mx.fast.rope is in interleaved format
+    out = mx.fast.rope(x, dims=x.shape[-1], traditional=True, base=base, scale=1.0, offset=offset)
+
+    # Reorder from interleaved to blocked layout
+    even = out[..., 1::2]  # odd indices -> first half
+    odd = out[..., 0::2]   # even indices -> second half
+    return mx.concatenate([even, odd], axis=-1)
```
awni: This doesn't look like it matches the original implementation. Which is from [here](https://github.com/karpathy/nanochat/blob/master/nanochat/gpt.py#L41-L49) right ?

I made a comparison you can use to do the MLX implementation. Should be faster as it avoids the slicing / concatenate. The idea is to use the fact that sin is odd and cos is even and negate the precomputed frequencies.


```python
import math
import mlx.nn as nn
import mlx.core as mx

# mx.fast.rope is in interleaved format
offset = 0
base = 1000.0
x = mx.random.normal(shape=(1, 4, 1, 32))

def fast_nanochat_rope(x):
    head_dim = x.shape[3]
    half_D = head_dim // 2
    freqs = -mx.exp(
        mx.arange(0.0, half_D, dtype=mx.float32) * (math.log(base) / half_D)
    )
    return mx.fast.rope(x, dims=x.shape[-1], traditional=True, base=None, freqs=freqs, scale=1.0, offset=offset)

def nanochat_rope(x):
    head_dim = x.shape[3]
    seq_len = x.shape[2]
    # stride the channels
    channel_range = mx.arange(0, head_dim, 2)
    inv_freq = 1.0 / (base ** (channel_range / head_dim))
    # stride the time steps
    t = mx.arange(seq_len)
    # calculate the rotation frequencies at each (time, channel) pair
    freqs = t[:, None] * inv_freq
    cos, sin = freqs.cos(), freqs.sin()
    cos, sin = cos[None, :, None, :], sin[None, :, None, :] # add batch and head dims for later broadcasting
    
    d = x.shape[3] // 2
    x1, x2 = x[..., :d], x[..., d:] # split up last time into two halves
    y1 = x1 * cos + x2 * sin # rotate pairs of dims
    y2 = x1 * (-sin) + x2 * cos
    out = mx.concatenate([y1, y2], axis=3) # re-assemble
    return out


print((fast_nanochat_rope(x) -  nanochat_rope(x)).abs().max())
```

awni: Concretely you'd want to precompute the frequencies in the SDPA class but as non-parameters so make them:
`self._freqs = ...`. And then use them in `apply_rope`.

dnakov: very nice, thank you for explaining! I think you meant traditional=False, which works! I don't notice a speed improvement but haven't really run any tests other than a few mlx.generate's. Most importantly, I think it actually fixes the repetition issue I was getting probably due to my prev incorrect implementation.

awni: Yes traditional false! Sorry!

awni (approved):
LGTM!

Issue #535, opened 2025-10-07
How to specify Iteration offset for training


kudla:
Hi guys.
Please  is it really possible to specify iterations offset somehow while using `--resume-adapter-file` to keep `0000*_adapters.safetensors` sequence consistent?

Thank you

awni:
I think you are looking for this: https://github.com/ml-explore/mlx-lm/pull/97

I think we could get something with that functionality landed.

zcbenz:
Sorry but I'm closing non-essential issues so it would be possible for us to actually work on important items.

PR #556, opened 2025-10-16
Feature Enhancement: Auto-detect & resume iteration from checkpoint files 

dotvignesh:
Implements feature enhancement request from #535:

Keep numbered adapter checkpoints monotonic when resuming with `--resume-adapter-file`.

**Behavior (no new flags):**

1. If the resume filename matches `NNNNNN_adapters.safetensors`, parse `NNNNNN` and continue from there (next save = `NNNNNN + save_every`).
2. Else, scan the target directory for the highest `*_adapters.safetensors` and continue from that number.
3. Else, start from 0. Unnumbered `adapters/adapters.safetensors` behavior is unchanged.


zcbenz:

PR #569, opened 2025-10-27
Add cache resume optimization for identical prompts

blightbow:
Add support for skipping prompt prefill when the KV cache already
contains the full prompt from a previous identical request. This
provides 2.7-15x speedup for retry scenarios (e.g., chat "swiping").

Implementation:
- Add can_resume_from_cache() to validate cache state
- Add resume_from_cache parameter to generate_step()
- Skip prefill loop when cache.offset >= prompt_length
- Maintain all integrity constraints (RoPE, masks, slicing)

The optimization is backwards compatible (defaults to False) and
does not manipulate cache.offset directly, avoiding potential
integrity issues with positional embeddings and attention masks.

Add comprehensive unit tests for the cache resume functionality:

- test_can_resume_from_cache(): Tests the validation function with
  various cache states (empty, sufficient, insufficient tokens)
- test_resume_from_cache_with_generate(): Tests end-to-end behavior
  with actual generation, verifying identical token outputs

The tests verify that:
1. can_resume_from_cache() correctly identifies when cache is valid
2. Cache resume produces identical generation outputs
3. Cache offset increases correctly during resumed generation
4. The optimization works with both KVCache and RotatingKVCache

Tested with Qwen3-235B model: 2.048s → 0.749s (2.73x speedup)
on 7-token prompt. Expected 6-15x on longer prompts (100-500 tokens).

### DIFF
--- a/tests/test_prompt_cache.py
+++ b/tests/test_prompt_cache.py
@@ -18,6 +18,7 @@ from mlx_lm.models.cache import (
     MambaCache,
     QuantizedKVCache,
     RotatingKVCache,
+    can_resume_from_cache,
     load_prompt_cache,
     make_prompt_cache,
     save_prompt_cache,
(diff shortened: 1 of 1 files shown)


awni:
Closing as we have more sophisticated cache reuse in `mlx_lm.server` as of https://github.com/ml-explore/mlx-lm/pull/625. Also this sort of logic should be handled outside the core generation to avoid complexity / unnecessary coupling.

PR #573, opened 2025-10-28
Make mlx-lm more type-checker friendly

tnadav:
This commit add type annotations for model load functions to allow type checkers to better understand the interaction with mlx-lm.
As a drive by, I removed mutable default values for the `load`, `load_tokenizer` and `load_model` functions.


awni:
Could you run the format hooks?

tnadav:
> Could you run the format hooks?

Sure! Thanks for the review

Issue #497, opened 2025-09-24
Qwen3-Omni support?


vade:
Hi!

Firstly thanks for MLX! 

Curious if MLX-LM has any plans for omnimodal support for models like Qwen3-Omni? I know its been about 20 seconds since its release, but you know <Ferris Bueller> life moves pretty fast... </Ferris Bueller>

Thanks again!

mcritz:
Subscribed to the issue. I also want to know if we can get Omni model support

Elvincth:
Subscribed to the issue.

Om-GG:
I'm keeping an eye on this one too. I found a model on HF that seems to be using mlx to run Omni (https://huggingface.co/abnormalmapstudio/Qwen3-Omni-30B-A3B-Instruct-mxfp4-mlx), but just now downloading it to test.

hg0428:
I tried it, it didn't load. Unsupported architecture, as expected.

Om-GG:
Yep, that was my experience too. I'm looking into contributing to make it possible but it feels a bit on the hard side of the spectrum. I'll keep updating on this thread if I can make it work!

reneleonhardt:
One option would be to look at other projects, for example vLLM:
https://github.com/vllm-project/vllm/pull/25550

You see some numbers and algorithms to deal with Qwen3-Omni, on which of it's predecessors it is based upon (Qwen3-VL which is already supported by mlx-lm), and you have a working inference engine to test your model against, but the code is totally different of course.

pherber3:
I took a stab at implementing a quantized + MLX version of qwen 3 omni here on huggingface:
https://huggingface.co/pherber3/Qwen3-Omni-30B-A3B-Instruct-4bit-mlx

as well as made a PR for adding qwen3_moe_omni TEXT GENERATION capability to mlx-lm here:
https://github.com/ml-explore/mlx-lm/pull/574

[...]

PR #574, opened 2025-10-29
Added support for qwen3_omni_moe text generation

pherber3:
Hacked something together this afternoon to get text generation working through mlx-lm for qwen3 omni. This ONLY supports text-to-text generation out of the box and does not let you input audio files or images directly. 

I made an mlx 4-bit quantized version of the model here that I've confirmed it works with. It also at least starts to generate with the base model from Qwen/Qwen3-Omni-30B-A3B-Instruct but my computer doesn't have the memory to actually test if it completes. There was another mlx 4-bit version [here](https://huggingface.co/abnormalmapstudio/Qwen3-Omni-30B-A3B-Instruct-mxfp4-mlx) but it never worked for me, I had to make my own which I've uploaded [here](https://huggingface.co/pherber3/Qwen3-Omni-30B-A3B-Instruct-4bit-mlx). Not really sure what the difference is between the two conversion methods used unfortunately other than weights being saved in different ways, maybe the sanitation method implementation is not quite right. 

You can confirm it works via run_all_tests (this is a wrapper class and it looked like the other wrapper classes were added there instead of writing a standalone test like a base class), or try it with `mlx_lm.generate --model pherber3/Qwen3-Omni-30B-A3B-Instruct-4bit-mlx --prompt "Your prompt"` 

Was not certain if adding support for HF base model formats is necessary in the wrapper or if it should only support MLX models also.

### DIFF
--- a/mlx_lm/models/qwen3_omni_moe.py
+++ b/mlx_lm/models/qwen3_omni_moe.py
@@ -1,0 +1,181 @@
+# Copyright © 2025 Apple Inc.
+
+from dataclasses import dataclass
+from typing import Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from . import qwen3_moe
+from .base import BaseModelArgs
+from .qwen3_moe import create_attention_mask
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    """
+    Arguments for Qwen3-Omni-MoE model.
+
+    Qwen3-Omni-MoE is a multimodal model that contains text, audio, and vision
+    components. This implementation only supports text generation.
+
+    Attributes:
+        model_type (str): The model type identifier
+        thinker_config (dict): Configuration for the text language model component
+    """
+
+    model_type: str
+    thinker_config: dict
+
+
+class Model(nn.Module):
+    """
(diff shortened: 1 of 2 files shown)


pherber3:
Closing request because true support will be added soon to mlx-vlm https://github.com/ml-explore/mlx-lm/issues/497#issuecomment-3468005048

rosmur:
@pherber3 Wait why is this closed? mlx-vlm is not an official repo.

pherber3:
> @pherber3 Wait why is this closed? mlx-vlm is not an official repo.

Hmm I did not see that honestly I thought it was under the same umbrella, I suppose in that case there could be some merit to keeping it open. I was not sure if it would be useful given the constraints on not being able to natively have multimodal inputs with this approach whereas it seemed like it would be possible in that repo. This would essentially just be using Qwen3-Omni as a text model unless you generate embeddings elsewhere as in my example. If that's useful as a component to other mlx repos then perhaps, but also given the prolific contributions/reviews of @Blaizzy to these official repos I would trust their implementation of things more than my ad-hoc attempt. I only discovered these mlx repos last week when getting this working on a mac as a prototype, so not entirely sure what the best practice would be in this situation.

PR #575, opened 2025-10-29
add MiniMax-M2 in supported models

sriting:
Link to the MiniMax M2 HF modal repo: https://huggingface.co/MiniMaxAI/MiniMax-M2

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,46 +236,19 @@ for more usage details.
 
 ### Supported Models
 
-`mlx-lm` supports thousands of Hugging Face format LLMs. If the model you want to
-run is not supported, file an
-[issue](https://github.com/ml-explore/mlx-lm/issues/new) or better yet,
-submit a pull request.
+`mlx-lm` supports thousands of LLMs available on the Hugging Face Hub. If the
+model you want to run is not supported, file an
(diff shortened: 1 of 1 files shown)


altaic:
As grateful as I am for MiniMax-M2 and the new support for it here, making a 230B model the first example model in the main readme doesn't sound like a great idea to me considering few machines will be able to run it. Nearly all models in that list are runnable with as little as 16GB RAM.

awni:
@altaic fair point.

I think that whole section of the README needs a revamp anyway. Most of those models are quite outdated. I'm contemplating a bit what to put there. One thought is to change it to point to some useful resources instead (like mlx-community models, benchmarks, etc). Maybe a list of all the model types we support in alphabetic order or something like that.

sriting:
@altaic @awni Fair point. 
Where do you think the MiniMax-M2 link fits best?

awni:
If we end up listing the models, then we could have links where we list them. If we don't include an explicit list of models then not sure.. open to suggestions.

mattharrison:
Excited to see if the 4bit quant fits on my 128G MBP...

awni:
I'm removing the list for now and just pointing people to the MLX Community on Hugging Face which I think is more useful / up-to-date.

awni (approved):
Ty!

PR #579, opened 2025-10-31
Added support for warmup and learning rate schedules

jyork03:
- Added "Learning rate schedule" argument group on mlx.lora containing the necessary arguments to provide warmup and learning rate schedules.
- Finished code implementation of learning rate schedules and tied it into the api.
- Added step-to-update-unit conversions to account for the fact that the optimizer's update() calls are throttled if gradient accumulation steps are configured.
- Added several new tests to confirm functionality.

**New args:**
```
--lr-schedule
--lr-schedule-steps
--lr-schedule-end
--lr-schedule-decay-rate
--lr-schedule-step-size
--lr-schedule-warmup-steps
--lr-schedule-warmup-init
```

### DIFF
--- a/mlx_lm/lora.py
+++ b/mlx_lm/lora.py
@@ -71,6 +71,12 @@ CONFIG_DEFAULTS = {
     "grad_checkpoint": False,
     "grad_accumulation_steps": 1,
     "lr_schedule": None,
+    "lr_steps": None,
+    "lr_end": None,
+    "lr_decay_rate": None,
+    "lr_step_size": None,
+    "warmup_steps": None,
+    "warmup_init": 0.0,
     "lora_parameters": {"rank": 8, "dropout": 0.0, "scale": 20.0},
     "mask_prompt": False,
     "report_to": None,
@@ -253,7 +259,7 @@ def build_parser():
         "--lr-schedule-warmup-init",
         dest="warmup_init",
         type=float,
-        default=0.0,
+        default=None,
         help="Initial learning rate during warmup.",
     )
     return parser


jyork03:
After taking another look at the existing codebase, I realized that I missed something.  Warmup and learning rate schedules were already configurable using a yaml file (as I'm sure you are all aware), but not with the CLI api alone.  For example:

```yaml
lr_schedule:
  warmup: 64
  warmup_init: 2e-7
  name: cosine_decay
  arguments: [2e-5, 2048]
```

The current state of my PR will break this contract for people using this existing configuration.

I still think there is some value in part of what I've written though.  Particularly the step-to-update-unit conversions to account for the fact that the optimizer's update() calls are throttled if gradient accumulation steps are configured.

For example, if `grad-accumulation-steps` is set to `32`, the configuration above would mean that `warmup` would last for `64*32` total steps and the `cosine_decay` would be scheduled over `32*2048` steps.  I think it's more intuitive for users to set those values over the expected actual steps, and not the value multiplied by their `grad-accumulation-steps` value.

I still think the following improvements could be made:

1. A more descriptive default for the `lr_schedule` on the `CONFIG_DEFAULT`.

Instead of:
```py
"lr_schedule": None,
```

Maybe something like the following:
```py
# name options: "cosine_decay", "linear_schedule", "exponential_decay", or "step_decay"
# arguments match positional values for the corresponding mlx scheduler:
# See: https://ml-explore.github.io/mlx/build/html/python/optimizers/schedulers.html
"lr_schedule": { "name": None, "arguments": [], "warmup": 0, "warmup_init": 0.0 },
```

2. Sane default arguments based on the chosen scheduler
3. Step-to-update-unit conversions to account for the fact that the optimizer's update() call is essentially throttled by the gradient accumulation steps.  e.g. setting warmup to `64`, would actually mean `64` *steps* whether you have gradient accumulation configured or not and changes to gradient accumulation won't require you to manually calculate and change your warmup and schedules.
4. Maybe some documented examples for the yaml configuration in the readme?

Since this PR includes some breaking changes, I'll close it and open a new one with these ideas in mind.

PR #583, opened 2025-11-01
Fixed typo in `load_adapters` that broke adapter loading

jyork03:
Fixed typo in `load_adapters` that broke adapter loading after a regression in a recent commit.

The function was defined as `load_adapeters` but all references were expecting `load_adapters`

Simple fix, easy to miss, but important.


awni:
😬  sorry that was my mistake.. and there's no test for it :\

PR #584, opened 2025-11-01
Fixed/improved behavior of the mask_prompt feature.

jyork03:
Without setting `add_generation_prompt=True`, the model/assistant turn header can be included, which forces loss to be calculated over more than just the model's output that we care about.

Before:
```
<start_of_turn>model
model output...<end_of_turn>
```

After:
```
model output...<end_of_turn>
```

I also introduced `_apply_chat_template_safe` to centralize defensive calls to `apply_chat_template` to account for some environments that don't support tools (added defensive measures for `add_generation_prompt` too just in case).

### DIFF
--- a/mlx_lm/tuner/datasets.py
+++ b/mlx_lm/tuner/datasets.py
@@ -8,34 +8,6 @@ from typing import Any, Dict, List
 from transformers import PreTrainedTokenizer
 
 
-def _apply_chat_template_safe(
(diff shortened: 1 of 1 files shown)


jyork03:
@awni Thanks for the merge!  I noticed the new commit introduced a runtime bug, but I don't think I can fix it on this PR after the merge so I opened a new one #591

PR #591, opened 2025-11-05
Fix: Remove call to deleted method

jyork03:
This PR fixes a runtime error introduced in  #584.

@awni It looks like `_apply_chat_template_safe` was removed in the followup commit, but a call to it was left in the `process` method of `CompletionsDataset`. This PR just removes that erroneous line.


awni (approved):
Oops, thanks for catching that!

jyork03:
No problem!

PR #589, opened 2025-11-03
Fix : mlx-server for chunked request (to support one-api, curl)

yiakwy-xpu-ml-framework-team:
## Description

Since in an agentic environment, multiple requests, or chunked requests send from model router (one-api) for example, we found that the server codes have been broken without chunked request support.

With this feature people can handle chunked in an agentic LLM flow

#### Verification in an agentic envrionment, where multiple concurrent calls made

mlx_test_img

#### curl for model router

For whom may refer to this PR and require a quick test with curl:

###### Agentic Entry

Our agentic model router (model can be any models handled in model router to the real model behind):

```
curl -v http://localhost:3000/v1/chat/completions \
 -H "Authorization: Bearer sk-${YOUR_KEY}" \
 -H "Content-Type: application/json" \
 -d '{"model": "gpt-oss-120b-MXFP4-Q4", "message": [{"role": "user", "content": "Once upon a time"}], "temperature": 0.8, "max_tokens": 1024, "stream": false}'
```

to replace usual test entry (model name must be the real model name) :

```
curl -v http://localhost:5001/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model": "gpt-oss-120b-MXFP4-Q4", "messages": [{"role": "user", "content": "Once upon a time"}], "temperature": 0.8, "max_tokens": 1024, "stream": true}'
```

Explanation : "http://localhost:3000/v1/chat/completions" is our model router to test various of models hosted in MacOS Studio:

It will automatically route models to the right services hosted by MLX (default to 5001)

###### The real request

```
echo -n '{"model": "gpt-oss-120b-MXFP4-Q4", "messages": [{"role": "user", "content": "Once upon a time"}], "temperature": 0.8, "max_tokens": 1024, "stream": true}' > payload.json

curl -v --request POST http://localhost:5001/v1/chat/completions \
  -H "Content-Type: application/json" \
  --data-binary @playload.json
```

###### simpler test

server:

> python -m mlx_lm.server --model "mlx-community/Qwen1.5-0.5B-Chat-4bit" --port 5001

client:

```
echo -n '{"model": "mlx-community/Qwen1.5-0.5B-Chat-4bit", "messages": [{"role": "user", "content": "Once upon a time"}], "temperature": 0.8, "max_tokens": 1024, "stream": false}' > payload.json

curl -v --request POST http://localhost:5001/v1/chat/completions \
  -H "Content-Type: application/json" \                                                         
  --data-binary @payload.json
```

#### python

[...]

PR #592, opened 2025-11-05
Fix: JSON parse error handling: avoid referencing stream before init

jyork03:
Fixes a bug where self.stream was referenced before being initialized when the request body fails to parse.

On JSONDecodeError, we now always return a non-stream JSON error response since streaming preference cannot be trusted without a parsed body.

This prevents potential exceptions during error handling.


awni (approved):
Looks good, thanks!

yiakwy-xpu-ml-framework-team:
@jyork03 @awni this has been fixed in https://github.com/ml-explore/mlx-lm/pull/589

PR #593, opened 2025-11-06
Fix for kimi k2

awni:
Support Kimi K2.


awni:
This adds a couple things:
1. Support for the native Kimi int4 quant. Somewhat inefficient since it's a linear quant and emulated with affine. But works well.
2. Support for benchmarking pipeline parallel models with `mlx_lm.benchmark`.

PR #595, opened 2025-11-07
Added optional accuracy reporting while training and evaluating models

jyork03:
**Summary**

Introduces an opt‑in report_accuracy flag to display token‑level accuracy during training reports and validation. Accuracy is computed from existing logits with no extra forward pass. When disabled, there is no accuracy compute overhead.

**Why**

For certain training tasks, users often want quick signal on training quality beyond loss. Accuracy provides a simple, intuitive metric for sanity checks and regressions.

**What Changed**

1. **New flag**: TrainingArgs.report_accuracy (default: False).
2. **Changed `default_loss` return**: default_loss now returns `(loss, ntoks, logits, targets, mask)` so that accuracy can be calculated conditionally outside. 
3. **Conditional accuracy**: Trainer step and evaluate compute accuracy only when the flag is enabled.
4. **Deduped logic**: Added a small helper to normalize loss outputs and compute accuracy only when requested.

**Performance Impact**

- **Disabled**: Should be zero added overhead.
- **Enabled**: Minimal

**API and compatibility**

- No breaking changes; defaults preserve prior behavior.
- Custom losses returning (loss, ntoks) continue to work; accuracy is computed only when the standard intermediates are returned.

**How to use**

- CLI: `--report-accuracy` boolean flag
- yaml config: `report_accuracy: true`

**Sample Logs**

- Train: `Train loss 2.345, Train acc 41.273%, ...`
- Val: `Val loss 2.210, Val acc 43.901%, ...`
- Test: `Test loss 2.240, ... Test acc 41.154%.`


zcbenz:

PR #598, opened 2025-11-08
Fix tool calling with openai server

code-ghalib:
Tool calling is currently broken for the openai style server. The `index` is a required field. The `id` is optional, but most client apps rely on a tool_call_id to match results particularly when executing tools in parallel. 

Some unit tests were failing locally, but seemed completely unrelated.


awni:
I don't see the index as a required field in the tool call response (see https://platform.openai.com/docs/api-reference/chat/create). Will add a uuid placeholder in #711

PR #601, opened 2025-11-11
add profiling control

yiakwy-xpu-ml-framework-team:
#### Usage

```
MTL_CAPTURE_ENABLED=1 python -m mlx_lm.server --model "mlx-community/Qwen1.5-0.5B-Chat-4bit" --port 5001 --log-level DEBUG
```

Start a server, then user can observe performance metrics with trace file.

Additional env parameters:

MLX_TRACE_FILE : trace file name
max_captured_steps : max generation steps (including 1 prefill step to initiate KV cache).

 #### Performance Analysis with MLX profiler
 
add profile control with 1 prefill + up to 3 full decode stages profiling data
 
截屏2025-11-11 上午8 22 44

#### Document

TBD

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -17,6 +17,8 @@ from typing import (
     Union,
 )
 
+import os
+
 import mlx.core as mx
 import mlx.nn as nn
 from mlx.utils import tree_reduce
@@ -640,6 +642,7 @@ def stream_generate(
     prompt: Union[str, mx.array, List[int]],
     max_tokens: int = 256,
     draft_model: Optional[nn.Module] = None,
+    max_captured_steps: int = 5,
     **kwargs,
 ) -> Generator[GenerationResponse, None, None]:
     """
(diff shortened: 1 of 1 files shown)


yiakwy-xpu-ml-framework-team:
@jyork03 @awni Could you have a look at it ?

---------------------------------------------
Please switch to https://github.com/ml-explore/mlx-lm/pull/603

yiakwy-xpu-ml-framework-team:
Note our system don't persist KV cache , which is required in agentic LLM to aggregate contexts by call of a search tool. We are going to resolve this issue later.

angeloskath:
I am not sure this needs to be a capability in the server. I would suggest to profile the workflow in a standalone script.

I will close for now but if you think that it is important for the server to be able to capture a trace then maybe open an issue to see if it is a common need.

yiakwy-xpu-ml-framework-team:
Yes it is important.

a) First, we have many models, adding this capture for every model is not convenient
b) Second, the mlx-lm is supported with distributed model, so later we should add api to allow profile on each instance. That is not convenient if we don't alow this capability in server side.
c) After some tries, it is very good to profil decode stage codes for models.

Could you help me to repoen it ? It seems that there is no button for me to operate here .

PR #602, opened 2025-11-11
Allow providing prompt caches in batched generation

angeloskath:
This PR enables returning individual prompt caches from batch generation and more importantly providing individual prompt caches when inserting sequences in `BatchGenerator`.

It only works with `KVCache` / `BatchKVCache` for now, will add `RotatingKVCache` later if we think this is in the right direction.

The idea is as follows, when prompts are provided together with KVCaches then process them with right padding and before decoding, roll each row of the KVCache the padding amount to make it a left-padded cache.

No perf degradation on my M2 air.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1132,7 +1132,7 @@ def batch_generate(
     prompt_caches: Optional[List[List[Any]]] = None,
     max_tokens: Union[int, List[int]] = 128,
     verbose: bool = False,
-    return_prompt_cache: bool = False,
+    return_prompt_caches: bool = False,
     **kwargs,
 ) -> BatchResponse:
     """
@@ -1142,10 +1142,15 @@ def batch_generate(
        model (nn.Module): The language model.
        tokenizer (PreTrainedTokenizer): The tokenizer.
        prompt (List[List[int]]): The input prompts.
+       prompt_caches (List[List[Any]], optional): Pre-computed prompt-caches
+          for each input prompt. Note, unlike ``generate_step``, the caches
+          won't be updated in-place.
        verbose (bool): If ``True``, print tokens and timing information.
           Default: ``False``.
        max_tokens (Union[int, List[int]): Maximum number of output tokens. This
           can be per prompt if a list is provided.
+       return_prompt_caches (bool): Return the prompt caches in the batch
+          responses. Default: ``False``.
        kwargs: The remaining options get passed to :obj:`BatchGenerator`.
           See :obj:`BatchGenerator` for more details.
     """
@@ -1159,11 +1164,12 @@ def batch_generate(
     with wired_limit(model, [generation_stream]):
         uids = gen.insert(prompts, max_tokens, caches=prompt_caches)
         results = {uid: [] for uid in uids}
-        prompt_caches = {uid: None for uid in uids}
+        prompt_caches = {}
         while responses := gen.next():
             for r in responses:
                 if r.finish_reason is not None:
-                    prompt_caches[r.uid] = r.prompt_cache
+                    if return_prompt_caches:
+                        prompt_caches[r.uid] = r.prompt_cache
                     if verbose:
                         fin += 1
                         print(
@@ -1172,15 +1178,13 @@ def batch_generate(
                         )
                 if r.finish_reason != "stop":
                     results[r.uid].append(r.token)
-                else:
-                    cache.trim_prompt_cache(r.prompt_cache, 1)
     if verbose:
         print(f"[batch_generate] Finished processing {fin}/{num_samples}")
 
     # Return results in correct order
     texts = [tokenizer.decode(results[uid]) for uid in uids]
     stats = gen.stats()
-    caches = [prompt_caches[uid] for uid in uids] if return_prompt_cache else None
+    caches = [prompt_caches[uid] for uid in uids] if return_prompt_caches else None
     if verbose:
         print(
             f"[batch_generate] Prompt: {stats.prompt_tokens} tokens, {stats.prompt_tps:.3f} tokens-per-sec"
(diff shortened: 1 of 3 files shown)


angeloskath:
Well it now works with RotatingKVCache as well.

The api is that a batch kv cache has a `prepare(lengths=..., right_padding=...)` and `finalize()` which is called before and after right-padded prompts.

On the high level APIs of `batch_generate` and `BatchGenerator` the difference compared to `generate` is that the caches are not actually edited in place. This is because they need to be batched and combined with each other and then subsequently extracted when generation has finished.

mlx_lm/generate.py
```diff
@@ -1057,8 +1129,10 @@ def batch_generate(
     model,
     tokenizer,
     prompts: List[int],
+    prompt_caches: Optional[List[List[Any]]] = None,
     max_tokens: Union[int, List[int]] = 128,
     verbose: bool = False,
+    return_prompt_cache: bool = False,
```
awni: Mind adding these to the docstring?

mlx_lm/generate.py
```diff
@@ -1083,24 +1157,30 @@ def batch_generate(
...
+                    if verbose:
+                        fin += 1
+                        print(
+                            f"[batch_generate] Finished processing {fin}/{num_samples} ...",
+                            end="\r",
+                        )
                 if r.finish_reason != "stop":
                     results[r.uid].append(r.token)
+                else:
+                    cache.trim_prompt_cache(r.prompt_cache, 1)
```
awni: What's the purpose of the trim here?

mlx_lm/generate.py
```diff
@@ -1083,24 +1157,30 @@ def batch_generate(
...
         while responses := gen.next():
             for r in responses:
-                if verbose and r.finish_reason != None:
-                    fin += 1
-                    print(
-                        f"[batch_generate] Finished processing {fin}/{num_samples} ...",
-                        end="\r",
-                    )
+                if r.finish_reason is not None:
+                    prompt_caches[r.uid] = r.prompt_cache
```
awni: I'm still digesting this part.. but it may not be a good idea to hold onto the prompt cache here unless explicitly requested. If you decode a lot of prompts and don't need the caches then you will use a lot of memory.

mlx_lm/generate.py
```diff
@@ -1036,7 +1106,9 @@ def _next(self):
             else:
                 finish_reason = None
                 keep_idx.append(e)
-            responses.append(self.Response(uid, t, logprobs[e], finish_reason))
+            if finish_reason is not None:
+                cache = batch.extract_cache(e)
+            responses.append(self.Response(uid, t, logprobs[e], finish_reason, cache))
```
awni, line 1111: Similar comment here. If we don't need to preserve the cache we should avoid holding onto it. 

For e.g. doing an eval with thousands of prompts this will take up a lot of RAM.

awni: This might mean the `BatchGenerator` needs to know the user wants the prompt cache back.

mlx_lm/models/cache.py
```diff
@@ -142,6 +146,12 @@ def meta_state(self, v):
     def is_trimmable(self):
         return False
 
+    def __len__(self):
+        return 0
+
+    def __bool__(self):
+        return True
```
awni, line 165: Is that used at all?

mlx_lm/generate.py
angeloskath: Yeah that is a good point. It should be stored only when requested.

mlx_lm/generate.py
angeloskath: I forgot to ask about that and am not too sure tbh.

The purpose is for the cache to contain the same things as the results. So if stop was encountered then we don't put the token in the results which means we shouldn't put the token in the cache isn't it so?

mlx_lm/generate.py
angeloskath: This should be fine since we don't really hold on to the caches. We just put them in the responses list only when some responses are finalized. So then the caller of `next()` is responsible for storing or deleting them (ie `batch_generate`).

mlx_lm/generate.py
awni: Yea this is an endless source of confusion for me. But I don't think you want to trim it, even though it's not in the result. Otherwise you would need to explicitly process the eos token in the beginning of the next set of tokens.

So if you do:
```
a b <end_of_turn> c d e
```

If you trim the <end_of_turn> from the cache then you would be doing `a b c d e`.

Usually with chat templates it will append the end of turn so even though it's not in the response it gets into the prompt on the next go around. For example:


```python
from mlx_lm import load

_, tokenizer = load("Qwen/Qwen3-0.6B")
messages = [{"role": "user", "content": "Tell a story"}]
    messages, add_generation_prompt=True, tokenize=False
)
print(prompt)


messages = [{"role": "user", "content": "Tell a story"}, {"role": "assistant", "content": "Here's a story"}, {"role": "user", "content": "thanks"}]
    messages, add_generation_prompt=True, tokenize=False
)
print(prompt)
```

The asisstant responds with "Here's a story" and we strip the `<|im_end|>`. But then on the next turn it gets added.

```
# First prompt
<|im_start|>user
Tell a story<|im_end|>
<|im_start|>assistant

# Follow on prompt
<|im_start|>user
Tell a story<|im_end|>
<|im_start|>assistant
Here's a story<|im_end|>
<|im_start|>user
thanks<|im_end|>
<|im_start|>assistant
```

mlx_lm/models/cache.py
angeloskath: Well that's a Python quirk. I will put a comment in the code for the reader to know as well.

When an object defines `__len__` then casting to bool checks whether `len(obj) != 0`. So we have code that does `cache or foo` which returns `foo` even though `cache` is a perfectly fine cache object.

awni: TIL

PR #603, opened 2025-11-11
Add profile

yiakwy-xpu-ml-framework-team:
## Description

This is the follow up of [PR](https://github.com/ml-explore/mlx-lm/pull/601) to add profiling control in server side.

## Next step

Once we verified multi-nodes setup, we will add restful api to enable this behavior in remote more easily.

cc @angeloskath


awni:
Rather than including this inside mlx-lm you can do it using the Python API outside mlx-lm.

Issue #548, opened 2025-10-10
Plans for batch KV cache?


shepardxia:
Hi there! I have been using the new batching methods and they bring some very nice performance boost. 

In the current implementation, the cache is one-off and thrown away after the batch request is complete. Are you planning on introducing persistent batch KV cache in the future, e.g. make BatchGenerator take in a batch cache object, similar to the past_key_values of HF transformers?

I understand that your immediate priority may be expanding batching support to more models, so was wondering whether the persistent cache may come after that.

awni:
At this point most models are covered. We are mostly focused on expanding more front-ends to use batching where appropriate (e.g. in the server, other types of evals, etc).

Exposing a way to persist the cache is not unreasonable. I'll mark this as a feature request.

shepardxia:
In that case, how long would it be before persistent cache is implemented? I am working on adding support for MLX-LM in [GenLM backend](https://github.com/genlm/genlm-backend), in addition to HF and vLLM. I was wondering whether I should go ahead and implement a custom MLX cache or wait for the release. 
Thanks a lot for the updates! The batching really helped speed performance.

awni:
Can you share more about what kind of API you are looking for? Are you expecting the `batch_generate` to persist the cache or  are you looking for a way to persist it yourself?

The way it works in single example generation is like so:
```python
kv_cache = make_cache(model)
# Updates kv cache in place:
result = generate(prompt, model, tokenizer, prompt_cache=kv_cache)
```

We could do something similar with batch generation. However, there are some design decisions to be made. Like should the KV cache be separate for each example or a single combined cache? I kind of lean towards the latter cause it's more flexible. But it also depends on the use case. Would be interested to know more about it.

shepardxia:
With regard to the KV cache design, I wanted to make sure I understand the designs you have in mind. Does separate KV cache for each example mean that the KV cache is kept alive for BatchGenerator objects, which then can then be reused by using the same BatchGenerator? If that's the case then a combined cache with lookups (i.e. a cache manager) would def be more flexible. 

We mainly use MLX-LM/HF/vLLM for logprobs computation rather than continuous generation. A persistent batch KV cache is important for performance since the prompts typically share a large part of their prefixes, which otherwise would be re-computed every time. In other words, we need the batch KV cache not for continuing an interrupted generation, but for precomputing frequent prefixes. 

In reality, rather than needing a persistent batch KV cache, I need APIs to:
[...]

Issue #499, opened 2025-09-24
did mlx_lm.server support batch_generate


liaoweiguo:
seems batch_generate provide best performance for qwen3-next

ghost:
It didn't. But I think it can.

A simple solution that I think of, is to wait one second for any more incoming requests, and if there are more requests, make them a batch to utilize the 5 times faster batch_generate. So 1 sec for 5 times faster, big win.

aniolekx:
I’m using the Aider benchmark to evaluate local language‑models for my master’s thesis.  The benchmark has a --threads option, which allows me to run multiple test cases in parallel ￼; increasing the thread count simulates multiple users and dramatically reduces the total benchmarking time.  However, the current mlx_lm.server processes requests one by one, so even with --threads set above 1 each request waits in line.  This forces me to run tests serially, which both slows down my experiments and fails to capture real multi‑user behaviour.

Support for batching or concurrent processing in mlx_lm.server would solve both problems: I could send multiple requests at once from the benchmark, have them processed together (or interleaved) by the model, and measure throughput and latency under realistic load. Implementing batching would make MLX‑LM more competitive with frameworks like vLLM and would enable serious benchmarking without resorting to external servers.

aniolekx:
Guys, is it possible to catch up with vLLM here? Could the M3 Ultra be the perfect device for private inference in a small company (support multiple requests form users without queueing)?

chimezie:
If I'm not mistaken, doesn't the OpenAI protocol support specifying batched requests? I recall this was the case from the last time I was looking into the mlx server as a basis to efficiently handle evaluation requests from lm-evaluation-harness.

If so, and assuming this isn't already the case, wouldn't a better compromise be to have the server support batching for requests made in this way, instead of doing yeoman work creating batches from serial requests?

How do other LLM services handle this?

awni:
> doesn't the OpenAI protocol support specifying batched requests?

Possibly yes. And if so that would be a trivial extension of the server as it should map directly to the batching API we added in mlx-lm recently.

> instead of doing yeoman work creating batches from serial requests?

Even for serial requests, large scale deployments have to do some form of continuous batching to use the hardware efficiently. So if you use the serial API to make lots of overlapping requests (which is quite reasonably depending on the use case, think several agents all running the same model locally), it should also support handling parallel requests more efficiently than it does now (which is fully sequential).

chimezie:
[...]

Issue #480, opened 2025-09-18
prompt trimming strategy is not very efficient


hkanyeeee:
prompt process seems to be MUCH slower
here is LM studio log using unsloth's GGUF:
"stats": {
"stopReason": "eosFound",
"tokensPerSecond": 57.36418276150394,
"numGpuLayers": -1,
"timeToFirstTokenSec": 0.406,
"totalTimeSec": 43.46,
"promptTokensCount": 7145,
"predictedTokensCount": 2493,
"totalTokensCount": 9638
}
llama_perf_context_print: prompt eval time = 2295.84 ms / 2551 tokens ( 0.90 ms per token, 1111.14 tokens per second)
llama_perf_context_print: eval time = 69676.65 ms / 4573 runs ( 15.24 ms per token, 65.63 tokens per second)
llama_perf_context_print: total time = 73473.23 ms / 7124 tokens
2025-09-18 07:08:38 【DEBUG】
Total prompt tokens: 7145
Prompt tokens to decode: 21

only decode 21 tokens

---------------
here is MLX version:
2025-09-18 07:03:24 【DEBUG】
【cache_wrapper】【INFO】: Trimmed 74 tokens from the prompt cache
2025-09-18 07:03:56 【DEBUG】
【cache_wrapper】【INFO】: Trimmed 2020 tokens from the prompt cache
2025-09-18 07:04:50 【DEBUG】
【cache_wrapper】【INFO】: Trimmed 3094 tokens from the prompt cache

"stats": {
"stopReason": "eosFound",
"tokensPerSecond": 62.67186512030397,
"numGpuLayers": -1,
"timeToFirstTokenSec": 3.184,
"totalTimeSec": 40.257,
"promptTokensCount": 5984,
"predictedTokensCount": 2523,
"totalTokensCount": 8507
}

trim triggered
TTFT is much longer

hkanyeeee:
for the coding agent like cline/roocode,this trimming strategy seems not very efficient

hkanyeeee:
UPDATE：The context computation time of MLX shows a strong positive correlation with context size. The 40-core GPU M4 Max is only usable with contexts below 60K. In contrast, GGUF’s context computation time does not increase as significantly.
And the ram that MLX uses is just too much, compares to flash attention

awni:
What model are you looking at? It's expected for most models that context processing time scales with context length. The only way you wouldn't see that is if you aren't using the full context (which you can do by limiting the context length).

[...]

Issue #259, opened 2025-06-28
Prompt caching returns different logits for repeat prompts


cbowdon:
Hi, thanks for your work on this awesome project. I have been using the intelligent prompt caching described in #137 and implemented in https://github.com/ml-explore/mlx-lm/commit/60c93b8caa9923d9da7ec712bb3a9d62c521e7a2. Surprisingly, it does not return the same logits when repeating a prompt.

This is problematic for applications that use the logits, because you will then get different logits for the same request. In my experience it is not usually unstable enough to affect token sampling.

Here's a reproducible example adapted from the existing prompt cache tests. The first test trims the cache back to the first generated token and passes - more or less exactly the same as the existing test. The second trims back one token further so we provide the last token of the prompt, as is done [here in server.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/server.py#L571). However, the test fails because the logits change.

This only appears for longer prompts. The second test would pass if the prompt was `"this is a prompt"` like it is in the existing tests.

## Passing test

This is almost exactly the same as the prompt cache test [here](https://github.com/ml-explore/mlx-lm/blob/main/tests/test_prompt_cache.py#L210) except:

- the prompt is longer
- more tokens are generated to avoid chance agreement
- using `assert` for convenience with an interpreter rather than `self.assert*`

```python
HF_MODEL_PATH = "mlx-community/Qwen1.5-0.5B-Chat-4bit"

def test_trim_cache_with_generate(MODEL=HF_MODEL_PATH):
    model, tokenizer = load(MODEL)
    prompt = tokenizer.encode(
        "this is a prompt that's a bit longer", return_tensors="mlx"
    )[0]

    prompt_cache = make_prompt_cache(model)

    # Generate one token so we process the full prompt
    last_tok, _ = next(generate_step(prompt, model, prompt_cache=prompt_cache))
    last_tok = mx.array([last_tok])

    # Generate n more tokens
    n = 10  # use a higher n to avoid chance agreement of toks
    results = zip(range(n), generate_step(last_tok, model, prompt_cache=prompt_cache))
    toks, all_logits = zip(*(r[1] for r in results))

    # To get back to the cache just after processing the prompt,
    # trim by n + 1 tokens
    trim_prompt_cache(prompt_cache, n + 1)

    # Generate the same thing again
    results = zip(range(n), generate_step(last_tok, model, prompt_cache=prompt_cache))
[...]

PR #610, opened 2025-11-13
Add continuous batching runtime with paged KV, APC, and bench evidence

Sohailm25:
## Proposed changes

- Wire an end-to-end continuous batching runtime behind `--enable-continuous-batching`: scheduler+runtime split, paged slot KV cache, prefix cache (APC) stats, paged attention backend shim, and slot generator updates so mid-decode admission plus server-lifetime KV reuse work for the OpenAI-compatible server.
- Add the array prefill runner + LLaMA graph overlay path so paged Metal kernels can overlap prefill/decode, plus the new attn backend shim that falls back cleanly when paged kernels/geometries aren’t available.
- Extend the bench harness (`bench/bench_continuous_vs_static.py`, `bench/prefill_profile.py`) and check in the evidence bundle under `benchmarks/` referenced by `PR_README.md`, showing dense vs paged baselines, continuous batching TTFT/throughput, and memory telemetry.
- Cover the new behavior with focused unit tests: paged slot KV cache, paged attention patch, LLaMA prefill graph, model-runner metrics, array runner overlays, scheduler/runtimes, and the new paged model cache tests.

This implements the goals captured in `agent-reference/prd.md`._

## Checklist

- [x] I have run `pre-commit run --all-files` *(ran pre-commit on the staged files; `--all-files` would rewrite untouched legacy files and benches)*
- [x] I have updated the necessary documentation *(PR_README.md already describes the feature set; no user-facing README changes needed)*
- [x] Tests: `PYTHONPATH=../mlx/python:. .venv/bin/python -m pytest tests/server_batched -vv`

DEPENDS ON CHANGES IN MLX HERE: [https://github.com/ml-explore/mlx/pull/2760](url)


Sohailm25:
Addresses these issues:

#499 
#548 
#178 
#259 
#480

awni:
See corresponding comment in https://github.com/ml-explore/mlx/pull/2760 (copied below):

> Thanks for the contribution! At the moment this looks a bit early stage for a PR. If you want to keep running with it, I'd consider refining in your own fork. Then you could start an issue or discussion about if it makes sense to contribute this back to MLX.

> Generally an interesting direction and I'd be curious to see more benchmarks around where something like this could help and by how much.

Sohailm25:
> See corresponding comment in [ml-explore/mlx#2760](https://github.com/ml-explore/mlx/pull/2760) (copied below):
> 
> > Thanks for the contribution! At the moment this looks a bit early stage for a PR. If you want to keep running with it, I'd consider refining in your own fork. Then you could start an issue or discussion about if it makes sense to contribute this back to MLX.
> 
> > Generally an interesting direction and I'd be curious to see more benchmarks around where something like this could help and by how much.

@awni broke it down to a smaller first pr, performed a bench, proved 6x TTFT reduction for continous batching!
https://github.com/ml-explore/mlx-lm/pull/629

WindChimeRan:
Is there any followups for this PR? I saw #629 was also closed, but there's still no support to paged attention or continuous batching. @Sohailm25

awni:
- continuous batching is supported
- paged attention is not supported .. we may add it in the future if it seems worth while.

PR #612, opened 2025-11-14
FIX: Making chat history stateful by moving messages list

Deekshith-Dade:
I found a bug in the mlx_lm.chat utility where the messages list was being re-initialized as empty inside the main while True: loop. This caused the chat to be stateless, forgetting the entire conversation history with every new user query.

This PR fixes this behavior by:
1. Moving the messages = [] initialization outside the while True: loop.
2. Adding the assistant's full_response to the messages list after each turn.
3. Updating the 'r' (reset) command to also clear this new persistent messages list.

This makes the chat utility work as a stateful, multi-turn conversation, which I believe is the intended functionality.

### DIFF
--- a/mlx_lm/chat.py
+++ b/mlx_lm/chat.py
@@ -4,6 +4,8 @@ import argparse
 
 import mlx.core as mx
 
+from mlx_lm import generate
+
 from .generate import stream_generate
 from .models.cache import make_prompt_cache
 from .sample_utils import make_sampler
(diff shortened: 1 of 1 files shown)


angeloskath:
The `messages` list is cleared because the state is held in the `prompt_cache`. I am closing this but if you notice a wrong behaviour feel free to open an issue with a reproduction.

PR #626, opened 2025-11-21
Batching in the server

angeloskath:
I feel I am slightly stepping on @trillionmonster 's toes but this PR has been some time coming given #602 and #625 . I haven't fully cleaned it yet, there is more refactoring to be done in the API handler and probably some cleaning, docs and tests all around.

The idea is the following: `ResponseGenerator` receives a `CompletionRequest` and  `GenerationArguments` and returns a `GenerationContext` and a generator object that yields `Response` objects. These all contain python values to make sure that only one thread is ever accessing any MLX objects.

The APIHandler can implement all the logic based on the above so no functionality is lost.

The `ResponseGenerator` can switch between batch generation and single generation as well as between different batch generation objects. This is done because in a single "batch" we currently use a single sampler.

I know it is not super fancy but here is a 4 way parallel generation on my M2 air with cache reuse and everything.

https://github.com/user-attachments/assets/d72e3c97-bf6a-46c5-aec8-c087d643429c

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1258,10 +1258,6 @@ class APIHandler(BaseHTTPRequestHandler):
             if gen.finish_reason is not None:
                 finish_reason = gen.finish_reason
 
-        # logging.debug(f"Prompt: {gen.prompt_tps:.3f} tokens-per-sec")
-        # logging.debug(f"Generation: {gen.generation_tps:.3f} tokens-per-sec")
-        # logging.debug(f"Peak memory: {gen.peak_memory:.3f} GB")
-
         if self.stream:
             response = self.generate_response(
                 segment, finish_reason, tool_calls=tool_calls


otarkhan:
I've been testing this PR on a Mac Studio (M3 Ultra) and found a significant performance regression compared to the main branch.

The Issue: Even with a single request (batch size = 1), GPU utilization drops from ~98-100% (main branch) to ~30% (this PR).

Analysis: powermetrics shows one CPU cluster (P2) is pinned at 100% while the GPU sits mostly idle. It appears the overhead introduced by the threaded ResponseGenerator loop (queue management, cache checks, Python-side batch construction) is higher than the inference time for a single token on high-end chips.

The Python control loop effectively cannot feed the M3 Ultra GPU fast enough, making the system CPU-bound rather than Memory-bound.

angeloskath:
@otarkhan Yep, that wasn't quite ready to merge (also tests not passing). I also added progress callbacks and early stopping so now the functionality is exactly as before. 

The source of the slowdown was two-fold.
1. Checking the request queue after every token slows things down a bit (only checked on my laptop).
2. The logprobs were not evaluated in the generation stream causing a dependency that slowed things down quite a bit.

Now I get no slow-down on my laptop; batch size 2 gets pretty much 2x speedup (2 * 38 tps vs 41 tps for llama 3B) . Will fix the tests and report some numbers for various model/machine pairs.

angeloskath:
Some numbers:

M2 Ultra - Qwen 30B/3B 4bit - Generating ~1500 tokens
| Batch size | Method | TPS |
|----------|----------|----------:|
| 1 | mlx_lm.generate | 89 |
| 1 | mlx_lm.server | 86 |
| 2 | mlx_lm.server | 141 |
| 2 | mlx_lm.benchmark | 145 |
| 4 | mlx_lm.server | 204 |
| 4 | mlx_lm.benchmark | 200 |

angeloskath:
More refactoring can be done in the server but I think this is ready to review. #625 should be merged first.

It supports everything that was previously supported and seamlessly switches from batching to non-batching. It also seems to scale the same as `mlx_lm.benchmark` which is nice.

mlx_lm/server.py
```diff
@@ -849,12 +1255,12 @@ def keepalive_callback(processed_tokens, total_tokens):
...
-            finish_reason = gen_response.finish_reason
+            if gen.finish_reason is not None:
+                finish_reason = gen.finish_reason
 
-        logging.debug(f"Prompt: {gen_response.prompt_tps:.3f} tokens-per-sec")
-        logging.debug(f"Generation: {gen_response.generation_tps:.3f} tokens-per-sec")
-        logging.debug(f"Peak memory: {gen_response.peak_memory:.3f} GB")
+        # logging.debug(f"Prompt: {gen.prompt_tps:.3f} tokens-per-sec")
+        # logging.debug(f"Generation: {gen.generation_tps:.3f} tokens-per-sec")
+        # logging.debug(f"Peak memory: {gen.peak_memory:.3f} GB")
```
awni: Should we remove that? I guess with threading, the logging is a bit trickier?

mlx_lm/server.py
```diff
@@ -372,21 +451,397 @@ def validate_draft_tokenizer(draft_tokenizer):
...
+        ):
+            return False
+        if args.logits.logit_bias is not None:
+            return False
+        if args.logits.repetition_penalty != 0:
+            return False
+        if args.logprobs > 0:
+            return False
+        if args.seed is not None:
+            return False
```
awni, line 500: Interesting.. I would have never thought of that. I guess we would need to put RNG keys in the batch like everything else.

mlx_lm/server.py
angeloskath: Yep. Forgot that. We might want to add it back again later ie some general server stats at least...

angeloskath:
I think the only subtle but important thing to notice is the `return sampled, list(logprobs)` in the batch `_step()`.

The reason is that we want the logprobs evaluated asynchronously in the generation stream but if we don't pre-slice then the slice in `self.Response(uid, t, logprobs[e], finish_reason, cache)` actually puts a slice op in the generation stream that messes up with generation when being accessed outside of the generation stream.

I 'll merge when the tests pass and we can iterate further after that.

otarkhan:
Followup: https://github.com/ml-explore/mlx-lm/pull/652

PR #633, opened 2025-11-25
Added TTFT logic into the existing flow

meghsat:
The TTFT logic involves wrapping the token generator with a timer that tracks the time taken to generate the first token.
Along with the existing metrics, TTFT will also be printed like this:
Prompt: 2048 tokens, 475.326 tokens-per-sec
**Time to first token**: 4.309 sec
Generation: 128 tokens, 29.802 tokens-per-sec
Peak memory: 27.527 GB

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+*.txt
 
 # PyInstaller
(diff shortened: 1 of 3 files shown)


awni:
You can get TTFT by computing `prompt_tokens / prompt_tps`

PR #652, opened 2025-12-03
Fix slow batch generation in server by setting wired_limit

otarkhan:
## Summary

- The server's batch generation path was missing the `wired_limit` setting that `stream_generate` uses, causing ~10x slower performance and ~30% GPU utilization on large models.
- When `wired_limit` is not set, model weights cannot stay pinned in GPU-accessible memory, causing frequent memory paging and GPU stalls.
- This fix sets `max_recommended_working_set_size` when creating the BatchGenerator and restores the previous limit when done, matching the behavior of `stream_generate`.

## Test plan

- [x] Tested with large model (MiniMax-M2-Q8) - GPU utilization went from ~30% to ~100%
- [x] Verified tokens/second improved ~10x to match sequential generation performance

## Before
https://github.com/user-attachments/assets/9055ea08-7201-43d4-9da1-f892bd92ac42

## After
https://github.com/user-attachments/assets/28f4b155-13aa-423e-a9ff-5f5c70a42283

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -948,6 +948,22 @@ class BatchGenerator:
 
         self.active_batch = None
 
+        if mx.metal.is_available():
+            self._old_wired_limit = mx.set_wired_limit(
+                mx.metal.device_info()["max_recommended_working_set_size"]
+            )
+        else:
+            self._old_wired_limit = None
+
+    def close(self):
+        if self._old_wired_limit is not None:
+            mx.synchronize(generation_stream)
+            mx.set_wired_limit(self._old_wired_limit)
+            self._old_wired_limit = None
+
+    def __del__(self):
+        self.close()
+
     def insert(
         self, prompts, max_tokens: Union[List[int], int, None] = None, caches=None
     ):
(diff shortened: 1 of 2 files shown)


otarkhan:
@awni 
Would you please take a look? I think this is an important followup change for https://github.com/ml-explore/mlx-lm/pull/626

otarkhan:
@angeloskath for visibility

awni:
We do need to set the wired limit but I don't want to do it this way in the server. It couples the generation and server in a way that would be good to avoid and adds complexity to the server logic which is already quite complex.

I wonder if it could make sense to set wired limit when you construct the batch generator object and unset it when it goes out of scope?

angeloskath (approved):
Looks great to me! Thanks for adding that :-)

awni:
Would you mind updating `batch_generate` to not set the wried limit since it's superfluous (e.g. [here](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/generate.py#L1199))?

otarkhan:
> Would you mind updating `batch_generate` to not set the wried limit since it's superfluous (e.g. [here](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/generate.py#L1199))?

Good catch! Updated

PR #665, opened 2025-12-08
Add MiniMax M2 tool call parser support

otarkhan:
## Summary

This PR adds support for MiniMax M2's XML-based tool calling format through a pluggable tool parser system.

### Background

While testing MiniMax-M2 locally with [OpenCode](https://github.com/opencode-ai/opencode) (an AI coding agent), I discovered that the model's tool calls were not being parsed correctly. MiniMax M2 uses a unique XML format for tool calls that differs from the standard JSON format used by other models like Qwen and Llama:

```xml
<minimax:tool_call>
<invoke name="function_name">
<parameter name="param1">value1</parameter>
<parameter name="param2">value2</parameter>
</invoke>
</minimax:tool_call>
```

The existing server implementation only supports JSON-formatted tool calls wrapped in `<tool_call>` tags, causing MiniMax's XML format to be passed through unparsed, which breaks agent workflows that depend on properly structured tool call responses.

### Changes

- **New `mlx_lm/tool_parsers/` module**: Pluggable architecture for tool call parsing
  - `base.py`: Abstract base class `ToolParser` with `ToolCall` and `ExtractedToolCallInformation` dataclasses
  - `minimax_m2_tool_parser.py`: XML parser for MiniMax M2's format using regex patterns
  - `__init__.py`: Factory function and helper utilities

- **New `--tool-call-parser` CLI flag**: Explicitly select the parser to use
  - Currently supports: `minimax_m2`
  - When not specified, uses original JSON parsing behavior (backward compatible)

- **Updated `tokenizer_utils.py`**: Added MiniMax tool call token detection (`<minimax:tool_call>`, `</minimax:tool_call>`)

- **Updated `server.py`**: Integration with the tool parser system
  - Uses selected parser for tool call extraction
  - Proper `finish_reason: "tool_calls"` handling per OpenAI spec
  - Converts tool call arguments format for chat templates

- **Comprehensive tests**: 15 tests for MiniMax parser functionality

### Usage

```bash
mlx_lm.server --model MiniMaxAI/MiniMax-M2 --tool-call-parser minimax_m2
```

### Test Plan

- [x] All existing server tests pass (13 tests)
- [x] New tool parser tests pass (15 tests)
- [x] Tested manually with MiniMax-M2 and OpenCode agent
- [x] Backward compatible - no flag uses original behavior

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -36,6 +36,13 @@ from ._version import __version__
 from .generate import BatchGenerator, stream_generate
 from .models.cache import can_trim_prompt_cache, make_prompt_cache, trim_prompt_cache
 from .sample_utils import make_logits_processors, make_sampler
+from .tool_parsers import (
+    TOOL_PARSERS,
+    ToolParser,
+    get_tool_parser,
+    parse_tool_calls_to_openai_format,
+    process_tool_call_arguments,
+)
 from .utils import load
 
 
@@ -136,7 +143,7 @@ def process_message_content(messages):
     Convert message content to a format suitable for `apply_chat_template`.
 
     The function operates on messages in place. It converts the 'content' field
-    to a string instead of a list of text fragments.
+    to a string instead of a list of text fragments, and processes tool_calls.
 
     Args:
         message_list (list): A list of dictionaries, where each dictionary may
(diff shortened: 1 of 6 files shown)


awni:
Thanks a lot for the contribution! I actually added tool call parsers in #711 in addition to supporting reasoning content so I'm going to close this. If you notice anything off there please leave a note!

PR #671, opened 2025-12-10
Fix for Devstral-2

inferencers:
Convert cache offset to int for mx.arange compatibility in attention scale

### DIFF
--- a/mlx_lm/models/ministral3.py
+++ b/mlx_lm/models/ministral3.py
@@ -36,13 +36,17 @@ class ModelArgs(BaseModelArgs):
             self.layer_types = ["full_attention"] * self.num_hidden_layers
 
 
-def _get_llama_4_attn_scale(
-    start: int, stop: int, beta: float, max_position_embeddings: int
-):
+def _get_llama_4_attn_scale(size, offset, beta: float, max_position_embeddings: int):
+    if isinstance(offset, mx.array) and offset.ndim > 0:
+        offset = offset[:, None]
+
     scaling = 1 + beta * mx.log(
-        1 + mx.floor(mx.arange(start, stop) / max_position_embeddings)
+        1 + mx.floor((mx.arange(size) + offset) / max_position_embeddings)
     )
-    return scaling[:, None]
+    if scaling.ndim == 2:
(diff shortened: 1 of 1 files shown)


awni:
Could you share the repro that made you send this fix? Were you running batch generation?

We probably don't want to convert the offset in the model as that could trigger a graph eval.. so I think it makes more sense to change the scale function to work with arrays (e.g. switch to `mx.arange(size) + offset`). Instead of using a start and stop.

inferencers:
Sure, it's uploading now: https://huggingface.co/inferencerlabs/Devstral-2-123B-Instruct-2512-MLX-6.5bit

For BatchKVCache case - Offsets differ per batch item due to left padding, can use min to cover all positions
offset = int(cache[0].offset.min())

But, will also try your suggestion now.

awni:
I forgot about the batch dimension. Will need to make sure to expand the dimensions directly in that case. I think something like:

```python
def _get_llama_4_attn_scale(
    size, offset, beta: float, max_position_embeddings: int
):
    scaling = 1 + beta * mx.log(
        1 + mx.floor((mx.arange(size) + offset) / max_position_embeddings)
    )
   if scaling.ndim == 2:
     return scaling[:, None, :, None]
    else:
      return scaling[:, None]
```

inferencers:
Looks good, just added a offset check batched offsets.

```
def _get_llama_4_attn_scale(
    size, offset, beta: float, max_position_embeddings: int
):
    if isinstance(offset, mx.array) and offset.ndim > 0:
        offset = offset[:, None]
    
    scaling = 1 + beta * mx.log(
        1 + mx.floor((mx.arange(size) + offset) / max_position_embeddings)
    )
    if scaling.ndim == 2:
        return scaling[:, None, :, None]
    else:
        return scaling[:, None]
```        
        
and call it like you suggested without start,stop
```
attn_scale = _get_llama_4_attn_scale(
            inputs.shape[1],
            offset,
            self.args.rope_parameters["llama_4_scaling_beta"],
            self.args.rope_parameters["original_max_position_embeddings"],
        ).astype(h.dtype)
```

PR #673, opened 2025-12-11
Fix batch inference for ministral3 and tokenizer return type normalization

zenyr:
# Issue

`mlx_lm serve` could not process recently released Devstral 2 123B, while `generate` worked.

# Changes

- ministral3: `_get_llama_4_attn_scale` now handles array offsets for batch processing
- server: `_tokenize` normalizes tokenizer output to `list[int]` (handles BatchEncoding, arrays)


zenyr:
Seems like #671 also aims for the same goal but with a different approach?

awni:
#671 fixes this

PR #674, opened 2025-12-11
fix: server busy-waiting during idle request polling

zenyr:
# Issue
`mlx_lm serve` takes up 100% cpu (for a single core) and often ignore ctrl+c.

# Goal
Prevent 100% CPU spin-loop when server idle by adding 0.1s timeout to queue.get().

# Changes
- `get_next_request()` now accepts optional timeout param
- idle state (no batch running): use 0.1s timeout via queue.get(timeout=0.1)
- batch processing: use non-blocking get_nowait() (avoids latency impact)

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -532,12 +532,15 @@ class ResponseGenerator:
 
         unprocessed_requests = []
 
-        def get_next_request():
+        def get_next_request(timeout=None):
             if unprocessed_requests:
                 return unprocessed_requests.pop()
             else:
                 try:
-                    return self.requests.get_nowait()
+                    if timeout is not None:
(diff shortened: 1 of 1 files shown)


mlx_lm/server.py
```diff
@@ -532,12 +532,15 @@ def _generate(self):
...
 
-        def get_next_request():
+        def get_next_request(timeout=None):
             if unprocessed_requests:
                 return unprocessed_requests.pop()
             else:
                 try:
-                    return self.requests.get_nowait()
+                    if timeout is not None:
+                        return self.requests.get(timeout=timeout)
```
awni, line 541: Why use a timeout? Why not just block until the queue has a request in this case?

awni (approved):
LGTM

PR #676, opened 2025-12-15
Model parallel generation

angeloskath:
Revamps the `distributed-layers` branch basically. Main things it adds is

- `sharded_load` in `utils.py`
- `--shard` in benchmark and chat
- sharding for a few models

### DIFF
--- a/mlx_lm/chat.py
+++ b/mlx_lm/chat.py
@@ -7,7 +7,7 @@ import mlx.core as mx
 from .generate import stream_generate
 from .models.cache import make_prompt_cache
 from .sample_utils import make_sampler
-from .utils import load, pipeline_load, sharded_load
+from .utils import load, sharded_load
 
 DEFAULT_TEMP = 0.0
 DEFAULT_TOP_P = 1.0


awni:
Just a thought.. since `shard` is somewhat ambiguous, what do you think about changing the API like so:

```
sharded_load(..., pipeline_group, tensor_group)
```

- Could get rid of `pipeline_load` / make it internal
- For now we could require that only one of them is set (though in the future maybe both at the same time is ok?). 
- Merge the sharded example and pipeline example into just sharded example. Add an argument `--pipeline` which switches it to be pipeline parallel or something like that.
- And the same for `mlx_lm.benchmark`

angeloskath:
Yeah those are great suggestions. I did all of them. I don't see a reason to not apply both sharding and pipelining already.

mlx_lm/chat.py
```diff
@@ -7,7 +7,7 @@
 from .generate import stream_generate
 from .models.cache import make_prompt_cache
 from .sample_utils import make_sampler
-from .utils import load
+from .utils import load, pipeline_load, sharded_load
```
awni:
```suggestion
from .utils import load, sharded_load
```

mlx_lm/examples/sharded_generate.py
```diff
@@ -43,16 +44,23 @@
...
 
     def rprint(*args, **kwargs):
         if rank == 0:
             print(*args, **kwargs)
 
-    model, tokenizer = pipeline_load(args.model)
+    model, tokenizer = sharded_load(args.model, pipeline_group, tensor_group)
 
     messages = [{"role": "user", "content": args.prompt}]
     prompt = tokenizer.apply_chat_template(messages, add_generation_prompt=True)
```
awni: The rebase here will add `return_dict=False` for compatibility with Transformers v5.

ivanfioravanti:
THIS IS AMAZING! 🤩

PR #677, opened 2025-12-15
Add server logging and stop generation on client disconnect

otarkhan:
- Add periodic stats logging (running requests, pending, avg tok/s)
- Add per-request completion logging with timing metrics
- Add request ID to HTTP logs and request-specific logs for correlation
- Stop generation when client disconnects (streaming) to avoid wasted compute

https://github.com/user-attachments/assets/46b1616f-5dc7-47b9-802d-674b6763e5f0


otarkhan:
@awni Would you be interested in merging such a change? I can fix the conflict if so

otarkhan:
Pinging @angeloskath for visibility

angeloskath:
Hi @otarkhan , thanks for the ping. I think we are interested in some of the features but possibly in a simpler implementation. I haven't had the time to take quite a close look yet.

otarkhan:
@angeloskath Please let me know which ones you'd be interested in and I'll update the PR with only them and try to simplify the implementation as much as I can.

Thanks

trevorgordon981:
Hit this on `mlx_lm.server` 0.31.3 serving GLM-5.1 (`glm_moe_dsa`, 433GB on disk).

Streaming a single chat completion to `/v1/chat/completions` with `stream:true`, then having the client close the connection mid-stream, takes the whole server process down with:

```
File ".../mlx_lm/server.py", line 1489, in handle_completion
    self.wfile.write(f"data: {json.dumps(resp)}\n\n".encode())
File ".../socketserver.py", line 845, in write
    self._sock.sendall(b)
BrokenPipeError: [Errno 32] Broken pipe
```

This PR's `except (BrokenPipeError, ConnectionResetError)` around `handle_completion` would catch it. For large MoE models the relaunch cost is ~60s of model load, so this turns any client timeout into a hard service interruption. Bumping for review.

zcbenz:

PR #678, opened 2025-12-15
support nemotron 3

awni:
```
mlx_lm.generate --model nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16 --prompt "Write a story about Einstein"
```

It's not fully functional yet.. Model loads and starts to generate reasonable text but it devolves.


awni:
Working now. There was a bug in the gated rms norm for the mamba layer in our nemotron h model.

andresy (approved):

PR #679, opened 2025-12-16
fix: gpt-oss modules_to_not_convert

mbarnson:
The current MLX conversion/quantization for gpt-oss models quantizes layers that OpenAI specifies should remain at full precision (modules_to_not_convert). This PR aligns MXFP4 quantization with OpenAI's specification, keeping attention, embeddings, routers, and lm_head at bfloat16 while MLP experts remain at MXFP4.

Models quantized with `--q-mode mxfp4` will be slightly larger than before, with potentially better output quality matching OpenAI's intended precision scheme.

# Support for MXFP4 conversion of gpt-oss fine-tunes

Loading unquantized bfloat16 gpt-oss fine-tunes fail. Fine-tunes of gpt-oss often convert to Huggingface safetensors format. Attempting to convert them would fail with errors like:

```
ValueError: Received 108 parameters not in model:
model.layers.0.mlp.experts.down_proj,
model.layers.0.mlp.experts.gate_proj,
...
```

## Root Cause

| Format | Tensor Names | Shape | Dim Order |
|--------|--------------|-------|-----------|
| MXFP4 | `gate_up_proj_blocks` | `(128, 5760, ...)` | `(experts, OUT, IN)` |
| Unquantized | `gate_up_proj` | `(128, 2880, 5760)` | `(experts, IN, OUT)` |

HuggingFace uses interleaved gate/up outputs ([modeling_gpt_oss.py](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_oss/modeling_gpt_oss.py)):
```python
gate, up = gate_up[..., ::2], gate_up[..., 1::2]
```

Unquantized weights needed: interleaved split + transpose to match `SwitchLinear`.

## Testing

| Model | Source | Result |
|-------|--------|--------|
| `openai/gpt-oss-120b` | MXFP4 (baseline) | Pass |
| `ArliAI/gpt-oss-120b-Derestricted` | ArliAI fine-tune | Pass |
| `ArliAI/gpt-oss-20b-Derestricted` | ArliAI fine-tune | Pass |
| `p-e-w/gpt-oss-20b-heretic` | Heretic fine-tune | Pass |
| Unit test | Synthetic tensors | Pass |

Tested across multiple independent fine-tunes on Huggingface and both model sizes (20b, 120b).

## Compatibility

- MXFP4 paths unchanged (guarded by `_blocks`/`_scales` checks)
- Early return if already sanitized
- Unit test added: `test_gpt_oss_sanitize_hf`

## MXFP4 Quantization Alignment

Added mode-specific quantization predicates to match OpenAI's intended MXFP4 scheme.

### Changes to `utils.py`

Added support for mode-specific predicates to help avoid quantizing layers not intended for quantization with mxfp4  (e.g., `quant_predicate_mxfp4`).

### modules_to_not_convert support in `gpt_oss.py`

Added `quant_predicate_mxfp4` property matching OpenAI's `modules_to_not_convert`:

| Component | OpenAI MXFP4 | Our Implementation |
|-----------|--------------|-------------------|
| `model.layers.*.self_attn` | Full precision | Full precision |
| `model.layers.*.mlp.router` | Full precision | Full precision |
| `model.embed_tokens` | Full precision | Full precision |
| `lm_head` | Full precision | Full precision |
| MLP experts | MXFP4 4-bit | MXFP4 4-bit |

The default `quant_predicate` (for non-MXFP4 modes) leaves routers at their existing behavior.

### Result

Models quantized with `--q-mode mxfp4` now match OpenAI's quantization strategy with full-precision layers.

### DIFF
--- a/mlx_lm/models/gpt_oss.py
+++ b/mlx_lm/models/gpt_oss.py
@@ -233,34 +233,58 @@ class Model(nn.Module):
         if any("gate_proj.weight" in k for k in weights.keys()):
             return weights  # already sanitized
 
+        # Explicit format detection: MXFP4 uses paired _blocks/_scales suffixes
+        has_blocks = any("_blocks" in k for k in weights.keys())
+        has_scales = any("_scales" in k for k in weights.keys())
+        if has_blocks != has_scales:
+            import warnings
+
+            warnings.warn(
+                f"gpt_oss: Unexpected weight format - has _blocks={has_blocks}, "
+                f"has _scales={has_scales}. Expected both or neither."
+            )
+        is_mxfp4 = has_blocks and has_scales
+
         new_weights = {}
         for k, v in weights.items():
             if "gate_up_proj" in k and "bias" not in k:
-                if "_blocks" in k:
+                if is_mxfp4 and "_blocks" in k:
                     v = v.view(mx.uint32).flatten(-2)
                     k = k.replace("_blocks", ".weight")
-                if "_scales" in k:
+                    gate_v = mx.contiguous(v[..., ::2, :])
+                    up_v = mx.contiguous(v[..., 1::2, :])
+                elif is_mxfp4 and "_scales" in k:
                     k = k.replace("_scales", ".scales")
-                new_weights[k.replace("gate_up_proj", "gate_proj")] = mx.contiguous(
-                    v[..., ::2, :]
-                )
-                new_weights[k.replace("gate_up_proj", "up_proj")] = mx.contiguous(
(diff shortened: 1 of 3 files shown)


awni:
For this kind of one-off conversion you can use the [`convert`](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/convert.py#L95-L97) API directly in Python and pass it a predicate which determines which layers are ok to quantize and which aren't.

Issue #682, opened 2025-12-16
ValueError: Tokenizer class TokenizersBackend does not exist ... for NVIDIA-Nemotron-3-Nano-30B-A3B-4bit


namakshenas:
I upgraded `mlx-ml` to `0.29.1` and am getting this error only for model `NVIDIA-Nemotron-3-Nano-30B-A3B-4bit`:

```
model, tokenizer = load("mlx-community/NVIDIA-Nemotron-3-Nano-30B-A3B-4bit")
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/mlx_lm/utils.py", line 326, in load
    tokenizer = load_tokenizer(
                ^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/mlx_lm/utils.py", line 279, in load_tokenizer
    return _load_tokenizer(
           ^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/mlx_lm/tokenizer_utils.py", line 463, in load
    AutoTokenizer.from_pretrained(model_path, **kwargs),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/transformers/models/auto/tokenization_auto.py", line 1137, in from_pretrained
    raise ValueError(
ValueError: Tokenizer class TokenizersBackend does not exist or is not currently imported.
```

awni:
Try:
```
pip install transformers==5.0.0rc1
```

namakshenas:
> pip install transformers==5.0.0rc1

Tried:

```
Traceback (most recent call last):
  File "./mlx_test.py", line 19, in <module>
    text = generate(
           ^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/mlx_lm/generate.py", line 761, in generate
    for response in stream_generate(model, tokenizer, prompt, **kwargs):
  File ".../.venv/lib/python3.11/site-packages/mlx_lm/generate.py", line 677, in stream_generate
    prompt = mx.array(prompt)
             ^^^^^^^^^^^^^^^^
ValueError: Invalid type transformers.tokenization_utils_base.BatchEncoding received in array initialization.
```

hallelulius:
[...]

PR #684, opened 2025-12-16
Fixes for transformers v5

awni:
The fix should be back-compatible. For some reason the default is changed  in Transformers v5.


awni:
Should close #682

PR #689, opened 2025-12-17
Bump to transformer v5

awni:
Transformers stopped supporting clean up tokenization spaces. So it's removed from our tokenizer decoder as well (so that the two are consistent). It's very much an edge case.. so I think it's fine to remove it.


angeloskath (approved):
Nice.

PR #690, opened 2025-12-18
Make MambaCache compatible with batch generation for nemotron-h

nikhilmitrax:
nemotron's batch generation was failing with `AttributeError: 'MambaCache' object has no attribute 'extract'`.

This MR adds the extract method to the parent object.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -35,7 +35,9 @@ from huggingface_hub import scan_cache_dir
 from ._version import __version__
 from .generate import BatchGenerator, stream_generate
 from .models.cache import (
+    ArraysCache,
     KVCache,
+    MambaCache,
     RotatingKVCache,
     can_trim_prompt_cache,
     make_prompt_cache,
@@ -541,7 +543,7 @@ class ResponseGenerator:
         ):
             return False
         for c in self.model_provider.cache_types:
-            if c not in (KVCache, RotatingKVCache):
+            if c not in (KVCache, RotatingKVCache, ArraysCache, MambaCache):
                 return False
         if args.seed is not None:
             return False


TomLucidor:
What does this do for Nemotron-3-Nano/Mamba2?

nikhilmitrax:
tested batch generation for Nemotron-3-Nano, Haven't tested, but I'm guessing the batch generation was broken for all hybrid models (`plamo2`, `granitemoehybrid`, `qwen3_next`, `mamba`, `mamba2`, `baichuan_m1`, `recurrent_gemma`, `kimi_linear`, `jamba`, `falcon_h1`)

`extract` by the way of the `Batch` dataclass has become a required method for caches, and `_BaseCache` does not seem define an interface for caches, so I'm not sure if this is the correct way of handling it, but this should do as a band-aid

angeloskath:
Hey that's cool thanks!

It is not quite complete because except for `extract` we also need to change the mask according to the lengths in `prepare`. Currently this will work for left padding but not for right padding. Basically it only works once with batch generation when we have no previous cache.

The change might be something as simple as 

```python
def make_mask(self, N: int):
        if self.cache[0] is None and self.left_padding is not None:
            idx = mx.arange(N)
            return mx.logical_and(idx >= self.left_padding[:, None], idx < self._lengths[:, None])
        else:
            return None
```

and then keeping the lengths around in `prepare()` like in `BatchRotatingKVCache`.

nikhilmitrax:
it now supports right-padded masks, added a small test case as well

(AI use disclosure: The tests and the main changes for the new commit were created using AI. I still take responsibility of the code the same way I would as if it was written by me)

angeloskath:
Hi @nikhilmitrax sorry for the slightly misleading previous comment. Your PR does implement what I wrote but that is unfortunately not all that is needed.

You see the cache is unfortunately slightly coupled with the model implementations. In this case looking more closely we see
1. If there already exists a cache we do not mask anything. That means that even for left-padding the mask will only be used the first time so if the padding exceeds the prompt processing chunk the output will still be incorrect.
2. During generation when N is usually 1 the model often doesn't use a mask. In this case it either should or it should only update the cache conditional to the lengths.

PR #691, opened 2025-12-18
Revert return dict and wrap apply_chat_template

awni:
This reverts a previous change I made using `return_dict=False` in `apply_chat_template` and instead wraps the method. 

I think it's preferable since it's simpler and doesn't break old examples we have everywhere (e.g. in all our models on Hugging Face).

Also I this is a useful step towards enabling custom chat templates / other tokenization (e.g. for DSV3.2).


angeloskath (approved):
Very good point! I should have thought of that :-/ ...

Issue #695, opened 2025-12-20
Missing sharding support in glm4_moe


blightbow:
I was making unit tests for parallelism in another software project and noticed that glm4_moe.py was implemented without sharding support in the forward pass, unlike deepseek_v3.

Pinging @awni as the only contributor to that model file. Is there an architectural reason for this, or is it just a matter of having time for implementation and testing? I can work on a PR that imitates the deepseek_v3 approach, but anyone could have done that and I want to make sure there wasn't a reason it's not there yet.

awni:
It’s definitely the latter, we just haven’t gotten around to it. Feel free to send a PR if you have time.

awni:
This was added in #698

PR #696, opened 2025-12-21
feat(utils): add weight_loader parameter for custom loading

blightbow:
Add optional `weight_loader` parameter to `load()`, `load_model()`, and `sharded_load()` functions, allowing custom weight loading strategies.

The parameter accepts a callable with signature:
    (file_path: str) -> Dict[str, Any]

When not provided, defaults to `mx.load` (existing behavior).

Use cases include:
- Distributed weight streaming (load from network instead of disk)
- Encrypted weight files
- Custom caching strategies
- Memory-mapped loading

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -160,6 +160,7 @@ def load_model(
     strict: bool = True,
     model_config: Optional[Dict[str, Any]] = None,
     get_model_classes: Callable[[dict], Tuple[Type[nn.Module], Type]] = _get_classes,
+    weight_loader: Optional[Callable[[str], Dict[str, Any]]] = None,
 ) -> Tuple[nn.Module, dict]:
     """
     Load and initialize the model from a given path.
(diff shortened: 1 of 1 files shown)


blightbow:
I've been evaluating methods for streaming layers from Rank 0 to Rank 1 workers and hit upon an idea. With RDMA, why not just stream the files into the memory of Rank 1 workers and bypass the need for local disk storage on Rank 1 entirely? The layers can be incrementally streamed and loaded one at a time with a custom loader. I could write a custom loader for my app that duplicates some of the work that mlx-lm is doing, but at that point I may as well just submit a PR. (and here we are)

I considered some minimal validation on the weight_loader that gets passed through, but `get_model_classes` has no validation either and I opted to maintain the current code conventions.

blightbow:
@awni I apologize for two @mentions in two days, but would you be willing to review this? I skimped on storage when I bought M3 Studios this year, and this will simplify the logistics of doing proper testing for distributed model PRs. (like we discussed with #695)

blightbow:
Too early to commit this, apologies for the noise. Starting over from scratch.

PR #698, opened 2025-12-23
shard glm

awni:
Tensor parallel sharding for GLM


awni:
Results are so-so. On 4 M3 Ultras I'm getting 24.6 tok/sec, up from 16 or so on a single M3 Ultra.

angeloskath (approved):
🚀

mzbac:
You work too hard, still merged stuff on Christmas Eve. Have a good Christmas break everyone 😊

PR #702, opened 2025-12-28
chore: add model-path param flag for convert API for better clarity

### LINKED ISSUE #675: [FEATURE REQUEST] - Allow Convert to use local model rather than just from Hugging Face
I do a lot of fine tuning with Unsloth.

It would be great if I could fine tune a model in Unsloth (where I have a more powerful set of hardware) export the model to my local file server as a safetensors format, then run the mlx_lm.convert script but pass in the path to the local safetensors file where the current '--hf-path' parameter is passed in.

jaycoolslm:
Fixes #675 

- Add --model-path flag to mlx_lm.convert as a clearer alias for --hf-path while keeping --hf-path fully supported for backwards compatibility.
- Update CLI examples/docs to use the new flag and add a brief compatibility note. No behavior changes beyond accepting the new flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,6 @@ For more options run:
 mlx_lm.convert --help
 ```
 
-`--hf-path` is still supported for backwards compatibility.
-
 You can upload new models to Hugging Face by specifying `--upload-repo` to
 `convert`. For example, to upload a quantized Mistral-7B model to the
 [MLX Hugging Face community](https://huggingface.co/mlx-community) you can do:


jaycoolslm:
I’ve kept this change minimal: just adding --model-path as a clearer alias and updating docs/examples, with --hf-path still fully supported. 

If you’re open to it after this lands, I can follow up with a slightly deeper internal rename so model_path is the primary value in the codepath (with hf_path as a legacy alias), but I didn’t want to bundle that here.

awni:
I think if we change this we should change it to `--model` for consistency with `mlx_lm.generate` and other CLI commands.

Would you mind updating that? Also I think you can have multiple names for an argument in the argparse line.. so you can remove all the logic detecting which is set.

jaycoolslm:
I will make those changes

jaycoolslm:
Done @awni

mlx_lm/convert.py
```diff
@@ -179,7 +179,12 @@ def configure_parser() -> argparse.ArgumentParser:
         description="Convert Hugging Face model to MLX format"
     )
 
-    parser.add_argument("--hf-path", type=str, help="Path to the Hugging Face model.")
+    parser.add_argument(
+        "--hf-path",
+        "--model",
+        type=str,
+        help="Path to the model. This can be a local path or a Hugging Face model hub identifier.",
```
awni:
```suggestion
        help="Path to the model. This can be a local path or a Hugging Face Hub model identifier.",
```

README.md
```diff
@@ -179,13 +179,15 @@ For more options run:
 mlx_lm.convert --help
 ```
 
+`--hf-path` is still supported for backwards compatibility.
+
```
awni:
```suggestion
```

PR #707, opened 2025-12-29
batch_generate fails with Phi3 (LongRoPE) when prompts have different lengths

vyaivanove:
## Description

`batch_generate` fails with `ValueError: [convert] Only length-1 arrays can be converted to Python scalars` when using Phi3 models with LongRoPE scaling and variable-length prompts

I've used Claude Opus 4.5 to generate the proposed fix. Looks pretty simple

## Environment

- mlx-lm version: 0.30.1
- mlx version: 0.30
- Python version: 3.14.2
- macOS version: 26.2
- Hardware: M4 Pro

## Reproduction

```
from mlx_lm import load, batch_generate

model, tokenizer = load("numind/NuExtract-1.5")  # fails with NuExtract-1.5 which is Phi3, works with NuExtract-1.5-tiny which is Qwen2.5-0.5B
sampler = make_sampler(temp=0.0)

# Prompts with different lengths
prompts = [
    "Hello",
    "Hello, how are you today?",
    "Hi",
]
tokenized = [tokenizer.encode(p) for p in prompts]

# This fails
result = batch_generate(model, tokenizer, prompts=tokenized, max_tokens=256, sampler=sampler)
```

### DIFF
--- a/mlx_lm/models/rope_utils.py
+++ b/mlx_lm/models/rope_utils.py
@@ -47,8 +47,7 @@ class SuScaledRoPE(nn.Module):
         self.dim = dims
 
         freqs = base ** (mx.arange(0, dims, 2, dtype=mx.float32) / dims)
-        self._short_freqs = mx.array(short_factor, dtype=mx.float32) * freqs
-        self._long_freqs = mx.array(long_factor, dtype=mx.float32) * freqs
+        self._freqs = mx.array(long_factor, dtype=mx.float32) * freqs
 
         def default_scale(factor):
             return math.sqrt(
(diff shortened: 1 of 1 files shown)


mlx_lm/models/rope_utils.py
```diff
@@ -65,7 +65,13 @@ def default_scale(factor):
 
     def __call__(self, x, offset: int = 0):
         seq_len = offset + x.shape[-2]
-        if seq_len > self.original_max_position_embeddings:
+
+        if isinstance(seq_len, mx.array):
+            seq_len_scalar = seq_len.max().item()
+        else:
```
awni: We shouldn't do a graph evaluation in each call to rope. This will incur quite a bit of latency and potentially break async evaluation.

I also don't love the alternative of doing a `mx.where(seq_len.max() > self.original..., self._long_freq, self._short_freqs)`.

awni: Honestly I think this implementation is wrong anyway. What it should do is use the short rope if the prompt+generation is going to be short and it should use the long rope if the prompt+generation is going to be long.

Switching when it reaches a threshold seems wrong. And then having some switch at different thresholds based on the batching seems even worse.

I think what we should do is simply revert this to always use the long rope (and assume the user is going to do a long generation). At some point we could maybe provide a toggle.. but for now that seems like the cleanest solution.

PR #709, opened 2025-12-30
Allow mxfp8 and nvfp4

awni:
- Adds `mxfp8` and `nvfp4` as options for quantization 
- `bits` and `group_size` default to `None` and a reasonable value is chosen based on the mode (same as MLX).

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -187,17 +187,23 @@ def configure_parser() -> argparse.ArgumentParser:
         "-q", "--quantize", help="Generate a quantized model.", action="store_true"
     )
     parser.add_argument(
-        "--q-group-size", help="Group size for quantization.", type=int, default=64
+        "--q-group-size",
+        help="Group size for quantization.",
+        type=int,
+        default=None,
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
👍

nightscape:
I read somewhere that `nvfp4` is more efficient than `mxfp4`, but I'm not sure in which dimension, speed or quality-per-bit.
Do you happen to know? Would that also apply to MLX?

awni:
If I had to guess it would be quality per bit. nvfp4 uses slightly higher bits-per-weight than mxfp4 (I think it's 4.5 vs 4.25) so it should be higher quality. But I think even accounting for that (e.g. if mxfp4 used the same group size) then nvfp4 would probably still be higher quality.

In terms of speed mxfp4 will be slightly faster since the scale is easier to decode and it has fewer scales.

reneleonhardt:
Out of all four currently possible quantizations, which is the "winner" in terms of quality vs. resources?
Better quality is preferred if it's not 2x slower or memory hungrier: nvfp4 or mxfp8?
1. affine
2. mxfp4
3. nvfp4
4. mxfp8

PR #711, opened 2025-12-30
Improve reasoning and tool call parsing in server

### LINKED ISSUE #607: mlx_lm.server crashes when tool_calls aren’t JSON (Codex/OpenCode glob payloads)
**Summary**
  When GLM 4.5 Air/GLM 4.6 invokes a tool (e.g., glob) in opencode or codex cli, it sends the tool_call payload in its own XML-ish format:

  glob
  <arg_key>pattern</arg_key>
  <arg_value>**/public-audit.md</arg_value>

  `python3 -m mlx_lm server …` assumes every tool_call is JSON and tries `json.loads`, which raises `JSONDecodeError` and tears down the request.

  **Environment**
  - macOS 26.1 (Apple Silicon)
  - Python 3.13.5
  - mlx-lm 0.28.3 (pip)
  - mlx 0.29.4

  **Steps**
  1. Start the stock server:

  python3 -m mlx_lm server \
  --model "/Users/.../GLM-4.5-Air-MLX-8bit" \
  --host 127.0.0.1 \
  --port 8080 \
  --trust-remote-code

  2. Connect with a Codex/OpenCode session and run a task that triggers `glob`/`bash` (e.g., “Find this file ‘public-audit.md’”).

  **Actual**
  Server logs:

  tool_call = json.loads(tool_text.strip())
  json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

  The handler crashes and the client loses its run.

  **Expected**
  Server should tolerate non-JSON tool_call payloads: try `json.loads`, but fall back to parsing `<arg_key>/<arg_value>` or raw text so the request completes. A local patch that wraps `parse_function` like this fixes it; would be  great to carry upstream.

awni:
This is not technically part of the OpenAI chat API spec. But seems to be the defacto standard (used by vllm, open router, etc). 

And it makes the server work better with upstream tools which use openAI compatible APIs. For example with OpenCode you get the reasoning traces nicely formatted:

Screenshot 2025-12-30 at 12 47 58 PM

I also added tool call parsing and redesigned how custom chat templates and tool call parsing works.

Closes #607

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -452,12 +452,27 @@ def _is_bpe_decoder(decoder):
     return isinstance(decoder, dict) and decoder.get("type", None) == "ByteLevel"
 
 
+def _infer_tool_parser(chat_template):
+    """Attempt to auto-infer a tool parser from the chat template."""
+    if not isinstance(chat_template, str):
+        return None
+    elif "<minimax:tool_call>" in chat_template:
+        return "minimax_m2"
+    elif "<start_function_call>" in chat_template:
+        return "function_gemma"
+    elif "<arg_key>" in chat_template:
+        return "glm47"
+    elif "<tool_call>\n<function=" in chat_template:
+        return "qwen3_coder"
+    elif "<tool_call>" in chat_template and "tool_call.name" in chat_template:
+        return "json_tools"
+    return None
+
+
 def load(
     model_path,
     tokenizer_config_extra: Optional[Dict[str, Any]] = None,
-    return_tokenizer=True,
     eos_token_ids=None,
-    tool_module=None,
 ) -> TokenizerWrapper:
     """Load a huggingface tokenizer and try to infer the type of streaming
     detokenizer to use.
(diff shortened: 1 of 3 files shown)


awni:
Also fixed a bug where we were keeping eos text in some cases at the end of the generation in the server.

mzbac:
Thanks for the great initiative. I was working on the similar idea of creating different tool parsers for each model. I'm glad I'm on the right path, can't wait to see this get merged and tested on the MiniMax M2.1 with open code on local stack :)

awni:
I just added a parser for Minimax M2.1 as well. The main thing to be aware of is that right now you have to manually modify the `tokenizer_config.json` to add the tool parser type - as in [here](https://huggingface.co/mlx-community/MiniMax-M2.1-4bit/blob/main/tokenizer_config.json#L496). 

I would like to make that easier to automate.. but it's a bit tricky to auto-detect it since it's the tool parser is different than the model family in some cases (e.g. qwen 3 coder vs qwen 3). Might also be good to add a way to specify the tool call parser as an option.

mzbac:
> I just added a parser for Minimax M2.1 as well. The main thing to be aware of is that right now you have to manually modify the `tokenizer_config.json` to add the tool parser type - as in [here](https://huggingface.co/mlx-community/MiniMax-M2.1-4bit/blob/main/tokenizer_config.json#L496).
> 
> I would like to make that easier to automate.. but it's a bit tricky to auto-detect it since it's the tool parser is different than the model family in some cases (e.g. qwen 3 coder vs qwen 3). Might also be good to add a way to specify the tool call parser as an option.

Thanks for the Minimax parser, I'll try it out, when I was working on my fork to integrate with opencode, I noticed that the tokenizer sometimes generates weird detokenized words, such as "-[space]word" instead of "-word." I'll try the PR branch to see if this issue occurs there as well.

kernelpool:
Just a minor observation. I've noticed `reasoning` is being used instead of `reasoning_content` which seems to be the direction some projects are going (e.g. [vLLM](https://docs.vllm.ai/en/stable/features/reasoning_outputs/)). However, a lot of chat templates still rely on reasoning_content, including [GLM-4.7](https://huggingface.co/mlx-community/GLM-4.7-4bit/blob/main/chat_template.jinja). Considering things like interleaved/persistent reasoning (where a client passes reasoning back to the server), would it be expected that the client still uses `reasoning_content` for the chat templates to function correctly?

awni:
Yes it's confusing. The input `messages` accept `reasoning_content` in the `assistant` role (which is passed directly to the chat template. We could remap that if needed.

The ouput response sets the `reasoning` field in the choices.

I wouldn't have designed it that way 😅 but that's what the downstream models and upstream tools expect. 

For persistent reasoning with GLM4.7 we have to rely on the upstream tool to pass back the reasoning content which I don't think is standard yet.. but that's something to look into.

ivanfioravanti:
> I just added a parser for Minimax M2.1 as well. The main thing to be aware of is that right now you have to manually modify the `tokenizer_config.json` to add the tool parser type - as in [here](https://huggingface.co/mlx-community/MiniMax-M2.1-4bit/blob/main/tokenizer_config.json#L496).
> 
> I would like to make that easier to automate.. but it's a bit tricky to auto-detect it since it's the tool parser is different than the model family in some cases (e.g. qwen 3 coder vs qwen 3). Might also be good to add a way to specify the tool call parser as an option.

@awni what if this logic is handled internally, starting from model family and managing exceptions from there? Because adding a line in tokenizer_config.json breaks other tools using mlx-lm with their own version of models like lm-studio or exo. 
Not the most elegant solution, but maybe less impactful on the ecosystem.

ivanfioravanti:
This PR is a game changer! Experimental as everything around tool calling, but opens doors to new scenarios. 
Thanks @awni 🙏🏻

awni:
I made a hybrid approach which tries to infer the tool parser but still adds the `tool_parser_type` to the `tokenizer_config.json`. Are you saying adding `tool_parser_type` as a field is breaking downstream tools? That would be somewhat surprising.. usually with these kinds of configs they should ignore fields that they don't use.

ivanfioravanti:
No no, I meant that they will have to add this additional field in their version of the models or tool parser won’t work out of the box if they want to leverage same mlx tool parser.

awni:
Gotcha! That should be taken care of now!

mzbac:
Just sharing some findings, may not be directly related to the PR. Somehow, Minimax 2.1 at a temperature of 1.0 picks up a lot of incorrect tokens in the file path, for example, leading spaces with a slash. However, when the temperature is set to 0, it produces a lot of repetitive tokens. In the end, I added file path normalization to remove extra spaces and used temperature 0 with a higher repetition penalty. It seems with these settings, it works fine on my local setup up to a 40k context. I'm not sure if there are some differences in the sampler between MLX and other inference frameworks, given that the model card suggests a temperature of 1.0. @ivanfioravanti

awni:
I don't think it's related to this PR. If you have a prompt or somethign that produces low quality outputs that would be useful to share in an issue. And also the model you are using (e.g. quantization).

mzbac:
> I don't think it's related to this PR. If you have a prompt or somethign that produces low quality outputs that would be useful to share in an issue. And also the model you are using (e.g. quantization).

done https://github.com/ml-explore/mlx-lm/issues/725 :)

PR #723, opened 2026-01-04
Add sanitize() method to olmo3 model

s-jag:
Add missing `sanitize()` method to filter out unused precomputed rotary frequencies (`rotary_emb.inv_freq`) from HuggingFace weights. 

This matches the pattern used in olmo2.py and other RoPE-based models.


awni:
Not necessary for this model

Issue #725, opened 2026-01-05
temperature setting issue for minimax 2.1


mzbac:
I noticed the sampler issue while experimenting with the Minimax model (4-bit and 6-bit quantization). The model card suggested setting the temperature to 1.0, which resulted in a lot of poor outputs, forcing me to set it to 0. ~~Upon further investigation, it appears that in the Transformers implementation, temperature scaling is bypassed when the temperature is set to 1.0.
https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/src/transformers/generation/utils.py#L1320-L1321,~~ 

ps: 
~~https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/sample_utils.py#L275 looks to me as disabled the temperature scaling but somehow it looks like 1.0 still have some impact of the sampling..just noticed temp 1.0 mlx still calling mx.random.categorical instead of mx.argmax which would be the issue.~~

awni:
Setting the temperature to 1.0 should not turn off sampling. That is an unusual behavior in Transformers if it's how it works. It also doesn't make sense if you turn off categorical sampling then top-k and top-p are also irrelevant (since they don't change the argmax). 

Were you also setting top-k and top-p when setting the temperature to 1.0? If not that might explain the issue.

Could you share some sample outputs of the strange behavior you are seeing? Maybe also include the version of transformers you have? I'm not able to reproduce it so far.

mzbac:
My observation is the temperature is set to 1.0 and top-k and top-p are applied, the model starts to generate a lot of incorrect hallucinations. e.g. incorrect file paths, wrong tool call parameters. However, when I set the temperature to 0.0 with the suggested top-k and top-p values, the output becomes stable, and I can get around 80k context without much error in the model output.

awni:
> My observation is the temperature is set to 1.0 and top-k and top-p are applied, the model starts to generate a lot of incorrect hallucinations. e.g. incorrect file paths, wrong tool call parameters.

Anything else you can share to help reproduce this would be great.. otherwise it's pretty difficult to debug.

Is it just via OpenCode or do you notice it with mlx-lm directly? 
What's your transformers version?
Any example of an input or output that is bad?

> However, when I set the temperature to 0.0 with the suggested top-k and top-p values,

The top-k and top-p don't have an effect there (if they do it's a bug)

awni:
[...]

PR #726, opened 2026-01-05
chore(sampler): using greedy sampling when temperature set to 1.0

mzbac:
address https://github.com/ml-explore/mlx-lm/issues/725..

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -43,7 +43,7 @@ def make_sampler(
         Callable[mx.array, mx.array]:
             A sampler which takes log-probabilities and returns tokens.
     """
-    if temp == 0:
+    if temp == 0 or temp == 1.0:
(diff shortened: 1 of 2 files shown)


awni:
I'm not following this - why should we ignore the temperature when it is set to 1.0?

mzbac:
I don't have full context, but since the GPT-O1 model, thinking model starts to limit the temperature (disabled for the thinking model), and Gemini 3 suggests a temperature of 1.0 https://ai.google.dev/gemini-api/docs/gemini-3#temperature , my understanding is that a temperature of 1.0 is a convention to disable it, making sampling go greedy.  
Ps: there is paper about this but I can't recall it.

awni:
I don't think that is a convention. A temperature of 0 would disable sampling (which mathematically is correct). A temperature of 1.0 should not disable sampling and wouldn't make sense in the context of what models suggest to set sampling parameters to.  (Conversation in #725 )

PR #728, opened 2026-01-05
Add a server benchmark for continuous batching

awni:
Add a basic benchmark for continuous batching and a new directory for one-off benchmarks.

### DIFF
--- a/benchmarks/server_benchmark.py
+++ b/benchmarks/server_benchmark.py
@@ -156,14 +156,8 @@ async def make_request(
 
                         try:
                             data = json.loads(data_str)
-                            if "choices" in data and data["choices"]:
-                                delta = data["choices"][0].get("delta", {})
-                                content = (
-                                    delta["content"]
-                                    if "content" in delta
-                                    else delta.get("reasoning_content", "")
-                                )
-                                if content:
+                            if choices := data.get("choices", False):
+                                if choices[0].get("finish_reason") != "length":
                                     tokens.append(time.perf_counter())
                         except json.JSONDecodeError:
                             continue


angeloskath:
Updated the benchmark to handle concurrency better. The output is as follows on an M2 Ultra.

```
$ python -m mlx_lm server --model mlx-community/Llama-3.2-3B-Instruct-4bit --max-tokens 8192 2>/dev/null 1>/dev/null &
$ python benchmarks/server_benchmark.py --conc 8 --total 24
Starting benchmark with 8 concurrency and 24 total requests...
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 24/24 [00:05<00:00,  4.18it/s]

Benchmark completed in 5.75 seconds
Successful requests: 24/24
Total tokens generated: 2400
Aggregate tokens/sec: 405.94

Time to First Token (seconds):
  Min: 0.4109 | Max: 0.5858 | Avg: 0.4656 | P95: 0.5734

Tokens per Second (per request):
  Min: 52.06 | Max: 74.00 | Avg: 61.73 | P95: 74.00

0.00 ████████████████ (168.2312925170068)
1.00 ████████████████████████████████████████████████ (514.4888888888888)
2.00 ████████████████████████████████ (344.9)
3.00 ██████████████████████████████████████████████████ (525.5)
4.00 ███████████████████████████████ (329.0)
5.00 █████████████████████████████████████████████████ (524.4)
6.00 █████████████████████████ (265.0)
7.00  (0)
```

The bar chart at the bottom represents the per-second tokens per second (weird sentence). Basically during the first second since the first request the server generated ~168 tokens, ~500 during the second etc. It fluctuates are requests are added and removed from the batch continuously. If generating more tokens then it stabilizes way more as expected.

It is a bit more involved so lmk if you want to revert and use the AsyncOpenAI client instead.

awni (commented):
I can't approve it but I approve :)

Issue #727, opened 2026-01-05
Support loading pre-quantized AWQ/GPTQ models from HuggingFace


ericcurtin:
Current Behavior

  When attempting to load a model like Qwen/Qwen3-4B-AWQ (quantized with AutoAWQ), mlx-lm fails with:

  ValueError: Received 756 parameters not in model:
  model.layers.0.mlp.down_proj.qweight,
  model.layers.0.mlp.down_proj.qzeros,
  model.layers.0.mlp.down_proj.scales,
  ...

  This happens because AutoAWQ stores weights in a packed format (qweight, qzeros, scales) that MLX's model architecture doesn't recognize.

  Expected Behavior

  mlx-lm should be able to:
  1. Detect when a model uses AutoAWQ/AutoGPTQ quantization format (via config.json quantization config)
  2. Either:
    - Dequantize the packed weights and load them (possibly re-quantizing to MLX format), or
    - Instantiate quantized linear layers that understand the packed format

  Context

  - mlx-lm already supports AWQ/GPTQ as quantization algorithms via mlx_lm.convert --q-method awq
  - There are many popular pre-quantized models on HuggingFace (e.g., TheBloke's models, official AWQ variants)
  - Users expect these models to "just work" since the quantization methods are listed as supported

  Workaround

  Currently users must either:
  1. Use non-quantized models
  2. Use MLX-native quantized models from mlx-community
  3. Re-quantize models themselves: mlx_lm.convert --hf-path Qwen/Qwen3-4B -q --q-method awq

  Environment

  - mlx-lm version: 0.20+
  - Tested model: Qwen/Qwen3-4B-AWQ

PR #730, opened 2026-01-05
Add AWQ/GPTQ weight transformation utilities

ericcurtin:
Add functions to transform AutoAWQ and GPTQ packed weights to MLX quantization format. This includes an unpacking function for AWQ/GPTQ weights and a transformation function that converts the quantization format from AutoAWQ/GPTQ to MLX. The transformation handles both symmetric and asymmetric quantization by computing biases from zero points. Also add a test case to verify the transformation works correctly.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -56,63 +56,24 @@ MODEL_REMAPPING = {
 MAX_FILE_SIZE_GB = 5
 
 
-def _unpack_awq_weights(
-    qweight: mx.array, bits: int, gemm_order: bool = True
-) -> mx.array:
-    """
-    Unpack AWQ/GPTQ quantized weights from packed uint32 format.
-
-    Args:
-        qweight: Packed weights of shape [out_features, in_features // pack_factor]
-            where pack_factor = 32 // bits
-        bits: Number of bits per weight (typically 4)
-        gemm_order: If True, apply AutoAWQ GEMM interleaved unpacking order.
-            AutoAWQ GEMM packs values with order_map = [0, 2, 4, 6, 1, 3, 5, 7],
-            meaning nibble i contains the value originally at position order_map[i].
-
-    Returns:
-        Unpacked weights of shape [out_features, in_features] as uint32
-    """
+def _unpack_awq_weights(qweight: mx.array) -> mx.array:
+    bits = 4
     pack_factor = 32 // bits
     out_features, packed_in = qweight.shape
     in_features = packed_in * pack_factor
-
-    # Create bit masks and shifts for unpacking
     mask = (1 << bits) - 1  # e.g., 0xF for 4-bit
-
-    # Vectorized unpacking: create shifts array and broadcast
-    shifts = mx.arange(pack_factor) * bits  # [0, 4, 8, 12, 16, 20, 24, 28] for 4-bit
-    unpacked = (
-        qweight[..., None] >> shifts
-    ) & mask  # [out_features, packed_in, pack_factor]
-
-    if gemm_order and bits == 4:
-        inverse_order = mx.array([0, 4, 1, 5, 2, 6, 3, 7])
-        unpacked = unpacked[..., inverse_order]
-
-    unpacked = unpacked.reshape(out_features, in_features)
-
-    return unpacked
+    shifts = mx.array([0, 4, 1, 5, 2, 6, 3, 7]) * bits
+    unpacked = (qweight[..., None] >> shifts) & mask
+    return unpacked.reshape(out_features, in_features)
 
 
 def _transform_awq_weights(
     weights: Dict[str, mx.array],
     quantization_config: Dict[str, Any],
 ) -> Tuple[Dict[str, mx.array], Dict[str, Any]]:
-    """
-    Transform AutoAWQ/GPTQ packed weights to MLX quantization format.
-
-    AutoAWQ format (weights are transposed):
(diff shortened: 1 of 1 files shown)


ericcurtin:
https://github.com/ml-explore/mlx-lm/issues/727

mlx_lm/utils.py
```diff
@@ -56,6 +56,139 @@
...
+                unpacked_zeros = _unpack_awq_weights(qzeros, bits)
+
+                # Ensure scales are in [out_features, n_groups] format for MLX
+                # AutoAWQ typically stores as [n_groups, out_features]
+                if scales.shape[0] == n_groups and scales.shape[1] == out_features:
+                    scales = scales.T  # Now [out_features, n_groups]
+                    unpacked_zeros = unpacked_zeros.T  # Now [out_features, n_groups]
+                elif scales.shape[0] == out_features and scales.shape[1] == n_groups:
+                    # Already in MLX format
+                    unpacked_zeros = unpacked_zeros.T
```
awni: Should there be a final `else` here which raises if any of the previous two failed? If not are we always in the first or second case (the second `elif` should just be an `else`)?

mlx_lm/utils.py
```diff
@@ -56,6 +56,139 @@
...
+
+    # Unpack each position
+    unpacked_list = []
+    for i in range(pack_factor):
+        shift = i * bits
+        unpacked = (qweight >> shift) & mask
+        unpacked_list.append(unpacked)
+
+    # Stack and reshape: [pack_factor, out_features, packed_in] -> [out_features, in_features]
+    unpacked = mx.stack(unpacked_list, axis=-1)  # [out_features, packed_in, pack_factor]
```
awni: I think this can be vectorized, something like:
- Make an `arange` of shifts for each position.
- Expand a new last dimension in the qweight 
- shift & mask
- reshape

mlx_lm/utils.py
```diff
@@ -56,6 +56,139 @@
...
+    group_size = quantization_config.get("group_size", 128)
+
+    new_weights = {}
+    processed_prefixes = set()
+
+    for key in list(weights.keys()):
+        if key.endswith(".qweight"):
+            prefix = key[:-8]  # Remove ".qweight"
+            if prefix in processed_prefixes:
+                continue
```
awni: Is this necessary.. when would you encounter the same weight twice?

mlx_lm/utils.py
```diff
@@ -56,6 +56,139 @@
...
+    quantization_config: Dict[str, Any],
+) -> Tuple[Dict[str, mx.array], Dict[str, Any]]:
+    """
+    Transform AutoAWQ/GPTQ packed weights to MLX quantization format.
+
+    AutoAWQ format:
+        - layer.qweight: Packed quantized weights (uint32), shape [out_features, in_features // pack_factor]
+        - layer.qzeros: Packed zero points (uint32), shape [n_groups, out_features // pack_factor]
+        - layer.scales: Quantization scales (float16/float32), shape [n_groups, out_features]
+        - layer.g_idx: Optional group indices (not currently supported)
```
awni: Is that is present should we raise an error? Right now it looks like it's just ignored.

awni:
Looks very cool! I left a few comments, please take a look and then we can get it merged.

mlx_lm/utils.py
ericcurtin: Changed elif to else - if scales isn't in [n_groups, out_features] format, we assume it's already in MLX format [out_features, n_groups].

mlx_lm/utils.py
ericcurtin: Replaced the loop with vectorized operations using mx.arange for shifts and broadcasting.

mlx_lm/utils.py
ericcurtin: Removed the processed_prefixes set and associated check since each .qweight key appears exactly once in the iteration.

mlx_lm/utils.py
ericcurtin: Added an explicit check at the start of _transform_awq_weights that raises a ValueError if any .g_idx key is found.

ericcurtin:
All comments addressed

awni:
I tried running this:
```
 mlx_lm.generate --model Qwen/Qwen3-4B-AWQ --prompt "Tell me a story"
```

and it did not work (error below). Were you able to run it?

```
  File "/Users/awni/mlx-lm/mlx_lm/utils.py", line 471, in load
    model, config = load_model(model_path, lazy, model_config=model_config)
                    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awni/mlx-lm/mlx_lm/utils.py", line 391, in load_model
    weights, quantization = _transform_awq_weights(weights, quantization_config)
                            ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awni/mlx-lm/mlx_lm/utils.py", line 164, in _transform_awq_weights
    biases = -unpacked_zeros.astype(scales.dtype) * scales
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
ValueError: [broadcast_shapes] Shapes (4096,20) and (20,4096) cannot be broadcast.
```

ericcurtin:
@awni I think it should work now

awni (approved):
Looks good, thanks!

Issue #732, opened 2026-01-06
ModuleNotFoundError: No module named 'mlx_lm.tool_parsers' in v0.30.1


sjug:
Created a fresh venv for `v0.30.1`, installed `mlx-lm==0.30.1`, tried to run server. Computer says no.

```
~/Dev/Python
_ uv venv mlx-lm
Using CPython 3.12.11
Creating virtual environment at: mlx-lm
Activate with: source mlx-lm/bin/activate

~/Dev/Python
_ source ~/Dev/Python/mlx-lm/bin/activate

~/Dev/Python via __ v3.12.11 (mlx-lm)
_ uv pip install "mlx-lm==0.30.1" --prerelease=allow
Using Python 3.12.11 environment at: mlx-lm
Resolved 32 packages in 6ms
Installed 32 packages in 57ms
 + anyio==4.12.0
 + certifi==2026.1.4
 + charset-normalizer==3.4.4
 + click==8.3.1
 + filelock==3.20.2
 + fsspec==2025.12.0
 + h11==0.16.0
 + hf-xet==1.2.1rc0
 + httpcore==1.0.9
 + httpx==0.28.1
 + huggingface-hub==1.2.3
 + idna==3.11
 + jinja2==3.1.6
 + markupsafe==3.0.3
 + mlx==0.30.1
 + mlx-lm==0.30.1
 + mlx-metal==0.30.1
 + numpy==2.4.0
 + packaging==25.0
 + protobuf==6.33.2
 + pyyaml==6.0.3
 + regex==2025.11.3
 + requests==2.32.5
 + safetensors==0.7.0
 + sentencepiece==0.2.1
 + shellingham==1.5.4
 + tokenizers==0.22.2
 + tqdm==4.67.1
 + transformers==5.0.0rc1
 + typer-slim==0.21.0
 + typing-extensions==4.15.0
 + urllib3==2.6.2

~/Dev/Python via __ v3.12.11 (mlx-lm)
_ mlx_lm.server --model mlx-community/MiniMax-M2.1-6bit --host 0.0.0.0
Traceback (most recent call last):
  File "/Users/sjug/Dev/Python/mlx-lm/bin/mlx_lm.server", line 4, in <module>
    from mlx_lm.server import main
  File "/Users/sjug/Dev/Python/mlx-lm/lib/python3.12/site-packages/mlx_lm/__init__.py", line 9, in <module>
    from .convert import convert
[...]

Issue #729, opened 2026-01-05
The mlx-community/DeepSeek-V3.2-4bit model fails to load in LM Studio (mlx_lm 0.30.1)


DiscoStew6082:
See this for details.  I don't know where to download mlx_lm 0.30.1 to manually test/debug if needed:
https://github.com/lmstudio-ai/lmstudio-bug-tracker/issues/1352

PR #733, opened 2026-01-06
Fix mlx-lm release

awni:
Closes https://github.com/ml-explore/mlx-lm/issues/732 and I think https://github.com/ml-explore/mlx-lm/issues/729 as well.


angeloskath (approved):
👍

PR #737, opened 2026-01-06
Add OPT model implementation

ericcurtin:
This commit adds a complete implementation of the OPT (Open Pre-trained
Transformer) model architecture. The implementation includes all core
components such as the attention mechanism, MLP layers, decoder layers,
and position embeddings with the characteristic OPT offset of 2.

Additionally, this commit introduces functionality to convert PyTorch
.bin weight files to safetensors format during model loading. This
enables seamless loading of OPT models from PyTorch checkpoints by
automatically converting them to the safetensors format used by MLX. The
conversion handles shared weights properly by cloning tensors to break
shared memory references. The download and loading utilities have been
updated to support this conversion workflow.

Signed-off-by: Eric Curtin <eric.curtin@docker.com>

### DIFF
--- a/mlx_lm/models/opt.py
+++ b/mlx_lm/models/opt.py
@@ -1,0 +1,248 @@
+# Copyright © 2026 Apple Inc.
+
+from dataclasses import dataclass
+from typing import Any, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+
+
+@dataclass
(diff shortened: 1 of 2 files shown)


ericcurtin:
https://github.com/vllm-project/vllm-metal/issues/55

ericcurtin:
Surprised this didn't auto kickoff a workflow, with many projects, once you get the first PR in, it auto-kicks off builds from then on, must be a default github setting

awni:
I think this model is like 4 years old and not really used. I don't think it's worth adding especially since it doesn't have a standard format.

PR #738, opened 2026-01-06
Fix sliding window batching

### LINKED ISSUE #735: [Feature Request] Add --no-batch flag to disable request batching in server

awni:
This fixes some issues with batching and the batched sliding window cache. Should close #735 

The code in the batched sliding window KV cache is getting pretty hairy.. I don't have any good suggestions for simplifying it.. but we may need to add some more tests to make sure the different cases are all correct.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1071,7 +1071,8 @@ class BatchGenerator:
             prompt_cache = _merge_caches(caches)
 
             for c in prompt_cache:
-                c.prepare(lengths=lengths, right_padding=padding)
+                # subtract one from lengths since we don't process the last token during prefill
+                c.prepare(lengths=[l - 1 for l in lengths], right_padding=padding)
 
             while inputs.shape[1] > 1:
                 n_to_process = min(self.prefill_step_size, inputs.shape[1] - 1)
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
Thanks!

PR #741, opened 2026-01-08
Support distributed inference in the server

angeloskath:
This proved to be more tedious than I expected originally due to various races in propagating requests and in general syncing the independent processes.

The main solution is to use the distributed comms themselves to propagate requests and sync the processes.

1. It is the fastest.
2. No extra sockets and servers in python required.
3. It gets put on the graph so it ensures the ordering of operations. Importantly, it needs to be put in the generation stream to avoid races with the generation stream.

Notable weirdness: The `TimeBudget` class. Before this PR we would keep generating until we had no more tokens to generate or half a second passed. Then we would check for new requests. The `TimeBudget` class enables the same behavior in a distributed setting by adaptively choosing a number of iterations that is as close to half a second as possible.

Sketch:

- Requests go to rank 0
- Rank 0 broadcasts the request to all nodes
- All nodes collaboratively generate according to the time budget
- Rank 0 broadcasts any request that exited early to all nodes
- Repeat

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -6,21 +6,17 @@ import json
 import logging
 import pickle
 import platform
-import select
 import socket
-import subprocess
 import time
 import uuid
 import warnings
 from collections import deque
 from dataclasses import dataclass, field
-from functools import partial
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from pathlib import Path
 from queue import Empty as QueueEmpty
 from queue import Queue
-from socketserver import BaseRequestHandler, ThreadingTCPServer
-from threading import Lock, Thread
+from threading import Thread
 from typing import (
     Any,
     Callable,


mlx_lm/server.py
```diff
@@ -663,6 +777,20 @@ def progress_callback(info):
...
+                            top_p=args.sampling.top_p,
+                            top_k=args.sampling.top_k,
+                            min_p=args.sampling.min_p,
+                            xtc_probability=args.sampling.xtc_probability,
+                            xtc_threshold=args.sampling.xtc_threshold,
+                            xtc_special_tokens=[
+                                tokenizer.eos_token_id,
+                                tokenizer.encode("\n"),
+                            ],
+                        ),
```
awni: nit:
```suggestion
                        sampler=_make_sampler(args, tokenizer),
```

awni: Also curious, why are you adding a default sampler? The sampler already gets added per prompt when you insert the prompt.

mlx_lm/server.py
```diff
@@ -736,8 +859,20 @@ def progress_callback(info):
...
                             uids_to_remove.append(r.uid)
 
-                    if uids_to_remove:
-                        batch_generator.remove(uids_to_remove)
+                uids_to_remove = self._share_object(uids_to_remove)
+                if uids_to_remove:
+                    with mx.stream(generation_stream):
+                        caches = batch_generator.remove(
+                            uids_to_remove, return_prompt_caches=True
+                        )
```
awni, line 849: Would it work if `batch_generator` did the removal in the `generation_stream` rather than using it in `server`?

awni: Hmm I see it's also used elsewhere.

mlx_lm/server.py
```diff
@@ -736,8 +859,20 @@ def progress_callback(info):
                         if result["ctx"]._should_stop:
                             uids_to_remove.append(r.uid)
 
-                    if uids_to_remove:
-                        batch_generator.remove(uids_to_remove)
+                uids_to_remove = self._share_object(uids_to_remove)
+                if uids_to_remove:
```
awni, line 845: It looks like this came out one-level so now we might keep processing examples in the batch that have already met the stop condition?

mlx_lm/server.py
angeloskath: Yeah good point that must have been an oversight on some refactoring or sth. Removing...

mlx_lm/server.py
angeloskath: Yes indeed. That is on purpose otherwise we need to communicate at every step because the communication pattern needs to be deterministic. Except for rank 0 the others don't know if the condition was met or not.

This doesn't affect the results because the thread handling the http already got the result, trimmed, returned it and anything else it needed to do. The generation thread just continues on until the time budget is exhausted and then removes anything that needed removing.

Btw the reason for not communicating at every step is not because the communication is not fast it is because this communication breaks the pipeline unfortunately. We could make a fixed size communication that would probably be enough and doesn't break the pipeline but that seemed like a good enough solution as well.

kernelpool:
Thanks for working on this! I ran into an issue with clients using temperature > 0 and each rank sampling independently with its own random state. It can be fixed by adding something like the following to `BatchGenerator._step()` before returning.

```python
if is_distributed:
    sampled = mx.distributed.all_sum(
        sampled if self._distributed_group.rank() == 0 else mx.zeros_like(sampled)
    )
```

angeloskath:
Oh great catch @kernelpool. I think it is better to sync up random states actually.

angeloskath:
> Sorry for the delay.

No worries. Will fix some rogue forgotten imports and merge it later today.

PR #743, opened 2026-01-08
Make cache list batchable

awni:
- Sets a `merge` property an cache types to make it easier to turn them into batch caches and check if they support batching.
- Implements cache list merging using the above property

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -172,10 +172,6 @@ class _BaseCache:
         obj.meta_state = meta_state
         return obj
 
-    @classmethod
-    def merge(_, caches):
-        return BatchKVCache.merge(caches)
-
 
 class ConcatenateKVCache(_BaseCache):
     """ConcatenateKVCache the simplest KV cache implementation.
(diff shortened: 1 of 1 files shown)


mlx_lm/models/cache.py
```diff
@@ -172,6 +172,10 @@ def from_state(cls, state, meta_state):
         obj.meta_state = meta_state
         return obj
 
+    @classmethod
+    def merge(_, caches):
```
angeloskath: This being in the `_BaseCache` will mean everything is batchable. Also we 'll have the issue where the QuantizedKVCache will return an unquantized one corrupting the cache and the ConcatenateKVCache will switch to a KVCache object.

Basically, this should be in the child classes to signify whether they do support batching ...

angeloskath (approved):
Looks great! I left one comment that will need fixing unless I misunderstood sth.

mlx_lm/models/cache.py
awni: Oops I mean to put that in `KVCache`. Good catch.

PR #748, opened 2026-01-09
Fix batch generation for IQuestLoopCoder model

kernelpool:
This one slipped through the cracks.

`cache[0].offset` is a scalar for regular `KVCache` but an array of per-batch offsets for `BatchKVCache.offset`, causing batch generation to fail.

### DIFF
--- a/mlx_lm/models/iquestloopcoder.py
+++ b/mlx_lm/models/iquestloopcoder.py
@@ -167,7 +167,6 @@ class IQuestLoopCoderModel(nn.Module):
         if cache is None:
             cache = [None] * (2 * len(self.layers))
 
-        offset = len(cache[0]) if cache[0] is not None else 0
         mask = create_attention_mask(h, cache[0])
         window_mask = create_attention_mask(
             h, cache[len(self.layers)], window_size=self.loop_window_size
@@ -176,6 +175,7 @@ class IQuestLoopCoderModel(nn.Module):
         loop1_kv = []
         for layer, c in zip(self.layers, cache):
             h_norm = layer.input_layernorm(h)
+            offset = c.offset if c is not None else 0
             q1, k1, v1 = layer.self_attn.get_qkv(h_norm, offset)
 
             if c is not None:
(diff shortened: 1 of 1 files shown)


mlx_lm/models/iquestloopcoder.py
```diff
@@ -167,7 +167,7 @@ def __call__(
         if cache is None:
             cache = [None] * (2 * len(self.layers))
 
-        offset = cache[0].offset if cache[0] is not None else 0
+        offset = len(cache[0]) if cache[0] is not None else 0
```
awni: I don't understand, why does the original version not work with batch generation?

kernelpool: The wrong value gets used for rope positioning. I noticed this when using the server and using OpenCode ("hello" and model responds with illegible text). It works when using mlx_lm.generate.

awni:
@kernelpool I'm noticing the model isn't getting detokenized properly even in single example generation. Are you seeing the same?

```
In ▁the ▁tw ilight ▁of ▁his ▁years , ▁Albert ▁Einstein , ▁the ▁man ▁who ▁had ▁res h aped ▁the ▁very ▁fabric ▁of ▁reality , ▁found ▁himself ▁in ▁a ▁small , ▁quiet ▁village ▁in ▁S witz er < | reserved _ token | > r land . ▁The ▁village ▁was ▁nest led ▁in ▁the ▁heart ▁of ▁the ▁Al ps , ▁surrounded ▁by ▁snow - c apped ▁peaks ▁and ▁s ere ne ▁lakes .
```

mlx_lm/models/iquestloopcoder.py
awni: `len(cache[0])` is not the right value though. On the BatchKVCache it gives the padded length which should be a bug. Using the offset should be correct.. so I think there must be a bug somewhere else.

awni:
Seems like switching from `transformers==5.0.0rc2` to `transformers==5.0.0rc1` is breaking the detokenization 🤔 .

mlx_lm/models/iquestloopcoder.py
awni: This is the problem btw:
```
import mlx.core as mx

class C:
    def __init__(self):
        self.arr = mx.array([0])

    def update(self, n):
        self.arr += n

c = C()
arr = c.arr
c.update(3)
print(arr)
```

awni (approved):
Thanks for catching this problem.

Issue #564, opened 2025-10-23
Add support for Nemotron Super 49b v1.5


MattBash17:
Add support for Nemotron Super 49b v1.5. The model has been around for a while, but proper mlx support is still lacking, often throwing errors like "Error in iterating prediction stream: AttributeError: 'NoneType' object has no attribute 'shape'" (in LM Studio).

awni:
Hugging Face link: https://huggingface.co/nvidia/Llama-3_3-Nemotron-Super-49B-v1_5

lazarust:
Working on this. Should have a PR this weekend.

lazarust:
PR: #756, is ready to review! Was a pretty quick and easy fix.

PR #756, opened 2026-01-14
Adds support for Nemotron Super 49b v1.5

lazarust:
Closes: #564

### DIFF
--- a/mlx_lm/models/nemotron-nas.py
+++ b/mlx_lm/models/nemotron-nas.py
@@ -342,11 +342,17 @@ class NemotronNASModel(nn.Module):
         h = self.embed_tokens(inputs)
 
         if cache is None:
-            cache = [None] * len(self.layers)
+            cache = [None] * self.num_attn_layers
 
         mask = create_attention_mask(h, cache[0])
 
-        for layer, c in zip(self.layers, cache):
+        cache_idx = 0
+        for layer in self.layers:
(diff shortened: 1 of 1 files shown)


mlx_lm/models/nemotron-nas.py
```diff
@@ -380,3 +381,12 @@ def sanitize(self, weights):
...
     def layers(self):
         return self.model.layers
+
+    def make_cache(self):
+        caches = []
+        for layer in self.layers:
+            if layer.self_attn is not None:
+                caches.append(KVCache())
+            else:
+                caches.append(_BaseCache())
```
awni: `_BaseCache` is an abstract base class.. I don't think we should use it as a placeholder like this. It will probably break in other places like batch generation and such.

Instead we should change the forward function to something like:

```python
        if cache is None:
            cache = [None] * self.num_attn_layers # <-- set that property on init

        mask = create_attention_mask(h, cache[0])
        i = 0
        for layer in self.layers:
            if layer.self_attn is not None:
              c = cache[i]
              i += 1
            else:
              c = None
            h = layer(h, mask, cache=c)
```

lazarust: Ah that makes sense. Fixed the PR!

awni (approved):
Looks good, thanks!

PR #758, opened 2026-01-14
feat: Add --revision parameter to all CLI tools

malibayram:
## Summary

This PR adds the `--revision` parameter to all CLI tools that load models from HuggingFace Hub, enabling users to specify a particular model version (branch, tag, or commit hash) when loading models.

## Motivation

When working with models on HuggingFace Hub, users often need to load a specific version of a model rather than the latest. The underlying `load()` function in `utils.py` already supports the `revision` parameter, but it was not exposed through the CLI tools. This change provides a consistent interface across all CLI entry points.

### CLI Tools Updated (10 files)
- `convert.py` - Added `--revision` argument to argument parser
- `fuse.py` - Added `--revision` argument and passed to `load()`
- `lora.py` - Added `--revision` argument and passed to `load()`
- `generate.py` - Added `--revision` argument and passed to `load()`
- `chat.py` - Added `--revision` argument and passed to `load()`
- `perplexity.py` - Added `--revision` argument and passed to `load()`
- `benchmark.py` - Added `--revision` argument and passed to `load()`
- `cache_prompt.py` - Added `--revision` argument and passed to `load()`
- `evaluate.py` - Added `--revision` argument to `MLXLM` class and CLI
- `server.py` - Added `--revision` argument and passed to all `load()` calls

### Tests
- Updated `tests/test_chat.py` to include `--revision` in the argument parser tests

### Documentation
- Updated `README.md` to mention the `--revision` option

## Example Usage

```bash
# Load a specific commit
mlx_lm.generate --model mlx-community/Llama-3.2-3B-Instruct-4bit --revision abc123 --prompt "hello"

# Load a specific branch
mlx_lm.convert --model mistralai/Mistral-7B-Instruct-v0.3 --revision main -q

# Load a specific tag
mlx_lm.chat --model mlx-community/Llama-3.2-3B-Instruct-4bit --revision v1.0
```

## Testing

- [x] All modified CLI tools accept `--revision` parameter (verified with `--help`)
- [x] Parameter is passed correctly through to `load()` function
- [x] Existing tests continue to pass
- [x] New test case added for `--revision` in `test_chat.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,17 @@
-## MLX LM 
+## MLX LM
 
 MLX LM is a Python package for generating text and fine-tuning large language
 models on Apple silicon with MLX.
 
 Some key features include:
 
-* Integration with the Hugging Face Hub to easily use thousands of LLMs with a
-  single command. 
-* Support for quantizing and uploading models to the Hugging Face Hub.
-* [Low-rank and full model
+- Integration with the Hugging Face Hub to easily use thousands of LLMs with a
+  single command.
+- Support for quantizing and uploading models to the Hugging Face Hub.
+- [Low-rank and full model
   fine-tuning](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/LORA.md)
   with support for quantized models.
-* Distributed inference and fine-tuning with `mx.distributed`
+- Distributed inference and fine-tuning with `mx.distributed`
 
 The easiest way to get started is to install the `mlx-lm` package:
 
@@ -53,10 +53,11 @@ mlx_lm.generate -h
 ```
 
 The default model for generation and chat is
-`mlx-community/Llama-3.2-3B-Instruct-4bit`.  You can specify any MLX-compatible
+`mlx-community/Llama-3.2-3B-Instruct-4bit`. You can specify any MLX-compatible
 model with the `--model` flag. Thousands are available in the
 [MLX Community](https://huggingface.co/mlx-community) Hugging Face
-organization.
+organization. You can also load a specific revision (branch, tag, or commit) of
+a model using `--revision`.
 
 ### Python API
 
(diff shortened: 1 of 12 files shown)


awni:
Thanks for the PR, but I don't think this case is common enough to merit adding this.

malibayram:
Hi @awni, thanks for the review. I understand the goal of keeping the CLI minimal, but I believe this addresses a major functional gap for anyone doing iterative training.

**The Use Case:**
I am training a model and pushing checkpoints to the Hub as separate commits (e.g. "Epoch 1", "Epoch 2", "Muon Optimized"). You can see my repo here: [alibayram/magibu-128k-multilingual/commits/main](https://huggingface.co/alibayram/magibu-128k-multilingual/commits/main)

**The Problem:**
I need to convert and test *each specific revision* on my local machine to find the best performing one. Currently, the CLI tools ([convert](cci:1://file:///Users/alibayram/Desktop/pythons/mlx-lm/mlx_lm/server.py:113:0-134:26), [evaluate](cci:1://file:///Users/alibayram/Desktop/pythons/mlx-lm/mlx_lm/lora.py:303:0-314:65), [chat](cci:1://file:///Users/alibayram/Desktop/pythons/mlx-lm/mlx_lm/server.py:113:0-134:26)) provide **no way** to target these specific versions—they all force me to use the latest `HEAD`.

This effectively makes the CLI unusable for evaluating training history or comparing versions. I am forced to write custom Python scripts for every single step just to pass `revision=...`.

**Why merge this?**
I can (and do) modify my local branch to unblock myself, so this isn't just about my personal usage. I am submitting this because *any* model builder using `mlx-lm` will inevitably hit this same wall when trying to evaluate or convert specific model checkpoints.

Adding `--revision` turns the CLI into a tool that supports the full lifecycle of a model (development -> versioning -> evaluation), rather than just a "demo" tool that only sees the latest version.

I hope this context helps clarify why this is such a common/necessary workflow for the community!

PR #761, opened 2026-01-14
Add gpt-oss sharding

Evanev7:
Add tensor sharding for GPT-OSS. This one we have tested more significantly, though let me know how best to deal with the custom _all_to_sharded

### DIFF
--- a/mlx_lm/models/gpt_oss.py
+++ b/mlx_lm/models/gpt_oss.py
@@ -284,21 +284,7 @@ class Model(nn.Module):
         N = group.size()
         R = group.rank()
 
-        # GPT_OSS bias is a 2d array not a 3d array, so we need to shard it slightly differently
-        segments: int = 1
-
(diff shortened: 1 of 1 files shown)


mlx_lm/models/gpt_oss.py
```diff
@@ -268,6 +279,66 @@ def sanitize(self, weights):
...
+        segments: int = 1
+
+        def _all_to_sharded(path: str, weight: mx.array):
+            if path.endswith("bias"):
+                return weight.ndim - 1, segments
+            return max(weight.ndim - 2, 0), segments
+
+        def _sharded_to_all(path: str, weight: mx.array):
+            if path.endswith("bias"):
+                weight /= n
```
awni: What is this little `n` supposed to be? I can't find it defined anywhere.

Evanev7: ah - meant to be big N

PR #762, opened 2026-01-15
fix: unused batch_size parameter for mlx_lm.evaluate

AndrewTan517:
Fix `--batch-size` parameter being unused by `mlx_lm.evaluate`. Changed so that `--batch-size` parameter is passed into `__init__()` of `MLXLM` class as parameter `batch_size` and assigned to attribute `self._batch_size`.


awni (approved):
Looks good, thanks!

PR #764, opened 2026-01-15
fix(falcon_h1): support tied embeddings and correct muP scaling

solarpunkin:
Fixes `ValueError: Missing 1 parameters: lm_head.weight` when loading Falcon-H1 models (e.g., `tiiuae/Falcon-H1-Tiny-90M-Instruct`).

### Changes:
*   **Tied Embeddings:** Conditional initialization of `lm_head` based on `tie_word_embeddings` config. Uses `embed_tokens.as_linear()` in the forward pass when tied.
*   **Scaling Correction:** Falcon-H1 uses Maximal Update Parameterization with different multipliers for embeddings (`~5.6`) and the output head (`~0.039`). Since `embed_tokens` is scaled permanently during load, this adds a correction factor (`lm_head_mult / embedding_mult`) to the logits to ensure mathematical correctness when weights are shared.

### Verification:
Validated locally with `tiiuae/Falcon-H1-Tiny-90M-Instruct`; model loads successfully and generates coherent text.

### DIFF
--- a/mlx_lm/models/falcon_h1.py
+++ b/mlx_lm/models/falcon_h1.py
@@ -470,8 +470,6 @@ class Model(nn.Module):
             if name.endswith("embed_tokens.weight"):
                 param *= args.embedding_multiplier
             elif name.endswith("lm_head.weight"):
-                if args.tie_word_embeddings:
-                    continue
                 param *= args.lm_head_multiplier
             elif name.endswith("q_proj.weight") or name.endswith("k_proj.weight"):
                 param *= args.attention_in_multiplier


mlx_lm/models/falcon_h1.py
```diff
@@ -464,6 +470,8 @@ def sanitize(self, weights):
             if name.endswith("embed_tokens.weight"):
                 param *= args.embedding_multiplier
             elif name.endswith("lm_head.weight"):
+                if args.tie_word_embeddings:
+                    continue
```
awni: There shouldn't be a parameter `lm_head` if the weights are tied right? So this check seems superfluous.

```suggestion
```

solarpunkin: Yes this check is indeed redundant. I'll remove it.

PR #768, opened 2026-01-17
Fix LongCat Flash extended context support

kernelpool:
This PR fixes support for variants of LongCat Flash using YarnRoPE, e.g. [LongCat-Flash-Thinking-2601](https://huggingface.co/meituan-longcat/LongCat-Flash-Thinking-2601).
Previously this model would produce garbled output after 8192 tokens.

Additional notes:
- The HuggingFace model is currently missing a `model_type` so this must be manually set to `longcat_flash` prior to using `mlx_lm.convert`.
- This PR needs the fix addressed in #767


awni (approved):
LGTM, thanks!

PR #774, opened 2026-01-19
moved more activation functions to activations module

Goekdeniz-Guelmez:
Found some more activation functions that should be moved as well


mlx_lm/models/activations.py
```diff
@@ -41,3 +41,64 @@ def __init__(
...
+    def __init__(self):
+        super().__init__()
+
+    def __call__(self, x, gate):
+        return swiglu_clamped(x, gate)
+
+
+@partial(mx.compile, shapeless=True)
+def _silu_mul(gate: mx.array, up: mx.array) -> mx.array:
+    return nn.silu(gate) * up
```
awni: That should just be the swiglu right?

Goekdeniz-Guelmez: Yes

Goekdeniz-Guelmez:
@awni, sorry that it took so long but this should be cleanly mergable now

mlx_lm/models/activations.py
```diff
@@ -41,3 +41,59 @@ def __init__(
...
+
+    glu_scaled = alpha * x_glu
+    sig = mx.sigmoid(glu_scaled)
+
+    out_glu = x_glu * sig
+    # Note we add an extra bias of 1 to the linear layer
+    return out_glu * (x_linear + 1)
+
+
+class SwiGLU(nn.Module):
```
angeloskath: This should be renamed to `SwiGLUClamped` because it is at the very least confusing as is.

angeloskath (changes requested):
Looks good, I think the `SwiGLU` should be renamed to match what it actually does and then we can merge.

mlx_lm/models/activations.py
Goekdeniz-Guelmez: on it!

Goekdeniz-Guelmez:
@angeloskath fixed, also moved the base SwiGlu in switch_layers to activations as well.

zcbenz:

PR #778, opened 2026-01-20
import logging as it throws no logging error in place of actual error

Maanas-Verma:


mlx_lm/utils.py
```diff
@@ -303,6 +303,7 @@ def load_model(
     weight_files = glob.glob(str(model_path / "model*.safetensors"))
 
     if not weight_files and strict:
+        import logging
         logging.error(f"No safetensors found in {model_path}")
```
awni: Let's just get rid of that log as it's unnecessary since we raise the error right after it.
```suggestion
```

PR #780, opened 2026-01-20
Update glm4_moe_lite To Store KV Latent in Cache

N8python:
Updates GLM4-MOE-Lite to not store full KV in cache and instead used compressed MLA latent. Tested w/ 6-bit version of model (I don't have RAM for bf16). This works because we can move the latent compression into the query matrix (basically put the queries in kvlatent space) and then project out of kvlatent space afterward.

KLDiv 0.042 w/ original implementation (computed over Tulu 3 SFT mixture). Benches below (M3 Max, Low Power Mode):

| Prompt/Gen | Trials | New prompt_tps | New generation_tps | New peak_memory (GB) | Old prompt_tps | Old generation_tps | Old peak_memory (GB) |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 256/256 | 5 | 417.558 | 34.120 | 24.636 | 447.922 | 35.137 | 24.863 |
| 8192/256 | 1 | 339.855 | 25.228 | 27.073 | 430.655 | 14.470 | 34.163 |
| 16384/256 | 1 | 208.088 | 20.098 | 29.287 | 291.520 | 8.626 | 43.574 |

Most important: Peak memory at 16384 prompt is down from 43.574 GB to 29.287 GB. Given model size is 22.68GB, Peak went from 20.89400 GB over model size to 6.607 GB over model size. That's a 3.16x reduction in memory overhead beyond model size. Also gets more than double the decode speed at long contexts! Cost is slight hit to prompt processing speed.

Full disclosure: The code + impl was primarily done by GPT-5.2-Codex, supervised by me.

### DIFF
--- a/mlx_lm/models/glm4_moe_lite.py
+++ b/mlx_lm/models/glm4_moe_lite.py
@@ -446,18 +446,6 @@ class Model(nn.Module):
         return self.lm_head(out)
 
     def sanitize(self, weights):
-        # Stack experts
-        for l in range(self.args.num_hidden_layers):
-            prefix = f"model.layers.{l}"
-            for n, m in [("w1", "gate_proj"), ("w2", "down_proj"), ("w3", "up_proj")]:
-                for k in ["weight", "scales", "biases"]:
-                    if f"{prefix}.mlp.experts.0.{m}.{k}" in weights:
-                        to_join = [
-                            weights.pop(f"{prefix}.mlp.experts.{e}.{m}.{k}")
-                            for e in range(self.args.n_routed_experts)
-                        ]
-                        weights[f"{prefix}.mlp.switch_mlp.{m}.{k}"] = mx.stack(to_join)
-
         def is_mpt_layer(key):
             subkeys = key.split(".")
             if len(subkeys) < 3:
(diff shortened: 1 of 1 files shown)


ivanfioravanti:
Top one! Gonna test this later today!

ivanfioravanti:
I did some tests. text generation and memory consumption much better! But prompt processing is slower. 
Here a chart. Data from Today 20th Jan are done using this PR.

comparison_chart

ivanfioravanti:
@N8python I did a PR towards your main that helps a little with smaller contexts.
comparison_chart

N8python:
Added. Essentially free boost to early PP.

ronaldmannak:
How is this different from the DeepSeek V3 MLA implementation? Can we improve DeepSeek as well or was that already done this way?

N8python:
This could be used to improve deepseek! Another PR could be made.

d-b:
Really hoping this PR works out. The one thing holding me back from using GLM-4.7-Flash at the moment is the high memory usage.

mlx_lm/models/glm4_moe_lite.py
```diff
@@ -136,29 +226,98 @@ def __call__(
...
+
+        offset_zero = False
+        if cache is not None:
+            offset = cache.offset
+            if isinstance(offset, mx.array):
+                offset_zero = bool(mx.all(offset == 0).item())
+            else:
+                offset_zero = offset == 0
+        use_fast_prefill = L > 1 and (cache is None or offset_zero)
+        if use_fast_prefill:
```
awni: Since we process the prompt in chunks I don't really think this is worth keeping. It only matters for the first 2048 tokens which will usually be quite fast anyway.

awni: If we get rid of it then it makes sense to serialize the kv projection weights in the split format. Right now it looks like they get serialized twice.

N8python: Ah, yeah. That makes sense. We could get rid of it and drop the fast prefill path. I agree.

N8python: Will you remove it or should I?

awni: I will do it.

mlx_lm/models/glm4_moe_lite.py
```diff
@@ -45,12 +45,88 @@ class ModelArgs(BaseModelArgs):
...
-    rope_traditional: bool = True
     attention_bias: bool = False
     attention_dropout: float = 0.0
     partial_rotary_factor: float = 1.0
     tie_word_embeddings: bool = False
     num_nextn_predict_layers: int = 1
+    quantization: Optional[Dict[str, Any]] = None
+
+
+class MultiLinear(nn.Module):
```
awni, line 56: I will probably refactor these out in a follow-on so other models which use MLA can use them.

N8python: Awesome. Makes sense.

PR #781, opened 2026-01-20
Fix GLM4 MoE Lite RoPE and MoE layer selection

### LINKED ISSUE #777: GLM 4.7 Flash bug
GLM Flash 4.7 is working fine for the first few hundred tokens of a response, then always seems to get stuck in a loop. Here is a response

(also note I was _not_ mocking the poor thing, just pointing out that it kept getting stuck in a loop

> The user is frustrated because my previous responses contained massive, malformed code blocks (likely due to a generation error or token limit issue where I started repeating the same function definition over and over, resulting in a "loop").
> 
> The user explicitly pointed out: "you’re still getting caught in a loop for some reason. I reckon MLX has a bug..."
> 
> My task is to:
> 
> Acknowledge the error (don't make excuses, just admit it).
> Provide a complete, clean Tetris game in HTML5/JS.
> Ensure the code is concise enough to fit within a single response without getting cut off or looping.
> The code needs actual functionality: drawSquare, moveDown (gravity), collision detection, rotation.
> Plan for the Code:
> 
> HTML: Canvas tetris, Sidebar with Score, Next Piece.
> CSS: Centering layout2;
> }
> function drawSquare(x) { // Standard drawing
> ctx.fillStyle = colors[color];
> if (color == 'violet') {
> } else { // Default fill, handled by switch case usually
> }
> ctx.fillRect;
> }
> 
> function drawBoard(board) { // Draw static blocks already placed
> 
> for(let 7; } else {
> ctx.fillStyle;
> }
> 
> function drawSquare, { // Standard drawing
> } if (7; } else {
> ctx.fillStyle;
> }
> function drawSquare, { // Standard drawing
> } if (7; } else {
> ctx.fillStyle;
> }
> function drawSquare, { // Standard drawing
> } if (7; } else {
> ctx.fillStyle;
> }
> function drawSquare, { // Standard drawing
> } if (7; } else {
> ctx.fillStyle;
> }
> function drawSquare, { // Standard drawing
> } if (7; } else {
> ctx.fillStyle;
> }
> function drawSquare, { // Standard drawing
> } if (7; } else {
> ctx.fillStyle;
> }
> 
> Wait, I need to stop hallucinating the bad code. The user is mocking me because my previous output was a repetitive, broken loop of if (7; } else { ctx.fillStyle;} function drawSquare....
> 
> I need to provide a working script. I will write it out carefully.
> 
> Tetris Logic:
> 
> board: 12x20 array.
> piece object with shape, color and position (y,x).
> Collision: Check if piece coordinates are inside board bounds or overlap existing blocks.
> Rotation:
> Create a new matrix for the rotated piece (transpose + reverse rows).
> Game Loop: setInterval or requestAnimationFrame. Since it's simple, a fixed interval (e.g., 100ms) is easier to read.
> Let's construct the final response:
> 
> Apologize for the loop.
> Provide a single, valid code block.
> Code Structure:

```
<!DOCTYPE html>
<html lang="en">
<head>...</head>

<script type='text/javascript'>
  // Constants
  const canvas = document.getElementById('tetris');
  context.fillStyle;
}

function drawBoard(board) { // Draw static blocks already placed

 for(let y = 7;
}

function drawSquare, {
} if (5; } else { // The loop is killing me. I will just write the code.
}

function draw;
} if (7; } else { // Stop. Just give them the code.
}

function draw;
} if (7; } else { // Stop. Just give them the code.
}

function draw;
} if (7; } else { // Stop. Just give them the code.
}

function draw;
} if (7; } else { // Stop. Just give them the code.
}
```

ivanfioravanti:
Tested: generate with temp=1.0 (same prompt) no longer repeats. This fixes #777

### DIFF
--- a/mlx_lm/models/glm4_moe_lite.py
+++ b/mlx_lm/models/glm4_moe_lite.py
@@ -51,6 +51,7 @@ class ModelArgs(BaseModelArgs):
     partial_rotary_factor: float = 1.0
     tie_word_embeddings: bool = False
     num_nextn_predict_layers: int = 1
+    quantization: Optional[Dict[str, Any]] = None
 
 
 class Glm4MoeLiteAttention(nn.Module):
@@ -101,6 +102,7 @@ class Glm4MoeLiteAttention(nn.Module):
             * (self.q_head_dim - self.qk_rope_head_dim + self.v_head_dim),
             bias=False,
         )
+        self._init_presplit_kv_b_proj()
 
         self.o_proj = nn.Linear(
             self.num_heads * self.v_head_dim,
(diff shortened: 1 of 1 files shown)


mlx_lm/models/glm4_moe_lite.py
```diff
@@ -45,12 +45,15 @@ class ModelArgs(BaseModelArgs):
...
     rope_scaling: Optional[Dict] = None
+    rope_parameters: Optional[Dict] = None
     rope_traditional: bool = True
+    rope_interleave: Optional[bool] = None
     attention_bias: bool = False
     attention_dropout: float = 0.0
     partial_rotary_factor: float = 1.0
     tie_word_embeddings: bool = False
     num_nextn_predict_layers: int = 1
+    mlp_layer_types: Optional[List[str]] = None
```
awni: None of these parameters exist in the config so I'm pretty confused as to how they could fix anything.

awni:
@ivanfioravanti is there a prompt that you ran that looped before this diff that did not after? I would like to test it out to see. The only thing I can see that would change anything here is the addition of 1e-20 to the denominator in the MoE (which did not fix looping for low precision + high temp when I added it yesterday).

ivanfioravanti:
Pretty simple: "Write a very long poem on LLMs, then review it and write a new version." 
I totally agree fix is 1e-20 that is present in many other models.
Params above are used on Transformer PR #43031, but effectively they are not in the config file so they are useless. 👀

ivanfioravanti:
I think this can be closed and the #780 used that is much much better from memory and generation.

ivanfioravanti:
Gonna do a new PR to clean these unused param from it for @N8python

ivanfioravanti:
All unused params removed.

awni:
I think we should merge the fix here and get it in rather than bundle it in #780 which may take some time to land.

awni:
Actually i see this PR got a little mixed.

PR #787, opened 2026-01-21
Fix: Added missing logging import causing `NameError`

g-aggarwal:
## Summary of changes
- Added missing import for [logging](https://docs.python.org/3/library/logging.html) in [utils.py](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/utils.py)

## Description
Fixes a `NameError` triggered when the library or external modules attempt to use the logging namespace.

## Validation
Verified the fix by successfully calling the logger through the utility namespace:
```
python -c "from mlx_lm import utils; print(utils.logging.root)"
```

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -5,6 +5,7 @@ import glob
 import importlib
 import inspect
 import json
+import logging
 import os
 import resource
 import shutil


awni:
I'm not following the purpose of this.. but I don't think we use logging anywhere in the utils so this should not be necessary. Feel free to comment if I missed something and we can reopen.

g-aggarwal:
@awni Apologies for the lack of context. Line 609 calls logging.set_verbosity_info(), which is a specific huggingface_hub utility method rather than a standard library call. Currently, this triggers a NameError. I have updated the PR to from huggingface_hub import logging to correctly resolve the reference and enable the intended verbosity control.

PR #788, opened 2026-01-21
Fix: add missing `huggingface_hub.logging` import to `utils.py`

g-aggarwal:
### Summary
- Add missing `huggingface_hub.logging` import to `mlx_lm/utils.py`.

### Description
Fixes a `NameError` on line **609** within the load function. The code calls `logging.set_verbosity_info()` without the import, which crashes the library during model initialization.

### Validation
Verified the fix by ensuring the huggingface_hub logging namespace is correctly resolved through the utility module:
```
python -c "from mlx_lm import utils; utils.logging.set_verbosity_info()"
```
PS - @awni this should make more sense

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -8,6 +8,7 @@ import json
 import os
 import resource
 import shutil
+from huggingface_hub import logging
 from pathlib import Path
 from textwrap import dedent
 from typing import (


awni:
It doesn't make sense, we import logging just above line 609: https://github.com/ml-explore/mlx-lm/blob/f3e8c601d95c3410152e778ea721e11fef893b53/mlx_lm/utils.py#L605.

Please file an issue with the command an error you are are getting and we can help debug it.

PR #791, opened 2026-01-21
add kimi tool parser

Evanev7:
Adds a tool parser based on the glm47 and vllm tool parser. Some points of note:

- Kimi-K2 has a <|start_tool_calls_section|> in which it makes multiple calls. I don't handle this, and I don't think the rest of the code intends to, but would a follow up PR aiming to address this kind of parallel tool calling be appreciated? I'm not sure how much you want PRs making larger changes to the repo.
- I couldn't actually get the loading logic correct with the full kimi-k2. I tested the parsing code manually with some custom ignores for the calls section, but the loading might need to change.

### DIFF
--- a/mlx_lm/tool_parsers/kimi_k2.py
+++ b/mlx_lm/tool_parsers/kimi_k2.py
@@ -1,4 +1,4 @@
-# kimi_k2_tool_parser.py
+# Copyright © 2026 Apple Inc.
 """
 Modified from:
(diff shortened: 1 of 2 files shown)


awni:
Looks great. Could you please add a test here https://github.com/ml-explore/mlx-lm/blob/main/tests/test_tool_parsing.py ?

awni (approved):
LGTM!

PR #792, opened 2026-01-21
fix: handle GLM 4.7 tool call fallbacks

jalehman:
## Summary
When using GLM 4.7 with OpenCode/Codex, the tool parser crashes if the model outputs tool calls in a format that doesn't match the expected `<arg_key>` regex pattern.

## Problem
```python
func_name = _func_name_regex.search(text).group(1)
# AttributeError: 'NoneType' object has no attribute 'group'
```

The regex `^(.*?)<arg_key>` doesn't match all GLM tool call formats (JSON, plain text, etc.).

## Solution
- Add JSON fallback parsing for common tool call shapes (`name+arguments`, `function+arguments`, nested `tool` objects)
- Add plain-text fallback (`name {json}` or `name key=value`)
- Return safe unknown tool with raw text if nothing matches
- Normalize argument values using tool schema types
- Add tests covering fallback paths

## Testing
- Added tests in `tests/test_tool_parsing.py`
- Verified fix works with OpenCode + GLM 4.7 Flash via mlx-lm server

Closes issue with tool parsing crashes when using coding agents with local GLM models.

### DIFF
--- a/mlx_lm/tool_parsers/glm47.py
+++ b/mlx_lm/tool_parsers/glm47.py
@@ -22,22 +22,21 @@ tool_call_start = "<tool_call>"
 tool_call_end = "</tool_call>"
 
 
-def _is_string_type(
-    tool_name: str,
-    arg_name: str,
-    tools: list[Any] | None,
-) -> bool:
+def _get_string_arg_names(tool_name: str, tools: list[Any] | None) -> set[str]:
     if tools is None:
-        return False
+        return set()
     for tool in tools:
-        func = tool["function"]
-        if func["name"] == tool_name:
-            params = func["parameters"]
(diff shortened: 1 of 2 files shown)


ivanfioravanti:
I tested this and works great! Without it I was not able to use ChromeDeveloper MCP 🤷🏻‍♂️
When will this be reviewed and merged?

vade:
Would be great to get this merged. Im running MLX backend on LMStudio, and with GLM serving for a local OpenClaw (dont judge me its weird and fun) agent, tool usage seems to break as OpenClaw's tools return JSON.

Honestly with LM Studios responses API and fixing tool calling with structured output, this is going to be incredible.

PR #793, opened 2026-01-21
feat: Migrate server to FastAPI

elie00:
This PR migrates the server implementation to FastAPI for better performance and validation.
    
    Changes:
    - Added `mlx_lm/server_fastapi.py`
    - Updated `setup.py` with `fastapi` and `uvicorn`
    - Validated basic endpoints

### DIFF
--- a/mlx_lm/logits_processors.py
+++ b/mlx_lm/logits_processors.py
@@ -1,0 +1,100 @@
+
+import mlx.core as mx
+import numpy as np
+import logging
+
+try:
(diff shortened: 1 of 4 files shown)


angeloskath:
Hi @elie00 , thanks for the PR. I am inclined to close this as it tries to do many things and there isn't a clear reason for either.

1. Should we move to FastAPI? Do you have any benchmark that shows it is better?
2. It adds outlines. I don't think it was our intention to do that...

PR #797, opened 2026-01-21
feat: Evaluation Harness Support

elie00:
This PR adds support for lm-evaluation-harness (MMLU, HellaSwag, etc.).
    
    Changes:
    - Added `mlx_lm/internal_eval.py`: MLXLM wrapper.
    - Added `mlx_lm/evaluate.py`: CLI entry point.

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -1,511 +1,55 @@
-# Copyright © 2024 Apple Inc.
-
-"""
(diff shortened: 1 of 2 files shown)


angeloskath:
I will close this since we already have eval harness support in `evaluate.py`. It also builds on the other PRs although it shouldn't as it doesn't seem to need anything from them.

PR #799, opened 2026-01-22
Fix batch cache filter on zero-length arrays

### LINKED ISSUE #798: BatchKVCache filter crash on zero-length cache when reusing prompt cache (second server_benchmark run)
Steps to reproduce the issue:
```
mlx_lm.server --model mlx-communty/GLM-4.7-Flash-4bit --temp 1 --top-p 0.95 --top-k 50 --min-p 0.01
python benchmarks/server_benchmark.py --conc 4 --total 16
python benchmarks/server_benchmark.py --conc 4 --total 16
``` 

Second run of benchmark triggers ValueError: [take] Cannot do a non-empty take from an array with zero elements from mlx_lm/models/cache.py in BatchKVCache.filter.

ivanfioravanti:
This fixes #798 
Tested with same flow defined in the issue and two tests added.

### DIFF
--- a/tests/test_prompt_cache.py
+++ b/tests/test_prompt_cache.py
@@ -481,6 +481,16 @@ class TestPromptCache(unittest.TestCase):
         self.assertEqual(cache_a.offset.tolist(), [6, 7, 6, 1, 4])
         self.assertEqual(cache_a.left_padding.tolist(), [2, 1, 2, 7, 4])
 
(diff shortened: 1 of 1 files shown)


awni:
Thanks for reporting and the fix! I'm closing in favor of https://github.com/ml-explore/mlx/pull/3046 which will keep things simpler here!

PR #809, opened 2026-01-26
Fix for Exception - MultiLinear.to_quantized() missing 'mode'

inferencers:
Add mode parameter to mixed_quant_predicate_builder as MLX now requires mode to be specified for nn.quantize class_predicate

How to reproduce: Try to convert GLM-4.7-Flash with a mixed quant predicate
MLX: v0.30.3

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -18,8 +18,9 @@ from .utils import (
 
 
 def mixed_quant_predicate_builder(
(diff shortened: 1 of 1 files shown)


awni:
Thanks for the fix. Since the mixed builder really only makes sense for `"affine"` I would change it to hardcode that mode in the function itself. And additionally if a mode other than affine is requested with a mixed quant, I would either print warning or raise an exception.

awni (approved):
LGTM, thanks!

PR #813, opened 2026-01-27
Add Kimi-K2.5

kernelpool:
WIP

Model: https://huggingface.co/moonshotai/Kimi-K2.5

Notes:
- The custom tokenizer (tokenization_kimi.py) needs an updated path for `bytes_to_unicode()` to work with Transformers 5: `from transformers.convert_slow_tokenizer import bytes_to_unicode`
- The custom tokenizer also requires tiktoken: `pip install tiktoken`

### DIFF
--- a/mlx_lm/models/kimi_k25.py
+++ b/mlx_lm/models/kimi_k25.py
@@ -1,60 +1,30 @@
 # Copyright © 2026 Apple Inc.
 
 from dataclasses import dataclass
-from typing import Any, Dict, Optional, Union
+from typing import Any, Optional, Union
 
 import mlx.core as mx
 import mlx.nn as nn
-from mlx.nn.layers.distributed import shard_inplace, shard_linear
+from mlx.utils import tree_flatten, tree_unflatten
 
 from .base import BaseModelArgs
-from .deepseek_v3 import DeepseekV3MLP, DeepseekV3Model
-
-
-@dataclass
-class TextArgs(BaseModelArgs):
-    vocab_size: int = 163840
-    hidden_size: int = 7168
-    intermediate_size: int = 18432
-    moe_intermediate_size: int = 2048
-    num_hidden_layers: int = 61
-    num_attention_heads: int = 64
-    num_key_value_heads: int = 64
-    n_shared_experts: Optional[int] = 1
-    n_routed_experts: Optional[int] = 384
-    routed_scaling_factor: float = 2.827
-    kv_lora_rank: int = 512
-    q_lora_rank: int = 1536
-    qk_rope_head_dim: int = 64
-    v_head_dim: int = 128
-    qk_nope_head_dim: int = 128
-    topk_method: str = "noaux_tc"
-    scoring_func: str = "sigmoid"
-    norm_topk_prob: bool = True
-    n_group: int = 8
-    topk_group: int = 4
-    num_experts_per_tok: int = 8
-    moe_layer_freq: int = 1
-    first_k_dense_replace: int = 1
-    max_position_embeddings: int = 262144
-    rms_norm_eps: float = 1e-6
-    rope_theta: float = 10000.0
-    rope_scaling: Dict = None
-    attention_bias: bool = False
+from .deepseek_v3 import DeepseekV3Model
+from .deepseek_v3 import Model as DeepseekV3LM
+from .deepseek_v3 import ModelArgs as TextConfig
 
 
 @dataclass
 class ModelArgs(BaseModelArgs):
-    text_config: Union[TextArgs, dict]
+    text_config: Union[TextConfig, dict]
     model_type: str = "kimi_k25"
 
     def __post_init__(self):
         if isinstance(self.text_config, dict):
-            self.text_config = TextArgs.from_dict(self.text_config)
+            self.text_config = TextConfig.from_dict(self.text_config)
 
 
 class LanguageModel(nn.Module):
-    def __init__(self, config: TextArgs):
+    def __init__(self, config: TextConfig):
         super().__init__()
         self.args = config
         self.model = DeepseekV3Model(config)
@@ -76,15 +46,6 @@ class Model(nn.Module):
         self.model_type = config.model_type
         self.language_model = LanguageModel(config.text_config)
 
-    @staticmethod
-    def sanitize_config(config):
-        if "quantization_config" not in config:
-            if "quantization_config" in config.get("text_config", {}):
-                config["quantization_config"] = config["text_config"][
-                    "quantization_config"
-                ]
-        return config
-
     def __call__(
         self,
         inputs: mx.array,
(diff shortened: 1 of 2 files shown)


mlx_lm/utils.py
```diff
@@ -551,6 +554,16 @@ def sharded_load(
...
+
+    # Evaluate parameters layer-by-layer for large models to avoid hangs
+    if hasattr(model, "model") and hasattr(model.model, "embed_tokens"):
+        mx.eval(model.model.embed_tokens.parameters())
+    if hasattr(model, "layers"):
+        for layer in model.layers:
+            mx.eval(layer.parameters())
+    if hasattr(model, "model") and hasattr(model.model, "norm"):
+        mx.eval(model.model.norm.parameters())
+    # Final eval for any remaining parameters (lm_head, etc.)
```
awni: I don't think we should add that. For conversion that don't require loading the full model in memory they should pass `lazy=True` to the load function (which is what `mlx_lm.convert` already does).

I'm curious what did command did this fix for you?

mlx_lm/models/kimi_k25.py
```diff
@@ -0,0 +1,204 @@
...
+        self.language_model = LanguageModel(config.text_config)
+
+    @staticmethod
+    def sanitize_config(config):
+        if "quantization_config" not in config:
+            if "quantization_config" in config.get("text_config", {}):
+                config["quantization_config"] = config["text_config"][
+                    "quantization_config"
+                ]
+        return config
```
awni: Let's move this logic to the higher level `load`. Having a `text_config` for vlms is common and I it seems reasonable to handle quantization in text configs.

mlx_lm/models/kimi_k25.py
```diff
@@ -0,0 +1,204 @@
...
+            for m in ["gate_proj", "down_proj", "up_proj"]:
+                for k in ["weight", "scales", "biases"]:
+                    if f"{prefix}.mlp.experts.0.{m}.{k}" in weights:
+                        to_join = [
+                            weights.pop(f"{prefix}.mlp.experts.{e}.{m}.{k}")
+                            for e in range(self.args.text_config.n_routed_experts)
+                        ]
+                        weights[f"{prefix}.mlp.switch_mlp.{m}.{k}"] = mx.stack(to_join)
+
+        return weights
```
awni: I think you can do this instead by:
1. Unflattening the dict of weights 
2. popping unused VLM weights
3. Calling `DeepSeekV3.sanitize`
4. Flatten weights

mlx_lm/models/kimi_k25.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def shard(self, group: Optional[mx.distributed.Group] = None):
+        group = group or mx.distributed.init()
+        N = group.size()
+        for layer in self.language_model.model.layers:
```
awni: Same here with a slight modification should be able to use `DeepSeekV3.shard`.

mlx_lm/models/kimi_k25.py
```diff
@@ -0,0 +1,204 @@
...
+    n_group: int = 8
+    topk_group: int = 4
+    num_experts_per_tok: int = 8
+    moe_layer_freq: int = 1
+    first_k_dense_replace: int = 1
+    max_position_embeddings: int = 262144
+    rms_norm_eps: float = 1e-6
+    rope_theta: float = 10000.0
+    rope_scaling: Dict = None
+    attention_bias: bool = False
```
awni: I don't think we need this. Use the `DeepSeekV3` model args.

mlx_lm/utils.py
kernelpool: I had to add this for `mlx.launch --verbose --backend jaccl --hostfile hosts-jaccl.json --env MLX_METAL_FAST_SYNCH=1 -- /Users/optimus/repo/mlx-lm/mlx_lm/examples/sharded_generate.py --model /Users/optimus/models/catalyst/Kimi-K2.5-4bit --prompt "Who is Albert Einstein?" -m 1024` to work. If I remove it, both machines get stuck at 100% GPU (ive also seen both machines load the models, but then no generation happens). Both machines are on macOS 26.2 and using mlx 0.30.3.

awni: That's very odd that this fixes it. This breaks lazy loading and will cause the `sharded_load` to load the full model for each process which seems like it should be worse.

But either way seems like there is a bug. Would you mind removing this and we will investigate a better fix?

Also how much RAM did each of your machines have?

kernelpool: Sure, no worries. Each has 512gb.

Evanev7:
Just adding a followup that the same issue has been found in EXO and that loading layer-by-layer also helped there for some tensor sharded models. 

(https://github.com/exo-explore/exo/pull/1291/changes#diff-1101baba71502dc7aaa2d4fc3bd721fdf9edc584cd3afc6db08b1271e927a085R650)

awni:
My first try running the model in it's native precision did not hang:
```
mlx.launch --verbose --backend jaccl --hostfile m3-ultra-jaccl.json --env MLX_METAL_FAST_SYNCH=1 -- /Users/awni/mlx-lm/mlx_lm/examples/sharded_generate.py --model moonshotai/Kimi-K2.5 --prompt "Who is Albert Einstein?" -m 1024
```

I'm running it on 2 M3 Ultra 512s. I'll try the MLX 4-bit  next.

awni:
@kernelpool or @Evanev7 do you have the following sysctl `iogpu.wired_limit_mb` set to anything?

Either way could you share the output of:
```
python -c "import mlx.core as mx; print(mx.metal.device_info())"
```

kernelpool:
1:
```
sysctl iogpu.wired_limit_mb
iogpu.wired_limit_mb: 516096

python -c "import mlx.core as mx; print(mx.metal.device_info())"
{'resource_limit': 499000, 'max_buffer_length': 373662154752, 'architecture': 'applegpu_g15d', 'memory_size': 549755813888, 'max_recommended_working_set_size': 541165879296, 'device_name': 'Apple M3 Ultra'}
```

2:
```
sysctl iogpu.wired_limit_mb
iogpu.wired_limit_mb: 507904

python -c "import mlx.core as mx; print(mx.metal.device_info())"
{'resource_limit': 499000, 'max_buffer_length': 373662154752, 'architecture': 'applegpu_g15d', 'memory_size': 549755813888, 'max_recommended_working_set_size': 532575944704, 'device_name': 'Apple M3 Ultra'}
```

awni:
Thanks! Nothing unusual there.. 

So far I haven't been able to repro this with MLX on two M3 Ultras with either the native model or 4-bit. They are on macOS 26.3 .. not sure if that makes a difference.

How consistent is the hang? Is it every time or just 1/N times? Does it go away if you reboot the machine?

kernelpool:
It's consistent across reboots. Happens every time. I added some debug prints to `shareded_load` and it never gets to "Synchronizing processes". Note that `--pipeline` works just fine, I'm only seeing this for tensor parallelism.

```
  print("[INFO] Loading model...")
  model, _ = load_model(model_path, lazy=True, strict=False)
  if tensor_group is not None:
      print("[INFO] Sharding model for tensor parallelism...")
      model.shard(tensor_group)
  if pipeline_group is not None:
      print("[INFO] Sharding model for pipeline parallelism...")
      model.model.pipeline(pipeline_group)
  print("[INFO] Evaluating model parameters...")
  mx.eval(model.parameters())

  # Synchronize processes to avoid timeout
  print("[INFO] Synchronizing processes...")
  mx.eval(mx.distributed.all_sum(mx.array(1.0), stream=mx.cpu))
  print("[INFO] Synchronization complete.")
  if return_config:
      return model, tokenizer, config
  else:
      return model, tokenizer
```

Output:

```
mlx.launch --verbose --backend jaccl --hostfile hosts-jaccl.json --env MLX_METAL_FAST_SYNCH=1 -- /Users/optimus/repo/mlx-lm/mlx_lm/examples/sharded_generate.py --model /Users/optimus/models/catalyst/Kimi-K2.5-4bit --prompt "Who is Albert Einstein?" -m 1024
[INFO] Running /Users/optimus/repo/mlx-lm/.venv/bin/python /Users/optimus/repo/mlx-lm/mlx_lm/examples/sharded_generate.py --model /Users/optimus/models/catalyst/Kimi-K2.5-4bit --prompt 'Who is Albert Einstein?' -m 1024 
[jaccl] Connection attempt 0 waiting 1000 ms
[09:43:39.785] [INFO] Loading weights...
[09:43:39.811] [INFO] Loading weights...
[09:43:40.037] [INFO] Loading model...
[09:43:40.072] [INFO] Loading model...
[09:43:40.090] [INFO] Loading weights...
[09:43:40.099] [INFO] Sharding model for tensor parallelism...
[09:43:40.106] [INFO] Evaluating model parameters...
[09:43:40.142] [INFO] Loading weights...
[09:43:40.151] [INFO] Sharding model for tensor parallelism...
[09:43:40.161] [INFO] Evaluating model parameters...
```

After the last log output, i can see the memory of both machines gradually increasing to around 67% while GPU is 100%. GPU then drops back to 0 while memory stays at 67% and nothing happens.

angeloskath:
Hi @kernelpool, sorry that you are experiencing this... it is pretty hard to debug.

Can you try disabling the wired collector using `sudo sysctl iogpu.disable_wired_collector=1`?

kernelpool:
> Hi @kernelpool, sorry that you are experiencing this... it is pretty hard to debug.
> 
> 
> 
> Can you try disabling the wired collector using `sudo sysctl iogpu.disable_wired_collector=1`? 

I tried setting this on both machines but it made no difference unfortunately.

kernelpool:
FWIW, I was doing the testing on a 4bit quant (so it had some additional layer quantization over native int4 which keeps some layers in bf16), using `-q --q-bits 4`. I tried without `-q` and now the model loads just fine.. That being said, I was able to minimize my previous workaround to the following to make the 4-bit quantized version work. Not sure why this "pre-warming" is necessary, but it seems to do the trick. Anyway, I'll just use the native version for now, and maybe create an issue if I run into the problem again.

```python
if hasattr(model, "model") and hasattr(model.model, "embed_tokens"):                        
    embed_params = model.model.embed_tokens.parameters()                                    
    mx.eval(list(embed_params.values()))
```

awni:
I did indeed try using the full q4 with the same conversion flags. It did not reproduce for me in that case. But there could be other differences in my environment that account for that as well.

Will try messing around with other settings.  But let's keep an eye on this.. once we can reproduce it consistently, it should be much easier to debug.

Evanev7:
> do you have the following sysctl `iogpu.wired_limit_mb` set to anything?

{'device_name': 'Apple M3 Ultra', 'max_recommended_working_set_size': 498216206336, 'memory_size': 549755813888, 'architecture': 'applegpu_g15d', 'max_buffer_length': 373662154752, 'resource_limit': 499000}

sysctl iogpu.wired_limit_mb
iogpu.wired_limit_mb: 0

PR #820, opened 2026-01-29
Fix NemotronH config compatibility with HuggingFace format

luqmannurhakimbazman:
HuggingFace's NemotronH config uses separate `time_step_min` and `time_step_max` fields, but mlx-lm expected a `time_step_limit` tuple. This caused loading failures since `time_step_limit` was required but never populated from the config.

- Make `time_step_limit` optional with default None
- Add `time_step_min` and `time_step_max` optional fields
- Add `__post_init__` to construct tuple from separate fields


awni (approved):
LGTM, thanks!

Issue #822, opened 2026-01-29
[Bug] ModuleNotFoundError: No module named 'mlx_lm.cli' after latest merge


Vlor999:
### Describe the bug
After pulling the latest changes from `main` which introduced the unified CLI entry point, running the command `mlx_lm` results in a `ModuleNotFoundError: No module named 'mlx_lm.cli'`. 

This happens even after performing a clean editable installation as recommended in the contribution guide.

### To Reproduce
1. Pull the latest changes from the repository.
2. Run `pip install -e .`
3. Run `mlx_lm --help`

**Error output:**
```text
Traceback (most recent call last):
  File ".../.venv/bin/mlx_lm", line 4, in <module>
    from mlx_lm.cli import main
ModuleNotFoundError: No module named 'mlx_lm.cli'
```

I would be interested to resolve this by adding the `cli.py` file into the repo

awni:
Sorry I missed a file in the PR. It should be fixed after #823. Thanks for flaggin.

PR #823, opened 2026-01-29
Actually add cli

awni:
Closes https://github.com/ml-explore/mlx-lm/issues/822

This was supposed to be part of https://github.com/ml-explore/mlx-lm/pull/816


angeloskath (approved):
Nice!

PR #824, opened 2026-01-29
Add ModernBERT

cTatu:


awni:
This is cool! I think it would be better fit for an embedding package like https://github.com/taylorai/mlx_embedding_models or https://github.com/Blaizzy/mlx-embeddings

PR #826, opened 2026-01-30
fix cli

solarpunkin:
Running `mlx_lm.generate --help` works, but by the same analogy `mlx_lm.quant.dwq --help` doesn't resolve. 

now `mlx_lm dwq --help` or `mlx_lm quant.dwq --help` or `mlx_lm.quant.dwq --help` also work.


awni:
Thanks .. but in the interest of keeping it simple.. let's use the current approach e.g.

```
mlx_lm.dwq --help 
```

That should give you help for the dwq quant.

solarpunkin:
Fair enough. However, the current help output is misleading. When a user runs `mlx_lm --help`, they see `quant.dwq`. Following the pattern of `mlx_lm.generate`,  they naturally try `mlx_lm.quant.dwq --help`, which fails. 

```bash
$ mlx_lm --help
The supported subcommand are ('benchmark', 'cache_prompt', 'chat', 'convert', 'evaluate', 'fuse', 'generate', 
'lora', 'manage', 'perplexity', 'quant.awq', 'quant.dwq', 'quant.dynamic_quant', 'quant.gptq', 'server', 'upload')

For help on an individual subcommand, pass --help to the subommand. For example: mlx_lm.generate --help
```
You may want to fix the typo

awni:
Oh I got your point. Do you want to update this PR / send a new one to fix that instead?

PR #829, opened 2026-01-30
server: support chat_template_kwargs and top_logprobs

percontation:
* Adds support for clients sending "chat_template_kwargs", matching other open source LLM servers. This is gated behind `--trust-client-kwargs` because transformers does not provide any safe way to do this.

* changes the server's logprobs response to better match the OpenAI chat api & other open source servers.

It's not perfect, there are some missing fields, but it's enough to allow mlx_lm.server to be a compatible backend for a pydantic-ai based app I'm working on.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -728,7 +728,11 @@ class ResponseGenerator:
                     and current_model == args.model
                     and is_batchable
                 ):
-                    prompt = self._tokenize(current_tokenizer, request, args)
+                    try:
+                        prompt = self._tokenize(current_tokenizer, request, args)
+                    except Exception as e:
+                        rqueue.put(e)
+                        continue
 
                     ctx = GenerationContext(
                         has_tool_calling=tokenizer.has_tool_calling,


mlx_lm/server.py
```diff
@@ -963,8 +971,10 @@ def do_POST(self):
         self.xtc_probability = self.body.get("xtc_probability", 0.0)
         self.xtc_threshold = self.body.get("xtc_threshold", 0.0)
         self.logit_bias = self.body.get("logit_bias", None)
-        self.logprobs = self.body.get("logprobs", -1)
+        self.logprobs = self.body.get("logprobs", False)
```
awni, line 1100: Is that still used at all or should it be removed?

awni: I guess we shouldn't remove it because it's part of the spec. But it looks like it's not checked properly?

mlx_lm/server.py
```diff
@@ -801,15 +807,17 @@ def progress(tokens_processed, tokens_total):
...
-                    top_indices = sorted_indices[: args.logprobs]
+                    top_indices = sorted_indices[: args.top_logprobs]
                     top_logprobs = gen.logprobs[top_indices]
-                    top_token_info = zip(top_indices.tolist(), top_logprobs.tolist())
-                    top_tokens = tuple(top_token_info)
-
+                    txts = tokenizer.convert_ids_to_tokens(top_indices)
+                    top_tokens = tuple(
+                        {"id": i, "token": s, "logprob": g}
+                        for i, s, g in zip(top_indices.tolist(), txts, top_logprobs.tolist())
```
awni: Could you refactor that into a little helper to avoid duplicating it above? It's getting a bit more involved.

mlx_lm/server.py
awni: Nvm I think I underand now. logprobs is a bool not an int. And top_logprobs is the int.

mlx_lm/server.py
```diff
@@ -1295,10 +1312,11 @@ def parse_tools(tool_calls):
 
             # Save the token and its logprob
             tokens.append(gen.token)
-            token_logprobs.append(gen.logprob)
+            if args.logprobs:
+                token_logprobs.append(gen.logprob)
 
             # If requested save the k top logprobs
-            if gen.top_tokens is not None:
+            if args.top_logprobs and gen.top_tokens is not None:
```
awni: if `args.top_logprobs` is set would `gen.top_tokens` ever be `None`?

mlx_lm/server.py
```diff
@@ -801,15 +807,17 @@ def progress(tokens_processed, tokens_total):
...
             ):
                 top_tokens = None
-                if args.logprobs > 0:
+                if args.top_logprobs > 0:
                     sorted_indices = mx.argpartition(
-                        -gen.logprobs, kth=args.logprobs - 1
+                        -gen.logprobs, kth=args.top_logprobs - 1
                     )
-                    top_indices = sorted_indices[: args.logprobs]
+                    top_indices = sorted_indices[: args.top_logprobs]
```
awni: Convert that to a list before passing to `top_indices` to avoid too many graph evaluations:

```suggestion
                    top_indices = sorted_indices[: args.top_logprobs].tolist()
```

mlx_lm/server.py
awni:
```suggestion
                        for i, s, g in zip(top_indices, txts, top_logprobs.tolist())
```

mlx_lm/server.py
percontation: Correct, it shouldn't be. As part of the other change, we'll now get empty-tuples instead of None anyways, so I changed this to just `if args.top_logprobs:`

mlx_lm/server.py
percontation: Yep! A bit redundant but 🤷‍♂️

percontation:
I believe I've implemented all your comments. Also I fixed the type annotations for what I was changing 😅.

mlx_lm/server.py
```diff
@@ -1791,6 +1804,11 @@ def main():
         help="""A JSON formatted string of arguments for the tokenizer's apply_chat_template, e.g. '{"enable_thinking":false}'""",
         default="{}",
     )
+    parser.add_argument(
+        "--trust-client-kwargs",
+        action="store_true",
+        help="Enable chat_template_kwargs, allowing clients to send arguments to apply_chat_template.",
+    )
```
awni: Is this necessary? Why not just always allow client kwargs?

percontation: You can pass weird stuff like `"chat_template_kwargs": {"tokenize": false}` and it breaks things because AFAICT the only way to ask transformers to apply these is as `**kwargs` to `apply_chat_template`. There's a lot of potentially unsafe kwargs that are interpreted by Python, and they vary by underlying model, so I just didn't implement any filtering. This seems like a security concern that warranted this feature being off by default.

percontation: I guess it should be safe enough to whitelist the most popular well-known jinja template args like "enable_thinking" and "reasoning_effort". If those were allowed even without `--trust-client-kwargs`, like 99% of people wouldn't need to set that flag.

Oh, or I guess we could instead have a user-configurable whitelist passed via an argument, where the default value is that same well-known key list.

percontation:
Fixed the black linter errors & a bug that crept into the rebase.

The additional new commit implements the default-chat_template_kwargs-whitelist idea. It's a little clunky, but, most people won't ever need to set `--client-chat-template-args`, and the ones that do will probably just be setting `--client-chat-template-args '*'`. Thoughts?

mlx_lm/server.py
percontation: I liked that ^ whitelist idea, I have added it in a new commit here for consideration.

The default whitelist came from looking at the user-configurable variables that `llama.cpp/models/templates/*.jinja` templates use.

percontation:
I'd suggest keeping the

```
                    try:
                        prompt = self._tokenize(current_tokenizer, request, args)
                    except Exception as e:
                        rqueue.put(e)
                        continue
```

That way, this codepath has the same behavior as `_serve_single` if a weird arg gets set (e.e. `"add_generation_prompt": true`).

(Or, I guess the other way to make them match would be to remove `_serve_single`'s `prompt = self._tokenize(tokenizer, request, args)` from its `try: ... except Exception as e: rqueue.put(e)` wrapper. 🤷‍♂️ )

PR #832, opened 2026-02-01
Add OpenAI-compatible Responses API and vision support

Szowesgad:
## What\n- Add OpenAI-compatible /responses API with storage, retrieval, deletion, and cancel handling\n- Add vision responses support via mlx-vlm fallback when torchvision is unavailable\n- Add Harmony parsing and improve streaming detokenizer behavior\n- Add docs, examples, and tests (including heavy vision tests)\n\n## Why\n- Enable OpenAI Responses API semantics on the mlx-lm server for parity with existing clients and tooling\n\n## How\n- Introduce responses_server normalization + vision helpers\n- Wire responses routing into server with SSE streaming events\n- Add parser and tokenizer utilities to handle Harmony outputs and spacing edge cases\n\n## Tests\n- Not run (local env)


angeloskath:
I think this is a bit out of scope in its goal (especially the vision part) and the harmony integration should be tied to the gpt-oss model (if it should be integrated at all).

For the above I am going to close the PR.

PR #834, opened 2026-02-02
Allow default creation of BatchRotatingKVCache instead of BatchKVCache in batch mode

christian-lms:
Currently, when empty caches are passed into `BatchGenerator.insert()`, `_make_cache` creates new `BatchKVCache`s by default. This makes it difficult to impose a maximum size on the cache (done with a `BatchRotatingKVCache`).

This PR allows an optional `max_kv_size` parameter to be set in the `BatchGenerator` constructor so that `_make_cache` will create a `BatchRotatingKVCache` with `max_size=self.max_kv_size` in the event of an empty cache, while defaulting to `BatchKVCache` otherwise. This does not change the default behavior of `BatchGenerator`.

Another fix is to pass the input caches to `_make_cache` instead of ignoring them, but this runs into problems when the caches are heterogeneous in type or (in the `RotatingKVCache` case) have `keep > 0`.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -883,7 +883,7 @@ class Batch:
         return [c.extract(idx) for c in self.cache]
 
 
-def _make_cache(model, left_padding):
+def _make_cache(model, left_padding, max_kv_size):
     """
     Convert a list of regular caches into their corresponding
     batch-aware caches.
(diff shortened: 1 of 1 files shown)


awni:
> but this runs into problems when the caches are heterogeneous in type or (in the RotatingKVCache case) have keep > 0.

That's a very good point.. I think you found the right solution. For now `max_kv_size` should be a global property on the batch generator since we don't support heterogenous values for it.

Issue #831, opened 2026-02-01
Resource limit (499000) exceeded with mlx_lm.server distributed.


ivanfioravanti:
I'm playing with mlx_lm.server distributed added in #741
I use this command:
```
mlx.launch \
      --backend jaccl \
      --env MLX_METAL_FAST_SYNCH=1 \
      --hostfile /Users/ifioravanti/github/mlx-lm/m3-ultra-jaccl.json  \
      /Users/ifioravanti/github/mlx-lm/.venv/bin/mlx_lm.server \
      --model mlx-community/GLM-4.7-Flash-8bit
``` 

Then I use OpenCode to do some local coding. Memory is under control, sometimes same commands work other times I get following exception.

```
Exception in thread Thread-2 (_generate):
Traceback (most recent call last):
  File "/Users/ifioravanti/.local/share/uv/python/cpython-3.13.11-macos-aarch64-none/lib/python3.13/threading.py", line 1044, in _bootstrap_inner
    self.run()
    ~~~~~~~~^^
  File "/Users/ifioravanti/.local/share/uv/python/cpython-3.13.11-macos-aarch64-none/lib/python3.13/threading.py", line 995, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ifioravanti/github/mlx-lm/mlx_lm/server.py", line 803, in _generate
    responses = batch_generator.next()
  File "/Users/ifioravanti/github/mlx-lm/mlx_lm/generate.py", line 1260, in next
    return self._next()
           ~~~~~~~~~~^^
  File "/Users/ifioravanti/github/mlx-lm/mlx_lm/generate.py", line 1217, in _next
    mx.async_eval(batch.y, batch.logprobs)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
```

angeloskath:
Hmm interesting. I will try to repro but it might be difficult. This points towards leaking some buffers or sth along those lines. Thanks for the report, will keep you posted (or maybe ask you more questions :-) ).

listenfree:
[...]

PR #839, opened 2026-02-03
DSV3 MLA

awni:
With the latest MLX there is a nice speedup for long context (the longer the context the more the speedup):
 
Context | prompt tok/s pre | prompt tok/s post | gen tok/s pre | gen tok/s post | memory GB pre | memory GB post
----- | ---- | ----| ----- | ---- | ---- | -----
4096 | 199.149 | 213.337 |  13.592 | 20.647 | 401.668 | 383.213
8192 | 176.519 | 187.632 | 9.750 | 18.852 | 424.215 | 388.343
16384 | 139.040 | 147.876 | 6.520 | 16.772 | 469.906 | 398.605

### DIFF
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -11,6 +11,7 @@ from mlx.nn.layers.distributed import shard_inplace, shard_linear, sum_gradients
 from .activations import swiglu
 from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
 from .cache import CacheList, KVCache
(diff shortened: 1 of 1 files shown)


awni:
Unfortunately the prefill time is slower.. but it's a net win on time. And the memory saving is also quite worth-while.

CC @kernelpool

awni:
Also closes #831

awni:
I made some changes so now it's actually faster for prefill. Note it could still be faster if we had a fused SDPA for 256 head dimension. Something to consider in the future.

PR #836, opened 2026-02-02
Add Step 3.5 Flash

kernelpool:
Model: https://huggingface.co/stepfun-ai/Step-3.5-Flash

Note: I've used regular KVCache with masking for the sliding window attention to support trimming

Tool calling uses same format as qwen3 coder.

Issue: #835

```
mlx_lm.generate --model /Volumes/WD_EXTRA/models/catalyst/Step-3.5-Flash-4bit --prompt "Who is Albert Einstein?" --trust-remote-code -m 1024
The tokenizer you are loading from '/Volumes/WD_EXTRA/models/catalyst/Step-3.5-Flash-4bit' with an incorrect regex pattern: https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503/discussions/84#69121093e8b480e709447d5e. This will lead to incorrect tokenization. You should set the `fix_mistral_regex=True` flag when loading this tokenizer to fix this issue.
==========
Hmm, the user is asking about Albert Einstein. This is a straightforward factual question about a well-known historical figure. The user likely wants a concise yet comprehensive overview of who Einstein was, covering his major contributions and legacy. 

I should start with his most iconic achievement—the theory of relativity—since that’s what he’s most famous for. Then I can mention his Nobel Prize for the photoelectric effect, which is less known but equally important. It’s also worth highlighting his broader impact beyond physics, like his advocacy for peace and civil rights, since that paints a fuller picture of his character. 

The user might be a student or someone casually curious, so I’ll keep it clear but avoid oversimplifying. I’ll structure it chronologically: early life, key scientific work, later activism, and legacy. No need for excessive detail—just the essentials that define his significance. 

Oh, and I should clarify that “E=mc²” is part of relativity, not a separate theory, since people often conflate them. Also, mentioning his emigration to the U.S. and his role in the atomic bomb letter adds historical context. The tone should be respectful but not overly academic—accessible to anyone.
</think>
Albert Einstein (1879–1955) was a German-born theoretical physicist who revolutionized modern physics and became one of the most famous scientists in history. Here’s a concise overview of his life and legacy:

---

### **Key Scientific Contributions**
1. **Theory of Relativity**  
   - **Special Relativity (1905):** Introduced the idea that the laws of physics are the same for all non-accelerating observers, and that the speed of light is constant. This led to the famous equation **E=mc²**, showing the equivalence of mass and energy.  
[...]

PR #840, opened 2026-02-03
Fix Step 3.5 Flash model conversion

kernelpool:
Fix to avoid applying the RMSNorm delta twice, at conversion and subsequently at load.
This simply reverts back to the original approach from [b8c4549](https://github.com/ml-explore/mlx-lm/pull/836/commits/b8c4549cee8f97d40f37a50218ff227bd496e5f4). Maybe theres a better way?

More info: https://github.com/ml-explore/mlx-lm/pull/836#issuecomment-3838891260

### DIFF
--- a/mlx_lm/models/step3p5.py
+++ b/mlx_lm/models/step3p5.py
@@ -406,6 +406,10 @@ class Model(nn.Module):
             (".share_expert.", ".mlp.share_expert."),
         ]
 
+        is_vanilla = any(
+            src in k and dst not in k for k in weights for src, dst in remappings
+        )
+
         new_weights = {}
         for k, v in weights.items():
             if ".mtp" in k:
@@ -421,9 +425,8 @@ class Model(nn.Module):
                     k = k.replace(src, dst)
                     break
 
-            if k.endswith(".weight") and "norm" in k:
-                if mx.mean(v).item() < 0.5:
-                    v = v + 1
+            if is_vanilla and k.endswith(".weight") and "norm" in k:
+                v = v + 1
 
             new_weights[k] = v
 


awni:
Yes you can just check if any of the original (un-mapped) keys are still in the weight keys. If they are it means it hasn't been converted yet to MLX format so it's safe to apply the +1.0

ghost:
Hi Guys,just some feedback. This time I run the 8bit uploaded by myself and also the 4bit model uploaded by kernelpool under many different chat situation rather than the given test command.

It looks like they all have repetition problems. They might (almost 100% when the question is long) repeat certain words forever, the words is context related. Some times a whole short sentence is repeated.

It's not production ready for now. Don't know why, I'm just a test user, sorry.

kernelpool:
@awni What do you think about the < 0.5 check? Otherwise we need to re-upload the existing models.

kernelpool:
> Hi Guys,just some feedback. This time I run the 8bit uploaded by myself and also the 4bit model uploaded by kernelpool under many different chat situation rather than the given test command.
> 
> 
> 
> It looks like they all have repetition problems. They might (almost 100% when the question is long) repeat certain words forever, the words is context related. Some times a whole short sentence is repeated.
> 
> 
> 
> It's not production ready for now. Don't know why, I'm just a test user, sorry.
> 
> 
> 
> 

What model parameters (temperature, etc) are you using?

awni:
> @awni What do you think about the < 0.5 check? Otherwise we need to re-upload the existing models.

I don't think we should do it that way. It's somewhat brittle to the mean of the weights and also breaks lazy loading to some extent.

I think we should just check the presence of a pattern in the weight keys to determine if it's already been converted. And I will re-uploading the models, that's not so difficult. (But others will have to reconvert or re-download).

ghost:
>What model parameters (temperature, etc) are you using?

I tried different temperatures, like 0.6, 1, they all behave the same way. top-p 0.95/1 also.

awni:
I will re-upload the models as soon as this lands.

ghost:
Yup I confirm the old models are not working for the new commit.(Output nonsense again.) reupload required.

PR #841, opened 2026-02-03
feat: enhance chat CLI with readline history, line editing, and distributed support

### LINKED ISSUE #818: [Improvement] Enhance `mlx_lm.chat` CLI experience (history, line editing, and formatting)
## Description: 
Currently, `mlx_lm.chat` provides a basic interactive loop, but lacks standard terminal features like:
Line editing: Arrow keys to move the cursor.
History: Up/Down arrows to cycle through previous prompts.
Rich formatting: Rendering model outputs in Markdown for better readability (especially for code blocks).

## Proposed Solution:
Integrate `readline` for persistent history and better line editing.
(Optional) Use `rich` for a cleaner UI (Markdown rendering, streaming text styling).

This would make `mlx_lm.chat` feel more like a polished tool rather than a debug script.

Vlor999:
## Description
This PR improves the `mlx_lm.chat` interactive interface by adding standard terminal features and ensuring compatibility across different execution environments.

## Changes
* **Persistent History**: Integrated readline to support command history (Up/Down arrows) saved to `~/.mlx_lm_chat_history`.
* **Line Editing**: Enabled full cursor movement (Left/Right arrows) and in-place text editing for prompts.
* **Distributed Support**: Implemented broadcast_string logic to ensure the chat loop remains synchronized across all ranks in multi-GPU/distributed mode.
* **Graceful Fallbacks**: Added safety checks and EOFError handling (Ctrl+D to exit) for a smoother terminal experience.
* **User Feedback**: Added an explicit warning message when model output is truncated due to `max_tokens` limits.

## Motivation
The current chat interface is a basic loop that lacks standard CLI ergonomics. Users cannot easily correct typos or recall previous instructions, which hinders the local testing workflow. These changes bring mlx-lm closer to the UX of modern LLM toolkits.

## Future Work / UI Enhancements
I have also prototyped an enhanced UI version using the rich library (see screenshot below) which supports Markdown rendering and syntax highlighting for code blocks. I kept this PR minimal to avoid adding new dependencies, but I am open to integrating a "soft dependency" version if the maintainers are interested in a more polished visual output.

Closes #818 

Capture d’écran 2026-02-03 à 13 34 08

### DIFF
--- a/mlx_lm/chat.py
+++ b/mlx_lm/chat.py
@@ -1,8 +1,15 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import os
+import readline
 
 import mlx.core as mx
+from rich.console import Console
+from rich.live import Live
+from rich.markdown import Markdown
+from rich.panel import Panel
+from rich.text import Text
 
 from .generate import stream_generate
 from .models.cache import make_prompt_cache
@@ -15,7 +22,33 @@ DEFAULT_XTC_PROBABILITY = 0.0
 DEFAULT_XTC_THRESHOLD = 0.0
 DEFAULT_SEED = 0
 DEFAULT_MAX_TOKENS = 256
+DEFAULT_RENDER_WINDOW_SIZE = 20
+DEFAULT_REFRESH_RATE = 10
 DEFAULT_MODEL = "mlx-community/Llama-3.2-3B-Instruct-4bit"
+DEFAULT_SYSTEM_PROMPT = (
+    "You are a helpful assistant. Your responses are rendered in a terminal with "
+    "Markdown support. Feel free to use Markdown formatting when appropriate: "
+    "**bold**, *italic*, `inline code`, code blocks with syntax highlighting "
+    "(```language), bullet lists, numbered lists, and headers."
+)
+
+
+def broadcast(s: str, group: mx.distributed.Group, src: int = 0) -> str:
+    """
+    Broadcast a string from the source rank to all other ranks in the group.
+    """
+    if group.size() == 1:
+        return s
+    if group.rank() == src:
+        data = mx.array(s.encode("utf-8"))
+        mx.eval(mx.distributed.all_sum(data.size, group=group))
+        mx.eval(mx.distributed.all_sum(data, group=group))
+    else:
+        size = mx.distributed.all_sum(0, group=group).item()
+        data = mx.distributed.all_sum(mx.zeros(size, dtype=mx.uint8), group=group)
+        mx.eval(data)
+        s = bytes(data).decode("utf-8")
+    return s
 
 
 def setup_arg_parser():
@@ -52,8 +85,8 @@ def setup_arg_parser():
     parser.add_argument(
         "--xtc-threshold",
         type=float,
-        default=0.0,
-        help="Thresold the probs of each next token candidate to be sampled by XTC",
+        default=DEFAULT_XTC_THRESHOLD,
+        help="Threshold the probs of each next token candidate to be sampled by XTC",
     )
     parser.add_argument(
         "--seed",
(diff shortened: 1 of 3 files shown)


angeloskath:
Hi! Thanks for the PR.

I think if we are gonna do this let's go all the way. How do you feel about PRing the full terminal UI?

The main difficulty stems from properly handling the chat when in distributed mode. Currently when run in distributed mode the input is simply broadcast by `mlx.launch`. This is very likely to break the UI (it breaks readline for instance).

Similarly, chat history is a bit less useful in my opinion. It also introduces synchronization issues (the way it is currently implemented the histories can be different among devices).

You can test how it works on distributed mode by launching locally. I usually run

```
mlx.launch -n 2 mlx_lm.chat
```

Vlor999:
Thank you for the feedback! I'm excited to bring the full terminal UI to `mlx-lm`.
I will refactor the PR to ensure that:
* All UI rendering (Rich) and input handling (Readline) is strictly isolated.
* Input is broadcasted correctly to other ranks to prevent UI breakage.
* I'll handle the history issue too.

I'll update the PR shortly.

Vlor999:
**Done**! I've squashed the commits and updated the PR based on your feedback.

Summary of changes:
* UI: Full rich integration for a proper terminal experience (Markdown, syntax highlighting).
* Distributed Mode: Implemented robust `broadcast_string` synchronization and isolated all UI/Input logic to rank 0.
* History: Switched to standard `readline` session history (removed the file-based logic).
* Dependencies: Added rich to `install_requires` in `setup.py`.

Ready for review! @angeloskath

mlx_lm/chat.py
```diff
@@ -1,8 +1,16 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import os
+import readline
 
 import mlx.core as mx
+import numpy as np
```
angeloskath: No need for numpy.

mlx_lm/chat.py
```diff
@@ -16,6 +24,60 @@
...
 DEFAULT_MODEL = "mlx-community/Llama-3.2-3B-Instruct-4bit"
+DEFAULT_SYSTEM_PROMPT = (
+    "You are a helpful assistant. Your responses are rendered in a terminal with "
+    "Markdown support. Feel free to use Markdown formatting when appropriate: "
+    "**bold**, *italic*, `inline code`, code blocks with syntax highlighting "
+    "(```language), bullet lists, numbered lists, and headers."
+)
+
+
+def broadcast_string(s: str, group: mx.distributed.Group, root: int = 0) -> str:
```
angeloskath: This function is unnecessarily slow and complicated. Something better would be

```python
def broadcast(s: str, group: mx.distributed.Group, src: int = 0) -> str:
    if group.size() == 1:
        return s
    if group.rank() == src:
        data = mx.array(s.encode("utf-8"))
        mx.eval(mx.distributed.all_sum(data.size, group=group))
        mx.eval(mx.distributed.all_sum(data, group=group))
    else:
        size = mx.distributed.all_sum(0, group=group).item()
        data = mx.distributed.all_sum(mx.zeros(size, dtype=mx.uint8), group=group)
        mx.eval(data)
        s = bytes(data).decode("utf-8")
    return s
```

the above doesn't need numpy, doesn't cast to list and generally is more readable. Communicate the size and data, collect the size and data.

mlx_lm/chat.py
```diff
@@ -93,12 +161,19 @@ def main():
 
     group = mx.distributed.init()
     rank = group.rank()
+    is_rank_0 = rank == 0
```
angeloskath: Total nitpick but why that over `rank == 0` ?

mlx_lm/chat.py
```diff
@@ -93,12 +161,19 @@ def main():
...
     pipeline_group = group if args.pipeline else None
     tensor_group = group if not args.pipeline else None
 
-    def rprint(*args, **kwargs):
-        if rank == 0:
-            print(*args, **kwargs)
+    console = Console(force_terminal=True, highlight=True) if (is_rank_0) else None
```
angeloskath: There is a lot of cases below where there is a fallback that will be unused because `console` will never be `None` for `rank == 0`. I would simply remove them to make everything simpler.

* `rprint` remains and uses `console.print`
* `print_help()` just creates the help and `rprint`s the `Panel`
* The header should use `rprint` no need for if checks there
* Only one generation loop no need to both print and console print
* etc

angeloskath (changes requested):
Looks great. A lot of this can be simplified to begin with. I left some comments.

I tested the distributed use a bit and the live markdown resorts to ellipsis quite too fast which is annoying cause you can't see what is being generated. Not sure what the fix is for that.

Vlor999:
Hi, @angeloskath 
I've fixed everything. Everything works properly. 
If there is still something that tweaks you, tell me and I'll change it.

Vlor999:
Hi @angeloskath @awni,

I just synced the branch with main to ensure there are no conflicts with the latest changes.

Everything looks good on my end. Let me know if there is anything else needed before merging!

Vlor999:
I've updated everything and sync with the main branch.
Tell me @angeloskath  if you want anything else before merging ?

Vlor999:
Hi team! 👋 
Just a gentle ping on this PR. Let me know if there's anything else you need from my side or if you'd like any changes to the implementation! 🚀

PR #842, opened 2026-02-04
Fix batch mamba

awni:
Fixes a couple things in batch generation:

1. Check if a request / model pair are batchable after loading the model
2. Get rid of `MambaCache` since it was causing issues trying to merge it in the `ArraysCache.merge` method. (Size parameter was interpreted as the left padding).

### DIFF
--- a/mlx_lm/models/baichuan_m1.py
+++ b/mlx_lm/models/baichuan_m1.py
@@ -8,7 +8,7 @@ import mlx.nn as nn
 
 from .activations import swiglu
 from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
-from .cache import CacheList, KVCache, MambaCache, RotatingKVCache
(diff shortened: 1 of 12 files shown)


mlx_lm/server.py
```diff
@@ -773,13 +771,9 @@ def progress_callback(info):
                     }
                     continue
 
-                # We have no batch and it actually is not a batchable request
-                # so serve single sequence at a time.
-                elif batch_generator is None and not is_batchable:
```
angeloskath, line 778: Out of curiosity why the change here?

awni: Because before we load the model it defaults to not batchable. So the first time we check `_is_batchable` it evaluates to false every time (assuming the model wasn't passed to the server but loaded from the request).

That means that for every first request we will use _serve_single and be in sequential mode even when it could have been batched.

Issue #814, opened 2026-01-27
RuntimeError: [metal::Device] Unable to build metal library from source


arthurhjorth:
I think it might be related to https://github.com/ml-explore/mlx-lm/issues/784. However, pulling the mlx repo, running `rm -rf ~/Library/Caches/mlx` and re-building did **not** solve it for me, so that's why I'm posting this. That is to say, it might be redundant, so obviously feel free to just close it with no comments if this doesn't help.

I get an error when using `outlines` to create structured outputs on the mlx-lm server.  This only happens with bf16 tensor LLMs, but works just fine on all other dtypes (as far as I know.) 

The error is here:

```
RuntimeError: [metal::Device] Unable to build metal library from source
mlx/backend/metal/kernels/utils.h:459:36: error: assigning to 'bfloat16_t' (aka 'bfloat') from incompatible type 'float'
out[batch * inp_shape[1] + elem] = bit ? inp[batch * inp_shape[1] + elem] : -INFINITY;
                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
mlx/backend/metal/kernels/utils.h:456:25: warning: comparison of integers of different signs: 'uint' (aka 'unsigned int') and 'const constant int' [-Wsign-compare]
uint bit = ((elem >> 5) < mask_shape[1]) &&
             ~~~~~~~~~  ^ ~~~~~~~~~~~~~
mlx/backend/metal/kernels/utils.h:456:25: warning: comparison of integers of different signs: 'uint' (aka 'unsigned int') and 'const constant int' [-Wsign-compare]
uint bit = ((elem >> 5) < mask_shape[1]) &&
             ~~~~~~~~~  ^ ~~~~~~~~~~~~~
mlx/backend/metal/kernels/utils.h:463:156: note: in instantiation of function template specialization 'custom_kernel_bitmask_apply_batched__bfloat16_t<bfloat>' requested here
template [[host_name("custom_kernel_bitmask_apply_batched__bfloat16_t")]] [[kernel]] decltype(custom_kernel_bitmask_apply_batched__bfloat16_t<bfloat16_t>) custom_kernel_bitmask_apply_batched__bfloat16_t<bfloat16_t>;
                                                                                                                                                           ^

```

A full stack trace can be seen in [this gist](https://gist.github.com/arthurhjorth/aa9522bfcbdc0c74426d38f00479497b).

mlx-versions and hashes

```
[...]

PR #845, opened 2026-02-05
Feat: Structured outputs

arthurhjorth:
This is my first contribution to a larger project, so I'm writing this with trepidation.

### Description
This PR adds support for structured outputs on the `mlx-lm.server` and for `BatchGenerator`. Structured outputs are achieved by creating a stateful logits_processor from a JSON schema using `outlines`'s `OutlinesCoreLogitsProcessor`. It keeps track of and forwards its own state for each token, so it works out of the box for how logits_processors are handled by `mlx-lm` and so aligns nicely with the recent additions of logits_processors to `BatchGenerator` (v 0.30.1).

### Changes
* `generate.py`:  `batch_generate()` accepted `logits_processors` as an argument, but never passed them on to `BatchGenerator`. So I made one tiny change in line 1299 to pass them in alongside prompts, tokens and caches.
* `structured.py` is where the `outlines` related stuff lives. There is some overhead when building each for the first time, so it is built around a cache of `OutlinesCoreLogitsProcessor` objects can be retrieved by tokenizer+schema or tokenizer+Type[BaseModel].
* `server.py`: Request handling now extracts JSON schema from requests, and if present gets a logit processor from the cache and passes that into generate calls.

Edit: by the way, I wrote pseudocode for the logits processors cache, but had an LLM write it for me. I then moved things around bit and added data classes since the LLMs code resulted in incomprehensible type hinting (List[Dict[str, List[int]...) etc. Just for full disclosure.

### Motivation
Structured outputs are fundamental for reliably working programatically with LLM outputs. For that reason, they're in demand for LLM-servers. My personal motivation for including it in batching is that most of my research focuses on text analyses of massive corpora, and I need to be able to specify precisely what I want the LLM to analyze/look for/label. Structured outputs is basically a requirement for that.

###  Final thoughts/comments:
Outlines has some issues. The current version of `outlines` installs an `outlines-core` that fails on fp16 models (#814). The only way to make it work is to force an update on only `outlines-core`, which makes it _work_ - but gives a pip error. I could see why you wouldn't want to include that as a requirement for `mlx-lm`. Maybe they will fix this, but as you see in #814, @awni [raised this with the outlines team four months ago](https://github.com/dottxt-ai/outlines/issues/1769), and it's still not fixed. Not super confidence inspiring.

The _minimal_ change would be to just keep the change to `generate` so logits_processors are at least passed into `BatchGenerator`. It feels like it might have been the intention anyway, since `logits_processors` are not used for anything else in there.

If you decide to go with the minimal change, I'm wondering if I should do another PR that instead focuses on aligning the ways in which logits_processors are passed into different kinds of generation functions, e.g. for `generate()` and `stream_generate()` you have to pass them in as kwargs:
```
g = generate(model, tokenizer, prompts[0], **{'logits_processors': logits_processors[0]})
```
whereas `batch_generate()` accept them as named args,

```
result = batch_generate(
    model, tokenizer, prompts, verbose=False, return_prompt_caches=True, logits_processors=logits_processors
)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,8 @@ setup(
         "protobuf",
         "pyyaml",
         "jinja2",
+        "outlines==1.2.9",
+        "outlines_core==0.2.14",
     ],
     packages=[
         "mlx_lm",


awni:
Indeed I don't think it makes sense to take outlines as a dependency here. But I would like to get the fix to the logits processors you made merged. Would you mind either sending that as a separate PR or removing the rest here?

arthurhjorth:
Weekend had started in Denmark and was away for the weekend, sorry!

Very happy to - I'll make a separate PR based on a fresh pull tomorrow and submit that.

And re. outlines dependency, totally understandable. I'll just keep the server in a repo of its own in case someone wants to use that. It might still be useful to some.

PR #848, opened 2026-02-06
refactor: use time.perf_counter() for duration measurements

m92y:
Refactor to use `time.perf_counter()` instead of `time.time()` for duration measurements.

Files modified:
- `mlx_lm/cache_prompt.py`
- `mlx_lm/perplexity.py`
- `mlx_lm/quant/dwa.py`
- `mlx_lm/server.py`

Rationale:

- Monotonicity: `time.time()` can be affected by system clock updates (NTP sync), whereas `time.perf_counter()` is a monotonic clock that never goes backward.
- Precision: Provides the highest resolution possible to calculate precise duration.
- Consistency: This aligns `mlx-lm` with the timing approach already adopted in the core MLX library (see [[ml-explore/mlx#2943](https://github.com/ml-explore/mlx/pull/2943)](https://github.com/ml-explore/mlx/pull/2943)).

**Note:** `time.time()` is intentionally preserved in `server.py` line 244 (`self.created = int(time.time())`) as this generates an epoch timestamp for API responses, not a duration measurement.

### DIFF
--- a/mlx_lm/cache_prompt.py
+++ b/mlx_lm/cache_prompt.py
@@ -112,11 +112,11 @@ def main():
     y = mx.array(prompt)
 
     # Process the prompt
-    start = time.time()
+    start = time.perf_counter()
     max_msg_len = 0
 
     def callback(processed, total_tokens):
-        current = time.time()
+        current = time.perf_counter()
         speed = processed / (current - start)
         msg = f"\rProcessed {processed:6d} tokens ({speed:6.2f} tok/s)"
         nonlocal max_msg_len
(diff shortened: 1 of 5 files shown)


mm65x (commented):
`share.py` also uses `time.time()` for duration measurement (lines 126 and 171: `start_time = time.time()` / `time.time() - start_time`). might want to include that for completeness.

m92y:
> `share.py` also uses `time.time()` for duration measurement (lines 126 and 171: `start_time = time.time()` / `time.time() - start_time`). might want to include that for completeness.

Good catch, thanks! Added refactor to `share.py` in the latest push.

zcbenz:

PR #849, opened 2026-02-06
server: add usage.prompt_tokens_details.cached_tokens to json response

percontation:
Adds an OpenAI compatible info field indicating how many prompt tokens from a request were found in a cache.

https://platform.openai.com/docs/api-reference/responses/object#responses-object-usage-input_tokens_details-cached_tokens

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -371,6 +371,7 @@ class GenerationContext:
     eos_token_ids: set
     stop_token_sequences: List[List[int]]
     prompt: List[int]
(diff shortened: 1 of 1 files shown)


angeloskath (commented):
Looks great!

I have one nitpick comment, how about `prompt_cache_length` everywhere instead of `prompt_ncache` and `prompt_cache_count`?

percontation:
👍. I'll standardize on "prompt_cache_count" because it sits beside the already-existing "prompt_token_count" and "completion_token_count" variables.

PR #850, opened 2026-02-06
server: log bad tool_calls from model

percontation:
Makes the server more robust against failed tool call parsing.

I was occasionally running into the issue in #792 and didn't want exceptions here to blow up the http response. I don't need this patch anymore now that that's been fixed, but, it might still make sense. Feel free to reject if this isn't how you want to handle Exceptions here.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1383,7 +1383,11 @@ class APIHandler(BaseHTTPRequestHandler):
                 return []
             result = []
             for tool_text in tool_calls:
-                parsed = ctx.tool_parser(tool_text, request.tools)
+                try:
+                    parsed = ctx.tool_parser(tool_text, request.tools)
+                except Exception:
+                    logging.exception(f"Unparsable tool call from model {tool_text=}")
+                    parsed = []
                 if isinstance(parsed, list):
                     result.extend(format_tool_call(tc) for tc in parsed)
                 else:


mlx_lm/server.py
```diff
@@ -1383,7 +1383,11 @@ def parse_tools(tool_calls):
                 return []
             result = []
             for tool_text in tool_calls:
-                parsed = ctx.tool_parser(tool_text, request.tools)
+                try:
+                    parsed = ctx.tool_parser(tool_text, request.tools)
+                except Exception:
+                    logging.exception(f"Unparsable tool call from model {tool_text=}")
+                    parsed = []
```
mm65x, line 1390: with `parsed = []` the failed tool call is silently dropped from the response. an alternative would be to `continue` (like #899 does) which has the same effect but is slightly clearer in intent - "skip this one, move on."

mm65x (commented):
note that #899 is essentially the same change with a narrower exception filter (`json.JSONDecodeError, ValueError, KeyError` instead of bare `Exception`). might be worth coordinating so you don't duplicate effort. the narrower catch is arguably better since a bare `except Exception` could mask unexpected bugs (e.g. `TypeError` from a wrong function signature), though `logging.exception` at least makes them visible in logs.

one other thing: when the parse fails, the tool call text is silently dropped — the client gets no indication a tool call was attempted. could be worth falling back to including it in the response `content` field instead, so the user at least sees what the model tried to produce.

zcbenz:

PR #855, opened 2026-02-07
Fix missing Hugging Face Hub dependency

sealad886:
## Summary\n- add `huggingface_hub` to runtime dependencies so CLI tools that call Hub APIs (including `mlx_lm.convert`) import cleanly\n\n## Testing\n- not run (dependency change only)


angeloskath:
I will close this as `huggingface-hub` is installed as part of transformers and even though `mode` could be set in the mixed quant fallback it is not a bug (affine is the default).

Feel free to showcase a bug and open another PR if there is such a problem.

PR #857, opened 2026-02-07
Make validation set optional in training process

Goekdeniz-Guelmez:
so that when loading a HF dataset (like: mlx-community/dclm-baseline-1.0-138k) that don't have the valid split, will still trains.


angeloskath (approved):
Thanks that's nice :-)

PR #863, opened 2026-02-08
Fix gradient computation in moe_gate_select function

Goekdeniz-Guelmez:
addreses [PR](https://github.com/Goekdeniz-Guelmez/mlx-lm-lora/issues/44) The value error: `[GatherMM] Cannot calculate VJP with respect to indices` is because the attempt to calculate gradients for the indices and then passed to the gather_mm operation.

### DIFF
--- a/mlx_lm/models/step3p5.py
+++ b/mlx_lm/models/step3p5.py
@@ -110,7 +110,7 @@ def moe_gate_select(gates, router_bias, top_k, routed_scaling_factor, norm_topk_
             mx.sum(topk_weights, axis=-1, keepdims=True) + 1e-20
         )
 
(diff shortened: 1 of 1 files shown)


awni:
Are you sure the model is in training mode? The gradient is always stopped when `model.training` is `True`: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/switch_layers.py#L186-L187

Goekdeniz-Guelmez:
it looks like the user of the PR in mlx-lm-lora is applying 4 bit quantisation on top of a already 5 bit quant. thats most likely the reason. so it has nothing to do with the indices in mlx-lm. I'll go ahead and close this PR.

PR #861, opened 2026-02-08
Adding Support for Qwen3.5

johnmai-dev:
Port of https://github.com/huggingface/transformers/pull/43830

TODO:
- [ ] Test Qwen3.5 Model https://huggingface.co/Qwen/Qwen3.5-9B-Instruct

PR #869, opened 2026-02-10
[MODEL] support qwen3.5 series w/o vision

JJJYmmm:
This PR adds model support for the upcoming Qwen3.5 models, including both dense and MoE variants.

It's a refine version of #861 by @johnmai-dev.

Reference HF implementation - https://github.com/huggingface/transformers/pull/43830

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -530,7 +530,7 @@ class TestModels(unittest.TestCase):
         self.model_test_runner(
             model, args.model_type, args.vocab_size, args.num_hidden_layers
         )
-    
+
     def test_qwen3_5_family_convert_then_load_norm_not_shift_twice(self):
         text_config = {
             "hidden_size": 8,


johnmai-dev:
Do we need to add support for qwen3_5_text and qwen3_5_moe_text?

https://github.com/huggingface/transformers/blob/42791a34fdeae197f60f11ace3807c81f44b0729/src/transformers/models/auto/modeling_auto.py#L356-L357

JJJYmmm:
> Do we need to add support for qwen3_5_text and qwen3_5_moe_text?
> 
> https://github.com/huggingface/transformers/blob/42791a34fdeae197f60f11ace3807c81f44b0729/src/transformers/models/auto/modeling_auto.py#L356-L357

imo it’s not need, because we’ll only release the vlm ckpts, so just following previous vlms e.g. qwen3vlmoe should be ok.

mlx_lm/models/qwen3_5.py
```diff
@@ -0,0 +1,376 @@
...
+            "model.norm.weight",
+            ".q_norm.weight",
+            ".k_norm.weight",
+        )
+        for k, v in weights.items():
+            if "conv1d.weight" in k and v.shape[-1] != 1:
+                weights[k] = v.moveaxis(2, 1)
+            if any(k.endswith(sfx) for sfx in norm_keys):
+                if v.ndim == 1:
+                    weights[k] = v + 1.0
```
awni: I think this is a bug. The sanitize function is called every time a mlx model is loaded so if you do convert the model (which will call sanitize) then run it (which will call sanitize) you will add 1.0 to these values twice.

Instead we should only apply this scaling once. An easy way to do that is to have a condition which can tell you if the model has already been sanitized. (For example if the "mpt" layer is in the weights or something).

JJJYmmm: got it, update the sanitize logic and add a test🫡

awni:
@JJJYmmm should we go ahead and merge this? Have you tested it on an actual model yet?

JJJYmmm:
I’ve tested it on preview ckpts, so it’s fine to merge now. I’ll check if it still works when the official version drops. 🤗

PR #871, opened 2026-02-11
Share model

angeloskath:
Mostly putting this to check if people think it's useful. We could also put it in `mlx_lm.manage`.

This PR introduces a utility called `mlx_lm.share` that can be used to broadcast models (or any folder/file) from 1 node to the others using `mlx.core.distributed`. It is often the case that we download a model on one node and then have to move it to others via external storage or something similar.

Thunderbolt 5 with JACCL actually saturates the write bandwidth of the disk so with `mlx_lm.share` I am getting 5-6 GB/s broadcast so no need for external storage to move things around. It would be strictly slower...

Examples

```
# Assumes we have ran
# mlx.distributed_config --hosts h1,h2,h3,h4,... --over thunderbolt \
#                        --backend jaccl (or jaccl-ring) --output hosts.json

mlx_lm.share --path mlx-community/Kimi-K2.5 --hostfile hosts.json

mlx_lm.share --path my-big-file.bin --hostfile hosts.json --dst /tmp/dst.bin
```

### DIFF
--- a/mlx_lm/share.py
+++ b/mlx_lm/share.py
@@ -2,10 +2,14 @@
 
 import argparse
 import os
+import pickle
 import sys
-from functools import partial
+import time
+from dataclasses import dataclass
+from functools import partial, total_ordering
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import Literal, Optional
 
 import mlx.core as mx
 from huggingface_hub.errors import LocalEntryNotFoundError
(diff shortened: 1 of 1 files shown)


mlx_lm/share.py
```diff
@@ -0,0 +1,214 @@
...
+        share_file(path, file, src, group)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Distribute a model to other nodes using MLX distributed."
+    )
+    parser.add_argument(
+        "--path", type=str, required=True, help="Path to the MLX model."
+    )
```
awni, line 218: nit: wdyt about renaming that to `--model` to be consistent with other command?

awni: Although if the intention is this can be used for anything (not just a model) than `--path` is probably better.

ivanfioravanti:
Nice! No more manual copy and paste across nodes!

PR #878, opened 2026-02-12
Add memory-efficient chunked conversion for large models

### LINKED ISSUE #876: Support memory-efficient chunked conversion for models larger than RAM
## Summary

`mlx_lm.convert` currently requires the entire quantized model to fit in memory during the save phase. This makes it impossible to convert very large models (e.g., 700B+ MoE models like GLM-5 at 744B/1.51TB) on any existing Apple Silicon hardware, even the Mac Studio M3 Ultra with 512GB unified memory.

## Problem

The conversion pipeline has three phases:

1. **Load** (`load_model` with `lazy=True`) -- loads weight references lazily. Memory-efficient.
2. **Quantize** (`nn.quantize`) -- walks each `Linear` module, calls `mx.quantize()` per layer. Potentially streamable.
3. **Save** (`save_model`) -- calls `tree_flatten(model.parameters())` collecting **all** quantized weights into a single dict, then `make_shards()` bins them into ~5GB chunks before writing.

The bottleneck is step 3. Even though `make_shards` splits weights into 5GB shards and the write loop clears references after each shard, all quantized weight references are collected upfront. For a 744B model at 4-bit quantization, this requires ~370GB of memory.

### Relevant code (`utils.py`):

```python
def save_model(save_path, model, *, donate_model=False):
    weights = dict(tree_flatten(model.parameters()))  # ALL weights collected here
    shards = make_shards(weights)                      # ALL shards planned here
    # ... write loop processes one shard at a time but all are in memory
```

## Proposed Solution

Add a `--chunked` flag to `mlx_lm.convert` that processes weights in shard-sized batches:

1. Load model lazily (existing behavior)
2. Read `model.safetensors.index.json` to get the source weight map
3. Plan output shards by grouping weight keys into ~5GB output chunks (estimating quantized sizes from shapes + target bits)
4. For each planned output shard:
   - Load only the source safetensor files needed for those weight keys
   - Apply `sanitize()` transforms for the relevant keys
   - Call `mx.quantize()` directly on quantizable weights
   - Write the output shard via `mx.save_safetensors()`
   - Free memory explicitly + `mx.metal.clear_cache()`
5. Write config, tokenizer, and index metadata

### Peak memory target

~1 shard size (~5GB) + overhead for `sanitize()` transforms (e.g., MoE expert stacking loads all experts for one layer, typically a few GB). Total peak: **~10-15GB** regardless of total model size.

### Key design considerations

- **`sanitize()` compatibility**: Model-specific `sanitize()` methods (e.g., MoE expert stacking) operate on subsets of weights. The chunked pipeline needs to batch weights by layer so sanitize transforms have all their inputs available within a single chunk.
- **Quantization predicates**: Mixed quantization recipes (`mixed_2_6`, etc.) use layer indices to vary bit widths. These can be pre-computed from config without loading the full model.
- **`cast_predicate`**: Some models (e.g., GLM MoE) exclude specific keys from dtype casting. This needs to be applied per-weight in the chunked path.

## Motivation

The largest Apple Silicon machines today have 512GB unified memory. Many recent MoE models exceed this:

| Model | Total Params | BF16 Size | 4-bit Quantized |
|-------|-------------|-----------|-----------------|
| GLM-5 | 744B | 1.51 TB | ~370 GB |
| DeepSeek-V3 | 671B | 1.34 TB | ~335 GB |

Even models that *barely* fit (e.g., DeepSeek-V3 at 4-bit on 512GB) leave no headroom for the conversion process overhead.

A chunked conversion would allow these models to be converted on machines with as little as 32-64GB of RAM, dramatically expanding the set of hardware that can produce MLX-format weights.

## Alternatives Considered

- **Disk-backed swap**: macOS virtual memory can extend beyond physical RAM, but the conversion becomes extremely slow due to SSD thrashing and is impractical for TB-scale models.
- **Cloud GPU conversion**: MLX requires Apple Silicon, so standard cloud GPUs (A100/H100) can't run `mlx_lm.convert`. Cloud Apple Silicon options max out at 128GB RAM currently.
- **External tool (e.g., convert via GGUF then to MLX)**: Adds quality loss from double conversion and doesn't address the core limitation.

## Testing Strategy

The feature could be validated on smaller MoE models that fit in RAM (e.g., Mixtral 8x7B) by comparing the chunked conversion output against the standard conversion output bit-for-bit.

iamadalek:
## Summary

- Adds `--chunked` flag to `mlx_lm.convert` that processes weights one transformer layer at a time, enabling conversion of models larger than available RAM
- New `mlx_lm/chunked_convert.py` module with streaming pipeline: safetensors header scanning, layer-based chunk planning, quantization plan building, per-chunk load/sanitize/cast/quantize/save with aggressive memory reclamation between chunks
- Produces output **identical** to the standard conversion path (verified by bit-for-bit comparison tests on synthetic models)
- Supports standard quantization (affine, mxfp4, nvfp4, mxfp8), mixed recipes, and dtype-only conversion
- Incompatible combinations (--dequantize, AWQ/GPTQ/DWQ, VL models, custom model_file) raise clear errors; single-file models fall back to standard conversion automatically

## Benchmark

Synthetic 8-layer Llama model (hidden=1024, vocab=32000, 642 MB total):

| Mode | Peak Memory | Ratio |
|------|------------|-------|
| Standard | 742 MB | 1x |
| Chunked | 164 MB | **4.5x reduction** |

The ratio scales with layer count. An 8-layer model processes ~1 layer at a time (plus embeddings/lm_head). A 32-layer model would see ~16x reduction. For the target use case (DeepSeek V3 at 671B), peak memory drops from ~370GB (full model) to ~29GB (one MoE layer).

## Test plan

- [x] Unit tests: safetensors header parsing, layer chunk planning, quantized size estimation, quant plan building, validation checks, chunked sanitize
- [x] Integration test: bit-for-bit comparison of chunked vs standard conversion on synthetic Llama model (4-layer, quantized)
- [x] Integration test: MoE expert stacking via chunked sanitize on synthetic Mixtral model (sentinel injection for non-zero layers)
- [x] Integration test: dtype-only chunked conversion
- [x] Error case: --chunked + --dequantize raises ValueError
- [x] CLI tests: --chunked flag registered in parser with correct default
- [x] Full test suite: all 176 tests pass (175 existing + 1 new)
- [x] Linting: black + isort clean

Closes #876

### DIFF
--- a/mlx_lm/chunked_convert.py
+++ b/mlx_lm/chunked_convert.py
@@ -1,0 +1,562 @@
+# Copyright © 2023-2024 Apple Inc.
+
+import copy
+import gc
+import glob
+import json
+import re
+import shutil
+import struct
+from collections import defaultdict
+from dataclasses import dataclass
+from math import prod
+from pathlib import Path
+from typing import Callable, Dict, List, Optional, Set, Tuple, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .convert import MODEL_CONVERSION_DTYPES, mixed_quant_predicate_builder
+from .tokenizer_utils import load as _load_tokenizer
+from .utils import (
+    MAX_FILE_SIZE_GB,
+    _download,
+    _get_classes,
+    create_model_card,
+    load_config,
+    save_config,
+    upload_to_hub,
+)
+
+
+@dataclass
+class WeightChunk:
+    layer_idx: Optional[int]
+    weight_keys: List[str]
+    source_files: Set[str]
+
+
+def read_safetensors_header(filepath: str) -> dict:
+    """Read tensor metadata from safetensors file header.
+
+    Returns {tensor_name: {"shape": [...], "dtype": "...", "size_bytes": N}}
+    """
+    with open(filepath, "rb") as f:
+        header_size = struct.unpack("<Q", f.read(8))[0]
+        header = json.loads(f.read(header_size))
+    header.pop("__metadata__", None)
+    return {
+        name: {
+            "shape": meta["shape"],
+            "dtype": meta["dtype"],
+            "size_bytes": meta["data_offsets"][1] - meta["data_offsets"][0],
+        }
+        for name, meta in header.items()
+    }
+
+
+def scan_source_tensors(model_path: Path, weight_map: dict) -> dict:
+    """Scan all source safetensors headers. Returns complete tensor catalog."""
+    catalog = {}
+    for source_file in sorted(set(weight_map.values())):
+        header = read_safetensors_header(str(model_path / source_file))
+        for name, meta in header.items():
+            meta["source_file"] = source_file
+            catalog[name] = meta
+    return catalog
+
+
+def plan_layer_chunks(weight_map: dict) -> List[WeightChunk]:
+    """Group weight keys by transformer layer index."""
+    layer_pattern = re.compile(r"(?:model\.layers|transformer\.blocks)\.(\d+)\.")
+    layer_groups = defaultdict(list)
+    non_layer = []
+
+    for key in sorted(weight_map.keys()):
+        m = layer_pattern.search(key)
+        if m:
+            layer_groups[int(m.group(1))].append(key)
+        else:
+            non_layer.append(key)
+
+    chunks = []
+    # Non-layer weights as a single chunk (embed, lm_head, norms)
+    if non_layer:
+        chunks.append(
+            WeightChunk(
+                layer_idx=None,
+                weight_keys=non_layer,
+                source_files={weight_map[k] for k in non_layer},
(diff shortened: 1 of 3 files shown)


awni:
This is not necessary.. it's already possible to convert a model larger than the RAM of the machine as the conversion is streamed in chunks already.

iamadalek:
> This is not necessary.. it's already possible to convert a model larger than the RAM of the machine as the conversion is streamed in chunks already.

Thanks for the quick review, Awni. I appreciate the clarification — I should have dug deeper into how MLX’s lazy evaluation handles the memory pressure during conversion before assuming the bottleneck was where I thought it was. Good to know the existing path already streams effectively.

PR #886, opened 2026-02-12
Fix save/load of CacheList

angeloskath:
As title.

It also changes the way the state of a cache list is represented (in a backwards incompatible way) but given that nobody could save and load it I don't think this will break anybody's code.


awni (approved):
Nice, thanks!!

PR #893, opened 2026-02-15
Add `mixed_2_3` and `mixed_2_4` quant predicates

spicyneuron:
Huge models like Kimi K2.5, even quantized to 3 bits or `mixed_3_4`, barely fit on a Mac Studio M3 512G.

This PR adds even more aggressive versions which hopefully strike a usable balance.

---

EDIT: Turns out K2.5 at `mixed_2_4` gets... loopy. So probably no hope for `mixed_2_3` either, at least for this particular case.

I'll leave this open in case folks think there's value in more options, or feel free to close.

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -23,7 +23,13 @@ def mixed_quant_predicate_builder(
     mode = "affine"
     high_bits = 6
 
-    if recipe == "mixed_2_6":
+    if recipe == "mixed_2_3":
+        low_bits = 2
+        high_bits = 3
+    elif recipe == "mixed_2_4":
+        low_bits = 2
+        high_bits = 4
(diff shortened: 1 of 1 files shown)


angeloskath:
Thanks for doing all the testing and sorry it doesn't work. Indeed <4 bits almost always requires some form of QAT unfortunately...

spicyneuron:
@angeloskath Got it working after a bit of trial and error! https://huggingface.co/spicyneuron/Kimi-K2.5-MLX-mixed-2.8-bit

Key was to keep MoE routing at full precision and attention at 8. Seems like MLX's default options all compress those beyond what the model can bear. Some better controls and defaults would go a long way, though it'd might look like "model family templates" more than general CLI knobs.

PR #898, opened 2026-02-16
Fix sharded rms norm in MiniMax M2.5

angeloskath:
Minimax M2.5 sharding is currently broken because the qk norm is done on the whole vector.

This PR fixes it but introduces 2 extra communications per attention layer which pretty much destroys decoding perf improvement. I have a more complicated version that does one communication and overlaps it with some computation which is better but not by much, we may need something more fundamental here.

Otoh this is now correct and the prompt processing scales quite nicely at 2.7x speedup across 4 nodes at 8k tokens. Adding to that the fact that the KV cache is also 1/4th the size it still makes sense to shard it for long agentic tasks or using multiple subagents.

### DIFF
--- a/mlx_lm/models/minimax.py
+++ b/mlx_lm/models/minimax.py
@@ -37,11 +37,20 @@ class ModelArgs(BaseModelArgs):
 @lru_cache
 def sharded_rms_norm(group):
     @mx.compile
-    def _inner_sharded_rms_norm(x, w, eps):
-        norm2 = x.square().sum(-1, keepdims=True)
+    def _cast_square_sum(x):
+        return x.astype(mx.float32).square().sum(-1, keepdims=True)
+
+    @mx.compile
+    def _normalize(x, norm2, w, eps):
         norm2 = mx.distributed.all_sum(norm2, group=group)
         norm = mx.rsqrt(norm2 / (x.shape[-1] * group.size()) + eps)
-        return x * norm * w
+        return (x.astype(mx.float32) * norm * w).astype(x.dtype)
+
+    # Split the compile so that x upcasting doesn't break the compile and we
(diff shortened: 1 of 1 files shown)


mlx_lm/models/minimax.py
```diff
@@ -33,6 +34,45 @@ class ModelArgs(BaseModelArgs):
...
+        group = group or mx.distributed.init()
+        self.weight = mx.ones((dims // group.size(),))
+        self.group = group
+        self.eps = eps
+
+    def _extra_repr(self):
+        return f"{self.weight.shape[0] * self.group.size()}, eps={self.eps}"
+
+    def __call__(self, x):
+        return sharded_rms_norm(self.group)(x, self["weight"], self.eps)
```
awni, line 72: I think that will make a new function every time so each call needs to be recompiled.

angeloskath: Oops yeah. Imported the lru cache but forgot to add it.

mlx_lm/models/minimax.py
```diff
@@ -33,6 +34,46 @@ class ModelArgs(BaseModelArgs):
     use_qk_norm: bool = True
 
 
+@lru_cache
+def sharded_rms_norm(group):
+    @mx.compile
+    def _inner_sharded_rms_norm(x, w, eps):
+        norm2 = x.square().sum(-1, keepdims=True)
+        norm2 = mx.distributed.all_sum(norm2, group=group)
+        norm = mx.rsqrt(norm2 / (x.shape[-1] * group.size()) + eps)
```
awni: Oh also x should probably be up cast prior to the sum for parity with `mx.fast.rms_norm`

PR #900, opened 2026-02-16
Add streaming AWQ to reduce memory required for quantization

spicyneuron:
This PR adds a streaming AWQ path that quantizes and writes model weights incrementally, so we don’t need to keep the full output in memory before saving. This is mainly useful for very large models that risk OOM'ing on the available hardware.

It's part of my saga to optimize Kimi K2.5 on Mac Studio, but this change is separate and self-contained.

Includes a few adjacent fixes (can remove / split into separate PRs, if preferred):

- Adds `--max-file-size-gb` and applies it to both paths.
  - Before: shard sizing was effectively fixed by path and not consistently configurable across AWQ save flows.
  - Now: both streaming and non-streaming honor the same shard-size limit, so output artifact sizing is predictable either way.

- Uses the adjusted distributed `num_samples` value when loading calibration data.
  - Before: we rounded `num_samples` for world-size divisibility, but still loaded calibration data with the original unadjusted value.
  - Result: mismatch between intended per-rank split assumptions and the actual loaded sample count.
  - Now: calibration loading uses the adjusted value directly.

- Gates non-streaming `save(...)` to rank 0.
  - Before: in distributed runs, multiple ranks could try to write the same output artifacts concurrently.
  - Now: only rank 0 writes model/tokenizer/config files, matching streaming behavior and avoiding write contention/corruption risk.

### DIFF
--- a/mlx_lm/quant/awq.py
+++ b/mlx_lm/quant/awq.py
@@ -784,7 +784,7 @@ def main():
     if (awq_config := AWQ_MODEL_CONFIGS.get(model_type, None)) is None:
         raise NotImplementedError(f"AWQ support for {model_type} models NYI.")
 
-    calibration_data = load_data(tokenizer, args.num_samples, args.sequence_length)
+    calibration_data = load_data(tokenizer, num_samples, args.sequence_length)
 
     calibration_data = dist_split(calibration_data, group)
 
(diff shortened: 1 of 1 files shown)


spicyneuron:
Closing, since streaming AWQ ended up being less necessary and useful than I expected. No reason to maintain two parallel code paths.

PR #902, opened 2026-02-16
Fix qwen3.5 casting to fp32

awni:
The already uploaded model needs to be reconverted with this. I am working on it.

CC @pcuenca @ivanfioravanti @Blaizzy


Blaizzy:
Thanks for the heads up @awni!

Issue #903, opened 2026-02-17
Caching doesn't seem to be working for Qwen3.5


dexloom:
Running lm_server with Qwen3.5 is completely different in terms of caching, it always processes whole prompt, cached tokens counter in response is always 0.

awni:
The mamba cache cannot be trimmed to so there are fewer cases where you get a cache. Basically if you are continuing a conversation caching should work. But if you using the same system prompt with a new query it will have to reprocess the system prompt.

Could you say more about your use case?

kernelpool:
I was running into the same issue (when using a coding agent) so I hacked together something quick as a workaround - maybe something like https://github.com/kernelpool/mlx-lm/commit/8e08dbb5f0fb410ea9c3af5e1f3a43f8aae6da54 could work?

dexloom:
> Could you say more about your use case?

Local MLX server running Qwen/Qwen3.5-397B-A17B MXFP4, custom agent, supporting jinja templates, I am pretty sure beginning of the conversation including system prompt remains the same, switching back to Minimax I have no problems. I've checked server logs to confirm as well.

mzbac:
> The mamba cache cannot be trimmed to so there are fewer cases where you get a cache. Basically if you are continuing a conversation caching should work. But if you using the same system prompt with a new query it will have to reprocess the system prompt.
> 
> Could you say more about your use case?

@awni I was checking the cache today,  noticed people with dual RTX 6000 pro can handle 190k context window with 4bit minimax m2.5 but I keep getting OOM on my m2 192GB, so far I think the prompt cache for thinking model may have some issues, I can see the the LRU keeps inserting cache entries for multi turn conversation for the same session, so far my findings:

for thinking models, the key also contains reasoning tokens that the client typically does not replay (because reasoning isn’t used when tokenizing the next request https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/server.py#L163) . 

In that case the sequences diverge much earlier:
  - cached key: P + R + A + EOS
  - next prompt: P + A + …

Shared prefix is basically just P, so the cache can only reuse up to the end of the previous prompt and you still re-prefill a lot of tokens. It’s not always a total miss (it’s a longer hit + trim), but it’s a bad hit for speeding up multi-turn.


I am not very familiar with the prompt cache implementation, but from the symptoms, it looks like when there is a longer cache hit and it is trimmed, a new cache entry is inserted into the LRU.

awni:
[...]

PR #906, opened 2026-02-18
Improve the cache size limits

### LINKED ISSUE #883: mlx_lm.server causes macOS kernel panic (IOGPUMemory crash) due to unbounded memory growth
### Summary

Running `mlx_lm.server` with Qwen3-Coder-30B-A3B-Instruct-8bit on a Mac Studio M3 Ultra (96GB) caused a full **macOS kernel panic** after the KV cache grew unboundedly during an agentic coding session (~58k+ tokens context). The system crashed and rebooted — not a process termination, but a kernel-level panic in `IOGPUMemory.cpp`.

### System Information

- **Hardware**: Mac Studio (Mac15,14) — Apple M3 Ultra, 28 cores (20P + 8E), 60 GPU cores, 96 GB unified memory
- **macOS**: 26.3 (25D125)
- **Kernel**: Darwin 25.3.0 (xnu-12377.81.4~5/RELEASE_ARM64_T6031)
- **Python**: 3.12
- **mlx-lm**: 0.30.6
- **mlx**: 0.30.6

### Model

`mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit` (~32GB on disk)

### Steps to Reproduce

1. Start `mlx_lm.server`:
   ```bash
   mlx_lm.server --model mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit --port 8080 --max-tokens 64000
   ```
2. Use an agentic coding tool (e.g., [OpenCode](https://opencode.ai)) that makes repeated `/v1/chat/completions` requests with growing conversation context
3. Let the context grow to ~58k+ tokens over multiple tool-call round trips
4. **System crashes** — full kernel panic, forced reboot

### Kernel Panic Details

**Panic string:**
```
panic(cpu 18 caller 0xfffffe00428d679c): "completeMemory() prepare count underflow" @IOGPUMemory.cpp:550
```

**Memory status at crash:**
| Metric | Value |
|--------|-------|
| Wired memory | **80.14 GB** (83.5% of 96GB) |
| Active memory | 7.12 GB |
| Free memory | **0.01 GB** (essentially zero) |
| `memoryPressure` | `false` (system didn't detect pressure!) |

**Python/MLX process memory at crash:**
```
PID 30289 (Python): residentMemoryBytes = 83,235,119,080 (~83.23 GB)
```

The MLX server process was consuming **83.23 GB** of the 96 GB system at the time of the kernel panic.

### Root Cause Analysis

1. **`mlx_lm.server` starts by wiring ~75% of RAM** via `mx.set_wired_limit(max_recommended_working_set_size)` — on a 96GB system, that's ~72GB. Wired memory **cannot be swapped** to disk.

2. **The KV cache grows without bound** — `mlx_lm.server` does not support `--max-kv-size` (see #615), so as the agentic tool sends longer and longer conversations, the KV cache keeps growing.

3. **Memory breakdown at crash:**
   - Model weights (8-bit): ~32 GB
   - KV cache (58k+ tokens, unbounded): ~10-20 GB
   - MoE activations (128 experts): ~3-5 GB
   - Metal buffer cache: ~2-4 GB
   - Total MLX process: **~83 GB**

4. **macOS couldn't recover** — because the memory was **wired** (locked in RAM), macOS's Jetsam OOM killer couldn't reclaim it. The GPU driver (`IOGPUMemory.cpp`) hit a "prepare count underflow" when it couldn't allocate memory, causing a **kernel panic** instead of a graceful process termination.

5. **`memoryPressure: false`** in the panic report shows that macOS didn't even detect memory pressure — the wired memory bypassed the normal memory pressure monitoring system entirely.

### Suggested Fixes

1. **Add `--max-kv-size` to `mlx_lm.server`** (already requested in #615) — this is the most critical fix. Without a KV cache cap, memory grows unboundedly with context length.

2. **Add `--memory-limit` flag to `mlx_lm.server`** — allow users to cap total memory usage via `mx.metal.set_memory_limit()`. The current default (wiring 75% of RAM) is dangerously aggressive for systems running other processes.

3. **Reduce the default wired limit** — `mx.set_wired_limit(max_recommended_working_set_size)` should not wire 75% of RAM by default. Consider 50-60%, or at minimum, make it configurable.

4. **Add memory monitoring with graceful shutdown** — the server should monitor its own memory usage and return an error response (e.g., HTTP 503) when approaching the limit, rather than growing until the kernel panics.

5. **Document the risk** — users should be warned that running large models (30B+) on systems with ≤128GB can cause kernel panics if context grows too large.

### Related Issues

- #615 — `mlx_lm.server` does not support `--max-kv-size`
- [mlx-lm#427](https://github.com/ml-explore/mlx-lm/issues/427) — System Crash while Running Local AI Models

### Panic Log

<details>
<summary>Key fields from panic-full-2026-02-12-011744.0002.panic</summary>

```json
{
  "bug_type": "210",
  "timestamp": "2026-02-12 01:17:44.00 -0300",
  "os_version": "macOS 26.3 (25D125)",
  "incident_id": "0BA6183D-4448-439B-B57E-EE3B2CD397AB"
}
```

**Panic string:**
```
panic(cpu 18 caller 0xfffffe00428d679c): "completeMemory() prepare count underflow" @IOGPUMemory.cpp:550
Debugger message: panic
Memory ID: 0xff
OS release type: User
OS version: 25D125
Kernel version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:47:03 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6031
```

**Memory status:**
```json
{
  "compressorSize": 47640,
  "compressions": 8141671,
  "decompressions": 4388400,
  "busyBufferCount": 0,
  "memoryPressureDetails": {
    "pagesWanted": 3027,
    "pagesReclaimed": 5807
  },
  "pageSize": 16384,
  "memoryPressure": false,
  "memoryPages": {
    "active": 466937,
    "throttled": 0,
    "fileBacked": 342561,
    "wired": 5251984,
    "purgeable": 69720,
    "inactive": 465071,
    "free": 973,
    "speculative": 54
  }
}
```

**Calculated:**
- Wired: 5,251,984 pages × 16,384 bytes = **80.14 GB**
- Free: 973 pages × 16,384 bytes = **0.01 GB**

**MLX Python process:**
```
PID 30289 (Python): residentMemoryBytes = 83,235,119,080 (~83.23 GB)
```

</details>

angeloskath:
Introduce limiting of KV cache by size to fix #883 and part of #903.

The behavior changes as follows:
- All caches expose an `nbytes` property
- `LRUPromptCache` takes a `max_bytes` argument and evicts KV caches so that it will always hold 1 KV Cache or less than `max_bytes`
- The batch generator exposes a `kv_cache_nbytes` property which returns the size of the caches held by the batch generator
- We ask the `LRUPromptCache` to trim its size such that the size held by the batch generator plus the `LRUPromptCache` will be less than a number (at beginning of generation so this can grow more, perhaps we should put that check elsewhere as well)

In a following PR I 'll add saving the prompt processing part of the KV cache so that we can have a working cache for hybrid models.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -803,12 +803,9 @@ class ResponseGenerator:
                     # just making sure we don't leave a reference around
                     del cache
 
-                    if (
-                        self.model_provider.cli_args.prompt_cache_total_bytes
-                        is not None
-                    ):
-                        total = self.model_provider.cli_args.prompt_cache_total_bytes
-                        active = batch_generator.kv_cache_nbytes
+                    if self.model_provider.cli_args.prompt_cache_bytes is not None:
+                        total = self.model_provider.cli_args.prompt_cache_bytes
+                        active = batch_generator.prompt_cache_nbytes
                         self.prompt_cache.trim_to(n_bytes=total - active)
                     continue
 
@@ -1771,10 +1768,7 @@ def run(
     handler_class=APIHandler,
 ):
     group = mx.distributed.init()
-    prompt_cache = LRUPromptCache(
-        model_provider.cli_args.prompt_cache_size,
-        model_provider.cli_args.prompt_cache_bytes,
-    )
+    prompt_cache = LRUPromptCache(model_provider.cli_args.prompt_cache_size)
     response_generator = ResponseGenerator(model_provider, prompt_cache)
     if group.rank() == 0:
         _run_http_server(host, port, response_generator)
@@ -1897,15 +1891,9 @@ def main():
         help="Maximum number of distinct KV caches to hold in the prompt cache",
     )
     parser.add_argument(
-        "--prompt-cache-inactive-bytes",
+        "--prompt-cache-bytes",
         type=int,
-        default=1 << 63,
-        help="Maximum size in bytes of the KV caches held in the prompt cache ie not being actively used",
-    )
-    parser.add_argument(
-        "--prompt-cache-total-bytes",
-        type=int,
-        help="If provided make a best effort to keep the prompt cache and active KV caches lower than this number",
+        help="Maximum size in bytes of the KV caches",
     )
     parser.add_argument(
         "--pipeline",


mlx_lm/generate.py
```diff
@@ -1043,6 +1043,13 @@ def remove(self, uids: List[int], return_prompt_caches: bool = False):
         if return_prompt_caches:
             return caches
 
+    @property
+    def kv_cache_nbytes(self):
```
awni: Maybe switch `prompt_cache_nybtes` for consistency with naming elsewhere.

mlx_lm/server.py
```diff
@@ -1847,6 +1890,23 @@ def main():
...
+        "--prompt-cache-bytes",
+        type=int,
+        default=1 << 63,
+        help="Maximum size in bytes of the KV caches held in the prompt cache",
+    )
+    parser.add_argument(
+        "--prompt-cache-total-bytes",
+        type=int,
+        help="If provided make a best effort to keep the prompt cache and active KV caches lower than this number",
+    )
```
awni, line 1897: I'm not quite clear on what is the difference between these two arguments?

awni:
> In a following PR I 'll add saving the prompt processing part of the KV cache

I was looking at this a bit yesterday and there is some subtlety here that makes it a bit harder than I was expecting, particularly for reasoning models. In a reasoning model you have something like:

Turn 1: 

(capital letters represent arbitrary text)

```
user: A
assistant: <think> B <\think> C
```

Turn 2:

```
user:  A
assistant: C
user: D
assistant: <think> E </think> F
```


In the first turn the prompt is:

```
user: A
assistant: <think>
```

In the second turn the prompt is:

```
user:  A
assistant: C
user: D
assistant: <think>
```

So you can see that even catching the cache right after prefill is not sufficient for an SSM model. It needs to be caught before the first `<think>` token.

angeloskath:
Yeah if you look at the branch `thinking-cache` I have a fix for that which isn't horrible imho.

Basically I look for the think end token and cache there. However to my surprise and annoyance it isn't enough. Tools like opencode for instance add tokens before the assistant start token. Like a system prompt reminder.

It can be extended easily to cache up to just before the assistant start token but now it feels more of a hack. But maybe it's ok 🤷‍♂️

mlx_lm/server.py
angeloskath: Yeah that is probably the most confusing part here.

The other two arguments, affect the prompt cache. But say you get a batch of 10 requests. Your active kv cache is also significant in size, not just the saved ones.

This attempts to take that into account. Basically when we fetch a cache and put it in batch, we limit the saved prompt caches such that their size plus active kv caches in the batch are less than this number.

awni: Makes sense! The naming is a bit confusing -> people might not know how to set this or the above argument.

Do you think it's reasonable to get rid of the max bytes on the prompt cache and just use this instead?

angeloskath: I made it one. Indeed it would be too confusing.

PR #911, opened 2026-02-19
Better caching in the server

angeloskath:
This aims to solve the problem of non-trimmable caches resulting in no KV cache, very common with linear attention and sliding window attention models.

The standard agentic interaction is kind of the following
1. System and user prompt
2. Thinking and generation
3. Tool
4. Repeat 2-3 until done
5. Repeat 1 (system prompt reminder)

For 2 & 3 we want to keep the thinking in the cache for 1 we don't because typically the thinking will be removed from the history.

This PR makes the following changes (they only apply to the batched generation for now):
- Requests that end in a user message are treated specially and saved as a checkpoint
- The cache does not extract on fetch but removes prefixes of trimmable caches on insert
- Checkpoint caches have higher priority than normal caches so we can have many 2-3 steps (above) with cache and without evicting the cache from step 1.

It isn't perfect and may require a bigger refactoring to improve but at least for now we can utilize the cache for non trimmable KVs as well.

I will add some benchmarks and leave it for comments/reviews before merging.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -61,8 +61,9 @@ class DummyModelProvider:
 
 
 class MockCache:
-    def __init__(self, value):
+    def __init__(self, value, is_trimmable: bool = True):
         self.value = value
+        self._is_trimmable = is_trimmable
 
     @property
     def nbytes(self):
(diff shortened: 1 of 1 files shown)


nastya236:
Thank you! Looking into it now!

PR #915, opened 2026-02-22
fix: avoid extra _step/async_eval call when n reaches max_tokens

### LINKED ISSUE #226: mlx_lm/generate extra _step/async_eval call
Isn't there one extra _step/async_eval call here when we hit max_tokens?

https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/generate.py#L404-L419

Consider a situation when max_tokens == 1

First, do this before the loop:

`    mx.async_eval(y, logprobs)`

This should kick off the eval of the only token we need to generate.

On the first iteration of while loop, we'll kick off another one:

```
        if n != max_tokens:
            next_y, next_logprobs = _step(y)
            mx.async_eval(next_y, next_logprobs)
```

as n is still 0 at this point.

Then we'll correctly yield y, increment n, and exit the loop on the next iteration. So, we always yield correct sequence of tokens, but seem to start async_eval for one more?

siddhipowar:
Fixes #226

When max_tokens == 1, the condition `n != max_tokens` evaluates 
to True on the first iteration (n=0), causing an unnecessary 
_step/async_eval call for a token that will never be used.

Changing the condition to `n + 1 < max_tokens` ensures the next 
token is only prefetched when it will actually be needed.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -451,7 +451,7 @@ def generate_step(
     mx.async_eval(y, logprobs)
     n = 0
     while True:
-        if n != max_tokens:
+        if n + 1 < max_tokens:
             next_y, next_logprobs = _step(y)
             mx.async_eval(next_y, next_logprobs)
         if n == 0:


awni:
See discussion in https://github.com/ml-explore/mlx-lm/issues/226

PR #917, opened 2026-02-22
Add 'mx.clear_cache()' to piecewise prompt processing in server.

N8python:
Prompts that were being newly added to the server didn't have cache-clearing enabled while being processed. This lead to massive memory hang.

For prompts of length ~16K on GLM 4.7 Flash 6bit, this took peak memory from 50+GB to a more reasonable 28GB.


ivanfioravanti:
It works! A single line of code with a mega impact.

N8python:
Can we merge?

PR #924, opened 2026-02-23
Add tokens to eval to avoid large graphs when they are not used

awni:
The `batch.tokens` arrays are not always used and this can create some big graphs for long generations.


awni:
Benchmarked a couple models and did not observe any change in generation speed.

PR #923, opened 2026-02-23
Hybrid cache for Qwen3.5

dexloom:
### Summary

This pull request includes new features, refactoring, and fixes to enhance the functionality of LRUPromptCache, particularly for branching conversation histories and prompt chaining.

### Changes Made
- **Checkpointing**: Periodic snapshots added to preserve cache states for branching use cases.
- **Boundary Caching**: Cache entries can now be stored at message boundaries for shared cache usage.
- **In-Place Updates**: LRU strategy refactored to reduce memory overhead with mutable cache objects.
- **Cache Persistence**: Fixes to ensure reused prefix prompts remain in the cache for hybrid models.
- **Cache Key Separation**: Improved handling of cache keys for accurate prompt chaining.

### Checklist
- [ ] Unit tests have been added/updated
- [ ] Documentation has been updated (if applicable)
- [ ] Changes match coding standards and guidelines

PR #929, opened 2026-02-25
Fix hybrid cache checkpoints for short conversations

omui-ai:
## Context

Builds on #923 by @dexloom — found three issues while testing that PR's checkpoint approach on real hardware (Qwen3.5-35B-A3B, M1 Max 64GB, multi-turn chat).

## What's wrong

With #923 as-is, multi-turn cache hit rate is **0% across all turns** for conversations under 8k tokens. Three issues combine to cause this:

### 1. No checkpoints for short conversations

`_find_checkpoint_positions` iterates `range(checkpoint_interval, tokens_len, checkpoint_interval)`. When `tokens_len < 8192` (i.e. most early conversation turns), the range is empty — no checkpoints are created. Without a checkpoint, the next turn can't reuse the cache for hybrid models because `can_trim_prompt_cache` returns `False` and the "longer" match path in `fetch_nearest_cache` is skipped.

**Fix:** Always checkpoint at the last message boundary, then add periodic checkpoints for long contexts.

### 2. No boundaries for single-message conversations

`_compute_message_boundaries` uses `range(1, len(messages))` which is empty when there's only one message. The first turn of every conversation never gets a boundary, so no checkpoint can be created even if issue 1 were fixed.

**Fix:** `range(1, len(messages) + 1)`.

### 3. All-or-nothing trim in snapshots

`_create_checkpoint_snapshot` uses `can_trim_prompt_cache(snapshot)` which returns `False` if *any* layer is non-trimmable. For hybrid models (KVCache + ArraysCache), this means KVCache layers don't get trimmed in snapshots either, leaving stale KV entries from the generation prompt.

**Fix:** Fall back to per-layer trim — KVCache layers get cleaned while ArraysCache layers stay as-is.

## Test results

Qwen3.5-35B-A3B (30 DeltaNet + 10 attention layers), 8-bit MLX, served via OpenAI-compatible API:

| Turn | Prompt tokens | Cached | Remaining | Hit rate |
|------|--------------|--------|-----------|----------|
| 1    | 17           | 0      | 17        | 0%       |
| 2    | 32           | 12     | 20        | 37%      |
| 3    | 47           | 27     | 20        | 57%      |

Savings scale upward as the conversation grows — only new messages need processing each turn.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -217,31 +217,37 @@ class LRUPromptCache:
     def _find_checkpoint_positions(self, tokens_len, boundary_positions=None):
         """Find positions where we should create checkpoint snapshots.
 
-        Creates checkpoints at message boundaries that fall near checkpoint_interval
-        boundaries (e.g., around 8k tokens). This ensures we snapshot at natural
-        conversation break points without creating too many snapshots.
+        Always creates a checkpoint at the last message boundary so the next
+        turn can resume from it.  For long contexts, also creates periodic
+        checkpoints at ~checkpoint_interval spacing aligned to message
+        boundaries.
 
         Returns list of positions (not including 0 or tokens_len).
         """
-        positions = set()
-
         if not boundary_positions:
             return []
 
-        # Find message boundaries near checkpoint intervals
-        # For each 8k interval, find the closest message boundary
-        last_checkpoint = 0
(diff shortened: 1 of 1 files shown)


angeloskath:
I will close this due to #911 but feel free to open another that doesn't cover it.

PR #930, opened 2026-02-25
fix: catch specific exceptions instead of bare except in to_json

haosenwang1018:
Replace bare `except:` with `except (UnicodeEncodeError, ValueError):` in `deepseek_v32.py`'s `to_json()`. The `json.dumps(ensure_ascii=False)` fallback only needs to handle encoding-related exceptions, not `KeyboardInterrupt` or `SystemExit`.


haosenwang1018:
Closing per review backlog policy; can resubmit in smaller chunks.

PR #931, opened 2026-02-25
Cast to model dtype on load

pcuenca:
When running inference from native transformers safetensors (without converting), casting to the model dtype is not performed. This may cause inadvertently slow inference when a float32 layer pollutes the forward pass.

Consider the recently released `Qwen/Qwen3.5-35B-A3B`:
```
mlx_lm.generate --model Qwen/Qwen3.5-35B-A3B --prompt "Why is the sky blue?" -m 200
```

* **Before**: ~3tps
* **After**: ~68 tps - consistent with the same model converted locally using `bfloat16`, without quantization.

This PR applies casting to both code paths.

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -7,13 +7,13 @@ from typing import Callable, Optional, Union
 import mlx.nn as nn
 
 from .utils import (
+    MODEL_CONVERSION_DTYPES,
     cast_model_dtype,
     dequantize_model,
     load,
     quantize_model,
     save,
     upload_to_hub,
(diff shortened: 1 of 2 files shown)


angeloskath:
I think the correct solution here is actually to fix the `Qwen3NextRMSNormGated` to cast `x` back to the input's dtype. 

Casting a model's parameters on load is generally not very compatible with models shipping with different precisions per layer/weight.

Let me know if you want to fix it or whether I should go ahead and fix it myself.

angeloskath:
I 'll close in favor of #951 let me know if there is any other issue.

PR #933, opened 2026-02-27
feat: add context_length to /v1/models response

lichengzhe:
## Summary

Include `context_length` in the `/v1/models` endpoint response by reading `max_position_embeddings` from each model's `config.json`.

## Motivation

Many OpenAI-compatible API gateways and proxy tools (e.g. [one-api](https://github.com/songquanpeng/one-api), [new-api](https://github.com/Calcium-Ion/new-api)) read `context_length` from the `/v1/models` response to automatically determine a model's context window size. Without this field, users must manually configure context limits for each model, which is error-prone and tedious.

Related: #966 (KeyError on context_length)

## Changes

- Added a `get_context_length()` helper that reads `max_position_embeddings` from a HuggingFace cached model's `config.json`
- Updated the HF cache model listing to include `context_length` when available
- Updated the local model path listing to also read and include `context_length`
- Handles models that nest the value under `text_config` (e.g. some vision-language models)
- Gracefully falls back (omits the field) if `config.json` is missing or unreadable

## Example response

```json
{
  "object": "list",
  "data": [
    {
      "id": "RepublicOfKorokke/Qwen3.5-35B-A3B-mlx-lm-mxfp4",
      "object": "model",
      "created": 1740000000,
      "context_length": 262144
    }
  ]
}
```

## Testing

Tested locally with `Qwen3.5-35B-A3B-mlx-lm-mxfp4` — confirmed `context_length: 262144` appears in the response.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1712,6 +1712,22 @@ class APIHandler(BaseHTTPRequestHandler):
             file_names = {f.file_path.name for f in repo.refs["main"].files}
             return all(f in file_names for f in files)
 
+        def get_context_length(repo):
+            """Read max_position_embeddings from model config.json."""
+            try:
+                for f in repo.refs["main"].files:
+                    if f.file_path.name == "config.json":
+                        with open(f.file_path) as fh:
+                            config = json.load(fh)
+                        for key in ("max_position_embeddings",):
+                            if key in config:
+                                return config[key]
+                            if "text_config" in config and key in config["text_config"]:
+                                return config["text_config"][key]
+            except Exception:
+                pass
+            return None
+
         # Scan the cache directory for downloaded mlx models
         hf_cache_info = scan_cache_dir()
         downloaded_models = [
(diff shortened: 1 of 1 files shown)


mm65x (commented):
the motivation makes sense — a lot of API gateways expect this field.

a few suggestions:

1. the config.json reading logic is duplicated between `get_context_length(repo)` and the local model path branch. could extract a small helper that takes a config dict and returns the context length:
```python
def _context_length_from_config(config):
    val = config.get("max_position_embeddings")
    if val is None:
        val = config.get("text_config", {}).get("max_position_embeddings")
    return val
```

2. `for key in ("max_position_embeddings",):` is a loop over one element — could just use the key directly, it's simpler to read.

3. instead of iterating `repo.refs["main"].files` to find `config.json`, you can use `repo.refs["main"].snapshot_path / "config.json"` directly — `probably_mlx_lm` already validated the file exists.

4. minor: `context_length` isn't part of the official OpenAI model object spec (`id`, `object`, `created`, `owned_by`). might be worth noting as a non-standard extension in a comment or the PR description, since some strict clients could trip on it.

mlx_lm/server.py
```diff
@@ -1712,33 +1712,63 @@ def probably_mlx_lm(repo):
             file_names = {f.file_path.name for f in repo.refs["main"].files}
             return all(f in file_names for f in files)
 
+        def get_context_length(repo):
+            """Read max_position_embeddings from model config.json."""
+            try:
+                for f in repo.refs["main"].files:
```
mm65x, line 1718: you could use `repo.refs["main"].snapshot_path / "config.json"` directly here instead of iterating all files - `probably_mlx_lm` already validated config.json exists.

mlx_lm/server.py
```diff
@@ -1712,33 +1712,63 @@ def probably_mlx_lm(repo):
...
+                    if f.file_path.name == "config.json":
+                        with open(f.file_path) as fh:
+                            config = json.load(fh)
+                        for key in ("max_position_embeddings",):
```
mm65x, line 1722: this loop only has one element, so it's effectively just an if statement. could use the key directly:

```suggestion
                        if "max_position_embeddings" in config:
                            return config["max_position_embeddings"]
                        if "text_config" in config and "max_position_embeddings" in config["text_config"]:
                            return config["text_config"]["max_position_embeddings"]
```

mlx_lm/server.py
```diff
@@ -1712,33 +1712,63 @@ def probably_mlx_lm(repo):
...
-            for repo in downloaded_models
-        ]
+            ctx_len = get_context_length(repo)
+            if ctx_len is not None:
+                model_info["context_length"] = ctx_len
+            models.append(model_info)
 
         if self.response_generator.cli_args.model:
             model_path = Path(self.response_generator.cli_args.model)
             if model_path.exists():
```
mm65x, line 1752: this is the same config lookup logic as `get_context_length` above. could extract a small helper to deduplicate:

```python
def _context_length_from_config(config):
    val = config.get("max_position_embeddings")
    if val is None:
        val = config.get("text_config", {}).get("max_position_embeddings")
    return val
```

then both call sites just load the config dict and call this.

zcbenz:

PR #934, opened 2026-02-27
feat: add --kv-bits CLI args for server KV cache quantization

lichengzhe:
## Summary

Expose KV cache quantization parameters as server CLI arguments: `--kv-bits`, `--kv-group-size`, and `--quantized-kv-start`.

## Motivation

`generate_step()` already supports KV cache quantization via `kv_bits`, `kv_group_size`, and `quantized_kv_start` parameters, and the CLI `mlx_lm.generate` exposes them. However, `mlx_lm.server` does not — users serving models via the OpenAI-compatible API have no way to enable KV cache quantization.

This is particularly useful for long-context serving scenarios where KV cache memory is the bottleneck. For example, serving a 35B MoE model with 262K context on a 128GB Mac can easily exhaust memory with multiple concurrent long-context requests. KV cache quantization (4-bit or 8-bit) significantly reduces memory per cache entry.

## Changes

- Added `--kv-bits` (choices: 4, 8), `--kv-group-size` (default: 64), `--quantized-kv-start` (default: 0) to the server argument parser
- Pass these through to `stream_generate()` in `_serve_single`

## Usage

```bash
mlx_lm.server --model my-model --kv-bits 4
mlx_lm.server --model my-model --kv-bits 8 --kv-group-size 128
```

## Benchmark

Tested with Qwen3.5-35B-A3B (mxfp4) on Mac Mini M4 Pro 64GB:

| Config | Aggregate t/s | Per-request avg t/s |
|--------|--------------|-------------------|
| Baseline (no quantization) | 84.5 | 69.3 |
| KV 4-bit | 79.0 | 66.2 |

Speed is comparable (slight overhead expected). The main benefit is memory savings for long-context workloads.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -983,6 +983,11 @@ class ResponseGenerator:
             logging.info(f"We have {ncaches} kv caches that take {nbytes/1e9:.2f} GB")
 
             # Process the prompt and generate tokens
+            kv_kwargs = {}
+            if self.cli_args.kv_bits is not None:
+                kv_kwargs["kv_bits"] = self.cli_args.kv_bits
+                kv_kwargs["kv_group_size"] = self.cli_args.kv_group_size
+                kv_kwargs["quantized_kv_start"] = self.cli_args.quantized_kv_start
             for gen in stream_generate(
                 model=model,
                 tokenizer=tokenizer,
@@ -994,6 +999,7 @@ class ResponseGenerator:
                 draft_model=draft_model,
                 num_draft_tokens=args.num_draft_tokens,
                 prompt_progress_callback=progress,
+                **kv_kwargs,
             ):
                 rqueue.put(
                     Response(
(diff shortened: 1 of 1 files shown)


mlx_lm/server.py
```diff
@@ -983,6 +983,11 @@ def progress(tokens_processed, tokens_total):
...
 
             # Process the prompt and generate tokens
+            kv_kwargs = {}
+            if self.cli_args.kv_bits is not None:
+                kv_kwargs["kv_bits"] = self.cli_args.kv_bits
+                kv_kwargs["kv_group_size"] = self.cli_args.kv_group_size
+                kv_kwargs["quantized_kv_start"] = self.cli_args.quantized_kv_start
             for gen in stream_generate(
                 model=model,
                 tokenizer=tokenizer,
```
mm65x, line 993: since `generate_step` already treats `kv_bits=None` as "no quantization", you could skip the conditional dict and just pass the args directly:

```python
for gen in stream_generate(
    ...
    kv_bits=self.cli_args.kv_bits,
    kv_group_size=self.cli_args.kv_group_size,
    quantized_kv_start=self.cli_args.quantized_kv_start,
):
```

removes the `kv_kwargs` dict and `**kv_kwargs` unpack. not a blocker though, current approach works fine.

mm65x (approved):
looks good, clean parity addition with `mlx_lm.generate`. `stream_generate` accepts `**kwargs` and forwards to `generate_step` which already handles `kv_bits=None` as no-op, so this works end to end.

zcbenz:

PR #935, opened 2026-02-27
fix: `convert()` uses incorrect defaults for quantization mode

spicyneuron:
Quantization modes like mxfp4 expect certain bits and group sizes. These are correctly set by default when running the CLI: `mlx_lm.convert --q-mode mxfp4` gets 4 bits and group size 32.

But from Python, calling `convert(model_path, quantize=True, q_mode="mxfp4")` without explicitly setting group size incorrectly applies 64 because its signature hardcodes affine defaults.

**Minimal Reproduction**

```python
import json, tempfile
from mlx_lm.convert import convert

    path = f"{d}/out"
    convert("Qwen/Qwen3-0.6B", mlx_path=path, quantize=True, q_mode="mxfp4")
    config = json.load(open(f"{path}/config.json"))
    print(f"group_size={config['quantization']['group_size']}  (expected 32)")

# main: ValueError: mxfp4 quantization requires group size 32 but got 64
# fixed: group_size=32
```

**Solution**

This PR defaults `convert()`'s defaults to `None` and resolves defaults from `q_mode` instead. This unifies the CLI and Python paths. Tests included.

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -9,7 +9,6 @@ import mlx.nn as nn
 from mlx.utils import tree_map_with_path
 
 from .utils import (
-    QUANT_MODE_DEFAULTS,
     dequantize_model,
     load,
     quantize_model,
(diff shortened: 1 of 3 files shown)


angeloskath:
Thanks for catching this but the only thing that is needed is to remove the values and default `None`. The rest is handled by `mx.quantize` automatically anyway.

Basically this should be a 2 line change we don't need the rest. Lmk know if you want to adjust it.

spicyneuron:
There was a separate issue with default ordering, but on second look, it might only be relevant in context of the `--q-override` PR.

Either way, this one has been simplified.

PR #936, opened 2026-02-27
fix(model definition): granitemoehybrid rope_parameters and time_step_limit

mshamrai:
# Support Granite configs from `transformers >= 5.0.0`

## Problem

Recent `transformers` updates changed the Granite config format, breaking MLX conversion:

1. `rope_theta` moved into:
   ```json
   "rope_parameters": {
     "rope_theta": 10000,
     "rope_type": "default"
   }
2. Default for `time_step_limit` was changed. Now it serialises as:
```json
 "time_step_limit": [
    0.0,
    {
      "__float__": "Infinity"
    }
  ],
```
The `"__float__": "Infinity"` cannot be deserialized. 

## Fix
1. Read `rope_theta` from `rope_parameters` when present.
2. Normalize `{ "__float__": "Infinity" } → float("inf")` during init.
3. Fully backward compatible with older configs.

### Links
https://github.com/huggingface/transformers/blob/main/src/transformers/models/granitemoehybrid/configuration_granitemoehybrid.py

### DIFF
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -135,3 +135,12 @@ def scaled_dot_product_attention(
             mask=mask,
             sinks=sinks,
         )
+
+
+def deserialize_float(x):
+    if isinstance(x, dict) and "__float__" in x:
+        if x["__float__"] == "Infinity":
+            return float("inf")
+        if x["__float__"] == "-Infinity":
+            return float("-inf")
+    return x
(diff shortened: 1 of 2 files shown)


angeloskath:
Sorry for getting to this late. Could you share an example where this breaks and is fixed with this PR?

angeloskath:
#953 is doing a more extensive config migration. I will close this and you can contribute to #953 if something is missing from there.

Thanks.

PR #937, opened 2026-02-27
lora: add --epochs training argument (with --iters compatibility)

dmitryryabkov:
## Summary
This PR adds an `--epochs` parameter to `mlx_lm.lora` so users can specify training length in dataset passes, while preserving existing `--iters` behavior.

## Why
`--iters` is sensitive to dataset size and batch size, which makes run configuration less intuitive and easier to misconfigure. `--epochs` provides a more stable and user-friendly way to express training duration.

## What changed
- Added `--epochs` to `mlx_lm.lora` CLI.
- Added validation and resolution logic:
  - `--epochs` and `--iters` are mutually exclusive.
  - `--epochs` is converted to iterations via:
    - `steps_per_epoch = len(train_set) // batch_size`
    - `iters = ceil(epochs * steps_per_epoch)`
  - Clear input validation for invalid values and too-small datasets.
- Kept trainer internals iteration-based for minimal code churn.
- Preserved backward compatibility:
  - If neither option is provided, behavior remains equivalent to prior default (`iters=1000`).
- Updated docs/examples:
  - `mlx_lm/LORA.md`
  - `mlx_lm/examples/lora_config.yaml`
- Added unit tests for new resolution/validation behavior in `tests/test_finetune.py`.

## Notes
- This change does not alter existing batch construction semantics (remainder samples are still dropped by current iterator behavior).

## Validation
- `python -m unittest tests/test_finetune.py` passed.
- `pre-commit run --all-files` passed.

### DIFF
--- a/mlx_lm/LORA.md
+++ b/mlx_lm/LORA.md
@@ -60,8 +60,12 @@ mlx_lm.lora \
     --model <path_to_model> \
     --train \
     --data <path_to_data> \
-    --iters 600
+    --epochs 3
 ```
+
+You can use either `--epochs` or `--iters` to control training length. They
+are mutually exclusive. If neither is specified, `mlx_lm.lora` defaults to
+`--iters 1000` for backward compatibility.
 
 To fine-tune the full model weights, add the `--fine-tune-type full` flag.
 Currently supported fine-tuning types are `lora` (default), `dora`, and `full`.
--- a/mlx_lm/examples/lora_config.yaml
+++ b/mlx_lm/examples/lora_config.yaml
@@ -28,6 +28,10 @@ num_layers: 16
 # Minibatch size.
 batch_size: 4
 
+# Use either epochs or iters (mutually exclusive).
+# Epochs to train for.
+# epochs: 3
+
 # Iterations to train for.
 iters: 1000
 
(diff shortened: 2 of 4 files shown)


angeloskath:
I will close this cause it is already open in #523 . Would you mind commenting there and merging your changes in that one if you think it makes sense? Still not quite sure we need this.

dmitryryabkov:
Thanks @angeloskath for looking!
> I will close this cause it is already open in #523 . Would you mind commenting there and merging your changes in that one if you think it makes sense?

OK, I'll take a look. I see that PR is like 5 month old and is way behind the main branch, so it will take time to get these two PRs on the same page.

> Still not quite sure we need this.

I'll give you a concrete example of a workflow which I used and the lack of this parameter really annoyed me: when fine-tuning it feels reasonable to run through the full data set some number of times (i.e. 2 times). To get the number of iters one needs to get the number of items in the dataset first, and then multiply it by the number of times to run through it. Not super-convenient, but not too bad either. But then, to get the most out of your machine you start playing with parameters like the batch size and the number of gradient accumulation steps. The first one, if changed, affects the number of iterations, while the second one does not. Then on top of it you start adjusting LoRA parameters (i.e. LoRA rank), which affects memory, which may affect the batch size, which affects the number of iterations again. So you have to remember to adjust the number of iterations every time, otherwise you end up with a training run multiple times larger or smaller than the previous one prior to the parameter changes. In my case I almost always forget to do it.
With number of epochs instead of number of iterations it is much easier to adjust all these other parameters while keeping the number of iterations over the data set the same. I'll reiterate this point again in the other branch and tag you once I've consolidated the changes. Stay tuned.

PR #941, opened 2026-03-02
Support KV cache quantization with continuous batching

ochafik:
## Summary

- Add `QuantizedKVCache.merge()` — dequantizes to float, delegates to `BatchKVCache.merge()`, enabling quantized caches to participate in batch generation
- Add `quantize_config` parameter to `BatchKVCache.extract()` — extracted caches can be re-quantized as `QuantizedKVCache` for memory-efficient LRU storage
- Wire `kv_bits`/`kv_group_size` through `BatchGenerator` so callers can opt into quantized cache storage
- Add `QuantizedKVCache.size()` for consistency with other cache types

Currently `QuantizedKVCache` and `BatchGenerator` are mutually exclusive (the `is_batchable` guard rejects `kv_bits`). This PR adds the missing `merge()` support so they can work together. The approach keeps batch computation in float (short-lived, needed for padding/concatenation) and only quantizes on extract (long-lived LRU storage), which is where memory savings matter most.

## Test plan

- [x] 5 new tests covering merge, extract with quantization, roundtrip, empty cache merge, and `size()`
- [x] All 25 tests pass (`tests/test_prompt_cache.py`)
- [x] No changes to existing test behavior

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -879,8 +879,8 @@ class Batch:
         for c, o in zip(self.cache, other.cache):
             c.extend(o)
 
-    def extract_cache(self, idx):
-        return [c.extract(idx) for c in self.cache]
+    def extract_cache(self, idx, quantize_config=None):
+        return [c.extract(idx, quantize_config=quantize_config) for c in self.cache]
 
 
 def _make_cache(model, left_padding, max_kv_size):
@@ -890,7 +890,7 @@ def _make_cache(model, left_padding, max_kv_size):
     """
 
     def to_batch_cache(c):
-        if type(c) is KVCache:
+        if type(c) is KVCache or type(c) is QuantizedKVCache:
             return BatchKVCache(left_padding)
         elif isinstance(c, ArraysCache):
             c.left_padding = mx.array(left_padding)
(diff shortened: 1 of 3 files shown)


angeloskath:
This is interesting albeit a bit weird.

Basically all the computation is done in fp16 but the storage is done quantized. Could be useful but generally speaking there should be a BatchQuantizedKVCache that also does the computation quantized a) so that perf doesn't degrade after every turn b) so that long context (largely memory bound) is faster.

adurham:
+1 on the underlying need — we're hitting the exact same `AttributeError` in production on a multi-node exo cluster running long-context inference (`QuantizedKVCache` + `BatchGenerator` event-log replay). Carrying a near-identical `merge()` patch as a fork bandaid.

Re @angeloskath's redesign comment: happy to help drive a `BatchQuantizedKVCache` if this PR has stalled. We have an in-flight quantized SDPA kernel on the mlx side (operates on packed K/V, no per-call dequantization) which would back the "compute stays quantized" half with a real kernel rather than a fallback. Would land naturally as a two-PR series — mlx kernel → `BatchQuantizedKVCache` using it.

Happy to either contribute on top of this PR or open a follow-up, whichever the maintainers prefer.

zcbenz:

PR #943, opened 2026-03-02
Proposal: `--prefill-step-size` as cmd line argument for speed/memory usage trade-off

Abioy:
Hi mlx,

Proposal: a new argument to balance infer speed and memory usage.

I'm running local llm with mlx-lm on my M2 Pro 32GB. MLX has a better infer speed comparing with llama.cpp, but use much more memory.
When the prompt length comes to 15K~20K ( with Qwen3.5-35B-A3B-4bit ), MLX would run out all memory and cause system stuck or crash almost every time even I limit the prompt cache size. How ever, llama.cpp could run in such case stably.

I do some benchmarks and found that the default `prefill_step_size`, 2048 in fact, might not be the best choice on all cases. Peak memory increases much faster then infer speed does with `prefill_step_size` from 512 to 2048. When I change the prefill_step_size to 1024, I can run the same LLM with 20K ctx on MLX as llama.cpp does, but faster.  No OOM and crash happens. 

So I think a new argument for `prefill step size` might be useful.

Benchmarks:  (dots with step size from 512 to 1024 then 2048 left to right)
1. pp10240<br/>
image<br/>
2. pp20480<br/>
image<br/>
3. pp30720<br/>
image<br/>

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -995,7 +995,7 @@ class ResponseGenerator:
                 draft_model=draft_model,
                 num_draft_tokens=args.num_draft_tokens,
                 prompt_progress_callback=progress,
-                prefill_step_size=self.model_provider.cli_args.prefill_step_size,
+                prefill_step_size=self.cli_args.prefill_step_size,
             ):
                 rqueue.put(
                     Response(


Abioy:
@angeloskath @awni PTAL, thanks

angeloskath:
Yep that is definitely useful. Thanks!

Fix the lint errors if you want and then I 'll merge.

Abioy:
@angeloskath done and have tested it locally, thanks

angeloskath:
Could you also fix the test?

Abioy:
> Could you also fix the test?

fixed.

PR #947, opened 2026-03-03
Add --served-model-name flag to server

otarkhan:
## Summary

- Add `--served-model-name` CLI flag to the server, allowing operators to configure a custom model name for the API (similar to vLLM's `--served-model-name`)
- When set, the served name appears in `/v1/models` listings and is accepted in request bodies alongside the original model path
- When not set, behavior is fully backward compatible

## Details

The implementation leverages the existing `default_model_map` aliasing mechanism in `ModelProvider`. The served name is registered as an additional alias that resolves to `"default_model"`, so requests using either the served name or the original `--model` path work seamlessly.

**`/v1/models` behavior:**
- With `--served-model-name`: the served name is listed as the model id (replaces the raw path)
- Without: existing behavior preserved; also fixes a minor gap where HF repo IDs (non-local paths) were not listed

## Test plan

- [x] New `TestServedModelName` test class with 3 tests:
  - `/v1/models` endpoint lists the served name
  - Completions using the served name succeed and echo it back in the response
  - Backward compatibility: `"default_model"` still works as before
- [x] All 17 existing + new tests pass

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -490,6 +490,10 @@ class ModelProvider:
         self.default_model_map = {}
         if self.cli_args.model is not None:
             self.default_model_map[self.cli_args.model] = "default_model"
+            if getattr(self.cli_args, "served_model_name", None) is not None:
+                self.default_model_map[self.cli_args.served_model_name] = (
+                    "default_model"
+                )
             self.load(self.cli_args.model, draft_model_path="default_model")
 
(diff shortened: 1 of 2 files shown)


otarkhan:
@angeloskath If you have a couple of minutes for this small change. This change should make it much easier to connect the server with clients

zcbenz:

PR #950, opened 2026-03-05
[Bugfix] RotatingKVCache negative dimension in _update_in_place

ricky-chaoju:
## Summary

Fix `ValueError: [full] Negative dimensions not allowed` in `RotatingKVCache._update_in_place`
when the cache offset exceeds `max_size`.

## Cause

`new_size` was computed as:

min(self.step, self.max_size - prev)

where `prev` is `self.offset`.

After the cache rotates, `offset` keeps increasing beyond `max_size`, which makes
`self.max_size - prev` negative and leads to the negative dimension error.

## Fix

Compute `new_size` using the current buffer size instead of `offset`:

current_size = 0 if self.keys is None else self.keys.shape[2]
new_size = min(self.step, self.max_size - current_size)

This prevents negative dimensions once the rotating cache has wrapped.

The fix applies to both:
- `RotatingKVCache`
- `QuantizedRotatingKVCache`

This occurs when serving models with sliding window attention
(e.g. `gpt-oss-20b`) under multi-turn conversation load via `vllm-metal`.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -473,7 +473,8 @@ class RotatingKVCache(_BaseCache):
             prev >= self.keys.shape[2] and self.keys.shape[2] < self.max_size
         ):
             v_head_dim = values.shape[3]
-            new_size = min(self.step, self.max_size - prev)
+            current_size = 0 if self.keys is None else self.keys.shape[2]
+            new_size = min(self.step, self.max_size - current_size)
             k_shape = (B, n_kv_heads, new_size, k_head_dim)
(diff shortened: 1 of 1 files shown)


angeloskath:
Do you mind sharing a reproduction of the error? I don't think it happens in mlx-lm server for instance...

ricky-chaoju:
You're right . after further investigation, this doesn't reproduce in mlx-lm's own server. The issue is triggered by vllm-metal's batched decode path, which calls BatchRotatingKVCache.extract and can produce a cache where offset > max_size but keys.shape[2] < max_size. On the next decode step, _update_in_place computes a negative new_size.
Thanks!

PR #951, opened 2026-03-06
Ensure normalization does not promote to fp32

angeloskath:
Fixes the issue identified in #931 and also matches more closely the implementation in HF transformers by doing the gating in fp32.


pcuenca:
Awesome, thank you!

PR #952, opened 2026-03-06
Fuse gate/up expert projections in SwitchGLU (1 fewer gather per layer)

BurntToastGPT:
# Fuse gate/up expert projections in SwitchGLU (1 fewer gather per layer)

## Problem

`SwitchGLU.__call__` issues two `gather_qmm` (or `gather_mm`) calls for `gate_proj`
and `up_proj` with identical input tensors and expert indices. Each call dispatches a
separate Metal kernel and reads the same input activations from DRAM.

## Solution

Concatenate `gate_proj` and `up_proj` weights along the output dimension (axis=1),
issue a single `gather_qmm`/`gather_mm` call, and split the result at the midpoint.

The change is confined to `SwitchGLU`: a new `_fuse_projections()` method builds a
cached fused weight tensor, and `__call__` dispatches to the fused path during
inference.

**Design decisions:**

- **Original submodules are preserved.** `gate_proj` and `up_proj` remain in the
  module tree. The fused cache is stored in private (`_`-prefixed) attributes,
  does not change parameter keys; `gate_proj`/`up_proj` remain the serialized
  parameters. Serialization, LoRA, and training are unaffected.
- **Inference only.** Fusion is skipped when `self.training` is True. The original
  two-call path is always used during training.
- **Lazy, one-time.** The cache is built on the first inference-mode forward pass
  and eagerly materialized with `mx.eval()`. Cache construction is idempotent;
  concurrent calls may build the cache more than once, but correctness is unchanged.
- **LoRA safe.** Fusion uses strict `isinstance` checks on `gate_proj` and
  `up_proj`. If LoRA adapters are applied (wrapping modules in `LoRALinear`),
  the type check fails and the layer safely falls back to the unfused two-call
  path. Adapters are never bypassed.
- **Validated.** Quantization config (group_size, bits, mode), shape, and bias
  configuration are checked before fusing. Any mismatch falls back to the original
  path.

## Why this is correct

Per-row quantization means each output row has independent scales and biases.
Concatenating along the output dimension stacks rows without altering any per-row
metadata. The split at the midpoint is a zero-copy slice. Outputs are numerically
equivalent; token-exact on all tested models.

This holds across quantization formats (4-bit affine, 6-bit affine, mxfp4). The
`mode` parameter is forwarded to `gather_qmm` unchanged.

The activation function (`self.activation`) is called after the gather and split,
with the same `(x_up, x_gate)` arguments as the original code. Models with custom
activations (e.g., GPT-OSS's clamped SwiGLU) work without modification.

## Note on Qwen3.5 MoE

Qwen3.5 MoE ships weights as fused `gate_up_proj` tensors. `sanitize()` in
`qwen3_5_moe.py` splits them into separate `gate_proj` and `up_proj` at load time.
Fusing at inference aligns with their native layout.

## Memory

Zero additional steady-state memory. After building the fused tensor via
`mx.concatenate`, the original `gate_proj.weight` and `up_proj.weight` are replaced
with zero-copy views (slices) into the fused tensor. MLX slices share the underlying
buffer, so only one copy of the data exists in memory. The original allocations are
freed when their reference count drops to zero.

Same approach for `scales`, `biases` (quantization), and `bias` (linear) tensors.
`state_dict()` still returns `gate_proj.weight` and `up_proj.weight` (they are real
arrays, just backed by the fused buffer). LoRA can still target `gate_proj` and
`up_proj` as separate submodules.

There is a transient memory spike during the first inference call while both the
original and fused tensors coexist before view replacement. This is bounded by
the size of one gate+up weight pair per layer and lasts for the duration of the
`mx.eval()` call.

## Benchmarks

All benchmarks: `mlx_lm.generate`, argmax decoding, 200 generation tokens.
Hardware: Mac Studio M3 Ultra, 512 GB unified memory. One warmup generation
(20 tokens) was run before each timed series to populate the KV cache allocator
and trigger fusion; warmup results are discarded.

### Headline models (N=10, full statistics)

| Model | Baseline (mean +/- std) | Fused (mean +/- std) | Improvement | p-value |
|-------|:-----------------------:|:--------------------:|:-----------:|:-------:|
| Qwen3-30B-A3B | 108.69 +/- 0.33 tok/s | 118.00 +/- 0.31 tok/s | **+8.6%** | < 0.001 |
| MiniMax M2.5 | 51.55 +/- 0.14 tok/s | 54.18 +/- 0.15 tok/s | **+5.1%** | < 0.001 |

Non-overlapping distributions -- every fused run beats every baseline run on both models.

### Per-layer SwitchGLU microbenchmark (isolated, 100 iterations)

| Model | Per-layer savings | Projected total | End-to-end measured |
|-------|:-----------------:|:---------------:|:-------------------:|
| Qwen3-30B-A3B | 15.3 us (4.8%) | 0.74 ms | 0.7 ms |
| MiniMax M2.5 | 27.0 us (8.0%) | 1.68 ms | 0.9 ms |

Qwen3 projected savings (0.74 ms) match end-to-end measured savings (0.7 ms) within 6%,
confirming the speedup comes from per-layer kernel dispatch reduction.

### Prefill vs decode

Decode benefits 4-5x more than prefill (8.6% vs 2.1% for Qwen3, 5.1% vs 1.5% for M2.5).
Expected: single-token decode is more kernel-dispatch-bound; prefill amortizes dispatch
over batched tokens.

### Cross-model validation (N=3)

| Model | Family | Experts | Quant | MoE Layers | Baseline tok/s | Fused tok/s | Improvement | Correct |
|-------|--------|---------|-------|------------|---------------|------------|-------------|---------|
| Qwen3-30B-A3B | Qwen | 128, top-8 | 4-bit | 48 | 110.6 | 120.4 | +8.9% | Token-exact |
| MiniMax M2.5 (456B) | MiniMax | 256, top-2 | 4-bit | 62 | 51.2 | 54.2 | +5.9% | Token-exact |
| GPT-OSS-120B | GPT-OSS | 128, top-4 | mxfp4 | 36 | 85.4 | 89.6 | +5.0% | Token-exact |
| Qwen3.5-122B-A10B | Qwen | 256, top-8 | 6-bit | 48 | 48.7 | 51.1 | +5.0% | Token-exact |
| OLMoE-1B-7B | OLMoE | 64, top-8 | 4-bit | 16 | 357.7 | 371.3 | +3.8% | Token-exact |
| Qwen3.5-397B-A17B | Qwen | 512, top-10 | 4-bit | 60 | 7.5 | 7.5 | +0.8% | Token-exact |

Tested across 6 models, 4 architecture families, 3 quantization formats. All
token-exact correct, all positive improvement.

The gain scales with the ratio of dispatch overhead to per-token time: models running
at 50-120 tok/s with 36-62 MoE layers see the largest improvement (5-9%).
Bandwidth-bound models (Qwen3.5-397B at 7.5 tok/s) see minimal gain because kernel
dispatch is a negligible fraction of the per-token budget.

## Test plan

- [x] New `tests/test_switch_layers.py` with 10 test cases:
  - Fused quantized output matches unfused reference
  - Fused non-quantized output matches unfused reference
  - Fused quantized with index sorting (indices.size >= 64)
  - Repeated forward passes produce consistent results
  - Non-quantized with linear bias
  - Fusion skipped during training
  - Quantization config mismatch falls back to unfused
  - Asymmetric bias config falls back to unfused
  - Parameter keys unchanged after fusion (state_dict compatibility)
  - Memory not doubled: views point into fused buffer, gate_proj.__call__ works with view weights
- [x] Existing `python -m unittest discover tests/` passes unchanged
- [x] Token-exact match verified on 7 production MoE models (table above)

### DIFF
--- a/mlx_lm/models/switch_layers.py
+++ b/mlx_lm/models/switch_layers.py
@@ -173,7 +173,97 @@ class SwitchGLU(nn.Module):
         self.down_proj = SwitchLinear(hidden_dims, input_dims, num_experts, bias=bias)
         self.activation = activation
 
+    def _fuse_projections(self):
+        """Build a cached fused gate+up weight tensor for inference.
+
+        Concatenates gate_proj and up_proj weights along the output
+        dimension so that one gather_qmm (or gather_mm) dispatch replaces
+        two.  The result is split at the midpoint -- a zero-copy slice.
+
+        The original submodules are kept for state_dict compatibility and
+        training.  The fused cache is only used during inference.
+        """
+        gp = self.gate_proj
+        up = self.up_proj
+
+        if isinstance(gp, QuantizedSwitchLinear) and isinstance(
+            up, QuantizedSwitchLinear
+        ):
+            # Require matching quantization config
+            if (
+                gp.group_size != up.group_size
+                or gp.bits != up.bits
+                or gp.mode != up.mode
+                or gp.num_experts != up.num_experts
+                or gp.input_dims != up.input_dims
+            ):
+                self._fused = False
+                return
+
+            # Require symmetric bias configuration
+            if ("bias" in gp) != ("bias" in up):
+                self._fused = False
+                return
+            if (gp.biases is not None) != (up.biases is not None):
+                self._fused = False
+                return
+
+            self._fused_weight = mx.concatenate([gp.weight, up.weight], axis=1)
+            self._fused_scales = mx.concatenate([gp.scales, up.scales], axis=1)
+
+            self._fused_biases = None
+            if gp.biases is not None:
+                self._fused_biases = mx.concatenate([gp.biases, up.biases], axis=1)
+
+            self._fused_bias = None
+            if "bias" in gp:
+                self._fused_bias = mx.concatenate([gp.bias, up.bias], axis=1)
+
+            arrays = [self._fused_weight, self._fused_scales]
+            if self._fused_biases is not None:
+                arrays.append(self._fused_biases)
+            if self._fused_bias is not None:
+                arrays.append(self._fused_bias)
+            mx.eval(*arrays)
+
+            self._fused_out_dims = gp.output_dims
+            self._fused_group_size = gp.group_size
+            self._fused_bits = gp.bits
+            self._fused_mode = gp.mode
+            self._fused = "qmm"
+
+        elif isinstance(gp, SwitchLinear) and isinstance(up, SwitchLinear):
+            # Require compatible shapes
+            if gp.num_experts != up.num_experts or gp.input_dims != up.input_dims:
+                self._fused = False
+                return
+
+            if ("bias" in gp) != ("bias" in up):
+                self._fused = False
+                return
+
+            self._fused_weight = mx.concatenate([gp.weight, up.weight], axis=1)
+
+            self._fused_bias = None
+            if "bias" in gp:
+                self._fused_bias = mx.concatenate([gp.bias, up.bias], axis=1)
+
+            arrays = [self._fused_weight]
+            if self._fused_bias is not None:
+                arrays.append(self._fused_bias)
+            mx.eval(*arrays)
+
+            self._fused_out_dims = gp.output_dims
+            self._fused = "mm"
+
+        else:
+            self._fused = False
+
     def __call__(self, x, indices) -> mx.array:
+        # Build fused weight cache on first inference-mode forward pass.
+        if not self.training and not hasattr(self, "_fused"):
+            self._fuse_projections()
+
         x = mx.expand_dims(x, (-2, -3))
 
         # When we have many tokens, then sort them to make sure that the access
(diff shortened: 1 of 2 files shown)


angeloskath:
I am gonna close this as it is not the right way to do it. Fusing the gate and up projection is indeed a potential optimization along the lines of fusing the qkv projection matrices and so on. It should not be handled in the SwitchGLU layer but rather the model layer whether the weights are fused or not.

MLX does have the sanitize method that actually converts weights so we could opt for fusing them and moving to a fused SwitchGLU. There is the question of doing this in a backwards compatible manner and also evaluating the effect more thoroughly.

Feel free to start an issue to discuss whether it makes sense and maybe a PR where this happens on a per model basis via sanitize and a fused switchglu.

BurntToastGPT:
Thanks for the feedback, @angeloskath! I'll open an issue to discuss the sanitize based approach and work on a new PR. Appreciate the guidance.

PR #953, opened 2026-03-06
Reformat models for the rope parameter args

Goekdeniz-Guelmez:
for the transformers v5

### DIFF
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -16,7 +16,7 @@ Klear team - Kuaishou Technology's `Klear`, AI21 Lab's `Jamba` IBM's `Granite Mo
 Meituan's `LongCat`, Nvidia's `Nemotron H`, Swiss-AI's `Apertus`, Nikity's `Lille130m`, 
 Alibaba Qwen's `Qwen3Next`, Tele-AI's `TeleChat3`, and Allenai's `OLMoE` and `Olmo 3`; 
 Helped add support for the following model architectures: 
-Alibaba Qwen's `Qwen3 & Qwen3MoE)`; Added support for the following training algorithms: 
+Alibaba Qwen's `Qwen3 & Qwen3MoE), Qwen3.5 & Qwen3.5MoE`; Added support for the following training algorithms: 
 `Full Weight Fine-Tuning`, and the `Muon` optimizer; 
 Added support for the following other features: 
 `Multiple Optimizers to choose for training`, and `reporting training metrics to WandB (Weights & Biases)`.
--- a/mlx_lm/models/Klear.py
+++ b/mlx_lm/models/Klear.py
@@ -1,7 +1,7 @@
 # Copyright © 2025 Apple Inc.
 
 from dataclasses import dataclass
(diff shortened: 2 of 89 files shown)


Goekdeniz-Guelmez:
@angeloskath  since im already on the house keeping, should I make the head dim argument on all models optional too? Maybe make the classes have correct input/output type definitions etc. mlx-lm-lora and other would benefit from it imo.

angeloskath:
Thanks @Goekdeniz-Guelmez that looks like a great start. We can migrate the head dim as well but let's finish the rope parameters migration.

I would make sure that if a head dim is optional the model uses it correctly eg `self.head_dim = args.head_dim or ...`.

Goekdeniz-Guelmez:
@angeloskath agreed! Let's first get rope params merged. Should be mergeable now

angeloskath:
@Goekdeniz-Guelmez removing the defaults tends to break things 😉 Running the tests would save us a round-trip

Goekdeniz-Guelmez:
@angeloskath Done!

angeloskath:
Nice, sorry for the back and forth from my side now but shouldn't the other rope parameters also be moved from `rope_parameters` if that is present? ie `rope_scaling` etc.

Goekdeniz-Guelmez:
no worries! yes, that would make thinks easier

Goekdeniz-Guelmez:
@angeloskath shoudl have all the params in and tests passed too.

Goekdeniz-Guelmez:
will be added the new models and then its mergeable @angeloskath

menisadi:
How does it compare to #1465  ?

Goekdeniz-Guelmez:
the existing model configs use the old version where the rope parameters where defined separately, the new one has them under the "rope_parameters" so every model config needs that update, #1465 this just fixes the registration definition of the tokeniser object itself

dahai80:
Thanks for this refactor - the right direction for transformers v5 `rope_parameters`.

One gap I hit while testing: with `rope_theta: float` left as a **required** field (no default), a config that carries `rope_theta` **only** nested under `rope_parameters` (no top-level `rope_theta`) still fails to construct. `BaseModelArgs.from_dict` filters params to the signature and calls `cls(**params)`, so `__post_init__` never runs:

```
TypeError: ModelArgs.__init__() missing 1 required positional argument: 'rope_theta'
```

Repro (mirrors `from_dict` + the #953 `ModelArgs`):
```python
import dataclasses
@dataclasses.dataclass
class ModelArgs:
    rope_theta: float                 # required, no default (current #953)
    rope_parameters: dict = None
    def __post_init__(self):
        if self.rope_parameters is not None:
            self.rope_theta = self.rope_parameters.get("rope_theta", 100000.0)
ModelArgs(**{"rope_parameters": {"rope_theta": 1000000.0}})   # -> TypeError before __post_init__
```

Making `rope_theta` optional lets `__post_init__` do the lift:
```python
rope_theta: Optional[float] = None
```
This also matches issue #1563 (Qwen3 conversion fails when rope_theta is nested under rope_parameters). Happy to open a follow-up PR if that's cleaner than amending here.

zcbenz:

PR #961, opened 2026-03-07
Fix iterate_batches KeyError by calling process() in dataset __getitem__

### LINKED ISSUE #766: Bug in iterate_batches() and train() in tuner/trainer.py
I was using the `train` function in `mlx-lm/tuner/trainer.py`, which uses the `iterate_batches` function and keeps erroring out at this line:
```
 # Sort by length:
    if isinstance(dataset, CacheDataset):
        len_fn = lambda idx: dataset.itemlen(idx)
    else:
        len_fn = lambda idx: len(dataset[idx][0]) # Error here
```
The error log is as follows:
```
 File "/Users/miniconda3/lib/python3.13/site-packages/mlx_lm/tuner/trainer.py", line 104, in iterate_batches
    idx = sorted(range(len(dataset)), key=len_fn)
  File "/Users/miniconda3/lib/python3.13/site-packages/mlx_lm/tuner/trainer.py", line 103, in <lambda>
    len_fn = lambda idx: len(dataset[idx][0])
                             ~~~~~~~~~~~~^^^
KeyError: 0
```
However, even though I prepare my dataset properly using `create_dataset` in `mlx-lm/tuner/datasets.py`, and tried different format to generate a `TextDataset`, `ChatDataset`, etc., it would not work because `iterate_batches` assumes that the dataset object calls `process` on each item before/within `__getitem__`. 

The error was gone when I added this line to my custom ChatDataset.
```
def __getitem__(self, idx: int):
        return self.process(self._data[idx])  # FIX
```

Not sure if this behaviour is by design since I couldn't find documentation on the best training methods using the MLX API (most were CLI-based), but I needed some more flexibility in specification so I opted to use the former.

ars667:
Fix KeyError: 0 in iterate_batches when using non-cached datasets

`iterate_batches` expects `dataset[idx]` to return processed tuples
`(tokens, offset)`, but `TextDataset`, `ChatDataset`, and
`CompletionsDataset.__getitem__` return raw dicts.

Fixed by wrapping the dataset in `CacheDataset` at the start of
`iterate_batches` if it has a `process` method and is not already
wrapped. This is consistent with how `CacheDataset` already handles
processing and caching, and avoids duplicating logic across dataset
classes.

Fixes #766

### DIFF
--- a/mlx_lm/tuner/datasets.py
+++ b/mlx_lm/tuner/datasets.py
@@ -30,7 +30,7 @@ class TextDataset:
         return (d, 0)
 
     def __getitem__(self, idx: int):
-        return self._data[idx]
+        return self.process(self._data[idx])
 
     def __len__(self):
         return len(self._data)
@@ -77,7 +77,7 @@ class ChatDataset:
             return (tokens, 0)
 
     def __getitem__(self, idx: int):
-        return self._data[idx]
+        return self.process(self._data[idx])
 
     def __len__(self):
         return len(self._data)
@@ -127,7 +127,7 @@ class CompletionsDataset:
         return (tokens, 0)
 
     def __getitem__(self, idx: int):
-        return self._data[idx]
+        return self.process(self._data[idx])
 
     def __len__(self):
         return len(self._data)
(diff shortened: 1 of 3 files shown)


ars667:
Hey @angeloskath, would love to get your thoughts on this approach when you have a moment.

zcbenz:

Issue #790, opened 2026-01-21
GLM 4.7 Flash KV Cache size


TrajansRow:
GLM 4.7 Flash performs great in MLX, but the KV Cache seems to be several times larger per token than I would expect for this model. This prevents 32GB hosts from using more than about 10000 tokens of context, even with a 4-bit quantized model. The vLLM folks recently resolved a similar issue in their project. Is it possible this affects MLX-LM also? https://github.com/vllm-project/vllm/pull/32759

TrajansRow:
Looks like this is addressed in #780 :)

Issue #615, opened 2025-11-16
Feature Request: Add max-kv-size Support to MLX HTTP Server


stperic:
Problem
mlx_lm.generate() supports max_kv_size parameter, but mlx_lm.server doesn't. This prevents users from controlling KV cache size for memory optimization and performance tuning.
Proposed Solution
Add max-kv-size support in one or more ways:
Option 1: Startup Parameter
bashpython -m mlx_lm.server --model <model> --max-kv-size 4096
Option 2: API Request Parameter
pythonresponse = client.chat.completions.create(
    model="local-model",
    messages=[...],
    extra_body={"max_kv_size": 4096}
)
Option 3: Both (Recommended)

Startup parameter sets default
API requests can override per-request

Use Cases

Memory optimization: Set smaller cache on limited RAM systems
Long context: Increase cache for document processing
Mixed workloads: Different cache sizes for different request types

Benefits

Better memory management
Performance tuning flexibility
Feature parity with mlx_lm.generate()
Aligns with other frameworks (vLLM, TGI, mlx-openai-server)

Related

Issue #1170: Context size configuration
Third-party mlx-openai-server already implements --context-length

mabaeyens:
This is already implemented and waiting for review in #1353, which adds `--max-kv-size` to the server alongside the KV quantization flags. Option 1 from your list (startup parameter); no per-request override.

Confirming the gap is still real on current main, in case anyone is unsure whether this aged out: `server.py` has no `--max-kv-size` argument, and `git log -S'max_kv_size' -- mlx_lm/server.py` finds nothing across the file's history. The plumbing underneath exists, `BatchGenerator.__init__` already accepts `max_kv_size`, so what is missing is only the CLI wiring, which is what #1353 does.

Worth watching that PR rather than opening anything new here.

PR #963, opened 2026-03-08
feat: Add importance-aware KV cache compression (CompressedKVCache)

iamadalek:
## Add importance-aware KV cache compression (`CompressedKVCache`)

cc @awni — this touches cache infrastructure in your domain. Would appreciate your input on the design decisions, particularly the compact-before-quantize ordering and the cross-layer coherent eviction approach.

### Problem

KV cache memory grows linearly with context length and becomes the primary memory bottleneck for long-context generation on consumer Apple Silicon hardware. A Qwen3-8B (4-bit) model at 8K context consumes 1,152 MB of KV cache alone — a significant portion of available unified memory on 16-32GB machines.

Users have raised this directly:
- #790 — GLM 4.7 Flash KV cache prevents 32GB hosts from exceeding ~10K tokens of context
- #615 — Feature request for `max-kv-size` to control KV cache memory in the server

Existing `QuantizedKVCache` reduces precision but does not reduce the number of cached tokens. This PR adds a complementary approach: evicting the least important tokens while preserving generation quality.

### Approach

`CompressedKVCache` is a new `_BaseCache` subclass that performs importance-aware token eviction using L2-norm scoring of key vectors, based on the KnormPress paper (Devoto et al., arXiv:2406.11430). The core insight from that paper is that key vectors with the highest L2 norms contribute least to attention and can be safely evicted.

When the cache exceeds a configurable budget, a compaction pass:

1. Computes L2 norms of key vectors, aggregated across GQA heads
2. Protects recent tokens from eviction (configurable window)
3. Applies cross-layer coherent eviction — shared indices across all layers via a single `argsort`, matching the KnormPress finding that layer-independent eviction degrades quality
4. Gathers the retained keys and values in-place

### Key design decisions

- **Subclasses `_BaseCache` directly** — follows the flat hierarchy convention used by `KVCache` and `QuantizedKVCache`
- **Tracks `_physical_idx` separately from `offset`** — RoPE positional encodings are baked into keys at write time, so eviction must not shift positions. `offset` continues to count total tokens seen (for correct RoPE on new tokens), while `_physical_idx` tracks the actual write position in the compacted array
- **`make_mask()` uses physical cache size** — prevents SDPA dimension mismatch after compaction
- **`compact()` before `quantize` ordering in `generate_step`** — eviction should happen on full-precision keys for best scoring accuracy; quantization applies to the survivors
- **Cross-layer coherent eviction** — single `argsort` reduces kernel launches from 2×N_layers to 2 total
- **`to_quantized()` raises `NotImplementedError`** — composing compression with quantization is deferred to a follow-up

### Usage

```bash
# CLI — compress to 50% budget at 8K context
mlx_lm.generate --model mlx-community/Qwen3-8B-4bit \
    --compact-kv-budget 4096 \
    --prompt "your long prompt here"
```

```python
# Python API

cache = make_prompt_cache(model, compact_kv_budget=4096)
generate(model, tokenizer, prompt=prompt, prompt_cache=cache)
```

### Benchmark results (Qwen3-8B-4bit, M4 Max 128GB)

**Memory reduction — 50% at all context lengths:**

| Context | Full Cache | Compressed (50% budget) | Reduction |
|---------|-----------|------------------------|-----------|
| 2K tokens | 288 MB | 144 MB | 50% |
| 4K tokens | 576 MB | 288 MB | 50% |
| 8K tokens | 1,152 MB | 576 MB | 50% |

**Compaction latency (8K tokens, 36 layers):** ~35ms wall-clock per compaction cycle (graph construction + Metal gather ops across 36 layers × 2 arrays). One-time cost per cycle, amortized over subsequent generation steps via hysteresis threshold.

**Quality preservation:** 10/10 prompts produce identical output with and without compression (greedy decoding). Consistent with KnormPress paper findings that 50% compression preserves output quality.

### Files changed

- `mlx_lm/models/cache.py` — `CompressedKVCache` class, `make_prompt_cache()` updated with `compact_kv_budget` parameter
- `mlx_lm/generate.py` — `maybe_compact_kv_cache()` hook in `generate_step`, `--compact-kv-budget` CLI argument
- `tests/test_compressed_cache.py` — 18 unit tests covering compaction, GQA, mask generation, edge cases
- `benchmarks/bench_compressed_cache.py` — Reproducible benchmark script

### Future work (not in this PR)

- Compose with `QuantizedKVCache` (compress then quantize for maximum memory savings)
- Batch generation support
- Server integration (`--compact-kv-budget` flag for `mlx_lm.server`, addressing #615)
- Alternative scoring functions (attention-weight-based, hybrid)

### References

- [A Simple and Effective L2 Norm-Based Strategy for KV Cache Compression](https://arxiv.org/abs/2406.11430) — Devoto et al., 2024

### DIFF
--- a/benchmarks/bench_compressed_cache.py
+++ b/benchmarks/bench_compressed_cache.py
@@ -1,0 +1,216 @@
+# Copyright © 2023-2024 Apple Inc.
+
+"""Benchmark CompressedKVCache on Qwen3-8B-4bit.
+
+Measures:
+1. Memory before/after compaction at 2K/4K/8K contexts
+2. Compaction latency on M4 Max
+3. Extraction quality preservation (coherent responses after eviction)
+"""
+
+import time
+
+import mlx.core as mx
+
+from mlx_lm import generate
+from mlx_lm.generate import maybe_compact_kv_cache
+from mlx_lm.models.cache import CompressedKVCache, KVCache, make_prompt_cache
+from mlx_lm.sample_utils import make_sampler
+from mlx_lm.utils import load
+
+MODEL_PATH = "mlx-community/Qwen3-8B-4bit"
+
+
+def measure_cache_memory(cache):
+    """Return total cache memory in MB."""
+    total_bytes = sum(c.nbytes for c in cache)
+    return total_bytes / (1024 * 1024)
+
+
+def benchmark_memory(model, tokenizer):
+    """Benchmark memory before/after compaction at various context lengths."""
+    print("\n=== Memory Benchmark ===")
+    print(
+        f"{'Context':<12} {'Full (MB)':<14} {'Compressed (MB)':<18} {'Reduction':<12}"
+    )
+    print("-" * 56)
+
+    for ctx_len in [2048, 4096, 8192]:
+        budget = ctx_len // 2
+
+        # Generate a long prompt by repeating tokens
+        base = "The quick brown fox jumps over the lazy dog. "
+        prompt_text = base * (ctx_len // 8)
+        prompt_tokens = tokenizer.encode(prompt_text)[:ctx_len]
+        prompt_arr = mx.array(prompt_tokens)
+
+        # Full cache baseline
+        full_cache = make_prompt_cache(model)
+        model(prompt_arr[None], cache=full_cache)
+        mx.eval([c.state for c in full_cache])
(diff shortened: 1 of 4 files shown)


zcbenz:

PR #966, opened 2026-03-08
Fix SuScaledRoPE input mutation

### LINKED ISSUE #939: SuScaledRoPE mutates x parameter
See https://github.com/ml-explore/mlx-lm/blob/834fac934c4e04de9b3d723e2b9287a2c60cfd4a/mlx_lm/models/rope_utils.py#L60

```python
class SuScaledRoPE(nn.Module):
...
    def __call__(self, x, offset: Union[int, mx.array] = 0):
        x[..., : self.dim] = self._scale * x[..., : self.dim]
        return mx.fast.rope(
```

This ends up mutating the `x` parameter (it updates the interior pointer through the reference).  Per @angeloskath it should be something like:

```python
y = x[...]
y[..., : self.dim] *= self._scale
return f(y)
```

eyupcanakman:
SuScaledRoPE currently mutates the input tensor by scaling the rotary slice in place. Copy the input before applying the scale and add a regression test for the non-mutating behavior.

Fixes #939

### DIFF
--- a/mlx_lm/models/rope_utils.py
+++ b/mlx_lm/models/rope_utils.py
@@ -58,6 +58,7 @@ class SuScaledRoPE(nn.Module):
         self._scale = long_mscale or (1.0 if factor <= 1.0 else default_scale(factor))
 
     def __call__(self, x, offset: Union[int, mx.array] = 0):
+        x = mx.array(x)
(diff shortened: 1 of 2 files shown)


angeloskath:
Thanks for the PR. I merged #1003 that did both cases. Sorry for not being quicker about it.

PR #967, opened 2026-03-08
Fix CompletionsDataset mask_prompt crash

### LINKED ISSUE #786: Problem with CompletionsDataset and mask-prompt
There seems to be a bug CompletionsDataset process with mask-prompt, currently it crashes. 
Probably messages[0] in line 119 is a problem.

eyupcanakman:
CompletionsDataset passes a single message dict to `apply_chat_template()` when `mask_prompt=True`, which crashes prompt masking. Pass the prompt-only message list instead and add a regression test for the prompt boundary.

Fixes #786


angeloskath (approved):
Thanks!

PR #968, opened 2026-03-08
Implement QuantizedRotatingKVCache and fix RotatingKVCache.to_quantized()

Aaryaman3:
`RotatingKVCache.to_quantized()` has always raised `NotImplementedError`.
This means using `--max-kv-size` and `--kv-bits` together crashes at the
first quantization step, making bounded-memory + quantized inference impossible.

## Changes

**`mlx_lm/models/cache.py`**

- `QuantizedRotatingKVCache`: new cache class storing keys/values as
  `[data, scales, biases]` quantized lists. Mirrors `RotatingKVCache`'s
  two-path design — `_update_in_place` for single-token decode (in-place
  writes into a pre-allocated quantized buffer), `_update_concat` for
  multi-token prefill (quantize-then-trim). Implements `state`,
  `meta_state`, `make_mask`, `trim`, `is_trimmable`, `empty`, `nbytes`
  so it is a drop-in wherever `RotatingKVCache` is expected.
- `RotatingKVCache.to_quantized()`: was `NotImplementedError`, now
  constructs a `QuantizedRotatingKVCache` from the existing float buffer,
  preserving `offset` and `_idx` so generation continues from the exact
  same position.

**`tests/test_models.py`**

10 new tests: in-place path with value correctness, concat path with
trimming, sink token preservation across rotations, wraparound behaviour
(`_idx` wraps to `keep` not `0`), `to_quantized()` pre-rotation and on
empty cache (regression for the `state` property crash), `maybe_quantize_kv_cache`
pipeline integration, save/load round-trip, `_temporal_order_q` correctness
after rotation, and key/value isolation.

## How attention works with this cache

`scaled_dot_product_attention` in `base.py` dispatches on `hasattr(cache, "bits")`.
`QuantizedRotatingKVCache` sets `self.bits`, so it routes to
`quantized_scaled_dot_product_attention`, which calls `mx.quantized_matmul`
for both the QK and AV products — a fused kernel that never materialises
float K/V tensors. Both memory and bandwidth savings apply.

## Ordering: quantize-then-rotate, not rotate-then-requantize

Each token is quantized once at write time and never dequantized again.
`_temporal_order_q` and `_trim_q` move `[data, scales, biases]` slices
around as opaque blocks — no arithmetic, no requantization. The in-place
decode path writes a freshly quantized token directly into the correct
buffer slot at `_idx`. The concat path reorders the already-quantized
circular buffer into temporal order before appending new quantized tokens.
There is no dequantize → reorder → requantize round-trip anywhere in the
implementation.

## Benchmarks

Tested on Mistral-7B-Instruct-v0.3-4bit, M-series, `max_kv_size=512`.

**Peak KV cache memory at 1500 tokens**

| Cache | Memory | vs RotatingKVCache |
|---|---|---|
| `KVCache` (fp16) | 198 MB | — |
| `RotatingKVCache` (fp16) | 38 MB | 1× |
| `QuantizedRotatingKVCache` (4-bit) | 8 MB | 4.75× |

**Memory over time (single-token decode, 5000 tokens)**

| Cache | Memory at 5000 tokens |
|---|---|
| `KVCache` (fp16) | ~640 MB (growing) |
| `QuantizedRotatingKVCache` (4-bit) | ~38 MB (flat) |

`QuantizedRotatingKVCache` reaches its plateau at `max_kv_size` and stays
there regardless of generation length. `KVCache` grows without bound.

**Output quality / early stopping**

Generation length was compared across all three cache types at matching
bit widths. `QuantizedRotatingKVCache` produces the same early-EOS token
counts as `QuantizedKVCache` at both 4-bit (~1045 tokens) and 8-bit
(~1202 tokens). There is no regression in early stopping relative to the
existing quantized cache — the behaviour is a property of `mx.quantized_matmul`
accumulating rounding error over long sequences, present in both
implementations equally.

## Testing
```bash
python -m pytest tests/test_models.py \
  -k "quantized_rotating or to_quantized or maybe_quantize" -v
```

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, List, Optional
 
 import mlx.core as mx
 import mlx.nn as nn
-from mlx.utils import tree_flatten, tree_map, tree_unflatten
+from mlx.utils import tree_flatten, tree_map, tree_reduce, tree_unflatten
 
 from .base import create_causal_mask
 
@@ -546,8 +546,19 @@ class RotatingKVCache(_BaseCache):
         self._idx -= n
         return n
 
-    def to_quantized(self, group_size: int = 64, bits: int = 4) -> QuantizedKVCache:
-        raise NotImplementedError("RotatingKVCache Quantization NYI")
+    def to_quantized(self, group_size: int = 64, bits: int = 4) -> "QuantizedRotatingKVCache":
+        quant_cache = QuantizedRotatingKVCache(
+            max_size=self.max_size,
+            keep=self.keep,
+            group_size=group_size,
+            bits=bits,
+        )
+        quant_cache.offset = self.offset
+        quant_cache._idx = self._idx
+        if self.keys is not None:
+            quant_cache.keys = list(mx.quantize(self.keys, group_size=group_size, bits=bits))
+            quant_cache.values = list(mx.quantize(self.values, group_size=group_size, bits=bits))
+        return quant_cache
 
     def make_mask(
         self, N: int, window_size: Optional[int] = None, return_array: bool = False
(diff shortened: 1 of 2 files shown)


angeloskath:
This is nice but I think that a lot of the complicated logic is duplicated. It should probably be done in a different way so that it is shared between quantized and unquantized kv cache implementations.

Aaryaman3:
Thanks for the feedback @angeloskath ! I see two options: extract just the type-agnostic methods (make_mask, trim, is_trimmable, size, update_and_fetch) into a _RotatingKVCacheBase that both classes inherit from, or have QuantizedRotatingKVCache inherit directly from RotatingKVCache, which is simpler but carries some type awkwardness since keys has incompatible types between parent and child. Which direction did you have in mind?

Aaryaman3:
@angeloskath  took a shot at the _RotatingKVCacheBase approach in the latest commit. Let me know if this is the direction you had in mind or if you'd prefer the direct inheritance route!

zcbenz:

PR #970, opened 2026-03-08
mlx jamba layers devision support

office883:


office883 (commented):
@codex

angeloskath:
I don't think this needs to be in mlx-lm as it is very model specific.

PR #971, opened 2026-03-09
Presence and frequency penalties

angeloskath:
Different versions of the repetition penalty.

### DIFF
--- a/tests/test_sample_utils.py
+++ b/tests/test_sample_utils.py
@@ -116,6 +116,64 @@ class TestSampleUtils(unittest.TestCase):
         new_probs = mx.softmax(apply_xtc(mx.log(probs), 0, 0.1, [0]), -1)
         self.assertTrue(mx.allclose(new_probs, probs))
 
+    def test_presence_penalty(self):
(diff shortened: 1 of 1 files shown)


mlx_lm/sample_utils.py
```diff
@@ -96,15 +109,20 @@ def make_logits_processors(
         values = mx.array(list(logit_bias.values()))
 
         def logit_bias_processor(_, logits):
-            logits[:, indices] += values
-            return logits
+            return logits.at[:, indices].add(values)
```
nastya236, line 112: If I remember correctly, you mentioned once that `.at[].add()` is needed to use `Scatter::Sum` instead of `Gather + Add + Scatter::None`, right?

angeloskath: Exactly what you wrote. The reason I changed it is because the pattern `Gather -> Add -> Scatter` can be significantly more inefficient when the `Gather` or `Add` depends on the `src` of scatter because it breaks donation and `Scatter` ends up doing a copy.

Issue #980, opened 2026-03-11
Prefix cache reuse is broken for all hybrid-architecture models (sliding window, SSM/Mamba)


skcadri:
# Prefix cache reuse is broken for all hybrid-architecture models (sliding window, SSM/Mamba)

## Summary

Prompt prefix caching — the mechanism that reuses computed KV states across requests sharing a common prefix — only works for **pure full-attention models**. Any model using sliding window attention, Mamba/SSM layers, or mixed attention types silently falls back to full prompt recomputation on every request. This makes multi-turn conversations unusably slow for the majority of modern open-weight models.

This is not a single bug but a **systemic gap** in how mlx-lm handles non-standard cache types. I'm filing this as a unifying issue because the symptoms are spread across many separate reports that all trace back to the same root causes.

## Empirical Evidence

Tested on Mac Studio M3 Ultra (512GB unified memory), LM Studio 0.4.6, mlx-engine. Three sequential requests with identical system prompts, different user messages, `max_tokens=30`:

### MiniMax M2.5 (pure attention, MoE) — caching works

| Request | Time | Speedup |
|---------|------|---------|
| 1 (cold) | 29.33s | — |
| 2 (warm) | 6.15s | 4.8x |
| 3 (warm) | 2.79s | 10.5x |

### GPT-OSS 120B (sliding_attention + full_attention hybrid) — no caching

| Request | Time | Speedup |
|---------|------|---------|
| 1 (cold) | 1.54s | — |
| 2 (warm) | 1.77s | none |
| 3 (warm) | 1.67s | none |

### Qwen 3.5 9B (attention + Mamba/SSM hybrid) — no caching

| Request | Time | Speedup |
|---------|------|---------|
| 1 (cold) | 5.02s | — |
| 2 (warm) | 7.76s | none (slower) |
| 3 (warm) | 8.00s | none (slower) |

MiniMax shows clear prefix reuse (29s → 3s). GPT-OSS and Qwen 3.5 show zero improvement — the full prompt is recomputed every turn.

## Root Cause Analysis

There are **two distinct failure modes**, both stemming from the assumption that all layers use identical, trimmable KV caches:

### 1. Sliding window models → RotatingKVCache can't be trimmed

Models like GPT-OSS 120B and Gemma 3 27B alternate between sliding window and full attention layers:

```
# GPT-OSS 120B config.json
"layer_types": ["sliding_attention", "full_attention", "sliding_attention", "full_attention", ...]
[...]

PR #982, opened 2026-03-11
feat(server): Add opt-in prompt cache warmup

spicyneuron:
This PR adds a `--prompt-cache-warmup` flag for speeding up the experience of multi-turn chats in `mlx_lm.server`.

More specifically, it uses idle time in between requests to pre-compute a cache checkpoint for the _next_ turn of a chat, instead of waiting for the user message.

## Problem

Previous work on cache checkpoints didn't fully address the difficult scenarios (untrimmable SSM layers, dynamic `<think>` placement, tool result rendering, etc:

- https://github.com/ml-explore/mlx-lm/issues/903
- https://github.com/ml-explore/mlx-lm/issues/980

Models with such features _always_ miss the final checkpoint and require reprocessing the entire previous assistant turn. That's much better than starting from scratch, but too often, still means waiting on tens of thousands of tokens per turn.

## Solution

At the end of an assistant turn, we already have everything needed to calculate the next stable cache prefix. Just tokenize the history with two placeholder user messages, then compare to find the overlap. This uses the model's actual chat template instead of heuristics and should handle any supported scenario.

Then, when the batch is idle, process a new checkpoint that's guaranteed to hit as long as the user continues the chat. If a new request interrupts warmup while it's in progress, we save a checkpoint immediately. If it was a continuation, it benefits from the partial progress. If not, the new request is not held up unnecessarily.

## Caveats

- This PR focuses on supporting the simple, single-user chat experience — one warmup slot, no draft model, no distributed mode. I briefly explored integrating warmup as a different batch row type, but this would've been messy without a bigger refactor. Likewise, draft model support seemed feasible, but would've expanded the scope.

- This PR also fixes two adjacent bugs:
  1. In `process_message_content()` tool arguments could be pre-parsed dicts rather than strings in some models (Qwen 3.5 and Kimi 2.5 in my testing) and would error.
  2. `make_prompt_cache` used `self.model_provider.model` (last loaded model) instead of the model actually assigned to the current batch. This could diverge if a request for a different model was served mid-batch.  

---

I've been testing this locally with a Qwen 3.5 powered Claude Code, and it makes a huge difference. I know it's a moderately large change though, so I'm happy to test more and make any adjustments.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -523,7 +523,9 @@ class ResponseGenerator:
         rq = request[0] if request is not None else Queue()
         return rq, *shareable
 
-    def _tokenize_chat(self, tokenizer, messages, tools, role_mapping, extra_args):
+    def _prepare_chat_prompt(
+        self, tokenizer, messages, tools, role_mapping, extra_args
+    ):
         if tokenizer.has_chat_template:
             process_message_content(messages)
             if tools and not tokenizer.has_tool_calling:
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -856,6 +856,10 @@ class ResponseGenerator:
                 cache, uncached = self.prompt_cache.fetch_nearest_cache(
                     self.model_provider.model_key, prompt
                 )
+                # Ignore tiny uncached lengths (close checkpoint already exists)
+                if len(uncached) < 10:
+                    return True
+
                 def progress(tokens_processed, total_tokens):
                     logging.info(
                         f"Prompt cache warmup progress: {tokens_processed}/{total_tokens}"
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -986,6 +986,11 @@ class ResponseGenerator:
             # We got a request
             if request is not None:
                 rqueue, request, args = request
+                if self.cli_args.prompt_cache_warmup:
+                    with self._prompt_cache_warmup_lock:
+                        warmup = self._prompt_cache_warmup
+                        if warmup is not None and warmup.model != args.model:
+                            self._prompt_cache_warmup = None
 
                 # Can it be added to the current batch?
                 if (
@@ -1161,13 +1166,17 @@ class ResponseGenerator:
                                 del batch_results[uid]
                     continue
 
-            # Prefer a newly arrived foreground request over best-effort warmup.
-            request = get_next_request(timeout=0)
-            if request is not None:
-                unprocessed_requests.append(request)
-                continue
+            if self.cli_args.prompt_cache_warmup:
+                with self._prompt_cache_warmup_lock:
+                    has_pending_warmup = self._prompt_cache_warmup is not None
+                if has_pending_warmup:
+                    # Prefer a newly arrived foreground request over best-effort warmup.
+                    request = get_next_request(timeout=0)
+                    if request is not None:
+                        unprocessed_requests.append(request)
+                        continue
 
-            self._run_prompt_cache_warmup()
+                    self._run_prompt_cache_warmup()
 
     def _serve_single(self, request):
         rqueue, request, args = request
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -653,61 +653,63 @@ class ResponseGenerator:
                 warmup.model.draft,
             )
             prompt = self._build_prefill_request(tokenizer, warmup)
-            if prompt is not None:
-                self.prompt_cache.log_cache_stats()
-                cache, uncached = self.prompt_cache.fetch_nearest_cache(
-                    self.model_provider.model_key, prompt
+            if prompt is None:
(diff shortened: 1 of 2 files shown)


skcadri:
Tested this locally against Qwen3.5-9B (hybrid attention + Mamba/SSM) on Mac Studio M3 Ultra. We filed [#980](https://github.com/ml-explore/mlx-lm/issues/980) tracking this class of cache reuse failures for hybrid models — this PR directly addresses the most painful symptom.

### Results

Checked out the branch, ran `mlx_lm.server` with `--prompt-cache-warmup`, and sent a 4-turn agentic conversation (system prompt + tool-call round-trips):

**With `--prompt-cache-warmup`:**

| Turn | prompt_tokens | cached_tokens | % cached | wall time |
|---|---|---|---|---|
| 1 (cold) | 454 | 0 | 0% | 2.71s |
| 2 | 627 | **474** | **76%** | 2.63s |
| 3 | 700 | **651** | **93%** | 2.60s |
| 4 | 779 | **722** | **93%** | 2.97s |

**Without warmup (same model, same conversation, via LM Studio):**

| Turn | prompt_tokens | wall time |
|---|---|---|
| 1 | 454 | 2.75s |
| 2 | 617 | 2.94s |
| 3 | 691 | 2.96s |
| 4 | 773 | 2.97s |

93% cache hit rate by turn 3 on a model where `can_trim_prompt_cache()` returns `False`. The timing gap is modest at ~700 tokens but would scale significantly with the 10K–40K token contexts typical in agentic workloads.

### What this solves well

The approach of prefilling during idle time and inserting the result as a "shorter" cache entry is clever — it completely sidesteps the non-trimmable cache issue ([`ArraysCache.is_trimmable() → False`](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/cache.py#L144)) without needing changes to the cache layer itself. The placeholder-message-diff technique for finding the prefix boundary is also elegant — works with any chat template without heuristics.

### Remaining gaps (for visibility, not blockers for this PR)

This fixes the most common user-facing pain (single-user multi-turn agentic chat), but a few scenarios still hit the underlying `can_trim_prompt_cache() → False` wall:

- **Concurrent/parallel requests**: multiple agents or users sharing a model with the same system prompt but different user messages. The trie finds the first request's cache as "longer", can't trim it, falls through to full recompute. Only one warmup slot, so this isn't addressed here. (This is the scenario [@Giustino98 identified](https://github.com/ml-explore/mlx-lm/issues/980#issuecomment-4047708644) in #980.)
- **Fast-typing users**: if the next message arrives before warmup completes, the benefit is reduced (though the partial checkpoint mechanism helps).
- **LM Studio / mlx-engine integration**: this is `mlx_lm.server`-only for now. LM Studio users (where most people encounter this) would need mlx-engine to adopt it separately.

None of these are reasons to hold this PR — they're just the remaining surface area for a future deeper fix (making `ArraysCache`/`RotatingKVCache` trimmable, or per-layer trim strategies in `fetch_nearest_cache`).

skcadri:
**Updated with larger context test** (~2K token system prompt, 5-turn agentic conversation growing to ~3.5K tokens, Qwen3.5-9B):

**With `--prompt-cache-warmup` (mlx-lm server, this branch):**

| Turn | prompt_tokens | cached_tokens | % cached | wall time |
|---|---|---|---|---|
| 1 (cold) | 2031 | 0 | 0% | 4.71s |
| 2 | 2480 | **2037** | **82%** | **2.07s** |
| 3 | 2954 | **2496** | **84%** | **2.78s** |
| 4 | 3408 | **2978** | **87%** | **3.78s** |
| 5 | 3529 | **3411** | **96%** | **3.43s** |

**Without warmup (LM Studio, same model, same conversation):**

| Turn | prompt_tokens | wall time |
|---|---|---|
| 1 | 2031 | 4.63s |
| 2 | 2521 | 2.38s |
| 3 | 3002 | 3.73s |
| 4 | 3476 | 4.18s |
| 5 | 3676 | 4.06s |

Without warmup, wall time grows linearly with context (full recompute every turn). With warmup, it stays lower and flatter — only processing the uncached suffix. At 3.5K tokens the gap is 15–25% per turn. At the 10K–40K token contexts typical in real agentic sessions, this would scale to seconds vs minutes.

spicyneuron:
@skcadri Thanks for the benchmarks. The speedup is most obvious (and needed) when a previous assistant turn includes multiple tool calls (ex: searching then reading multiple files).

This often hits 10,000+ tokens, which as you mentioned, amounts to minutes of wait time even on a M3 Ultra.

spicyneuron:
Closing. Maybe a bit too niche for this repo.

f1yn:
@spicyneuron on the contrary, I'm likely scraping by on your patch alone. Qwen3.5+ is basically unusable without it, and I hope they revisit this soon ♥️✨

PR #985, opened 2026-03-11
Fix tool call parsing for models with multi-token delimiters

### LINKED ISSUE #984: Tool calling not detected for models with multi-token delimiters (e.g. Trinity)
## Description

`mlx_lm.server` fails to detect tool-calling support for models that use `<tool_call>` / `</tool_call>` delimiters when these are tokenized as multiple sub-tokens rather than single special tokens. The model generates correct tool calls, but the server can't parse them into structured `tool_calls` responses.

This affects all MLX-converted Trinity models (and likely other models where tool call delimiters weren't preserved as special tokens during conversion).

## Environment

- **mlx-lm:** 0.31.1
- **macOS:** 15.5 (Apple M3 Max, 48 GB)
- **Python:** 3.12
- **Model:** `mlx-community/Trinity-Mini-8bit` (afmoe architecture)

## The Problem

There are two issues:

### 1. Vocab guard disables tool parsing (`tokenizer_utils.py:306-312`)

```python
# Disable tool calling if tool call tokens aren't in vocab
if (tool_call_start and tool_call_start not in vocab) or (
    tool_call_end and tool_call_end not in vocab
):
    self._tool_call_start = None
    self._tool_call_end = None
    self._tool_parser = None
```

The `_infer_tool_parser()` function correctly identifies the model as using `json_tools` (it finds `<tool_call>` and `tool_call.name` in the chat template). But then the vocab check disables it because `<tool_call>` is tokenized as multiple sub-tokens (`<`, `tool`, `_`, `call`, `>`) rather than existing as a single special token.

### 2. Streaming parser uses exact token matching (`server.py:1579`)

```python
elif ctx.has_tool_calling and gen.text == ctx.tool_call_start:
```

Even if the vocab guard is removed, the streaming parser checks if a single generated token's text equals `<tool_call>`. Since the model generates it as multiple tokens, no single `gen.text` ever matches.

## Steps to Reproduce

```bash
uvx --from "mlx-lm>=0.28.4" mlx_lm.server \
    --model mlx-community/Trinity-Mini-8bit --port 8080
```

```bash
  -d '{
    "model": "mlx-community/Trinity-Mini-8bit",
    "messages": [{"role": "user", "content": "What is the weather in Paris?"}],
    "tools": [{"type": "function", "function": {"name": "get_weather", "description": "Get weather", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}}],
    "max_tokens": 200
  }'
```

**Server logs:**
```
WARNING - Received tools but model does not support tool calling.
```

**Response:** The model generates a correct tool call in `content`, but `tool_calls` is empty and `finish_reason` is `stop` instead of `tool_calls`:
```json
{
  "content": "\n<tool_call>\n{\"name\": \"get_weather\", \"arguments\": {\"city\": \"Paris\"}}\n</tool_call>",
  "tool_calls": []
}
```

### 1. Relax the vocab guard to a warning

The vocab check should warn rather than disable, since the streaming parser can use text matching:

```python
if (tool_call_start and tool_call_start not in vocab) or (
    tool_call_end and tool_call_end not in vocab
):
    logging.warning(
        f"Tool call tokens ({tool_call_start}, {tool_call_end}) not found "
        "as single tokens in vocab. Tool parsing will use text matching."
    )
```

### 2. Use buffer-based matching in the streaming parser

Instead of exact token matching, accumulate text and check for substrings:

```python
# Before (exact single-token match):
elif ctx.has_tool_calling and gen.text == ctx.tool_call_start:
    made_tool_call = True
    in_tool_call = True

# After (buffer-based substring match):
elif ctx.has_tool_calling:
    text += gen.text
    segment += gen.text
    if ctx.tool_call_start in text:
        text = text[:text.index(ctx.tool_call_start)]
        segment = ""
        made_tool_call = True
        in_tool_call = True
```

Same pattern for the end tag detection in `in_tool_call`.

I've tested both changes locally and they produce correct structured `tool_calls` with `finish_reason: tool_calls` for Trinity-Mini. Opening a PR with the fix.

## Affected Models

Any model where:
- The chat template uses `<tool_call>` delimiters (detected by `_infer_tool_parser` as `json_tools`)
- But `<tool_call>` is not a single special token in the vocabulary

This includes all `mlx-community/Trinity-*` models and potentially other community conversions where special tokens weren't preserved.

juliensimon:
## Summary

Fixes #984 — tool calling detection and parsing fails for models where `<tool_call>` / `</tool_call>` delimiters are tokenized as multiple sub-tokens rather than single special tokens.

This affects all `mlx-community/Trinity-*` models (afmoe architecture) and potentially other community conversions where special tokens weren't preserved during MLX conversion.

### Changes

- **`tokenizer_utils.py`**: Changed the vocab guard from disabling tool calling to emitting a warning. The streaming parser uses text matching which works fine for multi-token delimiters — there's no reason to disable it entirely.

- **`server.py`**: Changed the streaming tool call parser from exact single-token matching (`gen.text == ctx.tool_call_start`) to buffer-based substring matching (`ctx.tool_call_start in text`). This correctly detects `<tool_call>` even when it arrives as multiple tokens (`<`, `tool`, `_`, `call`, `>`).

### Before

```
WARNING - Received tools but model does not support tool calling.
```
Response has tool call in `content` but `tool_calls` is empty, `finish_reason` is `stop`.

### After

Tool calls are correctly parsed into structured `tool_calls` with `finish_reason: tool_calls`.

## Test plan

- [x] Tested with `mlx-community/Trinity-Mini-8bit` — tool calls correctly detected and parsed
- [x] Verified backward compatibility: models with single-token delimiters still work (exact match is a subset of substring match)
- [x] No changes to tool parser logic itself, only detection and streaming extraction

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1576,16 +1576,23 @@ class APIHandler(BaseHTTPRequestHandler):
                     in_reasoning = False
                 else:
                     reasoning_text += gen.text
-            elif ctx.has_tool_calling and gen.text == ctx.tool_call_start:
-                made_tool_call = True
-                in_tool_call = True
             elif in_tool_call:
-                if gen.text == ctx.tool_call_end:
+                tool_text += gen.text
+                # Detect end tag: exact single-token match or substring in buffer
+                if gen.text == ctx.tool_call_end or ctx.tool_call_end in tool_text:
+                    tool_text = tool_text[: tool_text.index(ctx.tool_call_end)]
                     tool_calls.append(tool_text)
                     tool_text = ""
                     in_tool_call = False
-                else:
-                    tool_text += gen.text
+            elif ctx.has_tool_calling:
+                text += gen.text
+                segment += gen.text
+                # Detect start tag: exact single-token match or substring in buffer
+                if ctx.tool_call_start in text:
+                    text = text[: text.index(ctx.tool_call_start)]
+                    segment = ""
+                    made_tool_call = True
+                    in_tool_call = True
             else:
                 text += gen.text
                 segment += gen.text
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -303,13 +303,17 @@ class TokenizerWrapper:
                 self._think_end_id = vocab[think_end]
                 break
 
-        # Disable tool calling if tool call tokens aren't in vocab
+        # Warn if tool call tokens aren't in vocab but keep tool calling
+        # enabled — the streaming parser uses text matching which still works
+        # for multi-token delimiters (e.g. <tool_call> tokenized as subwords)
         if (tool_call_start and tool_call_start not in vocab) or (
             tool_call_end and tool_call_end not in vocab
         ):
-            self._tool_call_start = None
-            self._tool_call_end = None
-            self._tool_parser = None
+            logging.warning(
+                f"Tool call tokens ({tool_call_start}, {tool_call_end}) not "
+                "found as single tokens in vocab. "
+                "Tool call parsing will use text matching."
+            )
 
     def apply_chat_template(self, *args, tokenize=True, **kwargs):
         if self._chat_template is not None:


mlx_lm/server.py
```diff
@@ -1576,16 +1576,23 @@ def parse_tools(tool_calls):
...
-                    tool_text += gen.text
+            elif ctx.has_tool_calling:
+                text += gen.text
+                segment += gen.text
+                # Detect start tag: exact single-token match or substring in buffer
+                if ctx.tool_call_start in text:
+                    text = text[: text.index(ctx.tool_call_start)]
+                    segment = ""
+                    made_tool_call = True
+                    in_tool_call = True
```
mm65x, line 1595: in streaming mode, each token's text gets appended to `segment` and flushed to the client *before* the full tag is recognized. so if `<tool_call>` arrives as `<`, `tool`, `_`, `call`, `>`, the first four fragments get streamed as SSE chunks (because `in_tool_call` is still `False` at that point, so the `not in_tool_call` flush guard passes). `text` gets trimmed correctly once `>` completes the match, but the partial tag content is already sent.

you'd need to buffer text while there's a potential partial match instead of flushing it immediately.

mlx_lm/server.py
```diff
@@ -1576,16 +1576,23 @@ def parse_tools(tool_calls):
...
                     reasoning_text += gen.text
-            elif ctx.has_tool_calling and gen.text == ctx.tool_call_start:
-                made_tool_call = True
-                in_tool_call = True
             elif in_tool_call:
-                if gen.text == ctx.tool_call_end:
+                tool_text += gen.text
+                # Detect end tag: exact single-token match or substring in buffer
+                if gen.text == ctx.tool_call_end or ctx.tool_call_end in tool_text:
+                    tool_text = tool_text[: tool_text.index(ctx.tool_call_end)]
```
mm65x, line 1583: this breaks mistral tool calling. mistral uses `tool_call_end = ""` (empty string), and `"" in tool_text` is always `True`, so the very first token inside a tool call triggers the end condition. then `tool_text[:tool_text.index("")]` returns `""` and the entire tool call body is lost.

the original code handled this correctly because `gen.text == ""` is never true during normal generation, so it fell through to the post-loop flush at line 1637-1639.

needs a guard to skip the substring check when `tool_call_end` is empty:

```suggestion
                tool_text += gen.text
                # Detect end tag: exact single-token match or substring in buffer
                if ctx.tool_call_end and ctx.tool_call_end in tool_text:
                    tool_text = tool_text[: tool_text.index(ctx.tool_call_end)]
```

mm65x (changes requested):
the vocab guard change in tokenizer_utils.py looks good to me.

found two issues with the server.py streaming logic though, details inline. happy to help with a fix if you want - i was looking at the same issue and can put together a patch.

zcbenz:

PR #986, opened 2026-03-12
Clear cache trainer memory

N8python:
Simply adds a 'mx.clear_cache()' after each iteration of LORA training. This takes memory use (measured by `phys_footprint`) from '51.0 GiB' to '7.09 GiB' under the same conditions. MLX's native peak_memory wasn't detecting this extra usage, so it didn't immediately appear in the server - but I noticed in in mactop, so investigated + patched.

### DIFF
--- a/mlx_lm/tuner/trainer.py
+++ b/mlx_lm/tuner/trainer.py
@@ -17,6 +17,11 @@ from .callbacks import TrainingCallback
 from .datasets import CacheDataset
 
 
+def _clear_cache(threshold: int):
+    if mx.get_cache_memory() > threshold:
+        mx.clear_cache()
+
+
 def grad_checkpoint(layer):
     """
     Update all instances of type(layer) to use gradient checkpointing.
@@ -68,6 +73,12 @@ class TrainingArgs:
         default=1,
         metadata={
             "help": "Number of steps to accumulate gradients before applying an optimizer update."
+        },
+    )
+    clear_cache_threshold: int = field(
+        default=0,
+        metadata={
+            "help": "Clear the allocator cache between steps if it grows too large."
         },
     )
 
@@ -170,6 +181,7 @@ def evaluate(
     max_seq_length=2048,
     loss: callable = default_loss,
     iterate_batches: callable = iterate_batches,
+    clear_cache_threshold: int = 0,
 ):
     model.eval()
     all_losses = mx.array(0.0)
@@ -194,14 +206,13 @@ def evaluate(
         all_losses += losses * toks
         ntokens += toks
         mx.eval(all_losses, ntokens)
-        mx.clear_cache()
+        _clear_cache(clear_cache_threshold)
 
     all_losses = mx.distributed.all_sum(all_losses, stream=mx.cpu)
     ntokens = mx.distributed.all_sum(ntokens, stream=mx.cpu)
-    mx.eval(all_losses, ntokens)
-    mx.clear_cache()
+    avg_loss = (all_losses / ntokens).item()
 
-    return (all_losses / ntokens).item()
+    return avg_loss
 
 
 def train(
@@ -315,7 +326,7 @@ def train(
         n_tokens += toks
         steps += 1
         mx.eval(state, losses, n_tokens, grad_accum)
-        mx.clear_cache()
+        _clear_cache(args.clear_cache_threshold)
         train_time += time.perf_counter() - tic
 
         # Report training loss if needed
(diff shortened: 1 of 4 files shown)


N8python:
Reported numbers observed for this LORA config:
```
# The path to the local model directory or Hugging Face repo.
model: "Qwen2.5-0.5B-Instruct"

# Whether or not to train (boolean)
train: true

# The fine-tuning method: "lora", "dora", or "full".
fine_tune_type: lora

# The Optimizer with its possible inputs
optimizer: adam
optimizer_config:
   adam:
     betas: [0.9, 0.9999]
     eps: 1e-6
     bias_correction: true

# Directory with {train, valid, test}.jsonl files
data: "train"

# The PRNG seed
seed: 0

# Number of layers to fine-tune
num_layers: 24

# Minibatch size.
batch_size: 1

# Iterations to train for.
iters: 1000

# Number of validation batches, -1 uses the entire validation set.
val_batches: 0

# Adam learning rate.
learning_rate: 1e-4

# Services to report logs to (comma-separated): wandb, swanlab, or both ('wandb,swanlab').
report_to: wandb
project_name: "pii-extraction"

# Number of training steps between loss reporting.
steps_per_report: 1

# Number of training steps between validations.
steps_per_eval: 10000000

# Number of micro-steps to accumulate before each optimizer update.
grad_accumulation_steps: 1

# Load path to resume training with the given adapter weights.
resume_adapter_file: null

# Save/load path for the trained adapter weights.
adapter_path: "adapters-long"

# Save the model every N iterations.
save_every: 2500

# Evaluate on the test set after training
test: false

# Number of test set batches, -1 uses the entire test set.
test_batches: 100

# Maximum sequence length.
max_seq_length: 8192

# Use gradient checkpointing to reduce memory use.
grad_checkpoint: true

# LoRA parameters can only be specified in a config file
lora_parameters:
  # The layer keys to apply LoRA to.
  # These will be applied for the last lora_layers
  keys: ["self_attn.q_proj", "self_attn.v_proj", "self_attn.k_proj", "self_attn.o_proj", "mlp.gate_proj", "mlp.up_proj", "mlp.down_proj"]
  rank: 128
  scale: 0.25
  dropout: 0.0

# Schedule can only be specified in a config file, uncomment to use.
lr_schedule:
  name: cosine_decay
  warmup: 10 # 0 for no warmup
  warmup_init: 1e-5 # 0 if not specified
  arguments: [1e-4, 1000, 1e-5] # passed to scheduler

```

mlx_lm/tuner/trainer.py
```diff
@@ -194,9 +194,12 @@ def evaluate(
...
         ntokens += toks
         mx.eval(all_losses, ntokens)
+        mx.clear_cache()
 
     all_losses = mx.distributed.all_sum(all_losses, stream=mx.cpu)
     ntokens = mx.distributed.all_sum(ntokens, stream=mx.cpu)
+    mx.eval(all_losses, ntokens)
+    mx.clear_cache()
 
     return (all_losses / ntokens).item()
```
angeloskath:
```suggestion
    avg_loss = (all_losses / ntokens).item()
    mx.clear_cache()

    return avg_loss
```

No need for two evals here.

mlx_lm/tuner/trainer.py
```diff
@@ -312,6 +315,7 @@ def step(batch, prev_grad, do_update):
         n_tokens += toks
         steps += 1
         mx.eval(state, losses, n_tokens, grad_accum)
+        mx.clear_cache()
```
angeloskath: This clear cache will slow down training when the inputs are of fixed size. Perhaps this should be a flag that can be enabled optionally and documented at the memory pressure part in the readme.

angeloskath (commented):
The peak memory won't report the cache so indeed it may grow as needed which may make MLX a bit of a bad player with other apps when training something.

However, when the sizes don't change much the cache won't grow and clearing it at every iteration actually may hurt performance significantly since we have to reallocate all the memory needed for the intermediates.

I think it is better to have this behind a flag.

mlx_lm/tuner/trainer.py
N8python: The inputs are almost never of fixed size in LORA training. It should be a flag that's *on* by default, IMO.

N8python: But I'd be happy if this was merged even w/ a flag that is off by default.

N8python:
Alright. That feels like a reasonable compromise! I'd be happy w/ merge.

N8python:
One of the tests isn't passing... some issue w/ glm_4_moe?

angeloskath:
Yeah I know, I 'll look into it. Just fell a bit lower in the queue.

PR #988, opened 2026-03-12
fix(ssm): correct SSM decode precision

rsnow:
## Problem
The Metal decode kernel (ssm_update_kernel) produces increasingly incoherent output during autoregressive generation for SSM/Mamba models with aggressive learned dynamics. After ~15 tokens, output degenerates into gibberish. The non-kernel path (ssm_attn, used for prefill and seq_len > 1) produces correct output for the same models.

Two issues in ssm.py:

1. compute_dt applies an incorrect upper clamp
return mx.clip(dt, time_step_limit[0], time_step_limit[1])
The [reference implementation](https://github.com/huggingface/transformers/blob/main/src/transformers/models/nemotron_h/modeling_nemotron_h.py) applies a lower bound only — torch.clamp(dt, self.time_step_min). The upper clamp (typically 0.1) destroys learned time-step diversity for models where dt values naturally exceed this bound after softplus.

2. Recurrent state truncated to model dtype every decode step
o_state[idx] = static_cast<T>(state);  // T = bfloat16
The SSM recurrence state is computed in float32 inside the kernel but written back to memory in model dtype (bfloat16/float16) on every token. On the next step, the truncated state feeds back into the recurrence. For models with large A_log values, the resulting decay multipliers (dA = exp(A * dt)) amplify these quantization errors, and output degenerates within 15–20 tokens.

The ssm_attn path keeps state in float32 throughout, which is why it produces correct results.

## Fix
compute_dt: Replace mx.clip with mx.maximum to apply lower bound only, matching reference behavior.
Metal kernel: Remove static_cast<T>(state) so recurrent state stays in float32. Update the Python wrapper to pass float32 state and dt to the kernel, and set output_dtypes accordingly.

### Numerical tests (tests/test_ssm_decode.py):

- compute_dt enforces lower bound, does not apply upper clamp
- Kernel returns float32 state
- Single-step kernel output matches ssm_attn within tolerance
- 50-step recurrence does not diverge between kernel and ssm_attn
- ssm_update correctly routes to kernel for seq_len=1

## Performance / memory impact
Recurrent decode state is now stored in float32 instead of model dtype. For a model with 128 heads × 64 head_dim × 128 state_dim, this adds ~2MB per Mamba layer (~112MB total for 56 layers). This is a modest increase relative to model size and is consistent with how ssm_attn already handles state.

Decode throughput decreases from ~67 to ~48 tok/s due to the wider state bandwidth, but the original speed was only achievable with incorrect output.

## Reproducer
Any Mamba/SSM model with large learned A_log values and dt values exceeding the upper clamp will exhibit this issue. Nemotron-3 Super 120B-A12B (A_log up to 6.8, mean dt_raw ~4.26) is a clear reproducer. Models with more conservative dynamics (e.g., Mamba-2 Nano 30B) may not trigger visible degeneration because their precision requirements are less extreme, but they still benefit from the correctness fix.

## Other SSM models 
(Falcon-H1, Granite MoE Hybrid, Mamba2, PLaMo2) are unaffected by the dt change - their upper bounds are already unreachable - and benefit from the float32 state fix with no correctness regression, only a modest increase in state memory.

### DIFF
--- a/mlx_lm/models/ssm.py
+++ b/mlx_lm/models/ssm.py
@@ -46,7 +46,7 @@ def make_ssm_kernel():
             auto idx = d_idx * Ds + s_idx;
             auto dB_by_x = x_ * dt_ * static_cast<float>(B_[s_idx]);
             auto state = dA * i_state[idx] + dB_by_x;
-            o_state[idx] = static_cast<T>(state);
+            o_state[idx] = state;
             acc += state * C_[s_idx];
         }
         acc = simd_sum(acc);
@@ -79,14 +79,19 @@ def ssm_update_kernel(
     n, _, h, d = hidden_states.shape
     input_type = hidden_states.dtype
     hb, ds = B.shape[-2:]
-    dt = compute_dt(dt, dt_bias, time_step_limit)
+    # Use float32 for dt and recurrent state to maintain numerical
+    # stability in the decode kernel. The SSM recurrence is sensitive
+    # to precision loss, especially for models with large A_log values.
+    dt = compute_dt(dt, dt_bias, time_step_limit).astype(mx.float32)
+    if state.dtype != mx.float32:
+        state = state.astype(mx.float32)
     return _ssm_kernel(
         inputs=[hidden_states, A_log, B, C, D, dt, state],
         template=[("T", input_type), ("Dh", d), ("Ds", ds), ("H", h), ("G", h // hb)],
         grid=(32, d, h * n),
         threadgroup=(32, 8, 1),
         output_shapes=[(n, 1, h, d), state.shape],
-        output_dtypes=[input_type, input_type],
+        output_dtypes=[input_type, mx.float32],
     )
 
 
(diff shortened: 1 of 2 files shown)


angeloskath:
Hi could you provide a reproduction of the problem? Any code or command to show the issue would be appreciated.

rsnow:
Sure thing

**Reproducer:**

```bash
mlx_lm.generate --model <path-to-nemotron-3-super-120b> --prompt "What is 2+2?" --max-tokens 50
```

Model: `nvidia/Nemotron-3-Super-120B-A12B-Instruct` (any quant - 4-bit, 8-bit, or BF16 all reproduce) -> Output degenerates after ~15 tokens.

**Workaround proof:** Force the `ssm_attn` path for `seq_len=1` (bypass the Metal kernel) -> coherent output throughout.

**Numerical tests:** `tests/test_ssm_decode.py` in this PR demonstrates both issues - the dt upper clamp destroying learned dynamics, and the kernel state diverging from `ssm_attn` over 50 steps due to bfloat16 truncation.

angeloskath:
Hi @rsnow, I will close this in favor of #992 that also adds support for the model and fixes the upper limit in a more general way.

If you have any example that shows degradation of the output by keeping the state in bf16 could you share it in that PR?

PR #995, opened 2026-03-13
Strip whitespace from tool call function names

atmin:
Some models (e.g. GLM-4.5-Air) emit tool call function names with trailing newlines, causing OpenAI-compatible clients to fail with "tool not found" errors.

### DIFF
--- a/mlx_lm/tool_parsers/glm47.py
+++ b/mlx_lm/tool_parsers/glm47.py
@@ -12,7 +12,7 @@ from typing import Any
 
 import regex as re
 
-_func_name_regex = re.compile(r"^(.*?)<arg_key>", re.DOTALL)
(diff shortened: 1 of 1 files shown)


angeloskath:
Wouldn't that be better in the glm47 tool parser? Does it happen with json tool calls?

I don't think the server should arbitrarily change the tool name even if it is stripping whitespace.

atmin:
Moved the fix to the glm47 parser. The regex `^(.*?)<arg_key>` with DOTALL was capturing trailing whitespace in the function name. E.g., GLM's tag-based tool call format looks like this:
```
bash
<arg_key>command</arg_key>
<arg_value>ls -la</arg_value>
```

michalk8:
closed via #1756

PR #998, opened 2026-03-14
Add thinking budget support for reasoning models

chopchop-jiahao:
## Summary

Adds `--thinking-budget N` CLI flag and `thinking_budget` API parameter to limit the number of tokens a model can generate inside `<think>` blocks. Designed for reasoning models (Qwen3, Qwen3.5) that use `<think>`/`</think>` special tokens.

### ThinkingBudgetProcessor

A logits processor that enforces a per-turn token budget on `<think>` blocks:

- **Budget enforcement**: After N thinking tokens, forces `</think>` by masking all other tokens
- **EOS masking after `</think>`**: Prevents empty responses by masking EOS for one step after think close
- **EOS masking during thinking**: Prevents premature stop (EOS before `</think>`)
- **Re-entry lockout** (`_thinking_done`): Prevents spurious `<think>` mid-tool-call from restarting the budget cycle

### Server routing fix

Handles `<tool_call>` emitted inside `<think>` blocks (before `</think>`). This is an upstream bug: the existing routing accumulates `<tool_call>` into `reasoning_text` instead of transitioning to tool call state.

### Bug fix: `</think>` token leak corrupting tool call JSON

**Observed in**: Qwen3.5-4B-OptiQ-4bit with `--thinking-budget 512` during SWE-bench evaluation (django-11099 task).

**Symptom**: When generating parallel tool calls, the model emits a `<think>` token ID mid-JSON argument string. ThinkingBudgetProcessor then forces `</think>` inside the JSON, corrupting it:

```
<tool_call>{"name": "run_command", "arguments": {"command": "ls -la /app/django/db/models/</think>

<function=run_command>
<parameter=command>ls -la /app/django/db/models/ 2>&1
```

The second tool call's JSON is broken by `</think>`, causing the model to fall back to text-based `<function=...>` format which the server can't parse.

**Root cause**: ThinkingBudgetProcessor had no awareness of tool call boundaries. When the budget was exceeded, it forced `</think>` regardless of whether generation was inside a `<tool_call>` block.

**Fix**: Made ThinkingBudgetProcessor tool-call-aware:
- Accepts optional `tool_call_start_id` / `tool_call_end_id` token IDs
- Tracks `<tool_call>`/`</tool_call>` boundaries via `_in_tool_call` state
- When inside a tool call: pauses budget enforcement AND masks `</think>` in logits (prevents both forced and spontaneous emission)
- Budget enforcement resumes after `</tool_call>`

This keeps the fix entirely within the processor — no changes needed to server.py's routing state machine.

### Usage

```bash
mlx_lm.server --model mlx-community/Qwen3.5-9B-4bit --thinking-budget 512
```

```python
# API
{"thinking_budget": 512, ...}
```

### Test coverage

- 77 tests across 3 test files
- `test_thinking_budget.py`: processor state machine, budget enforcement, EOS masking, tool-call awareness
- `test_server_routing.py`: all routing state machine paths (REASONING/NORMAL/TOOL_CALL states)
- `test_think_token_leak_in_tool_calls.py`: regression tests for `</think>` corruption pattern, with/without tool call IDs

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -634,6 +634,12 @@ def _make_logits_processors(args, tokenizer=None, prompt=None):
                 args.thinking_budget,
                 prompt,
                 eos_token_ids=frozenset(tokenizer.eos_token_ids),
+                tool_call_start_id=getattr(
+                    tokenizer, "tool_call_start_id", None
+                ),
+                tool_call_end_id=getattr(
+                    tokenizer, "tool_call_end_id", None
+                ),
             )
         )
     elif args.thinking_budget is not None and tokenizer is not None:
--- a/mlx_lm/thinking_budget.py
+++ b/mlx_lm/thinking_budget.py
@@ -37,6 +37,12 @@ class ThinkingBudgetProcessor:
     ``finish_reason=stop`` with no content.  One step is the minimal
     intervention: after the first non-EOS token, autoregressive momentum
     carries generation forward.
+
+    Tool-call awareness: when the model enters a ``<tool_call>`` block, budget
+    enforcement is paused until ``</tool_call>`` closes it.  This prevents
+    forced ``</think>`` from landing inside JSON arguments and corrupting tool
+    calls — the budget will be enforced on the next thinking token after the
+    tool call completes.
     """
 
     def __init__(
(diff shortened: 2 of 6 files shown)


chopchop-jiahao:
Closing — after further investigation, the underlying tool call generation issue (model not emitting EOS after tool calls) makes this feature incomplete. Switching to a different serving backend for now.

PR #999, opened 2026-03-14
Fix/prefix cache hybrid models

Giustino98:
Fixes the prompt cache trimming failure for sliding window models (e.g., Gemma 3, GPT-OSS) reported in #980.

### The Bug
Currently, `RotatingKVCache` and `BatchRotatingKVCache` fail to trim as soon as the sequence length exceeds the window size (when `offset >= max_size` and the buffer wraps around). Because `can_trim_prompt_cache` requires *all* layers to be trimmable, a single wrapped rotating cache forces the server to drop the entire prompt and trigger a full recompute on divergent-suffix requests (like tool calls in agentic workloads).

### The Fix
This PR enables trimming for wrapped rotating caches by aligning the architecture of `RotatingKVCache` to match `BatchRotatingKVCache`:

* **Separation of Physical vs Logical Offset:** `RotatingKVCache` previously used a single `offset` for both the physical buffer size and the absolute sequence position (RoPE). Trimming a wrapped buffer broke RoPE embeddings. We introduced `_offset` to track the physical tensor size (used for attention masks/slicing) while keeping `offset` as the absolute chronological position (used for RoPE generation).
* **Buffer Linearization:** `trim(n)` now physically unrolls the ring buffer into temporal order via `_temporal_order()` and slices off the most recent tokens.
* **API Contract Consistency:** `trim(n)` now correctly clips the physical deletion to the available history (`min(_offset, n)`) but strictly rolls back the absolute `offset` by `n` to ensure all transformer layers remain perfectly synchronized in time.

### Note on Mamba/SSM (ArraysCache)
This PR strictly addresses rotating KV caches. It does *not* touch `ArraysCache` (Qwen 3.5 / Mamba).
Since Mamba's recurrent state is an incompressible mathematical hash of the sequence, we cannot simply "trim" it chronologically. Forcing a local state reset for Mamba layers while keeping Attention layers intact would cause the model to silently hallucinate (as the server would only pass the new suffix tokens to an empty Mamba layer). Fixing Mamba prefix-caching requires a server-side checkpointing approach in `LRUPromptCache` at the prefix boundary, which is out of scope for this specific cache-level fix.

### Testing
* Ran the reproduction script provided in #980 locally using a sliding window model (`gpt-oss`).
* Verified that divergent-suffix requests now successfully trim the cache and skip the full prompt recompute, yielding the expected inference speedup.
* Measured KL Divergence between full-recompute and trimmed-generation to guarantee that the `_offset` vs `offset` separation accurately preserves RoPE positional embeddings (KL ~ 0.1).
* Passed local `pytest tests/` for regression checks.

### [UPDATE]
Thanks for @dae report

Slowdown happened because server.py assumed is_trimmable=True meaning cache could be trimmed back infinitely. With a ring buffer, this caused the server to aggressively delete valid intermediate caches, only to realize later it couldn't actually trim the longer cache back that far.

I've updated the PR to make the server aware of physical buffer limits:

`server.py` (insert_cache): Only evict intermediate caches if the trim distance fits within the physical buffer (`<= min(c.size() for c in prompt_cache)`). This keeps shorter prefixes alive as fallbacks.

`server.py` (fetch_nearest_cache): Explicitly check the result of trim_prompt_cache. If it fails to remove the requested amount of tokens, discard and fallback to a shorter cache instead of returning a broken state.

`cache.py` (extract): Added a quick keys is None guard to prevent TypeErrors during batch merging after evictions.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -538,12 +538,18 @@ class RotatingKVCache(_BaseCache):
         )
 
     def is_trimmable(self):
-        return self.offset < self.max_size
+        return True
 
     def trim(self, n):
         n = min(self.offset, n)
-        self.offset -= n
-        self._idx -= n
+        if n >= self.max_size or n == self.offset:
+            self.keys = None
+            self.values = None
+            self.offset = 0
+            self._idx = 0
+        else:
+            self.offset -= n
+            self._idx = (self._idx - n) % self.max_size
         return n
 
     def to_quantized(self, group_size: int = 64, bits: int = 4) -> QuantizedKVCache:
@@ -1240,12 +1246,19 @@ class BatchRotatingKVCache(_BaseCache):
         self.rotated = bool(v[3])
 
     def is_trimmable(self):
-        return self._offset < self.max_size
+        return True
 
     def trim(self, n):
         n = min(self._offset, n)
-        self._offset -= n
-        self._idx -= n
+        if n >= self.max_size or n == self._offset:
+            self.keys = None
+            self.values = None
+            self._offset = 0
+            self._idx = 0
+            self.rotated = False
+        else:
+            self._offset -= n
+            self._idx = (self._idx - n) % self.max_size
         self.offset -= n
         return n
 
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1359,6 +1359,8 @@ class BatchRotatingKVCache(_BaseCache):
 
     def extract(self, idx):
         cache = RotatingKVCache(self.max_size)
+        if self.keys is None:
+            return cache
         padding = self.left_padding[idx].item()
         offset = self.offset[idx].item()
         cache.keys = self.keys[idx : idx + 1]
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -541,8 +541,9 @@ class RotatingKVCache(_BaseCache):
         return True
 
     def trim(self, n):
-        n = min(self.offset, n)
-        if n >= self.max_size or n == self.offset:
+        effective = min(self.offset, self.max_size)
+        n = min(effective, n)
+        if n == effective:
             self.keys = None
             self.values = None
             self.offset = 0
@@ -1252,8 +1253,9 @@ class BatchRotatingKVCache(_BaseCache):
         return True
 
     def trim(self, n):
-        n = min(self._offset, n)
-        if n >= self.max_size or n == self._offset:
+        effective = min(self._offset, self.max_size)
+        n = min(effective, n)
+        if n == effective:
             self.keys = None
             self.values = None
             self._offset = 0
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -548,8 +548,11 @@ class RotatingKVCache(_BaseCache):
             self.offset = 0
             self._idx = 0
         else:
-            self.offset -= n
-            self._idx = (self._idx - n) % self.max_size
+            new_size = effective - n
+            self.keys = self._temporal_order(self.keys)[..., :new_size, :]
+            self.values = self._temporal_order(self.values)[..., :new_size, :]
+            self.offset = new_size
+            self._idx = new_size
         return n
 
     def to_quantized(self, group_size: int = 64, bits: int = 4) -> QuantizedKVCache:
@@ -1257,8 +1260,12 @@ class BatchRotatingKVCache(_BaseCache):
             self._idx = 0
             self.rotated = False
         else:
-            self._offset -= n
-            self._idx = (self._idx - n) % self.max_size
+            new_size = effective - n
+            self._temporal_order()
+            self.keys = self.keys[..., :new_size, :]
+            self.values = self.values[..., :new_size, :]
+            self._offset = new_size
+            self._idx = new_size
         self.offset -= n
         return n
 
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -413,6 +413,7 @@ class RotatingKVCache(_BaseCache):
         self.keys = None
         self.values = None
         self.offset = 0
+        self._offset = 0
         self.max_size = max_size
         self._idx = 0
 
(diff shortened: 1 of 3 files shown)


dae:
Testing with gpt-oss, this seems to cause a big slowdown even in the single reader case compared to the main branch. I notice it reports 0 tokens in cache after each request.

Giustino98:
@dae thanks for your report, I'm currently investigating it

Giustino98:
@dae 
Problem should be fixed, if you have time re-take the test, thanks

Example for parallel requests to mlx-community/gpt-oss-20b-MXFP4-Q4
127.0.0.1 - - [16/Mar/2026 10:50:43] "POST /v1/chat/completions HTTP/1.1" 200 -
2026-03-16 10:50:43,164 - INFO - KV Caches: 10 seq, 0.06 GB, latest user cache 79 tokens
127.0.0.1 - - [16/Mar/2026 10:50:43] "POST /v1/chat/completions HTTP/1.1" 200 -
2026-03-16 10:50:43,225 - INFO - KV Caches: 10 seq, 0.06 GB, latest user cache 79 tokens
127.0.0.1 - - [16/Mar/2026 10:50:43] "POST /v1/chat/completions HTTP/1.1" 200 -
2026-03-16 10:50:43,287 - INFO - KV Caches: 10 seq, 0.06 GB, latest user cache 79 tokens
127.0.0.1 - - [16/Mar/2026 10:50:43] "POST /v1/chat/completions HTTP/1.1" 200 -
2026-03-16 10:50:43,349 - INFO - KV Caches: 10 seq, 0.06 GB, latest user cache 79 tokens
127.0.0.1 - - [16/Mar/2026 10:50:43] "POST /v1/chat/completions HTTP/1.1" 200 -

As you can see there is 79 tokens reusage

This is the script I used for testing:
```
import` requests
import concurrent.futures
import time

URL = "http://127.0.0.1:8080/v1/chat/completions"

prompts = [
    "Raccontami la storia di Roma in 300 parole.",
    "Spiega la meccanica quantistica a un bambino.",
    "Scrivi una poesia sul mare.",
    "Qual è la differenza tra Python e C++?",
    "Dammi la ricetta per la carbonara perfetta."
]

def fai_richiesta(prompt, user_id):
    print(f"[User {user_id}] Richiesta inviata...")
    start = time.time()
    
    payload = {
        "messages": [{"role": "user", "content": prompt}],
        "max_tokens": 150,
        "temperature": 0.7
    }
    
    response = requests.post(URL, json=payload)
    
    end = time.time()
    print(f"[User {user_id}] Risposta ricevuta in {end-start:.2f}s!")

start_totale = time.time()
with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
    for i, prompt in enumerate(prompts):
        executor.submit(fai_richiesta, prompt, i)

print(f"Test completato in {time.time() - start_totale:.2f} secondi.")
```

dae:
That worked better, but tripped up after a while (testing pr999+1000 together):

```
Exception in thread Thread-2 (_generate):
  File "/Users/dae/.local/share/uv/python/cpython-3.12.10-macos-aarch64-none/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
[...]
  File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/generate.py", line 1257, in _next
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/generate.py", line 1126, in _process_prompts
    prompt_cache = _merge_caches(caches)
                   ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/generate.py", line 922, in _merge_caches
    batch_cache.append(caches[0][i].merge([c[i] for c in caches]))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/models/cache.py", line 603, in merge
    return BatchRotatingKVCache.merge(caches)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/models/cache.py", line 1403, in merge
    keys[i : i + 1, :, p : p + n] = c._temporal_order(c.keys)
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: [broadcast_shapes] Shapes (1,8,212,64) and (1,8,128,64) cannot be broadcast.
```

Use case is browser extension sending strings to translate in 8 parallel requests.

Giustino98:
> That worked better, but tripped up after a while (testing pr999+1000 together):
> 
> ```
> Exception in thread Thread-2 (_generate):
> Traceback (most recent call last):
>   File "/Users/dae/.local/share/uv/python/cpython-3.12.10-macos-aarch64-none/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
>     self.run()
> [...]
>   File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/generate.py", line 1257, in _next
>     batch = self._process_prompts(prompts)
>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/generate.py", line 1126, in _process_prompts
>     prompt_cache = _merge_caches(caches)
>                    ^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/generate.py", line 922, in _merge_caches
>     batch_cache.append(caches[0][i].merge([c[i] for c in caches]))
>                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/models/cache.py", line 603, in merge
>     return BatchRotatingKVCache.merge(caches)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/dae/Local/python/mlx/tmp/mlx-lm/mlx_lm/models/cache.py", line 1403, in merge
>     keys[i : i + 1, :, p : p + n] = c._temporal_order(c.keys)
>     ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
> ValueError: [broadcast_shapes] Shapes (1,8,212,64) and (1,8,128,64) cannot be broadcast.
> ```
> 
> Use case is browser extension sending strings to translate in 8 parallel requests.

Yes, this problem seems unrelated with this PR since it is present also in master branch. I will try to address it in another PR. Thanks.

zcbenz:

PR #1000, opened 2026-03-14
fix rotating kv cache merge with wrapped _idx

### LINKED ISSUE #983: BatchRotatingKVCache.merge() crashes on concurrent requests with different prompt lengths
## Description

`mlx_lm.server` crashes when handling concurrent requests to models that use `RotatingKVCache` (e.g. models with sliding window attention). The crash occurs in `BatchRotatingKVCache.merge()` when trying to merge caches with different `_idx` values, producing a shape mismatch.

This is reproducible with any client that sends multiple requests concurrently (e.g. OpenCode, which fires a short system probe and a full prompt simultaneously).

## Environment

- **mlx-lm:** 0.31.1
- **mlx:** 0.24.2
- **macOS:** 15.5 (Apple M3 Max, 48 GB)
- **Python:** 3.12
- **Model:** `mlx-community/Trinity-Nano-Preview-8bit` (afmoe architecture, uses RotatingKVCache)

Also reproduced with `mlx-community/Trinity-Mini-8bit`.

## Steps to Reproduce

1. Start the server:
```bash
uvx --from "mlx-lm>=0.28.4" mlx_lm.server \
    --model mlx-community/Trinity-Nano-Preview-8bit \
    --port 8080
```

2. Send two concurrent requests with different prompt lengths:
```bash
# In parallel (e.g. using & or a client that sends concurrent requests)
  -d '{"messages": [{"role": "user", "content": "Hi"}], "max_tokens": 10}' &

  -d '{"messages": [{"role": "system", "content": "You are a helpful assistant. '"$(python3 -c "print('x ' * 5000)")"'"}, {"role": "user", "content": "Hello"}], "max_tokens": 10}' &

wait
```

The first request succeeds, but the second crashes the server's generate thread.

## Error

```
Exception in thread Thread-2 (_generate):
  File ".../mlx_lm/server.py", line 948, in _generate
    responses = batch_generator.next()
  File ".../mlx_lm/generate.py", line 1329, in next
    return self._next()
  File ".../mlx_lm/generate.py", line 1257, in _next
  File ".../mlx_lm/generate.py", line 1126, in _process_prompts
    prompt_cache = _merge_caches(caches)
  File ".../mlx_lm/generate.py", line 922, in _merge_caches
    batch_cache.append(caches[0][i].merge([c[i] for c in caches]))
  File ".../mlx_lm/models/cache.py", line 580, in merge
    return BatchRotatingKVCache.merge(caches)
  File ".../mlx_lm/models/cache.py", line 1364, in merge
    keys[i : i + 1, :, p : p + c._idx] = c._temporal_order(c.keys)
ValueError: [broadcast_shapes] Shapes (1,2,3935,128) and (1,2,2048,128) cannot be broadcast.
```

## Analysis

In `cache.py:1364`, `BatchRotatingKVCache.merge()` assumes all caches being merged have compatible `_idx` values. When two requests have very different prompt lengths (e.g. 519 tokens vs 10,081 tokens), their `RotatingKVCache` entries end up with different `_idx` values (e.g. 3935 vs 2048), and the slice assignment fails because the shapes don't match.

The issue is in the merge logic — it allocates a target tensor based on one cache's dimensions but then tries to copy data from another cache with a different `_idx`.

## Workaround

Disabling batching in `server.py` avoids the crash:

```python
# In ModelProvider.load(), after the is_batchable check:
self.is_batchable = False
```

This forces sequential request processing. The `--prompt-cache-size 1` and `--decode-concurrency 1` flags do not prevent the crash because the server still attempts to batch concurrent requests.

## Expected Behavior

Concurrent requests with different prompt lengths should either:
- Be merged correctly (padding/truncating caches to compatible shapes), or
- Fall back to sequential processing when cache shapes are incompatible

mm65x:
fixes #983.

`BatchRotatingKVCache.merge()` uses `c._idx` to size the target slice when
copying temporally-reordered keys/values. but `_idx` is the circular buffer
write cursor, not the number of valid tokens. after the buffer wraps (offset >
max_size), `_idx` resets and can be much smaller than the actual valid length.

`_temporal_order()` always returns data with sequence length `c.size()`
(`min(offset, max_size)`), so the slice width needs to match that.

the crash shows up when the server batches concurrent requests with different
prompt lengths - a short request has a small unwrapped cache while a long
request has a wrapped cache with a different `_idx`, and the merge hits a shape
mismatch.

added a test that merges a wrapped and unwrapped cache to cover this case.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1361,8 +1361,9 @@ class BatchRotatingKVCache(_BaseCache):
         for i, (p, c) in enumerate(zip(padding, caches)):
             if c.keys is None:
                 continue
-            keys[i : i + 1, :, p : p + c._idx] = c._temporal_order(c.keys)
-            values[i : i + 1, :, p : p + c._idx] = c._temporal_order(c.values)
+            n = c.size()
+            keys[i : i + 1, :, p : p + n] = c._temporal_order(c.keys)
+            values[i : i + 1, :, p : p + n] = c._temporal_order(c.values)
 
         cache = cls(caches[0].max_size, padding)
         cache.keys = keys
--- a/tests/test_prompt_cache.py
+++ b/tests/test_prompt_cache.py
@@ -29,7 +29,6 @@ HF_MODEL_PATH = "mlx-community/Qwen1.5-0.5B-Chat-4bit"
 
 
 class TestPromptCache(unittest.TestCase):
-
     @classmethod
     def setUpClass(cls):
         cls.test_dir_fid = tempfile.TemporaryDirectory()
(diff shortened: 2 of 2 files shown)


Giustino98:
#999 already addresses this problem since trimming has been enabled for BatchRotatingKVCache in all cases

mm65x:
thanks for the heads up. i see that #999 makes `is_trimmable()` always return `True` for `BatchRotatingKVCache`, which means the trim path is available and `merge()` would be hit less often in the server's cache lookup.

that said, `merge()` itself is still buggy, as it uses `c._idx` where it should use `c.size()`, which produces wrong results whenever a cache has wrapped. even if the server avoids calling it after #999, the method is part of the public API and could be called from other contexts. seems worth fixing regardless.

angeloskath:
Fixed in #1072.

PR #1001, opened 2026-03-14
handle missing content-length header in server

### LINKED ISSUE #621: POST requests to mlx_lm.server without Content-Length header cause server error
**Description:**
`mlx_lm.server` raises an error if a `POST` request is sent to it without a `Content-Length` header. This is problematic as not all clients automatically include this header in their requests.

```
File "/opt/homebrew/Cellar/mlx-lm/0.28.3/libexec/lib/python3.14/site-packages/mlx_lm/server.py", line 292, in do_POST
    content_length = int(self.headers["Content-Length"])
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

**Expected Behavior:**
`mlx_lm.server` should handle the `POST` request gracefully and read as much data as is available from the request body, even if the `Content-Length` header is not provided.

**Relevant Code:**
```python
# server.py
def do_POST(self):
    ...
    content_length = int(self.headers["Content-Length"])
    raw_body = self.rfile.read(content_length)
```

**Minimal Solution:**
A simple approach would be to provide `-1` as default value if the `Content-Length` header is missing, so that read as much as possible is read.

```python
# server.py
def do_POST(self):
    ...
    content_length = int(self.headers.get("Content-Length", "-1"))
    raw_body = self.rfile.read(content_length)
```

mm65x:
fixes #621.

`do_POST` calls `int(self.headers["Content-Length"])` directly, which
crashes with `TypeError` when the header is missing. can happen with
clients that don't set Content-Length.

returns 411 Length Required when the header is absent, and 400 Bad
Request when it's present but not a valid integer. matches the existing
error response pattern in `do_POST`.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -172,7 +172,6 @@ def process_message_content(messages):
 
 
 class LRUPromptCache:
-
     @dataclass
     class CacheEntry:
         prompt_cache: List[Any]
@@ -357,7 +356,7 @@ class LRUPromptCache:
             else 0
         )
         logging.info(
-            f"KV Caches: {ncaches} seq, {nbytes/1e9:.2f} GB, latest user cache {ntok} tokens"
+            f"KV Caches: {ncaches} seq, {nbytes / 1e9:.2f} GB, latest user cache {ntok} tokens"
         )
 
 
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Looks great thanks!

PR #1003, opened 2026-03-14
avoid mutating input in SuScaledRoPE and YarnRoPE

### LINKED ISSUE #939: SuScaledRoPE mutates x parameter

mm65x:
fixes #939.

`SuScaledRoPE.__call__` and `YarnRoPE.__call__` scale the input with
in-place assignment (`x[..., :dim] = scale * x[..., :dim]`), which
mutates the caller's tensor. this can cause subtle bugs when the same
tensor is reused downstream (e.g. residual connections).

copies the input with `x[...]` before the in-place update, per
@angeloskath's suggestion. also fixed the same pattern in `YarnRoPE`
which had the identical issue.

checked all other RoPE classes and model files in the repo -
`DeepseekV2YarnRotaryEmbedding` already does `self.mscale * x` (no
mutation), and no other in-place scaling patterns exist.

added tests for both classes verifying the input is not mutated.

### DIFF
--- a/mlx_lm/models/rope_utils.py
+++ b/mlx_lm/models/rope_utils.py
@@ -58,6 +58,7 @@ class SuScaledRoPE(nn.Module):
         self._scale = long_mscale or (1.0 if factor <= 1.0 else default_scale(factor))
 
     def __call__(self, x, offset: Union[int, mx.array] = 0):
+        x = x[...]
         x[..., : self.dim] = self._scale * x[..., : self.dim]
         return mx.fast.rope(
             x,
@@ -71,7 +72,6 @@ class SuScaledRoPE(nn.Module):
 
 
 class Llama3RoPE(nn.Module):
-
     def __init__(
         self,
         dims: int,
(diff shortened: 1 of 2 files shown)


angeloskath:
Thanks, I 'll merge this but keep the issue open as I am preparing a better fix which is to use the `x.at[]` notation for this which is more correct and also allows for more optimizations.

So this should actually look as follows

```python
x = x.at[..., :self.dims].multiply(self._scale)
```
much cleaner in my opinion.

angeloskath (approved):

PR #1008, opened 2026-03-16
Inserting logits processors into BatchGenerator in batch_generate

arthurhjorth:
Tiny change: `logits_processors` are currently included in the signature in `generate.batch_generate()` but were unused. To actually use them in `BatchGenerator`, they need to be included when inserting prompts. This PR does that.

P.s. I said I'd do this a long time ago (https://github.com/ml-explore/mlx-lm/pull/845). Apologies for the delay, life hit hard and got in the way.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1337,7 +1337,6 @@ def batch_generate(
     max_tokens: Union[int, List[int]] = 128,
     verbose: bool = False,
     return_prompt_caches: bool = False,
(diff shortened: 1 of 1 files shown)


arthurhjorth:
By the way, in the old PR I mentioned that the API for passing in logits_processors with generate-calls is inconsistent. E.g. in `batch_generate()` they are a named arg, in single `generate()` you have to pass them in with `**kwargs`, i.e.

```
g = generate(model, tokenizer, prompts[0], **{'logits_processors': logits_processors[0]})
```

 If you'd like to make this consistent, I'd be happy to add those changes to this PR too.

PR #1014, opened 2026-03-17
Fix the rope mutation in a more natural way

angeloskath:
To be merged after https://github.com/ml-explore/mlx/pull/3266 lands in a release.


zcbenz (approved):
👍

nastya236 (commented):
Amazing change:)

PR #1019, opened 2026-03-18
Refactor LRUPromptCache

### LINKED ISSUE #1013: Suggestion: move `LRUPromptCache` from `server.py` to `models/cache.py`
`LRUPromptCache` is a general-purpose KV cache management utility. It handles prefix-trie lookup, LRU eviction, byte-budget tracking, and checkpoint support. Its usefulness is not limited to the standalone server: any inference engine that embeds mlx-lm can benefit from cross-turn KV caching.

Currently, importing it requires pulling in all of `mlx_lm.server` as a side effect (FastAPI, starlette, uvicorn, pydantic, …). This is a heavy dependency for embedders that only need the cache logic.

As example: I am working on adding cross-turn KV caching to [mlx-engine](https://github.com/lmstudio-ai/mlx-engine) (LM Studio's MLX backend) via `LRUPromptCache`, and the current import from `mlx_lm.server` pulls in the full server stack as an unintended side effect (see [this fork](https://github.com/AirRunner/mlx-engine/blob/feat/vlm-sequential-kv-cache/mlx_engine/cache_wrapper.py#L14)).

## Proposal

Move `LRUPromptCache` (and its inner classes `CacheEntry`, `CacheOrder`, `SearchResult`) to `mlx_lm/models/cache.py`, alongside `make_prompt_cache`, `trim_prompt_cache`, and `can_trim_prompt_cache`.

`LRUPromptCache`'s only non-stdlib dependencies are `can_trim_prompt_cache` and `trim_prompt_cache`, which already live in `models/cache.py`. Its remaining dependencies are pure stdlib (`copy`, `collections.deque`, `dataclasses`). `models/cache.py` does not import from `server.py`, so moving the class creates no import cycle.

angeloskath:
Refactor the LRUPromptCache into the `models.cache`. Closes #1013 .

In that spirit this PR also 
- Refactors the trie out of the lru cache
- Prepares the LRUPromptCache for a system prompt cache. Basically the LRUPromptCache is a multi-tier LRU cache where each tier is more and more persistent than the others.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -514,7 +514,7 @@ class TestLRUPromptCache(unittest.TestCase):
         self.assertEqual(c, [MockCache("test3")])
         self.assertEqual(t, [])
 
-        cache.insert_cache(model, [4, 5], [MockCache("test4")], checkpoint=True)
+        cache.insert_cache(model, [4, 5], [MockCache("test4")], cache_type="user")
         c, t = cache.fetch_nearest_cache(model, [2, 3])
         self.assertEqual(c, None)
         self.assertEqual(t, [2, 3])


Thump604:
Clean refactor. The N-tier `CacheOrder` generalization is the right move — I've been running a system prompt KV cache in vllm-mlx (PR #175) on a 122B MoE and the tiered eviction concept maps directly to what I needed: system prompts that survive across turns while user/assistant caches rotate. One thing to flag: the old `insert_cache` had a `len(self._lru) > 1` guard on the `max_bytes` eviction loop that's been dropped — this means a single large cache entry that exceeds `max_bytes` will now be immediately evicted after insertion, where previously it would have been kept as the last entry. Might be intentional but worth a note.

mlx_lm/models/cache.py
```diff
@@ -1381,3 +1383,224 @@ def nbytes(self):
...
+
+        def remove(self, model: Any, tokens: List[Any]):
+            for cache_type in self._ordering:
+                try:
+                    self._lrus[cache_type].remove((model, tokens))
+                    break
+                except ValueError:
+                    pass
+
+        def pop(self):
```
nastya236, line 1509: A couple of questions about the eviction logic in `CacheOrder.pop()`:
1. With the default ordering `["assistant", "user", "system"]`, we evict from the assistant queue until its size drops below the user queue, then start evicting user entries. What's the reasoning behind this? 
2. Would it be correct to say that when the cache is trimmable, `insert_cache` calls `pop_prefixes`, so in that case, the trie effectively have just one branch, making the eviction ordering irrelevant?

mlx_lm/server.py
```diff
@@ -655,6 +464,11 @@ def stop_and_join(self):
     def join(self):
         self._generation_thread.join()
 
+    def _log_cache_stats(self):
```
nastya236, line 467: The old version logged the latest user cache token count. The new `_log_cache_stats` only logs count and bytes. Just curious how important is logging the token count?

mlx_lm/models/cache.py
```diff
@@ -1381,3 +1383,224 @@ def nbytes(self):
...
+            tok = tokens[i - 1]
+            if len(node) > 0:
+                break
+            del parent[tok]
+        return value
+
+    def pop_prefixes(self, model: Any, tokens: List[int]):
+        values = []
+        current = self._trie[model]
+        for i in range(len(tokens) - 1):
```
nastya236, line 1437: Here -1 because we want to pop prefixes but not exact match because we don't want to pop something we just inserted, right?

nastya236:
Thank you! Very helpful refactoring.
This PR:
- Moves `LRUPromptCache` and related classes (`CacheEntry`, `CacheOrder`), and all search logic out of the `server.py` module into `models/cache.py` (where is belongs)
- introduces `PromptTrie` that encapsulates prefix tree handling that was in `LRUPromptCache`. 
Minor changes:
- moves `log_cache_stats()` to `_log_cache_stats()` of `ResponseGenerator` so it is no longer a part of cache 
- moves `parse_size()` to `utils.py`

I left couple of questions mainly for myself to understand better the code.

mlx_lm/models/cache.py
angeloskath: 1. Yep exactly this. This is done to ensure that the assistant doesn't kick out of the cache the user message. The reason we don't want the assistant to kick out the user message is that most (if not all) chat templates remove the thinking tokens before the last user message. Even more so they could insert a system message reminder and so on. The interaction looks like this (let's assume 1 slot per cache type)

1st message
```
<system> <--- system cache
<user> <--- user cache
<thinking>
<assistant>
<tool> <--- assistant cache
<thinking>
<assistant>
<tool> <--- assistant cache
```

You can see that the 2nd assistant cache will evict the user cache (actually the system cache but bear with me cause it is better for the example) if we simply treat them as a single LRU cache.

Moreover the chat template will remove all the thinking when a second user message arrives which renders the assistant caches useless (if not trimmable).

2nd message
```
<system> <--- system cache
<user> <--- user cache
<assistant>
<tool>
<assistant>
<tool>
<system reminder>
<user> <--- user cache
...
```

Since the messages changed after the user message we have to process all the messages up to the next user message which we put in the cache. You can see that we lost the two assistant cache entries, they are useless to us after the new user message.

2. Yes. When the cache is trimmable the user cache and system cache will always be included so we only need to hold the different diverging branches. `pop_prefixes` ensures that.

mlx_lm/server.py
angeloskath: Yeah not sure. I was kind of using it for debugging but it isn't too useful either. Perhaps we 'll expose some more stats later 🤷‍♂️

mlx_lm/models/cache.py
angeloskath: Yep `pop_prefixes` removes strict prefixes not including this key.

PR #1020, opened 2026-03-18
Fix flaky test

angeloskath:
As title (hopefully).

Unclear why this was flaky. It never failed locally and changing to normal and different size seems to never fail on CI either so 🤞


angeloskath:
> Thank you! Interesting why is it failing with uniform?

Switching to normal from uniform makes the values (especially for the gradients) a bit bigger which helps with some edge cases where the values are tiny. Basically it is hard to have a small relative error when the expected value is 1e-8.

But I think there is something else going on as well cause it required changing the shape as well. I have tried it replicating it on several different machines etc with no success so I 'll merge to unblock the rest of the PRs.

PR #1021, opened 2026-03-18
Add Mamba3

Goekdeniz-Guelmez:


Goekdeniz-Guelmez:
Do some weights on HF exist to test?

PR #1022, opened 2026-03-18
fix: handle ast.literal_eval failure in qwen3_coder tool parser

LeoCeasar:
## Problem

When a model generates a parameter value for an `object`/`array`/`dict`/`list` type that is neither valid JSON nor a valid Python literal, the current code raises an unhandled exception that crashes the entire request:

```
File "mlx_lm/tool_parsers/qwen3_coder.py", line 73, in _convert_param_value
    return json.loads(param_value)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 50 (char 49)
```

The `json.loads` already has a `try/except` fallback to `ast.literal_eval`, but `ast.literal_eval` itself can raise `ValueError` or `SyntaxError` for malformed values, which propagates up uncaught.

## Fix

Wrap `ast.literal_eval` in a `try/except (ValueError, SyntaxError)` and fall back to returning the raw string value. This allows the request to complete gracefully rather than crashing.

The change also consolidates the `ast.literal_eval` call so both the object/array branch and the default branch share the same error-handled path.

## Reproduction

Using a Qwen3.5 model with tool calling enabled, the model occasionally generates XML tool call parameters containing values that are syntactically invalid for both `json.loads` and `ast.literal_eval` (e.g. unquoted strings, trailing commas, or other LLM output artifacts).

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/new_model.md
+++ b/.github/PULL_REQUEST_TEMPLATE/new_model.md
@@ -1,0 +1,46 @@
+## Requirements
+
+- ☑️ I understand it is strictly prohibited to use AI to write PR description
+- AI usage disclosure: 
+
+## Model information
+
+- Name: 
+- Homepage: 
+- Reference implementation: https://github.com/huggingface/transformers/tree/main/src/transformers/models/____
+
+## Supported checkpoints
+
+- https://huggingface.co/___/___
+
+## Verification
+
+Run mlx-lm command:
+
+```console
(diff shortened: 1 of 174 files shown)


ronaldslc:
this patch is still needed, similar to https://github.com/ml-explore/mlx-lm/pull/1004

grzracz:
Still needed - bug causing downstream issues in wrappers https://github.com/jundot/omlx/issues/882

michalk8:
Seems like this was already on main, closing this.

PR #1024, opened 2026-03-19
Fix missing cache advance from qwen 3.5

angeloskath:
As title. The side-effects from this bug would only be visible during batched generation unfortunately...


nastya236:
thank you!

PR #1028, opened 2026-03-20
Fix A_log precision in mamba.py

### LINKED ISSUE #565: Numerical instability with BF16 models
Hi there, I noticed the outputs of BF16 models produce different logprobs when:
1. same input is computed in a batch vs. unbatched. 
2. comparing against HF outputs. 

Case 1 came from different padding. For example, the output logprobs would be similar if unpadded in both cases. I understand that such instability is inevitable when using padding for inference. 

For case 2, I used [tiiuae/falcon-mamba-7b-instruct](https://huggingface.co/tiiuae/falcon-mamba-7b-instruct), with mlx_lm/models/mamba.py backend in MLX-LM. I believe that the difference came from how A_log is not explicitly set to float32 as it was in mlx_lm/models/mamba2.py. The model, when loaded in HF, has all parameters in bfloat16 except A_log which are in float32.

Is this a known behavior?

eyupcanakman:
fixes #565.

mamba.py computes mx.exp(self.A_log) without casting to float32 first. When the model is loaded in bf16, the exponential loses precision and logprobs diverge from HuggingFace. mamba2.py, plamo2.py, and gated_delta.py all cast A_log to float32 at the usage site. Apply the same pattern here.

### DIFF
--- a/mlx_lm/models/mamba.py
+++ b/mlx_lm/models/mamba.py
@@ -133,7 +133,7 @@ class MambaBlock(nn.Module):
         conv_out = self.conv1d(x_full)
         new_conv_cache = x_full[:, -(K - 1) :, :]
         x = nn.silu(conv_out)
-        A = -mx.exp(self.A_log)
+        A = -mx.exp(self.A_log.astype(mx.float32))
         current_state = state_cache
         y = []
         for t in range(T):


angeloskath:
`A_log` is actually stored in float32 so no need to cast. Is there a saved model that has an issue that you can point to? Given that #565 is also stale and potentially unrelated I will close this and if you encounter an issue file an issue and we can reopen this if needed.

PR #1029, opened 2026-03-20
Fix CacheDataset.itemlen returning wrong length

### LINKED ISSUE #596: When the LoRA-invoked trainer sorts a CacheDataset by length, it doesn't appear to change ordering
It looks like `train_model` in lora.py creates a `CacheDataset` for each in dataset (for me, they're both `ChatDataset` instances).

Those are ultimately fed into tuner/trainer.py, where it sorts by length, and the following occurs:

```
if isinstance(dataset, CacheDataset):
        len_fn = lambda idx: dataset.itemlen(idx)
```

I noticed that `dataset.itemlen(idx)` in `trainer.py` appears to return 1 for all indices of my chat dataset, since they're all dictionaries of length 1 (the only key is `messages`).  Consequently, the dataset is not sorted by length.

Perhaps something should ensure that the dataset is pre-tokenized and not persisted into a `CacheDataset` as a dictionary?

If I instead use `len_fn = lambda idx: len(dataset[idx][0])` as my length function then it seems to work as intended.

Thanks for your work on this -- great library!

eyupcanakman:
fixes #596.

CacheDataset.itemlen calls len(self._data[idx]) which returns the number of dict keys (always 1 for ChatDataset), not the actual token count. Length-based batch sorting in the trainer is effectively disabled because every item looks the same length.

Changed to len(self[idx][0]) which processes the item through the tokenizer and returns the real sequence length.

### DIFF
--- a/mlx_lm/tuner/datasets.py
+++ b/mlx_lm/tuner/datasets.py
@@ -161,7 +161,7 @@ class CacheDataset:
         self._proc_data = [None] * len(data)
 
     def itemlen(self, idx: int):
-        return len(self._data[idx])
+        return len(self[idx][0])
 
     def __getitem__(self, idx: int):
         if self._proc_data[idx] is None:


axiom-of-choice:
I hit this same bug independently while diagnosing a fine-tune, and confirmed your fix is correct. Adding measured impact in case it helps this move, since a one-line diff without numbers is easy to defer.

**The bug affects all three dataset classes, not just `ChatDataset`.** `TextDataset`, `ChatDataset` and `CompletionsDataset` all return the raw record from `__getitem__`, so `len()` counts dict keys in every case:

```
current itemlen : [1, 1, 1, 1, 1]      <-- dict key count
true lengths    : [12, 402, 82, 1802, 26]
with this fix   : [12, 402, 82, 1802, 26]
```

**Measured cost of the missing sort.** 8786 rows of reasoning traces, batch 4, `max_seq_length` 2048:

| Batch order | Padded tokens | Real tokens | Waste | Mean batch width |
|---|---|---|---|---|
| current (sort disabled) | 14.16M | 10.61M | 25.1% | 1612 |
| with this fix | 10.76M | 10.61M | 1.4% | 1225 |

That is **31.6% extra compute per pass**, 3.40M wasted tokens. Row lengths were min 420, p50 1140, max 2049. The size of the win scales with length variance, so a uniform-length dataset loses nothing and a long-tailed one loses more; this number is one dataset, not a universal claim.

**No double tokenization**, which is the first thing I would have asked about. `process()` is called once per row during the sort and the later batch accesses are all cache hits:

```
process() calls: 5 for sort, 0 more on reuse (cache hit)
```

Worth stating explicitly in the PR description: the fix front-loads tokenization of the whole dataset to sort time instead of lazily per batch. Total work per pass is unchanged, but it moves the cost to startup and holds every tokenized row in memory at once. `CacheDataset` already allocates `_proc_data = [None] * len(data)` and caches every row it touches, so a full pass ends up holding everything anyway; the fix changes *when*, not *whether*. On a very large dataset that is still a real consideration, and a maintainer might prefer caching only the lengths.

**One thing likely blocking this merge:** the diff also reverts `rprint` back to `print` in three places in `load_custom_hf_dataset` and `load_dataset`, and drops the `from ..cli_ui import rprint` import. That looks like a stale branch rather than intent — `rprint` was introduced after this branch was cut. Rebasing on `main` and keeping only the `itemlen` line would make this a clean one-line change.

Also related: #766 is the same root cause reported from the API side, where `len(dataset[idx][0])` raises `KeyError: 0` on a dataset that isn't a `CacheDataset`. This fix addresses the `CacheDataset` path; that one may be worth a mention in the description.

Verified on mlx-lm at e5baded, mlx 0.32.0, M2 Pro. Happy to share the measurement script if useful.

zcbenz:

PR #1030, opened 2026-03-20
Fix per-request enable_thinking toggle in server

### LINKED ISSUE #914: Adjusting thinking/reasoning on a per-message basis?
I'm currently using GLM-5 with mlx-lm.server. There are occasions where I want to keep thinking disabled just for quick responses, but sometimes I want to temporarily enable thinking for the next completion. As far as I can tell by the way mlx-lm.server is designed, the chat template argument used to control thinking is immutable. It's either set to the default or overridden at startup time by a command-line argument.

Would it be reasonable to add the ability to override some/all chat template arguments via API call?

For comparison, Ollama uses a custom API argument "think", that does things internally to map to chat template arguments. You can arbitrarily enable/disable thinking on a per-message basis.

This could be generalized to just allow arbitrary kwarg replacements.

eyupcanakman:
fixes #914.

Re-implemented against the batch generation refactor (#1072) and Gemma4 multi-token changes (#1114). Same fix, new code.

PR #829 added chat_template_kwargs support so clients can send enable_thinking per request, but has_thinking was still read from the static tokenizer property during response handling. The prompt generation respected the override, but the state machine, prompt checkpointing, and GenerationContext did not.

Added _has_thinking() that checks per-request kwargs first, then CLI --chat-template-args, then falls back to tokenizer.has_thinking. Updated all five call sites: _tokenize initial state, _tokenize tail segment, _make_state_machine transitions, and both GenerationContext constructions.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -450,6 +450,18 @@ class ResponseGenerator:
                 f"- {cache_type}: {n_sequences} sequences, {n_bytes / 1e9:.2f} GB"
             )
 
+    def _has_thinking(self, tokenizer, args):
+        """Resolve effective thinking state for this request.
+
+        Priority: per-request kwargs > CLI args > tokenizer default.
+        """
+        if args.chat_template_kwargs and "enable_thinking" in args.chat_template_kwargs:
+            return args.chat_template_kwargs["enable_thinking"]
+        cli_args = self.model_provider.cli_args.chat_template_args
+        if "enable_thinking" in cli_args:
+            return cli_args["enable_thinking"]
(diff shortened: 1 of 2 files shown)


Thump604:
The priority chain (per-request kwargs > CLI args > tokenizer default) is the right design and fixes a real bug — setting `enable_thinking: false` per-request currently still triggers think-token stripping.

Heads up: this conflicts with PR #1006 which also modifies `_compute_prompt_checkpoint` (same function signature, both add `args` parameter). Whichever lands first, the other needs a rebase.

zcbenz:

PR #1031, opened 2026-03-20
Use model's generation_config.json for default sampling parameters

### LINKED ISSUE #140: Use a model's default generation config if it exists

eyupcanakman:
Fixes #140.

Models like Phi-4 ship a generation_config.json with sampling defaults (temperature, top_p, etc.) but mlx-lm only read eos_token_id from it. Now generate, chat, and server read these values and use them when the user does not specify an explicit override.

The priority chain is: user CLI arg > generation_config.json > hardcoded default.

The server resolves defaults lazily per request so model hot-swapping picks up the new model's config correctly.

Also adds --min-p and --top-k CLI args to chat.py (already present in generate.py) so generation_config values for those keys are not silently ignored.

### DIFF
--- a/mlx_lm/chat.py
+++ b/mlx_lm/chat.py
@@ -37,18 +37,10 @@ def setup_arg_parser():
         type=str,
         help="Optional path for the trained adapter weights and config.",
     )
-    parser.add_argument(
-        "--temp", type=float, default=None, help="Sampling temperature"
-    )
-    parser.add_argument(
-        "--top-p", type=float, default=None, help="Sampling top-p"
-    )
-    parser.add_argument(
-        "--min-p", type=float, default=None, help="Sampling min-p"
-    )
-    parser.add_argument(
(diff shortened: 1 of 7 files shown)


Thump604:
The three-tier priority (per-request > generation_config.json > hardcoded defaults) and the lazy `resolve_default()` on ModelProvider are well designed. One edge case: some HF configs set `do_sample: true` with `temperature: 1.0`, which effectively means "use default sampling." The current `do_sample: false` -> `temp: 0.0` mapping is correct, but when `do_sample: true` the code probably shouldn't inject the config's temperature since 1.0 is just the HF default, not an intentional override.

zcbenz:

Issue #956, opened 2026-03-07
Discussion: Fuse gate/up expert projections via sanitize for MoE inference speedup


BurntToastGPT:
Following up on #952, where @angeloskath suggested handling gate/up projection fusion at the model layer via `sanitize` rather than in `SwitchGLU.__call__`.

## The Optimization

SwitchGLU currently calls `gather_qmm` twice per MoE layer with identical input and expert indices — once for `gate_proj`, once for `up_proj`. Concatenating these weights and using a single `gather_qmm` call eliminates one kernel dispatch per layer per token.

## Measured Results

Benchmarked on Mac Studio M3 Ultra, 512GB, N=10 runs, mean ± std:

| Model | Family | MoE Layers | Baseline tok/s | Fused tok/s | Improvement |
|-------|--------|------------|---------------|------------|-------------|
| Qwen3-30B-A3B | Qwen | 48 | 108.7 ± 0.3 | 118.0 ± 0.3 | +8.6% |
| MiniMax M2.5 (456B) | MiniMax | 62 | 51.6 ± 0.1 | 54.2 ± 0.2 | +5.1% |
| GPT-OSS-120B | GPT-OSS | 36 | 85.4 | 89.6 | +5.0% |
| Qwen3.5-122B-A10B | Qwen | 48 | 48.7 | 51.1 | +5.0% |
| OLMoE-1B-7B | OLMoE | 16 | 357.7 | 371.3 | +3.8% |
| Qwen3.5-397B-A17B | Qwen | 60 | 7.5 | 7.5 | +0.8% |

Token-exact correctness confirmed on all models.

## Proposed Approach (per @angeloskath's feedback)

Handle fusion at the model layer through `sanitize`, similar to how QKV projections are fused:

1. In each MoE model's `sanitize()`, keep `gate_proj` and `up_proj` weights concatenated instead of splitting them (several models like Qwen3.5 already ship weights as fused `gate_up_proj` — the current sanitize splits them apart)
2. Create a `FusedSwitchGLU` variant (or modify `SwitchGLU`) that accepts pre-fused weights and does a single `gather_qmm` + split

## Questions for maintainers

- Should this be a new `FusedSwitchGLU` class, or a mode within the existing `SwitchGLU`?
- Should fusion be opt-in per model, or default for all MoE models that use `SwitchGLU`?
- For models that ship weights already fused (e.g. Qwen3.5 `gate_up_proj`): should `sanitize` simply stop splitting them?
[...]

PR #1032, opened 2026-03-21
Fuse gate/up expert projections in SwitchGLU

Thump604:
## Summary

Add `fuse_gate_up` option to `SwitchGLU` that performs a single `gather_qmm` call with concatenated gate+up weights instead of two separate calls. Eliminates one kernel dispatch per MoE layer per token.

Follows the approach proposed in #956 by @BurntToastGPT: handle fusion at the model layer via `sanitize`.

## Measured Results (from #956)

| Model | Improvement |
|-------|------------|
| Qwen3-30B-A3B | +8.6% |
| Qwen3.5-122B-A10B | +5.0% |
| MiniMax M2.5 (456B) | +5.1% |
| GPT-OSS-120B | +5.0% |
| OLMoE-1B-7B | +3.8% |

## Changes (10 files, 6 model families)

**Core:**
- `switch_layers.py`: `SwitchGLU` gets `fuse_gate_up=False` parameter. When True, creates single `gate_up_proj` SwitchLinear (2x hidden_dims). Forward pass auto-detects via `"gate_up_proj" in self`.

**Models with already-fused checkpoint weights (stop splitting):**
- `qwen3_5_moe.py`, `qwen3_vl_moe.py`: Sanitize keeps fused weights as `gate_up_proj.weight`
- `llama4.py`: Same pattern (contiguous split + swapaxes)

**Models with per-expert weights (stack + concatenate):**
- `olmoe.py`: Stack per-expert gate/up, concatenate into gate_up_proj (handles quantized weights/scales/biases)
- `mixtral.py`: Same pattern with w1/w2/w3 naming
- `minimax.py`: Same pattern (FP8 dequant then fuse)

**Constructor updates:**
- `qwen3_next.py`, `qwen3_moe.py`: Pass `fuse_gate_up=True`

**Sharding:**
- `qwen3_5.py`: Handles both fused and unfused paths

Backward compatible — `fuse_gate_up=False` (default) preserves existing behavior for all other models using SwitchGLU. GPT-OSS excluded (interleaved weights need separate handling).

## Test plan

- [ ] Verify token-exact output matches unfused path on Qwen3.5-122B
- [ ] Benchmark tok/s before/after on M2 Ultra
- [ ] Verify quantized models load correctly with fused weights
- [ ] Verify Mixtral, OLMoE, MiniMax load and generate correctly

### DIFF
--- a/mlx_lm/models/llama4.py
+++ b/mlx_lm/models/llama4.py
@@ -155,7 +155,7 @@ class MoE(nn.Module):
         self.top_k = args.num_experts_per_tok
         assert self.top_k == 1, "Only 1 expert per token supported"
         self.num_experts = args.num_local_experts
-        fuse = getattr(args, "fuse_gate_up", True)
+        fuse = getattr(args, "fuse_gate_up", False)
         self.experts = SwitchGLU(
             args.hidden_size,
             args.intermediate_size,
@@ -300,7 +300,7 @@ class Model(nn.Module):
         weights = {k: v for k, v in weights.items() if not to_remove(k)}
 
         # Rename expert weights for SwitchGLU
-        fuse = getattr(self.args.text_config, "fuse_gate_up", True)
+        fuse = getattr(self.args.text_config, "fuse_gate_up", False)
         for l in range(self.args.text_config.num_hidden_layers):
             prefix = f"language_model.model.layers.{l}.feed_forward.experts"
             if f"{prefix}.gate_up_proj" in weights:
--- a/mlx_lm/models/minimax.py
+++ b/mlx_lm/models/minimax.py
@@ -165,7 +165,7 @@ class MiniMaxSparseMoeBlock(nn.Module):
         self.num_experts_per_tok = args.num_experts_per_tok
 
         self.gate = nn.Linear(args.hidden_size, args.num_local_experts, bias=False)
-        fuse = getattr(args, "fuse_gate_up", True)
+        fuse = getattr(args, "fuse_gate_up", False)
         self.switch_mlp = SwitchGLU(
             args.hidden_size,
             args.intermediate_size,
@@ -313,7 +313,7 @@ class Model(nn.Module):
         if "model.layers.0.block_sparse_moe.experts.0.w1.weight" not in weights:
             return weights
 
-        fuse = getattr(self.args, "fuse_gate_up", True)
+        fuse = getattr(self.args, "fuse_gate_up", False)
         for l in range(self.args.num_hidden_layers):
             prefix = f"model.layers.{l}"
             w1_key = f"{prefix}.block_sparse_moe.experts.0.w1.weight"
--- a/mlx_lm/models/mixtral.py
+++ b/mlx_lm/models/mixtral.py
@@ -105,7 +105,7 @@ class MixtralSparseMoeBlock(nn.Module):
         # gating
         self.gate = nn.Linear(self.hidden_dim, self.num_experts, bias=False)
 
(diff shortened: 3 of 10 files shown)


Thump604:
**Update: Metal OOM on memory-constrained setup**

Testing on M2 Ultra 128GB with Qwen3.5-122B-A10B (5-bit, ~82GB weights), the fused `gather_qmm` with 2x output dimension causes `kIOGPUCommandBufferCallbackErrorOutOfMemory` on the first inference request. Reverting to unfused (two separate `gather_qmm` calls) works fine.

The 122B model leaves only ~46GB headroom for KV cache + Metal scratch + OS. The single larger `gather_qmm` likely requires more intermediate buffer space than two smaller calls combined.

The original benchmarks in #956 were on M3 Ultra 512GB where scratch memory isn't a constraint. This suggests the optimization needs a memory-aware fallback — perhaps fuse only when headroom is sufficient, or cap the fused dimension.

Will investigate whether `MLX_METAL_CACHE_LIMIT` or intermediate buffer sizing is the bottleneck.

Thump604:
**Update: OOM fix implemented**

Added config-driven `fuse_gate_up` (commit 801ba8c):
- `fuse_gate_up` is now read from model config via `getattr(args, "fuse_gate_up", True)` — defaults to `True` for the +5% speedup
- Set `"fuse_gate_up": false` in `config.json` for memory-constrained setups (e.g. 122B on M2 Ultra 128GB) to use separate gate/up projections
- The unfused path allows Metal to deallocate scratch memory between the two `gather_qmm` kernels
- Also fixed a bug: `qwen3_next.py` and `qwen3_moe.py` sanitize produced unfused weight names while the model init expected fused — this would break loading from HuggingFace format

All 10 model files now support both paths, controlled by a single config field.

kernelpool:
I tried this on some of my existing MiniMax DWQ quants (e.g. catalystsec/MiniMax-M2.5-4bit-DWQ) but unfortunately they no longer work with these changes, so you may want to look at the impact on existing (quantized) models.

Thump604:
Thanks @kernelpool for the report! Fixed in e7de34b.

The issue was that `fuse_gate_up` defaulted to `True`, which changed the expected weight names from `gate_proj`/`up_proj` to `gate_up_proj`. Existing quantized models have the unfused format, and quantized weights can't be concatenated (scales/biases are per-group).

**Fix:** Default is now `False` — existing models work unchanged. To opt in to the +5% speedup, set `"fuse_gate_up": true` in your model's `config.json` (should be done at quantize/convert time for new models).

Thump604:
@angeloskath @awni — this PR has been open since March 20 with no maintainer review. A community member (kernelpool) found a breaking-change issue with existing quantized models, which we fixed the same day — default is now `False` (opt-in only, zero impact on existing models).

Is there a concern with the approach? Happy to rework if needed. The fused path gives ~5% speedup on SwitchGLU MoE models for users who opt in via config.

Thump604:
Checking in on this one. The breaking-change issue kernelpool found has been resolved (default is now False, opt-in only). The PR is a config-driven performance optimization with no behavioral change unless explicitly enabled.

Happy to provide additional benchmarks or address any concerns.

Thump604:
Closing this because I’m narrowing active upstream work to vllm-mlx and I’m no longer in a position to maintain/rebase this PR properly. The underlying idea may still be useful, but I don’t want to leave an unowned open PR hanging while my production runtime no longer depends on it. If I need this again, I’ll reintroduce it from a fresh branch with current benchmarks/tests.

PR #990, opened 2026-03-13
Native MTP speculative decoding (Qwen3.5/3.6 reference implementation)

AirRunner:
## Summary

Qwen3.5 checkpoints ship with a built-in Multi-Token Prediction head (`mtp_num_hidden_layers: 1` in config) that predicts token t+2 from the backbone hidden state at t and the embedding of token t+1. This PR adds support for using it as a native speculative decoding mechanism. No separate draft model needed, at minimal extra compute (1 extra transformer layer).

## Changes

- **`mlx_lm/generate.py`**: MTP generation loop with draft/verify and probabilistic acceptance, `--mtp` CLI flag
- **`mlx_lm/models/cache.py`**: `rollback_state` slot for conv/SSM snapshot on draft rejection
- **`mlx_lm/sample_utils.py`**: `p_draw` parameter added to `apply_xtc` to share the XTC draw across draft and verify
- **`mlx_lm/models/qwen3_5.py`**: MTP head module, `self.norm` moved to `TextModel` to expose pre-norm hidden states for MTP, `n_confirmed` parameter for SSM rollback, `sanitize`: norm +1 shift now triggered only on raw HF checkpoints (unsanitized conv1d), not on presence of MTP weights, and MoE gate weights at 8-bit group=64
- **`mlx_lm/models/qwen3_5_moe.py`**: MTP checkpoint sanitization for MoE variants + handling both Qwen3.5 and Qwen3.6 (fused `gate_up_proj`)
- **`mlx_lm/server.py`**: `--mtp` flag, dynamic MTP/batch switching + fix `xtc_special_tokens` construction
- **`tests/test_mtp.py`**: 10 unit tests

## How it works

Each backbone forward pass returns both logits and pre-norm hidden states. The MTP head fuses `pre_fc_norm_hidden(h_t)` and `pre_fc_norm_embedding(embed(t+1))` via a linear projection, runs one full-attention transformer layer, and produces draft logits through the shared `lm_head`.

The generation loop verifies drafts by feeding `[confirmed_tok, draft_tok]` to the backbone with `n_confirmed=1`. This causes `GatedDeltaNet` to snapshot its conv/SSM state after the confirmed token. On acceptance, both tokens are emitted. On rejection, the SSM state is rolled back to the snapshot and KV caches are trimmed.

## Results (Qwen3.6-27B 4-bit, M4 Pro)

[...]

PR #1033, opened 2026-03-21
feat: Drop unsupported weights by default during model load

spicyneuron:
This makes `mlx-lm` drop unsupported weights by default when loading models that contain extra parameters (for example `vision_tower` weights from multimodal checkpoints).

## Problem

`mlx-lm` doesn't support vision, but models converted with [mlx-vlm](https://github.com/Blaizzy/mlx-vlm) still contain a perfectly usable text model.

Currently, attempting to load such models results in a hard crash:

```
  File "~/.local/share/uv/tools/mlx-lm/lib/python3.13/site-packages/mlx/nn/layers/base.py", line 185, in load_weights
    raise ValueError(
        f"Received {num_extra} parameters not in model: \n{extras}."
    )
ValueError: Received 333 parameters not in model:
language_model.vision_tower.blocks.0.attn.proj.bias,
language_model.vision_tower.blocks.0.attn.proj.weight,
...
```

That's especially painful right now because MLX quantizations on Hugging Face can still be a bit thin, and sometimes the only available checkpoint is a vision model. Even when that's not the case, needing to keep two versions of the same model in order to use both `mlx-lm` and `mlx-vlm` takes up a lot of storage.

## Solution

This PR makes `load_model` filter out weights that don't exist on the instantiated MLX model before loading. Unsupported extra weights are skipped and logged, while real incompatibilities like missing supported weights still fail as before.

The main change is in [`mlx_lm/utils.py`](https://github.com/ml-explore/mlx-lm/pull/1033/changes#diff-4161ef481328117f3973ece656bf81e7d339b455d40ca630199785ff02d88451).

## Alternatives

1. I initially implemented this as an opt-in `--drop-unknown-weights` flag, but after testing locally I couldn't really find a case where I would want this disabled. The change felt narrow in scope and general enough in benefit that it made more sense as the default behavior. Easy to revert if folks feel differently.

2. Add a `--disable-strict-load` style flag that toggles `strict=False`. I think that goes too far, since it would also hide incompatibilities. The current approach still preserves hard failures for missing supported weights.

### DIFF
--- a/mlx_lm/_version.py
+++ b/mlx_lm/_version.py
@@ -1,3 +1,3 @@
 # Copyright © 2023-2025 Apple Inc.
 
-__version__ = "0.31.2"
+__version__ = "0.31.3"
--- a/mlx_lm/benchmark.py
+++ b/mlx_lm/benchmark.py
@@ -73,11 +73,6 @@ def setup_arg_parser():
         default=0,
         help="Delay between each test in seconds (default: 0)",
     )
-    parser.add_argument(
-        "--drop-unknown-weights",
-        action="store_true",
-        help="Drop weights not present in the instantiated model.",
-    )
     return parser
 
 
@@ -99,11 +94,7 @@ def main():
 
     if group.size() > 1:
         model, tokenizer, config = sharded_load(
-            model_path,
-            pipeline_group,
-            tensor_group,
-            return_config=True,
-            drop_unknown_weights=args.drop_unknown_weights,
+            model_path, pipeline_group, tensor_group, return_config=True
         )
     else:
         model, tokenizer, config = load(
@@ -111,7 +102,6 @@ def main():
             return_config=True,
             tokenizer_config={"trust_remote_code": True},
             model_config={"quantize_activations": args.quantize_activations},
-            drop_unknown_weights=args.drop_unknown_weights,
         )
 
     # Empty to avoid early stopping
--- a/mlx_lm/benchmark.py
+++ b/mlx_lm/benchmark.py
@@ -73,6 +73,11 @@ def setup_arg_parser():
         default=0,
         help="Delay between each test in seconds (default: 0)",
     )
+    parser.add_argument(
+        "--drop-unknown-weights",
+        action="store_true",
+        help="Drop weights not present in the instantiated model.",
+    )
     return parser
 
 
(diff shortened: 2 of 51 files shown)


Thump604:
Heads up on a use case this would break: models with extra weight keys that are loaded by separate builders.

MTP (multi-token prediction) weights in PR #990 are stored as `mtp.*` keys alongside the base model weights. They're not in the base model's parameter tree because they're loaded separately by a dedicated builder (`text_model_from_vlm.py` in our case, `mtp_generate.py` in #990). With this PR, `tree_flatten(model.parameters())` wouldn't include `mtp.*`, so they'd all get dropped before `load_weights()` runs.

Same pattern applies to any pipeline where extra weights are co-located in the safetensors but consumed by a secondary model (adapter weights, reward heads, draft model weights, etc.).

A couple of options that would preserve the cleanup benefit without breaking these cases:
- A `keep_prefixes` parameter: `load_model(..., keep_prefixes=["mtp."])`
- Only drop + warn (current behavior) but behind a flag like `--strict-weights` that defaults to off
- Log the unknown keys as a warning but don't drop them

The warning-without-dropping approach is probably the safest default. Users who want strict behavior can opt in.

angeloskath:
Interesting. However we do have `strict=True`. Passing `strict=False` would allow loading and ignoring the unmatched weights.

Why not simply "expose" this parameter to the CLIs ?

spicyneuron:
@angeloskath I mentioned that as an alternative. `strict=False` would work but would also disable more sensible, desirable validation (missing parameters, tensor shape, types, etc). I'm not sure if there's ever a good reason for disabling those?

For the narrow case of "model has extra weights" though, I'm finding the "drop by default" behavior to be intuitive.

spicyneuron:
@Thump604 Thanks for the heads up. I'll take a look at what it'd take to warn without dropping.

EDIT: I took a look at #990, and it seems like "loaded separately by a dedicated builder" might be out of date? Current approach seems to making `mtp` a first class route. So as long as the keys survive `sanitize()`, they should be preserved by this PR. Let me know if I'm overlooking something.

spicyneuron:
@angeloskath Thoughts or feedback on this? I've been serving all local models through this without issue in the meantime.

Let me know if you're onboard with this approach or would prefer the `strict=False` instead.

zcbenz:

Issue #1015, opened 2026-03-17
generate() crashes on Metal OOM instead of recovering gracefully


alexpekarovsky:
## Problem

When using `mlx_lm.generate()` in a long-running process (daemon, server, agent), a Metal GPU out-of-memory error terminates the entire process with an uncaught C++ exception:

```
libc++abi: terminating due to uncaught exception of type std::runtime_error:
[METAL] Command buffer execution failed: Insufficient Memory
(00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
```

This happens because `generate()` and `stream_generate()` have **no try-except around the generation loop**. The `wired_limit` context manager sets limits but doesn't catch OOM — it only resets the limit in `finally`.

## Our Experience

We run Llama-3.1-8B-Instruct-4bit as a persistent subprocess for real-time inference (~50 requests/hour, 24/7). After ~14 hours of sustained inference, the Metal buffer cache fragments enough that the next generation triggers SIGABRT.

### Root Cause

- `mx.clear_cache()` releases buffers but doesn't defragment the allocator
- Over many inferences, the allocator holds fragmented regions that can't satisfy new allocations
- `mx.set_memory_limit()` caps total usage but can't prevent fragmentation within the cap
- The generation loop has no recovery path — it crashes the process

### Our Workaround (Application-Level)

We wrapped `mlx_lm.generate()` with OOM recovery in our calling code:

```python
def generate_response(model, tokenizer, prompt, max_tokens):
    import gc
    import mlx.core as mx
    from mlx_lm import generate as mlx_gen

    mx.clear_cache()  # pre-inference cleanup

    try:
        result = mlx_gen(model, tokenizer, prompt=prompt, max_tokens=max_tokens, verbose=False)
        return result
    except RuntimeError as e:
        if "memory" in str(e).lower() or "allocation" in str(e).lower():
            # OOM recovery: clear cache, GC, retry with fewer tokens
            mx.clear_cache()
            gc.collect()
            result = mlx_gen(model, tokenizer, prompt=prompt,
                             max_tokens=max(16, max_tokens // 2), verbose=False)
            return result
        raise
    finally:
        mx.clear_cache()
        if inference_count % 10 == 0:
            gc.collect()
```

We also do:
- **Periodic process recycling** (every 24h) to reset Metal allocator state entirely
- **Configurable `mx.set_memory_limit()` and `mx.set_cache_limit()`** to bound usage
[...]

PR #948, opened 2026-03-04
Improve OOM handling and add memory controls for long-running coding agent sessions

dmitryryabkov:
This PR improves mlx_lm.server stability under Metal memory pressure by adding memory limit controls and improves failure behavior for long-context, multi-turn workloads (like coding agents).

## Problem

In long-running chat sessions, active KV memory can grow until Metal OOM is hit. In practice this led to:

  - process aborts (kIOGPUCommandBufferCallbackErrorOutOfMemory)
  - unstable behavior near memory limits
  - poor user experience for tuning memory safety vs quality

The existing prompt cache size controls (`--prompt-cache-bytes`) help, but do not fully address active in-flight KV growth.

The problem was most obvious when using coding agents with mlx_lm.server.

### 1) Prompt context limit (primary unbounded memory growth control)

  - Add `--max-prompt-tokens`
  - Add `--prompt-overflow-policy` with:
      - `error` (reject oversized prompts)
      - `truncate`
  - Add `--prompt-keep-tokens` for beginning of the prompt token retention in `truncate` mode

  This is the key mechanism to stop memory growth over multiple interactions.

  ### 2) Active memory limit controls

  - Add `--max-active-kv-bytes` (admission control based on projected KV growth)
  - Add `--max-active-memory-bytes` (runtime guard using `mx.get_active_memory()` during prefill/progress)
  - Add `--prefill-step-size` to reduce prefill memory spikes by processing prompt in smaller chunks

  ### 3) Fixed active KV window option

  - Add `--max-kv-size` in server path (single and batched paths) to enable rotating-cache bounded working set where needed.
  - Document potential quality tradeoff.

  ### 4) Better OOM failure handling

  - Detect Metal OOM-like failures and map them to HTTP 503 instead of generic 404.
  - Use 500 for non-OOM server-side failures.
  - Harden batched generation failure propagation so failures are surfaced cleanly per request.

  ### 5) Documentation

  SERVER.md updated with:

  - all server memory-control flags in one section
  - explanation of how `--max-active-memory-bytes`, `--max-active-kv-bytes`, and `--max-kv-size` interact
  - practical sizing order and example values (tested on my 36GB machine, the only environment I have at the moment)

## Relationship to other in-flight / recent PRs

  - Complements cache limiting work from:
      - #906 (cache size limits)
      - #910 (friendlier cache limit UX)
  - Directly addresses the issues highlighted in:
      - #854 (server crashes on Metal OOM)
[...]

PR #1034, opened 2026-03-21
Handle Metal OOM gracefully in mlx_lm.server with structured errors

### LINKED ISSUE #854: mlx_lm.server crashes on Metal OOM instead of returning an HTTP error
## Description

When the KV cache grows beyond available GPU memory during a long conversation, `mlx_lm.server` crashes with an uncaught C++ exception instead of returning an HTTP error response and staying alive.

The server should catch the Metal out-of-memory error and return an HTTP 503 (or 500) so the client can handle it gracefully (e.g. retry, truncate context, or inform the user).

## Steps to Reproduce

1. Start `mlx_lm.server` with a large model that leaves limited headroom for KV cache (e.g. `Qwen3-Coder-Next-4bit` on a 96GB Mac — the model uses ~50GB, leaving ~46GB)
2. Send a series of multi-turn chat completion requests with tool calls, where context grows with each turn (~3900 tokens per request)
3. After ~12 requests, the KV cache exhausts available Metal memory

## Crash Output

```
127.0.0.1 - - [07/Feb/2026 01:28:11] "POST /v1/chat/completions HTTP/1.1" 200 -
libc++abi: terminating due to uncaught exception of type std::runtime_error: [METAL] Command buffer execution failed: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
[1]    68916 abort      mlx_lm.server --model mlx-community/Qwen3-Coder-Next-4bit --port 8080
/opt/homebrew/Cellar/python@3.12/3.12.12_2/Frameworks/Python.framework/Versions/3.12/lib/python3.12/multiprocessing/resource_tracker.py:279: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

## Expected Behavior

The server should catch the Metal OOM exception at the request handler level and return an HTTP 503 error response (e.g. `{"error": "Out of memory — context too large"}`) while keeping the server process alive for subsequent requests.

## Environment

- **Machine:** Apple Silicon Mac, 96GB unified memory
- **macOS:** Darwin 25.2.0
- **Python:** 3.12.12 (Homebrew)
- **mlx-lm:** 0.30.6
- **Model:** `mlx-community/Qwen3-Coder-Next-4bit`

Aristide021:
Classify generation failures in `mlx_lm.server` and return structured errors instead of crashing or misreporting as 404.

- Detect Metal/MLX OOM errors and map them to HTTP 503
- Map other generation exceptions to HTTP 500
- Return structured JSON error payloads for non-stream responses
- Emit terminal SSE error event + `[DONE]` for stream responses
- Keep server alive after OOM
- Defer non-stream 200 headers until success response is ready
- Add OOM regression tests (stream + non-stream) in `tests/test_server.py`
- Document OOM behavior and mitigation knobs in `mlx_lm/SERVER.md`
- Add extra OOM marker coverage (`insufficient memory for buffer`)
- Log classified Metal OOM events for operator visibility

Closes #854
Refs #1015

Aware of #948 (broader memory controls); this PR is intentionally scoped to crash-to-response handling and can merge independently.


Thump604:
The OOM detection markers look correct for Apple Silicon. The main paths MLX raises on unified memory exhaustion are:

1. `"failed to allocate"` from `allocator.cpp` when `MTLDevice allocateBuffer` returns nil -- this is the most common path and you've got it covered.
2. `"Metal error: command buffer execution failed due to out of memory"` from command buffer submission failure -- also covered.

One gap: when `mx.metal.set_memory_limit()` is active, MLX can throw `"Attempting to allocate X bytes which is greater than the maximum allowed buffer size"` (from the `metal::malloc` limit check). The `"failed to allocate"` marker wouldn't match that. Worth adding `"attempting to allocate"` or `"maximum allowed buffer size"` to the marker list.

The deferred-200 pattern for non-streaming is a good fix. The streaming error path handling (pre-stream vs mid-stream headers) is also correct.

Minor: the error response includes `retry_after: 30` which is a reasonable default, but memory recovery on Apple Silicon (unified memory, no separate GPU eviction) really depends on whether other processes release memory. A shorter default (5-10s) might give a better user experience for transient spikes.

zcbenz:

PR #1039, opened 2026-03-22
Fix prompt cache leak between conversations

kernelpool:
I ran into an issue where KV caches leak data into new conversations (using GLM-5), and believe this addresses it. 

After the checkpoint save, the remaining prompt tokens were fed through the model twice (once for prefill, then all of them again through _step). This inflated the KV cache offset, so trimming it for a new conversation would leave stale tokens from the previous one.

Example:

Screenshot 2026-03-24 at 1 13 35 pm

### DIFF
--- a/tests/test_generate.py
+++ b/tests/test_generate.py
@@ -632,6 +632,48 @@ class TestGenerate(unittest.TestCase):
         model = qwen3_next.Model(args)
         self._continued_generation_test_helper(model)
 
+    def test_prompt_checkpoint_cache_offset(self):
+        import copy
+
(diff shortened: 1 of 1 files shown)


angeloskath:
Thanks for flagging this. I am working on revamping the batch generation in general to support multiple checkpoints and ensure the cache key always contains all the processed tokens etc.

angeloskath:
Should be fixed in #1072 . Feel free to test it and close it is fine on my testing at least.

kernelpool:
Ran some tests, LGTM!

PR #1040, opened 2026-03-22
skip xtc sampling when threshold is zero

### LINKED ISSUE #1036: XTC threshold=0.0 default causes destructive sampling
## Summary

When a user sets `xtc_probability` without explicitly setting `xtc_threshold`, the default `xtc_threshold=0.0` causes `apply_xtc()` to mask all tokens except the least probable one. This is almost certainly not the intended behavior.

## Root cause

In `sample_utils.py:17`, `xtc_threshold` defaults to `0.0`:

```python
def make_sampler(
    ...
    xtc_probability: float = 0.0,
    xtc_threshold: float = 0.0,   # <-- here
    ...
```

The activation gate at `sample_utils.py:55` only checks `xtc_probability`:

```python
if xtc_probability > 0.0:
    sampling_methods.append(
        lambda x: apply_xtc(x, xtc_probability, xtc_threshold, xtc_special_tokens)
    )
```

Inside `apply_xtc()` at `sample_utils.py:280-281`:

```python
```

When `xtc_threshold=0.0`, all softmax probabilities are > 0.0, so `mx.where(probs > 0.0, probs, mx.inf)` returns probs unchanged. `.min()` then yields the smallest probability in the distribution. The mask selects every token with probability above that minimum, meaning all tokens except the single least probable one get masked to `-inf`.

With `xtc_probability=0.5`, this means 50% of token samples will be forced to select from only the least probable token(s).

## Steps to reproduce

```python

# User sets probability but relies on default threshold
sampler = make_sampler(temp=0.7, xtc_probability=0.5)
# Result: XTC is active with threshold=0.0, which is maximally aggressive
```

## Same default in server.py

`server.py:1233` also defaults to `0.0`:

```python
```

So API callers that send `{"xtc_probability": 0.5}` without `xtc_threshold` hit the same issue.

## Suggested fix

Default `xtc_threshold` to `0.1` (consistent with the XTC author's recommended range) in both `make_sampler()` and the server request parser. This way, setting only `xtc_probability` produces reasonable behavior out of the box.

Alternatively, if `xtc_threshold=0.0` is meant to disable XTC independently of `xtc_probability`, the activation gate at `sample_utils.py:55` should also check `xtc_threshold > 0.0`.

mm65x:
fixes #1036.

`make_sampler()` activates XTC when `xtc_probability > 0.0` but doesn't
check `xtc_threshold`. with the default `threshold=0.0`, all softmax
probabilities pass the threshold, so `apply_xtc()` masks everything
except the least probable token.

added `xtc_threshold > 0.0` to the activation gate so threshold zero
disables XTC, consistent with how probability zero already does.

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -52,7 +52,7 @@ def make_sampler(
         sampling_methods.append(lambda x: apply_top_p(x, top_p))
     if min_p != 0.0:
         sampling_methods.append(lambda x: apply_min_p(x, min_p, min_tokens_to_keep))
-    if xtc_probability > 0.0:
+    if xtc_probability > 0.0 and xtc_threshold > 0.0:
         sampling_methods.append(
             lambda x: apply_xtc(x, xtc_probability, xtc_threshold, xtc_special_tokens)
         )
--- a/tests/test_sample_utils.py
+++ b/tests/test_sample_utils.py
@@ -2,7 +2,13 @@ import unittest
 
 import mlx.core as mx
 
-from mlx_lm.sample_utils import apply_min_p, apply_top_k, apply_top_p, apply_xtc
+from mlx_lm.sample_utils import (
+    apply_min_p,
+    apply_top_k,
+    apply_top_p,
+    apply_xtc,
+    make_sampler,
+)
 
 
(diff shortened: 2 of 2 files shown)


zcbenz:

PR #1052, opened 2026-03-25
Fix BatchRotatingKVCache.merge() OOB when prompt exceeds max_size

LxYuan0420:
This PR is:
- To fix BatchRotatingKVCache.merge() writing past the merged buffer when the prompt length exceeds max_size, and add regression tests covering the prefill and autoregressive wrap-around paths

Context:
merge() used c._idx (raw prompt length) to size the destination slice, but the merged buffer is shaped to max_size. When prompt_length > max_size, this caused an out-of-bounds write. The fix uses c.size() (min(offset, max_size)) and takes the trailing n entries from _temporal_order() output.


angeloskath:
Thanks for the PR! This fix is already contained in #1072 so I will close this.

PR #1054, opened 2026-03-25
Fix IndexError in CacheOrder.pop() on empty cache

lyonsno:
## Summary

- `LRUPromptCache.trim_to(n_bytes=0)` can crash with `IndexError` if the byte-eviction loop runs after `_n_bytes` has drifted out of sync with the actual cache contents. The `while self._n_bytes > n_bytes` loop had no guard against an empty `CacheOrder`, so `pop()` could call `popleft()` on an empty deque.
- Added `len(self._lru) > 0` guard on the byte-eviction loop in `trim_to`, and an explicit empty check in `CacheOrder.pop()` for a clearer error if a future caller bypasses the guard.

## Test plan

- [x] `test_trim_to_zero_bytes_on_empty_cache` — `trim_to(n_bytes=0)` on empty cache no longer raises
- [x] `test_trim_to_zero_bytes_evicts_all` — `trim_to(n_bytes=0)` on populated cache evicts all entries cleanly

Found during production testing on Qwen3.5-122B (M2 Ultra 128GB) while reviewing #1042.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -349,6 +349,10 @@ class LRUPromptCache:
         while self._n_bytes > n_bytes and len(self._lru) > 0:
             model, tokens = self._lru.pop()
             self._delete(model, tokens)
+        if self._n_bytes > n_bytes:
+            raise RuntimeError(
+                "LRUPromptCache byte accounting drifted out of sync with cache order"
+            )
 
     def log_cache_stats(self):
         ncaches, nbytes = len(self), self.nbytes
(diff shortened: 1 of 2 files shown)


lyonsno:
Closing this stale draft to avoid keeping an old conflicted cache patch open while I prepare a separate, narrower server seed fix.

This cache-order/accounting issue may still be worth revisiting, but the code has moved since this draft was opened and it should be reworked against current  if pursued.

lyonsno:
Small typo in the close note above: I meant current `main`.

PR #1055, opened 2026-03-25
lfm2: strip lm_head.weight for tied embeddings

ykhrustalev:


ykhrustalev:
cc @Blaizzy

Blaizzy (commented):
Could you share what model weights have lm_head? 

Because this model by default shouldn't have lm_head because it has tied word embeddings.

ykhrustalev:
@Blaizzy you are right, the checkpoint I am using is little different, I will check and come back.

zcbenz:

PR #1059, opened 2026-03-26
[Experimental] Add TurboQuantKVCache: PolarQuant KV cache compression at 2-4 bits

### LINKED ISSUE #1060: PolarQuant KV cache compression (TurboQuant, ICLR 2026)
KV cache is the memory bottleneck for long context inference on Apple Silicon. The existing QuantizedKVCache uses affine quantization at 4/8 bits but doesn't go lower. PolarQuant (from Google's TurboQuant paper, ICLR 2026) offers a path to 3 bit KV cache with near lossless quality.

The algorithm is simple: apply a fixed random orthogonal rotation to each KV vector, then quantize each coordinate with a precomputed Lloyd-Max optimal scalar quantizer. The rotation makes the coordinate distribution predictable (Gaussian), so the quantizer is data oblivious. No calibration data needed.

I built a standalone implementation and benchmarked it on Apple Silicon with mlx-lm models:

**Quality (logit cosine similarity vs FP16 KV cache):**

| Model | 3-bit | 4-bit |
|---|---|---|
| Llama 3.2-3B (head_dim=128) | 0.988 | 0.997 |
| Qwen3-4B (head_dim=128) | 0.957 | 0.995 |
| Llama 3.2-1B (head_dim=64) | 0.823 | 0.974 |

Top-1 token accuracy is 4/4 at 4-bit across all models and prompts tested.

**Memory:** 4.6x compression at 3-bit, 3.8x at 4-bit (head_dim=128). Indices are bit-packed into uint32.

**Speed:** Decode is currently 0.5x FP16 because of the dequantize-on-fetch overhead. A fused Metal kernel for dequant-matmul would fix this, similar to how mx.quantized_matmul works for the existing QuantizedKVCache.

The implementation is a single file (~200 lines, pure mlx.core, no numpy) that subclasses _BaseCache. I have a draft PR at #1059 with 7 tests that pass alongside the existing 20.

Full benchmark methodology, results across 4 models, and the standalone implementation: https://github.com/rachittshah/mlx-turboquant

Paper: https://arxiv.org/abs/2504.19874

Would there be interest in adding this as an experimental cache type? Happy to adjust the approach based on feedback.

rachittshah:
Closes #1060

Adds PolarQuant KV cache compression (from Google's TurboQuant paper, ICLR 2026) as an opt-in experimental feature. The algorithm applies a fixed random orthogonal rotation to KV vectors then quantizes each coordinate with a precomputed Lloyd-Max optimal scalar quantizer. Data oblivious, no calibration needed.

Everything lives in a single new file `mlx_lm/models/turboquant.py` (~200 lines, pure mlx.core, no numpy). Nothing is imported unless `--turbo-kv-bits` is passed or `to_turboquant()` is called.

### Changes

| File | What |
|---|---|
| `models/turboquant.py` | New file. TurboQuantKVCache + PolarQuant helpers |
| `models/cache.py` | Lazy class resolver in `load_prompt_cache` (+10 lines), `to_turboquant()` on KVCache (+13 lines) |
| `generate.py` | `--turbo-kv-bits` CLI arg, `maybe_turboquant_kv_cache()` (+14 lines) |
| `tests/test_prompt_cache.py` | 7 new tests (+127 lines) |

No new dependencies. Codebook constants are inlined. Rotation matrix generated via `mx.linalg.qr`.

### Results

Tested on Apple Silicon across Llama 3.2-1B/3B and Qwen3-1.7B/4B. Logit cosine similarity vs FP16 KV cache:

| Model (head_dim) | 3-bit | 4-bit | 3-bit compression |
|---|---|---|---|
| Llama 3.2-3B (128) | 0.988 | 0.997 | 4.6x |
| Qwen3-4B (128) | 0.957 | 0.995 | 4.6x |
| Llama 3.2-1B (64) | 0.823 | 0.974 | 4.0x |

Top-1 token accuracy is perfect at 4-bit across all models and prompts. Full benchmark and methodology at https://github.com/rachittshah/mlx-turboquant/blob/main/REPORT.md

### Known limitations

Decode throughput is ~0.5x FP16 due to the dequantize-on-fetch path. A fused Metal kernel (analogous to `mx.quantized_matmul`) would close this gap but is out of scope for this PR.

Some models degrade below 4-bit (Qwen3-1.7B drops at 3-bit). Recommend 4-bit as default.

### Usage

```bash
python -m mlx_lm.generate --model mlx-community/Llama-3.2-3B-Instruct-4bit \
    --turbo-kv-bits 4 --prompt "The capital of France is"
```

```python
from mlx_lm.models.turboquant import TurboQuantKVCache
cache = [TurboQuantKVCache(bits=4) for _ in range(num_layers)]
# or convert an existing cache
tq = kv_cache.to_turboquant(bits=3)
```

### Tests

All 27 pass (20 existing + 7 new). Tests cover basic ops, quality, trim, save/load, conversion, model integration, and memory.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -204,6 +204,14 @@ def setup_arg_parser():
         default=DEFAULT_QUANTIZED_KV_START,
     )
     parser.add_argument(
+        "--turbo-kv-bits",
+        type=int,
+        help="[Experimental] Number of bits for TurboQuant KV cache "
+        "compression (2-4). Uses PolarQuant for data-oblivious compression. "
+        "Default: no TurboQuant compression.",
+        default=None,
+    )
+    parser.add_argument(
         "--draft-model",
         type=str,
         help="A model to be used for speculative decoding.",
@@ -300,6 +308,15 @@ def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_
             prompt_cache[e] = c.to_quantized(group_size=kv_group_size, bits=kv_bits)
 
 
+def maybe_turboquant_kv_cache(prompt_cache, turbo_kv_bits):
+    """Convert KV caches to TurboQuant compression (experimental)."""
+    if turbo_kv_bits is None:
+        return
+    for e, c in enumerate(prompt_cache):
+        if hasattr(c, "to_turboquant") and not hasattr(c, "turbo_bits"):
+            prompt_cache[e] = c.to_turboquant(bits=turbo_kv_bits)
+
+
 def generate_step(
     prompt: mx.array,
     model: nn.Module,
@@ -313,6 +330,7 @@ def generate_step(
     kv_bits: Optional[int] = None,
     kv_group_size: int = 64,
     quantized_kv_start: int = 0,
+    turbo_kv_bits: Optional[int] = None,
     prompt_progress_callback: Optional[Callable[[int, int], None]] = None,
     input_embeddings: Optional[mx.array] = None,
 ) -> Generator[Tuple[mx.array, mx.array], None, None]:
@@ -339,6 +357,9 @@ def generate_step(
         kv_group_size (int): Group size for KV cache quantization. Default: ``64``.
         quantized_kv_start (int): Step to begin using a quantized KV cache.
            when ``kv_bits`` is non-None. Default: ``0``.
+        turbo_kv_bits (int, optional): Number of bits for TurboQuant KV cache
+          compression (2-4). Uses PolarQuant for data-oblivious compression.
+          None implies no TurboQuant compression. Default: ``None``.
         prompt_progress_callback (Callable[[int, int], None]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
         input_embeddings (mx.array, optional): Input embeddings to use instead of or in
@@ -379,6 +400,11 @@ def generate_step(
         kv_bits=kv_bits,
     )
 
+    turboquant_cache_fn = functools.partial(
+        maybe_turboquant_kv_cache,
+        turbo_kv_bits=turbo_kv_bits,
+    )
+
     sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
 
     def _model_call(input_tokens: mx.array, input_embeddings: Optional[mx.array]):
@@ -412,6 +438,7 @@ def generate_step(
                     logits = processor(tokens, logits)
 
             quantize_cache_fn(prompt_cache)
+            turboquant_cache_fn(prompt_cache)
 
             logprobs = logits - mx.logsumexp(logits, keepdims=True)
             sampled = sampler(logprobs)
@@ -435,6 +462,7 @@ def generate_step(
                 ),
             )
             quantize_cache_fn(prompt_cache)
+            turboquant_cache_fn(prompt_cache)
             mx.eval([c.state for c in prompt_cache])
             prompt_processed_tokens += n_to_process
             prompt_progress_callback(prompt_processed_tokens, total_prompt_tokens)
@@ -1526,6 +1554,7 @@ def main():
         kv_bits=args.kv_bits,
         kv_group_size=args.kv_group_size,
         quantized_kv_start=args.quantized_kv_start,
+        turbo_kv_bits=args.turbo_kv_bits,
         draft_model=draft_model,
         num_draft_tokens=args.num_draft_tokens,
     )
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -74,8 +74,18 @@ def load_prompt_cache(file_name, return_metadata=False):
     arrays = tree_unflatten(list(arrays.items()))
     cache_metadata = tree_unflatten(list(cache_metadata.items()))
     info, metadata, classes = cache_metadata
+    def _resolve_cache_class(name):
+        if name in globals():
+            return globals()[name]
+        # Lazy-load experimental cache types
+        if name == "TurboQuantKVCache":
+            from .turboquant import TurboQuantKVCache
+
+            return TurboQuantKVCache
+        raise KeyError(f"Unknown cache class: {name}")
+
     cache = [
-        globals()[c].from_state(state, meta_state)
+        _resolve_cache_class(c).from_state(state, meta_state)
         for c, state, meta_state in zip(classes, arrays, info)
     ]
     if return_metadata:
@@ -387,6 +397,25 @@ class KVCache(_BaseCache):
                 self.values, group_size=group_size, bits=bits
             )
         return quant_cache
+
+    def to_turboquant(self, bits: int = 4):
+        """Convert to TurboQuant compressed cache (experimental).
+
+        Uses PolarQuant for data-oblivious KV cache compression at 2-4 bits.
+        See :class:`~mlx_lm.models.turboquant.TurboQuantKVCache` for details.
+
+        Args:
+            bits (int): Quantization bits per coordinate (2-4). Default: ``4``.
+        """
+        from .turboquant import TurboQuantKVCache
+
+        tq_cache = TurboQuantKVCache(bits=bits)
+        if self.keys is not None:
+            tq_cache.update_and_fetch(
+                self.keys[..., : self.offset, :],
+                self.values[..., : self.offset, :],
+            )
+        return tq_cache
 
     def make_mask(self, *args, **kwargs):
         return create_attention_mask(*args, offset=self.offset, **kwargs)
--- a/mlx_lm/models/turboquant.py
+++ b/mlx_lm/models/turboquant.py
@@ -1,0 +1,249 @@
+"""
+TurboQuant KV cache compression (experimental).
+
+PolarQuant from "TurboQuant: Redefining AI Efficiency with Extreme
+Compression" (Google, ICLR 2026, https://arxiv.org/abs/2504.19874).
+
+Data-oblivious KV cache quantization at 2-4 bits per coordinate via
+random orthogonal rotation followed by Lloyd-Max optimal scalar
+quantization. No calibration data needed.
+"""
+
+import math
+
+import mlx.core as mx
+
+from .cache import _BaseCache, create_attention_mask
+
+# fmt: off
+# Lloyd-Max optimal centroids and boundaries for N(0,1).
+# Scaled by 1/sqrt(head_dim) at runtime.
+_CENTROIDS = {
+    2: [-1.5104, -0.4528,  0.4528,  1.5104],
+    3: [-2.1519, -1.3439, -0.7560, -0.2451, 0.2451, 0.7560, 1.3439, 2.1519],
+    4: [-2.7331, -2.0698, -1.6189, -1.2570, -0.9431, -0.6573,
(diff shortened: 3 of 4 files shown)


Euda1mon1a:
I built fused Metal kernels that compute attention directly from packed PolarQuant indices — eliminates the dequantize-on-fetch step entirely. On M4 Pro:

- **Decode at 1K+ tokens**: ~2x faster than the naive dequant path
- **End-to-end at 600 tokens** (Qwen3.5-35B): 0.97x standard FP16 speed with 4.6x compression
- Lazy quantization: zero overhead below 512 tokens of context

The kernels fuse codebook lookup into the Q@K^T and scores@V matmuls so the packed indices are never materialized as full FP16 arrays. Tested on Llama-3.2-3B and Qwen3.5-35B (including hybrid linear/standard attention).

Repo with benchmarks and tests: https://github.com/Euda1mon1a/polarquant-metal

Happy to adapt this into a companion PR if there's interest in integrating it.

rewulff:
## Server-side support: a thread-stream fix unblocks `mlx_lm.server`

First off — thanks for this. The MLX quantised-KV story has lagged behind llama.cpp for a while, and the PolarQuant approach you ported is the first thing I’ve seen that makes aggressive KV compression realistic on Apple Silicon without a full fused-kernel rewrite. I came across this PR while trying to push context length on a 35B MoE inside 48 GB of unified memory, so the motivation lines up nicely with what you describe in the PR text.

Built the branch on an M4 Pro (macOS 26.3.1, Python 3.14, mlx-lm 0.31.2 from this tree) and pointed it at `unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit` with `--turbo-kv-bits 3`. `mlx_lm generate` just worked — including on Qwen3.6’s gated DeltaNet attention, which the PR description doesn’t claim to support but does, because your duck-typed check (`hasattr(c, "to_turboquant")`) is the right abstraction.

Then I tried `mlx_lm.server` and hit a wall on the first request:

```
File ".../mlx_lm/generate.py", line 1135, in _process_prompts
    mx.eval([c.state for c in prompt_cache])
```

### Root cause

`generation_stream` is created at module load (in the import thread), but MLX streams are thread-local. `mlx_lm.server` runs generation inside a dedicated worker (`Thread-1 (_generate)`), and the existing `with mx.stream(generation_stream):` wrap in `BatchGenerator.next()` tries to activate a stream that simply doesn’t exist in the worker thread.

Stock `KVCache` dodges this by accident: its `.state` property returns already-evaluated slices, so `mx.eval([c.state ...])` is effectively a no-op. `TurboQuantKVCache` quantises on-demand inside `update_and_fetch`, so the same `mx.eval` call forces pending GPU ops to run — and that’s where the missing stream bites. It’s a latent bug that your cache just happens to expose cleanly.

### Suggested fix (minimal)

Create a worker-local stream on first use in `BatchGenerator.next()`:

```python
def next(self):
    # Module-level generation_stream is bound to the import thread;
    # BatchGenerator runs in a dedicated worker thread (mlx_lm.server).
    # Create a thread-local stream on first use so GPU-backed cache ops
    # (e.g. TurboQuantKVCache) work inside the worker.
    if getattr(self, "_worker_stream", None) is None:
        self._worker_stream = mx.new_stream(mx.default_device())
    with mx.stream(self._worker_stream):
        return self._next()
```

One change, no behavioural difference for stock caches, and it closes the latent thread-ownership hole regardless of whether TurboQuant is in play.

### Server CLI integration (nice-to-have)

To actually exercise the path from `mlx_lm.server`, I mirrored the `--turbo-kv-bits` flag from `generate.py` and added a tiny helper that upgrades caches right after `make_prompt_cache(...)`:

```python
def _maybe_turboquant(cache_list, turbo_kv_bits):
    if not turbo_kv_bits:
        return
    for e, c in enumerate(cache_list):
        if hasattr(c, "to_turboquant") and not hasattr(c, "turbo_bits"):
            cache_list[e] = c.to_turboquant(bits=turbo_kv_bits)
```

Called at the two spots in `server.py` where `cache = make_prompt_cache(self.model_provider.model)` (and the draft-model path) produces a new cache. Plus one `parser.add_argument("--turbo-kv-bits", type=int, default=None, ...)` at the bottom of the argparse block. Same UX as your CLI flag, so users don’t have to learn anything new.

### Real-world validation

With both patches applied:

```
mlx_lm server \
  --model .../Qwen3.6-35B-A3B-UD-MLX-4bit \
  --turbo-kv-bits 3 \
  --chat-template-args '{"enable_thinking":false}'
```

Boots cleanly, answers OpenAI-compatible requests, stays stable across an agent session. The more interesting test was driving an OpenCode agent through a full landing-page generation: 425 lines of HTML with styling and inline JS, in German, followed by a shell tool call to open the file. Done in under three minutes, no regressions, no drift. That’s the kind of workload where a compressed KV actually matters — you want to keep the full conversation context around without blowing past the unified-memory ceiling.

For comparison: the same model on llama.cpp with `q8_0` KV and ctx 131k was eating ~82% of our Metal budget and occasionally OOM’ing under sustained load. Your 3-bit path brings that down dramatically while Qwen3.6’s output quality stays indistinguishable at the single-shot level. We haven’t done a proper perplexity sweep yet, but for coding-agent work there’s no visible degradation.

Happy to open a follow-up PR with these two diffs if that’s easier than threading them into this one. Either way — this is really nice work, and it unlocks a category of workloads on Apple Silicon that previously forced people back to llama.cpp. Thanks again.

rewulff:
### Follow-up: multi-skill agent orchestration survives the 3-bit path

Quick update after a day of using the patched setup in anger. Ran a full agent session with OpenCode where the orchestrator (Qwen3.6-35B-A3B at `--turbo-kv-bits 3`) had to chain four separate skills and tools in one conversation:

1. HTML landing-page generation from scratch (425 lines, German copy, CSS + inline JS)
2. Shell tool call to open the file in a browser
3. Delegation to a local image-generation skill (external 2-minute `mflux` run — correctly parsed stdout, kept the output path, waited it out)
4. Follow-up edit on the original HTML: read back the 425-line file, inject a new CSS block, patch the `<section class="hero">` to reference the generated image

All context preserved across turns, no drift, no dropped tool calls. That’s the workload class where compressed KV actually matters — you want to keep a long running conversation in-budget without OOM-dancing. It held up cleanly.

One minor observation worth mentioning in case it’s not already known: `--chat-template-args '{"enable_thinking":false}'` is respected on complex turns but appears to leak through on trivial ones (single-word "hello" still emitted a few thinking tokens with the Qwen3.6 chat template). Doesn’t affect tool use, doesn’t affect the KV path, just a cosmetic note in case other reviewers see the same.

Again — happy to open a follow-up PR with the two diffs (the thread-local stream fix in `BatchGenerator.next()` + the `--turbo-kv-bits` wiring in `server.py`) whenever it’s useful. Thanks for the work.

rewulff:
Hi @rachittshah — follow-up on our earlier comments (Thread-stream fix + server-side `--turbo-kv-bits` hook). After a broader benchmark on M4 Pro 48 GB with Qwen3.6-35B-A3B, we think there's a **design-level issue** in how the TurboQuant conversion is sequenced in `generate_step()` that makes the feature effectively unusable on long contexts via the CLI / `stream_generate` path.

### Observation

On the rebased fork (`feat/turboquant-kv-cache` rebased onto upstream `main @ ed1fca4`, mlx-lm 0.31.3):

| Test                                       | Context | Peak memory   | Result                                       |
|--------------------------------------------|---------|---------------|----------------------------------------------|
| Sanity, single-shot, `--turbo-kv-bits 3`   | 8k      | 10.5 GB       | PASS (needle retrieved)                      |
| Single-shot, `--turbo-kv-bits 3`           | 131k    | —             | **Metal OOM** (`kIOGPUCommandBufferCallbackErrorOutOfMemory`) |
| Short prompt (23 tok), no turbo, Gemma-4   | —       | 20.7 GB       | PASS, 58 t/s generation                      |
| Short prompt (23 tok), `--turbo-kv-bits 3`, Gemma-4 | —  | 21.1 GB       | PASS, **28 t/s** generation (2× slowdown)    |

Both symptoms — OOM on long context and generation slowdown on short context — point to the same root cause.

### Root cause

In `mlx_lm/generate.py` around line 443 (rebased), inside `generate_step`'s `_step` closure:

```python
def _step(input_tokens):
        logits = _model_call(input_tokens=input_tokens[None], ...)
        # ... logits processing ...
        quantize_cache_fn(prompt_cache)      # q8 conversion
        turboquant_cache_fn(prompt_cache)    # turbo conversion — only AFTER the step
```

The **first** `_step` invocation does the full-prompt prefill — `_model_call` fills a default `KVCache` (fp16) for the entire prompt. The turbo conversion runs *after* the forward pass is complete.

Implications:

1. **Peak-memory doubling during prefill.** For ctx 131k on Qwen3.6-35B (GQA: 2 KV heads, head_dim 256, 40 layers), fp16 K+V ≈ 21 GB. Adding model weights (~19 GB, 4-bit) and activations overshoots the 48 GB budget, even though the *post-conversion* 3-bit cache would only be ~4 GB.
2. **Per-step check overhead during generation.** `turboquant_cache_fn` iterates the cache in every `_step`. After the first conversion it's just a `hasattr` check per layer, but the TurboQuant cache still participates in each decode step — for Gemma-4 at short context, we see ~2× slower generation than the non-turbo path.

### Proposed fix

Lift the TurboQuant conversion out of `_step` and call it once, before any forward pass, in `generate_step`:

```python
def generate_step(..., turbo_kv_bits=None, ...):
    # ... existing setup ...

    # PRE-convert cache before any forward pass
    if turbo_kv_bits is not None:
        maybe_turboquant_kv_cache(prompt_cache, turbo_kv_bits)

    # ... then existing _step WITHOUT the turboquant_cache_fn(prompt_cache) call
```

Keep `quantize_cache_fn` (q8) inside `_step` — that one has threshold semantics via `quantized_kv_start` and is legitimately step-dependent. TurboQuant conversion is binary (once converted, stay converted), so lifting it is clean.

### Evidence the fix works

Three workarounds in our repo already do pre-conversion and all run reliably:

1. **Server patch** (our `server.py` change, see our earlier PR comment): `_maybe_turboquant(cache, turbo_kv_bits)` is called immediately after `make_prompt_cache()`, before `stream_generate`. Production usage on Port 1236 with ctx up to 131k has been stable for a week at ~31 GB peak.
2. **Perplexity benchmark script**: Pre-convert empty cache, feed prompt tokens in manual chunks. 6 runs × 50k tokens on 48 GB completed without OOM, peak 21.4–22.6 GB. Delta of perplexity to fp16-KV: +1.0% to +2.1% — quality is fine; the issue is only *how* the conversion is sequenced.
3. **Direct ctx-131k+ needle test with pre-converted cache**: Same Qwen3.6-35B-A3B setup that OOM'd via the `stream_generate(turbo_kv_bits=3)` path — after switching to pre-convert + manual chunked prefill + manual decode loop, ctx 131k / 180k / 230k / 262k all run successfully, needle retrieved correctly in each. Peak real-memory (Activity Monitor "Memory" column, which includes Metal / IOAccelerator unified-memory that Python `psutil.rss` misses) lands around 30–35 GB on a 48 GB M4 Pro, with some swap activity at the upper end. The clean A/B — OOM at 48 GB via the default path vs. stable 30-ish GB via pre-convert — isolates the post-step conversion as the cause.

We'd be happy to open a PR with the fix if you'd like. Let us know if there's a reason the conversion was placed post-step that we're missing.

zcbenz:

PR #1066, opened 2026-03-27
Fix gated delta kernel precision

### LINKED ISSUE #1061: Qwen3.5-35B-A3B-4bit can emit malformed tool-call output around 20k prompt tokens.
I am seeing malformed tool-call output with `Qwen3.5-35B-A3B-4bit`.

  The important part is that this is reproducible with a direct `mlx_lm.stream_generate(...)` call, without going through any OpenAI-compatible server wrapper and without any downstream parser.

  ## Environment

  - `mlx_lm==0.31.1`
  - model: mlx-community/Qwen3.5-35B-A3B-4bit
  - prompt length in this repro: `20043` tokens
  - tool calling enabled
  - sampling used in the repro:
    - `temperature=0.7`
    - `top_p=0.95`
    - `top_k=0`
    - `min_p=0.0`
    - `repetition_penalty=1.3`
    - `presence_penalty=0.0`
    - `frequency_penalty=0.0`
    - `max_tokens=32768`
    - `seed=3`
  - chat template kwargs:
    - `{"enable_thinking": false}`

  ## What I attached

  - [request.json](https://github.com/user-attachments/files/26272812/request.json)
    - direct replay bundle for `mlx_lm`
    - contains normalized `messages`, `tools`, `sampling`, and the fully rendered `prompt`
  - [test_script.py](https://github.com/user-attachments/files/26272811/test_script.py)
    - minimal script
    - just loads the model and streams the raw output
    - no parsing, no post-processing, no OpenAI server layer

  ## Reproduction

  Edit the model path on line 7 of request.json.

  Run:

  ```bash
  python test_script.py request.json
 ```

  The script simply does:

  - load(model_path)
  - mx.random.seed(seed)
  - stream_generate(...)
  - print raw text to stdout

  ## Expected behavior

  The model should either:

  - emit a valid, fully closed tool-call block, or
  - answer in plain text

  It should not emit an incomplete / malformed tool-call block.

  ## Actual behavior

  With the attached prompt and seed=3, the direct raw generation produces malformed output like this:

  好嘞！我这就用 `agent-browser` 来查：

  <tool_call>
  <function=browser>

  This is a direct mlx_lm replay, so this is not caused by an external OpenAI-compatible server wrapper or a client-side parser.

  ## Additional observation

  This specific attached request is the last clean request before malformed tool-call content started polluting later history.

  After one malformed tool-call appears in history, later turns become much more likely to produce even worse corrupted blocks, for example:

  - unclosed <tool_call>
  - unclosed <function=...>
  - stray tags like </arg_key>
  - malformed parameter sections such as binaryNameFromEntryPointMap

  So there seem to be two separate issues:

  1. the first malformed tool-call can already happen from a clean context
  2. once it happens, feeding that malformed output back into history causes strong cascading contaminatione

  ### As a comparison, I tested the GGUF version and the exact same issue does not occur. 

  ## Secondary note about mlx_lm.server

  I also tested the same context through mlx_lm.server in streaming mode.

  In that path, the client only received the leading assistant text, then the stream terminated without [DONE], and the server raised:

  ValueError: No function provided.

  inside the tool parser.

  That server behavior is secondary, though. The main issue here is that the direct raw mlx_lm generation already produces malformed tool-call output.

kernelpool:
Adds Kahan compensation and unrolls loops. Fixes #1061

mlx-community/Qwen3.5-35B-A3B-4bit

| Context | Before pp | After pp | Δ pp | Before tg | After tg | Δ tg | Before Mem | After Mem | Δ Mem |       
  |---------|----------|---------|------|----------|---------|------|-----------|----------|-------|             
  | 0.5k | 2040.1 | 2039.2 | -0.0% | 82.1 | 83.6 | +1.8% | 20.24 GB | 20.24 GB | 0.0% |                          
  | 1k | 2574.4 | 2551.9 | -0.9% | 81.7 | 84.1 | +2.9% | 20.57 GB | 20.57 GB | 0.0% |                            
  | 2k | 2830.8 | 2818.4 | -0.4% | 82.2 | 85.8 | +4.4% | 21.40 GB | 21.40 GB | 0.0% |                            
  | 4k | 2905.8 | 2889.3 | -0.6% | 85.0 | 84.8 | -0.3% | 21.66 GB | 21.65 GB | 0.0% |                            
  | 8k | 2845.5 | 2827.4 | -0.6% | 84.2 | 84.3 | +0.2% | 22.02 GB | 22.02 GB | 0.0% |                            
  | 16k | 2627.5 | 2614.6 | -0.5% | 83.3 | 84.1 | +1.0% | 22.70 GB | 22.70 GB | 0.0% |                           
  | 32k | 2245.7 | 2235.4 | -0.5% | 80.3 | 79.9 | -0.4% | 23.96 GB | 23.96 GB | 0.0% |                           
  | **Avg** | **2581.4** | **2568.0** | **-0.5%** | **82.7** | **83.8** | **+1.4%** | | | |                      
  | **PPL** | **4.14** | **4.14** | **0.0%** | | | | | | |

### DIFF
--- a/mlx_lm/models/gated_delta.py
+++ b/mlx_lm/models/gated_delta.py
@@ -59,37 +59,44 @@ def _make_gated_delta_kernel(has_mask=False, vectorized=False):
         {g_setup}
         auto beta_ = beta + b_idx * T * Hv;
 
-        for (int t = 0; t < T; ++t) {{
-          if ({mask_source}) {{
-            float kv_mem = 0.0f;
-            for (int i = 0; i < n_per_t; ++i) {{
-              auto s_idx = n_per_t * dk_idx + i;
-              state[i] = state[i] * {g_access};
-              kv_mem += state[i] * k_[s_idx];
-            }}
-            kv_mem = simd_sum(kv_mem);
-
-            auto delta = (v_[dv_idx] - kv_mem) * beta_[hv_idx];
-
-            float out = 0.0f;
-            for (int i = 0; i < n_per_t; ++i) {{
-              auto s_idx = n_per_t * dk_idx + i;
-              state[i] = state[i] + k_[s_idx] * delta;
-              out += state[i] * q_[s_idx];
-            }}
-            out = simd_sum(out);
-            if (thread_index_in_simdgroup == 0) {{
-              y[dv_idx] = static_cast<InT>(out);
-            }}
-          }}
-          // Increment data pointers to next time step
-          q_ += Hk * Dk;
-          k_ += Hk * Dk;
-          v_ += Hv * Dv;
-          y += Hv * Dv;
-          {g_advance}
-          beta_ += Hv;
+        #define BODY() {{ \
+            float kv_mem = 0.0f, kv_c = 0.0f; \
+            for (int i = 0; i < n_per_t; ++i) {{ \
+              auto s_idx = n_per_t * dk_idx + i; \
+              state[i] *= {g_access}; \
+              auto p = state[i] * k_[s_idx]; \
+              auto a = p - kv_c; \
+              auto b = kv_mem + a; \
+              kv_c = (b - kv_mem) - a; \
+              kv_mem = b; \
+            }} \
+            kv_mem = simd_sum(kv_mem); \
+            auto delta = (v_[dv_idx] - kv_mem) * beta_[hv_idx]; \
+            float out = 0.0f; \
+            for (int i = 0; i < n_per_t; ++i) {{ \
+              auto s_idx = n_per_t * dk_idx + i; \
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Kahan compensated summation for the kv_mem accumulation is the right fix. On Qwen3.5-122B with 36 GDN layers, precision drift in the recurrence state is a real problem at long context. This addresses the root cause of #1061.

The +2-4% tg improvement from loop unrolling is a nice bonus. No memory impact, no prefill regression.

Tested the concept independently on M2 Ultra 128GB — the GDN recurrence does accumulate error over thousands of steps without compensation.

kernelpool:
Closed this since this was somewhat experimental

PR #1074, opened 2026-03-30
feat: QuantizedRotatingKVCache + KVSplit (K/V different bits)

### LINKED ISSUE #191: Possibility for KVSplit support in mlx-lm?
I recently learned about KVSplit (https://github.com/dipampaul17/KVSplit), a technique where you quantize the keys and values to different precisions. (Their implementation uses patched llama.cpp in Apple Silicon). Apparently, 8-bit quantization for keys and 4-bit quantization for values is significantly more efficient for basically the same performance. Is this feature easy to implement here?

deceptech-packet-ninja:
## Summary

Two related features for quantized KV cache:

1. **QuantizedRotatingKVCache**: combines bounded-size rotation (`RotatingKVCache`) with quantized storage (`QuantizedKVCache`). Fixes the `NotImplementedError` in `RotatingKVCache.to_quantized()`. Enables `--max-kv-size` + `--kv-bits` together.

2. **KVSplit**: allows different quantization bits for keys vs values. Keys need higher precision (they determine attention routing) while values tolerate lower precision. `bits` and `group_size` accept `(key, value)` tuples.

## Motivation

**QuantizedRotatingKVCache**: Users cannot combine `--max-kv-size` (bounded context) with `--kv-bits` (quantized cache). This crashes with `NotImplementedError("RotatingKVCache Quantization NYI")`. This means bounded-context inference cannot benefit from KV cache quantization, leaving memory savings on the table. Related to #883 (OOM from unbounded KV cache growth).

**KVSplit**: Research shows keys need higher precision than values for attention quality. The maintainer noted in #191 that this would be "pretty straightforward." For example, `kv_bits=(8, 4)` gives 8-bit keys (accurate routing) with 4-bit values (memory savings), a practical sweet spot.

## Changes

**`mlx_lm/models/cache.py`**:
- `QuantizedKVCache`: added `key_bits`, `value_bits`, `key_group_size`, `value_group_size` with backward-compatible `bits`/`group_size` properties. Updated `update_and_fetch` to quantize K/V with their respective params. Updated `meta_state` with backward compat for the old 3-field format.
- `RotatingKVCache.to_quantized()`: now returns `QuantizedRotatingKVCache` instead of raising `NotImplementedError`.
- `QuantizedRotatingKVCache` (new): inherits from `RotatingKVCache`, reuses `update_and_fetch`, `size`, `make_mask`, `is_trimmable`, `trim` from parent. Overrides `_trim`, `_temporal_order`, `_update_concat`, `_update_in_place`, `state`, `meta_state`, `nbytes` for quantized tuple storage via `tree_map`.

**`mlx_lm/models/base.py`**:
- `quantized_scaled_dot_product_attention`: added `value_group_size` and `value_bits` optional params (default to key params for backward compat).
- `scaled_dot_product_attention`: passes separate K/V params via `getattr` with fallbacks.

**`mlx_lm/generate.py`**: added `QuantizedRotatingKVCache` to imports.

## Test plan

All configurations tested on Qwen2.5-0.5B-Instruct-4bit and Llama-3-8B-Instruct-4bit:

- [x] Standard BF16 — unchanged behavior
- [x] `kv_bits=4` — quantized cache works
- [x] `kv_bits=8` — quantized cache works
- [x] `kv_bits=(8, 4)` — KVSplit: keys 8-bit, values 4-bit
- [x] `kv_bits=(4, 2)` — KVSplit: keys 4-bit, values 2-bit
- [x] `max_kv_size=128` — rotating cache works
- [x] `max_kv_size=128, kv_bits=4` — **previously crashed**, now works
- [x] `max_kv_size=128, kv_bits=(8, 4)` — rotating + KVSplit works
- [x] `meta_state` round-trip with old 3-field format (backward compat)
- [x] `meta_state` round-trip with new 5-field format (KVSplit)
- [x] 100 decode steps past `max_size` with `keep=4` — rotation logic correct
- [x] `to_quantized()` after cache has wrapped — preserves offset/idx
- [x] Repeated `to_quantized()` calls (idempotent — returns self)

Closes #191. Addresses #968 (maintainer feedback: inherits from RotatingKVCache instead of duplicating logic).

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -31,6 +31,7 @@ from .models.cache import (
     CacheList,
     KVCache,
     QuantizedKVCache,
+    QuantizedRotatingKVCache,
     RotatingKVCache,
     load_prompt_cache,
 )
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -69,7 +69,15 @@ def quantized_scaled_dot_product_attention(
     mask: Optional[mx.array],
     group_size: int = 64,
     bits: int = 8,
+    value_group_size: Optional[int] = None,
+    value_bits: Optional[int] = None,
 ) -> mx.array:
+    # Default value params to key params for backward compat
+    if value_group_size is None:
+        value_group_size = group_size
+    if value_bits is None:
+        value_bits = bits
+
     B, n_q_heads, L, D = queries.shape
     n_kv_heads = q_keys[0].shape[-3]
     n_repeats = n_q_heads // n_kv_heads
@@ -96,7 +104,11 @@ def quantized_scaled_dot_product_attention(
             scores += mask
     scores = mx.softmax(scores, axis=-1, precise=True)
     out = mx.quantized_matmul(
-        scores, *q_values, transpose=False, group_size=group_size, bits=bits
+        scores,
+        *q_values,
+        transpose=False,
+        group_size=value_group_size,
+        bits=value_bits,
     )
 
     if n_repeats > 1:
@@ -123,8 +135,10 @@ def scaled_dot_product_attention(
             values,
             scale=scale,
             mask=mask,
-            group_size=cache.group_size,
-            bits=cache.bits,
+            group_size=getattr(cache, "key_group_size", cache.group_size),
+            bits=getattr(cache, "key_bits", cache.bits),
+            value_group_size=getattr(cache, "value_group_size", cache.group_size),
+            value_bits=getattr(cache, "value_bits", cache.bits),
         )
     else:
         return mx.fast.scaled_dot_product_attention(
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, List, Optional
 
 import mlx.core as mx
 import mlx.nn as nn
-from mlx.utils import tree_flatten, tree_map, tree_unflatten
+from mlx.utils import tree_flatten, tree_map, tree_reduce, tree_unflatten
 
 from .base import create_causal_mask
 
@@ -234,8 +234,42 @@ class QuantizedKVCache(_BaseCache):
         self.keys = None
         self.values = None
         self.offset = 0
-        self.group_size = group_size
-        self.bits = bits
+        # Support KVSplit: bits/group_size may be (key, value) tuples
+        if isinstance(bits, (tuple, list)):
+            self.key_bits, self.value_bits = bits
+        else:
+            self.key_bits = self.value_bits = bits
+        if isinstance(group_size, (tuple, list)):
+            self.key_group_size, self.value_group_size = group_size
+        else:
+            self.key_group_size = self.value_group_size = group_size
+
+    # Backward-compatible properties
+    @property
+    def bits(self):
+        if self.key_bits == self.value_bits:
+            return self.key_bits
+        return (self.key_bits, self.value_bits)
+
+    @bits.setter
+    def bits(self, v):
+        if isinstance(v, (tuple, list)):
+            self.key_bits, self.value_bits = v
+        else:
+            self.key_bits = self.value_bits = v
+
+    @property
+    def group_size(self):
+        if self.key_group_size == self.value_group_size:
+            return self.key_group_size
+        return (self.key_group_size, self.value_group_size)
+
+    @group_size.setter
+    def group_size(self, v):
+        if isinstance(v, (tuple, list)):
+            self.key_group_size, self.value_group_size = v
+        else:
+            self.key_group_size = self.value_group_size = v
 
     def update_and_fetch(self, keys, values):
         B, n_kv_heads, num_steps, k_head_dim = keys.shape
@@ -243,15 +277,15 @@ class QuantizedKVCache(_BaseCache):
         prev = self.offset
 
         if self.keys is None or (prev + num_steps) > self.keys[0].shape[-2]:
-            el_per_int = 8 * mx.uint32.size // self.bits
             new_steps = (self.step + num_steps - 1) // self.step * self.step
             shape = (B, n_kv_heads, new_steps)
 
-            def init_quant(dim):
+            def init_quant(dim, bits, gs):
+                el_per_int = 8 * mx.uint32.size // bits
                 return (
                     mx.zeros((*shape, dim // el_per_int), dtype=mx.uint32),
-                    mx.zeros((*shape, dim // self.group_size), dtype=keys.dtype),
-                    mx.zeros((*shape, dim // self.group_size), dtype=keys.dtype),
+                    mx.zeros((*shape, dim // gs), dtype=keys.dtype),
+                    mx.zeros((*shape, dim // gs), dtype=keys.dtype),
                 )
 
             def expand_quant(x):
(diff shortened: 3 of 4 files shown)


Thump604 (approved):
I run Qwen3.5-122B-A10B (5-bit, ~82GB weights) on M2 Ultra 128GB in production with continuous batching and MTP. KV cache quantization is on my critical path for 1M context support.

I independently built a similar asymmetric K8/V4 cache class and tested it extensively. A few findings that may be useful:

**Architecture:** My implementation dequantized the full cache to FP16 before passing to SDPA, which caused a 54% latency regression at 100K context. This PR's approach of keeping K/V in quantized tuple form and routing through `mx.quantized_matmul` is the right architecture since dequantization is kernel-fused. Nice work.

**Hadamard rotation:** I tested applying a Walsh-Hadamard transform before quantization to distribute outlier energy. At K8, it provides negligible quality improvement (1.01x MSE ratio vs no rotation). At K4, it helps but K4 fails on Qwen3.5 models due to high K tensor kurtosis. So no rotation is needed for the K8/V4 path this PR implements.

**Request:** Latency benchmarks at 100K+ context comparing FP16 KVCache vs QuantizedKVCache(bits=(8,4)) would strengthen this PR. I can run these on my M2 Ultra if helpful.

Looks good to me.

deceptech-packet-ninja:
## Benchmark Results

Ran latency benchmarks per @Thump604's request. Machine: Apple Silicon, 32GB unified memory.

**Model:** `mlx-community/Meta-Llama-3-8B-Instruct-4bit` (32 layers, 8 KV heads, D=128)

### Decode Latency vs Context Length

| Context | FP16 (ms) | FP16 (t/s) | K8/V4 (ms) | K8/V4 (t/s) | Q4 (ms) | Q4 (t/s) | K8/V4 vs FP16 |
|--------:|----------:|-----------:|-----------:|------------:|--------:|---------:|--------------:|
| 1,024   | 49.0      | 20.4       | 51.9       | 19.3        | 53.8    | 18.6     | 0.94x         |
| 4,096   | 55.9      | 17.9       | 59.8       | 16.7        | 54.1    | 18.5     | 0.93x         |
| 16,384  | 80.0      | 12.5       | 75.4       | 13.3        | 74.6    | 13.4     | **1.06x**     |
| 32,768  | 125.4     | 8.0        | 123.6      | 8.1         | 92.4    | 10.8     | **1.01x**     |

**Key finding:** K8/V4 crosses over to faster-than-FP16 at ~16K context, where attention becomes memory-bandwidth-bound. At 32K it's essentially even. Q4 is consistently faster at long contexts (1.36x at 32K) since it reads the least data.

### Estimated KV Cache Memory

| Context | FP16 | K8/V4 | Q4 | K8/V4 savings |
|--------:|-----:|------:|---:|--------------:|
| 1K      | 128 MB  | 80 MB   | 48 MB  | 38% |
| 4K      | 512 MB  | 320 MB  | 192 MB | 38% |
| 16K     | 2.0 GB  | 1.3 GB  | 768 MB | 38% |
| 32K     | 4.0 GB  | 2.5 GB  | 1.5 GB | 38% |
| 64K     | 8.0 GB  | 5.0 GB  | 3.0 GB | 38% |
| 128K    | 16.0 GB | 10.0 GB | 6.0 GB | 38% |

At 128K context on a 32GB machine with a ~4.5GB model, FP16 KV would use 16GB (total 20.5GB, tight). K8/V4 uses 10GB (14.5GB total, comfortable). Q4 uses 6GB (10.5GB, plenty of headroom).

### Output Quality

All three configs produce identical output for short generations:

```
FP16:  Gravity is a force that pulls objects towards each other. It is what keeps us on the ground and what...
K8/V4: Gravity is a force that pulls objects towards each other. It is what keeps us on the ground and what...
Q4:    Gravity is a force that pulls objects towards each other. It is what keeps us on the ground and what...
```

@Thump604 — if you're able to run these on your M2 Ultra with Qwen3.5-122B at 100K+ context, that would complement these results nicely. The longer the context, the more the bandwidth savings should show.

Thump604:
Great data. The crossover at 16K matches what I would expect — that is where attention shifts from compute-bound to memory-bandwidth-bound, and reading less data (quantized) starts winning.

The 38% memory savings at K8/V4 on Llama-3 (8 KV heads, D=128) is consistent. For Qwen3.5-122B (2 KV heads, D=256, only 12 of 48 layers use KV cache), the per-token savings are smaller in absolute terms but the headroom matters at 128K+ where every GB counts.

I am running context sweep benchmarks on the 4B right now (different work). Once that finishes I will run 122B K8/V4 vs FP16 at 32K/64K/100K and post the results here.

Thump604:
I built a similar K8/V4 asymmetric cache for Qwen3.5-122B on M2 Ultra 128GB. A few findings that may be useful:

- Hadamard rotation before quantization provides negligible quality improvement at K8. Not worth the overhead.
- The K8/V4 split is the right granularity. K needs higher precision for attention routing; V tolerates aggressive quantization.
- With mx.compile fusion for the dequant+SDPA path, the latency overhead is manageable (~1.8x baseline vs ~2x without fusion).

The KVSplit tuple API (`bits=(8,4)`) is cleaner than what I had. Closing out my local approach in favor of this.

One note for hybrid models (Qwen3.5 with GatedDeltaNet + Attention): only the attention layers use KV cache. The 36 GatedDeltaNet layers use fixed-size ArraysCache (not per-token). So the actual memory savings from KV quantization are limited to the 12 attention layers. On the 122B at 128K context, that is still ~1.8GB savings (from ~3GB to ~1.2GB for pure KV), which matters when the model weights already use 82GB of 128GB.

aiii-dot-id:
This PR changes `generation_stream` from a thread-local stream to a
single shared stream:

```diff
-generation_stream = mx.new_thread_local_stream(mx.default_device())
+generation_stream = mx.new_stream(mx.default_device())
```

That breaks any threaded server wrapper around `stream_generate`. Repro
with `mlx_lm.server` itself (which uses `http.server.ThreadingHTTPServer`):

```bash
python -m mlx_lm.server --model <any-mlx-model> --kv-bits 8
# in another terminal:
curl -sN http://127.0.0.1:8080/v1/chat/completions \
  -d '{"model":"default_model","messages":[{"role":"user","content":"hi"}],"max_tokens":4}'
```

The first request crashes with:

```
```

raised from `mx.synchronize(generation_stream)` in `wired_limit`'s
`finally` (and from `mx.eval([c.state for c in prompt_cache])` during
prefill).

Cause: `mx.new_stream(...)` binds a stream to the importing thread; in
worker threads the corresponding default GPU stream isn't attached yet,
and `mx.synchronize` rejects it. `mx.new_thread_local_stream(...)` lazily
attaches per-thread on first use, which is what the `mlx_lm.server`
threading model requires.

Restoring the thread-local form (one-line revert) makes this PR work
cleanly under threaded serving in our 6-model fleet (M3 Ultra 512 GB),
including the asymmetric `KVCache=9 + RotatingKVCache=39` mixed-cache
shape this PR is specifically designed to enable. Decode TPS, per-layer
cache types, and `--kv-bits 8` quant all behaved as expected once that
single line was reverted.

Suggested fix:

```diff
-generation_stream = mx.new_stream(mx.default_device())
+generation_stream = mx.new_thread_local_stream(mx.default_device())
```

Happy to send this as a follow-up commit on the branch if useful.

AirRunner:
@aiii-dot-id This is simply a rebase issue, the branch predates the introduction of `new_thread_local_stream`.

zcbenz:

PR #1078, opened 2026-03-31
Fix PromptTrie.pop_prefixes() off-by-one when pruning immediate prefixes

LxYuan0420:
This PR is:
- To fix `PromptTrie.pop_prefixes()` when a cached prefix is exactly one token shorter than the new insertion, so redundant prefix entries are pruned correctly.
- To preserve empty-token cache behavior when pruning root prefixes for non-empty inserts.
- To make `PromptTrie.search()` return a clean miss for empty-token lookups after root eviction instead of raising.
- To add regression coverage for immediate-prefix pruning, empty-token insertion, and empty-token lookup after root eviction.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1459,7 +1459,7 @@ class PromptTrie:
             index += 1
 
         # Got an exact match
-        if last_index == len(tokens) - 1:
+        if last_index == len(tokens) - 1 >= 0:
             return PromptTrieResult(model, tokens, None, None, 0)
 
         # Check if we found a prefix at any point


angeloskath:
Hi could you describe the bug this is fixing? Is it the ability to handle empty tokens list?

> To fix PromptTrie.pop_prefixes() when a cached prefix is exactly one token shorter than the new insertion, so redundant prefix entries are pruned correctly.

This suggests that we have an off by one error which would simply need to change the loop to omit the -1.

LxYuan0420:
The main bug is not the empty-token case.

The failing case is when a cached prefix is exactly one token shorter than a new trimmable insert. For example, if [1,2] is already cached and we then insert [1,2,3]. Before this patch, pop_prefixes() checked the current node before descending to the next one. In this example, it checked the root and [1], then moved to [1,2] and stopped. So [1,2] was reached, but never checked or removed.

So this is not just removing the -1 from the loop. The issue was the order of operations, not only the loop bound.

Minimal repro:
```python
cache.insert_cache(model, [1, 2], [MockCache("ab")])
cache.insert_cache(model, [1, 2, 3], [MockCache("abc")])

# before fix: [1,2] stays in the cache
# after fix: [1,2] is removed, so only one entry remains
assert len(cache) == 1
assert cache.nbytes == 3
```

The empty-token handling is a separate edge case in the same area, and the two extra tests cover that.

angeloskath:
> So this is not just removing the -1 from the loop
It is, as you can see :-)

PR #1084, opened 2026-04-01
perf: accumulate tokens in list instead of repeated concat

matteocelani:
## Summary

Replace O(n^2) token concatenation with O(n) list accumulation in `generate_step` when `logits_processors` are active.

## Problem

When `logits_processors` are provided (repetition penalty, presence penalty, frequency penalty, logit bias), `generate_step` accumulates the token history via `mx.concat([tokens, input_tokens])` at every generation step. For n tokens, this copies 1+2+3+...+n elements = O(n^2) total work.

## Solution

Use a Python `list.append()` to collect tokens, then a single `mx.concat(token_list)` when the processors need the full history. The concat still runs every step (processors need the full array), but each call only creates one new array from the list — no repeated copying of the growing prefix.

This also removes the `nonlocal tokens` binding and the `None` check, simplifying the code.

## Results

Micro-benchmark with `repetition_penalty=1.1`, prompt=512 tokens, generation=256 tokens, 3 trials:

| Implementation | tok/s | 
|---|---|
| Before (repeated concat) | 39.5 |
| After (list + concat) | 40.9 |
| **Improvement** | **+3.5%** |

The improvement grows with longer sequences as the O(n^2) cost dominates.

Standard benchmark (Llama-3.2-3B-Instruct-4bit, p=512, g=128, n=5, temp=0 / no processors):

| Metric | Before | After |
|---|---|---|
| prompt_tps | 453.231 | 453.005 |
| generation_tps | 51.338 | 51.324 |
| peak_memory | 2.340 GB | 2.340 GB |

No regression on the default (temp=0, no processors) path.

## Test plan

- [x] Full test suite passes (170/170)
- [x] pre-commit clean
- [x] No benchmark regression

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -361,7 +361,7 @@ def generate_step(
             "Either input_embeddings or prompt (or both) must be provided."
         )
 
-    tokens = None
+    token_list = []
 
     # Create the KV cache for generation
     if prompt_cache is None:
@@ -390,8 +390,6 @@ def generate_step(
             return model(input_tokens, cache=prompt_cache)
 
     def _step(input_tokens: mx.array, input_embeddings: Optional[mx.array] = None):
-        nonlocal tokens
-
         with mx.stream(generation_stream):
             logits = _model_call(
                 input_tokens=input_tokens[None],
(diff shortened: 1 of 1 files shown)


angeloskath:
Concatenating individual arrays should actually be significantly worse. It is also not better asymptotically, it still has to copy 1 + 2 + 3 ... items. It just does it in a different way.

I will close this but feel free to open an issue demonstrating the problem before PRing a fix.

PR #1088, opened 2026-04-01
fix: make generation_stream thread-local (MLX #3348 compat)

Thump604:
## Summary

Make `generation_stream` thread-local to work with MLX's thread-local CommandEncoder (ml-explore/mlx#3348).

## Problem

`generation_stream = mx.new_stream(mx.default_device())` is created at module import on the main thread. When generation runs in a worker thread (e.g., vllm-mlx BatchedEngine uses `asyncio.to_thread`), MLX's thread-local CommandEncoder cannot find the stream:

```
```

This breaks any server that dispatches generation to a thread pool, which is the standard pattern for async inference servers.

## Fix

Replace the module-level `mx.Stream` with a `threading.local()`-backed factory function. Each thread creates its own stream on first use.

```python
# Before (module-level, single thread only):
generation_stream = mx.new_stream(mx.default_device())

# After (per-thread):
def generation_stream() -> mx.Stream:
    s = getattr(_generation_stream_storage, "stream", None)
    if s is None:
        s = mx.new_stream(mx.default_device())
        _generation_stream_storage.stream = s
    return s
```

Call sites change from `generation_stream` to `generation_stream()`. 7 call sites updated.

Backward compatible: on MLX versions without thread-local command encoders, per-thread streams work identically.

## Test plan

- [x] Verified with vllm-mlx BatchedEngine (thread pool dispatch) on MLX with #3348
- [x] No behavior change on pre-#3348 MLX
- [x] 1 file, 21 insertions, 9 deletions

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -218,8 +218,20 @@ def setup_arg_parser():
     return parser
 
 
-# A stream on the default device just for generation
-generation_stream = mx.new_stream(mx.default_device())
+# Per-thread generation stream. Thread-local because MLX's CommandEncoder
+# is thread-local (ml-explore/mlx#3348) — a stream created on one thread
+# cannot be used from another.
+_generation_stream_lock = __import__("threading").Lock()
+_generation_stream_storage = __import__("threading").local()
+
+
+def generation_stream() -> mx.Stream:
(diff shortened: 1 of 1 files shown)


angeloskath:
Closing in favor of #1090 which does a more comprehensive fix.

PR #1090, opened 2026-04-02
Thread local generation stream

angeloskath:
This PR requires #3355 from MLX core.

- Uses a thread local stream for the generation stream (see https://github.com/ml-explore/mlx/pull/3355)
- Allows setting the stream to use in the batch generator
- Refactors the model provider to load the default model in the generation thread

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -6,6 +6,7 @@ import copy
 import functools
 import json
 import sys
+import threading
 import time
 from collections import deque
 from dataclasses import dataclass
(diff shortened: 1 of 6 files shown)


mlx_lm/generate.py
```diff
@@ -395,7 +396,7 @@ def _model_call(input_tokens: mx.array, input_embeddings: Optional[mx.array]):
     def _step(input_tokens: mx.array, input_embeddings: Optional[mx.array] = None):
         nonlocal tokens
 
-        with mx.stream(generation_stream):
+        with mx.stream(generation_stream.stream):
```
andresy: not sure about the naming?

angeloskath: Yeah good point. I guess I was being lazy yesterday. Moved `ThreadLocalStream` to C++ and made it implicitly convert to stream. So now we can use it wherever we would use a stream and that looks exactly like before.

andresy (approved):

PR #1093, opened 2026-04-02
Add gemma 4

Blaizzy:


weklund-agent:
Hey @Blaizzy — great work getting this up so fast, the full model implementation looks solid (MoE, KV sharing, PLE, all the norm variants).

Wanted to flag #1103 by @0xSoftBoi which targets your `gemma4` branch with some fixes and additions that build on top of this:

- **Sanitizer bug fix** — the multimodal wrapper's `sanitize` was double-prepending `model.` to language model weights, causing `ValueError` when loading checkpoints like `mlx-community/gemma-4-e2b-it-4bit`
- **PLE embedding split** — splits the single large per-layer embedding into individual `nn.Embedding` per layer to stay under Metal's 4GB buffer limit
- **Tool call parser** — `function_gemma4.py` for Gemma 4's native `<|tool_call>...<tool_call|>` JSON format, plus `_infer_tool_parser` integration
- **Additional tests** — PLE split sanitize, MoE + k_eq_v combined test matching the real 26B-A4B config

Just a friendly heads up :)

Blaizzy:
Hey @weklund-agent @0xSoftBoi 

Feel free to send a PR to my branch or I can apply cherry pick and apply the fixes here, that way we don't have duplicate work :)

0xSoftBoi:
Hey @Blaizzy @angeloskath — great work getting this moving so fast. I've been working on Gemma 4 support in #1103 and wanted to flag a few things I've landed there that might be useful to cherry-pick or integrate here:

1. **Multimodal sanitizer fix** — The current sanitizer double-prepends `model.` to weights that already contain it, which prevents loading multimodal checkpoints like `mlx-community/gemma-4-e2b-it-4bit`. Fix is a 2-line change in `gemma4.py`.

2. **Gemma 4 tool parser + auto-detection** — A dedicated `function_gemma4` parser for the `<|tool_call>...<tool_call|>` format with `<|"|>` quote unescaping, plus detection logic in `tokenizer_utils.py` that checks for both opening and closing delimiters. This closes #1096.

3. **PLE split with Metal buffer guard** — Splits HF's combined `[vocab, num_layers * ple_dim]` embedding tensor into per-layer chunks at sanitize time to stay under the 4GB Metal buffer limit. Includes tests for both the split path and the already-split passthrough.

4. **Test coverage** — 255 lines of new tests: MoE variant (SwitchGLU routing), k_eq_v (shared K/V projections), PLE sanitize split, and a combined MoE + k_eq_v test matching the real 26B-A4B config.

All tests passing, pre-commit green, verified inference on 4-bit E2B (~50 tok/s) and LoRA fine-tuning (loss 2.66 → 0.54 over 200 iterations).

Happy to submit any of this as a PR against your branch, cherry-pick commits, or help however gets Gemma 4 landed fastest. Let me know what works best.

0xSoftBoi:
Hey @Blaizzy — PR is up against your branch: https://github.com/Blaizzy/mlx-lm/pull/12

Rebased clean on top of `pc/add-gemma-4`, no conflicts, single commit. Should be a straightforward merge whenever you're ready.

For context — I originally opened #1103 because I was trying to fine-tune Gemma 4 with LoRA and kept hitting the sanitizer bug and Metal buffer crashes. Ended up fixing everything I ran into and figured I'd contribute it back. Appreciate you being open to integrating it!

angeloskath:
Hi @0xSoftBoi ! Can you share which devices encountered the 4GB limit.

angeloskath:
Feel free to PR the tool parser (and tests) separately as well cause I am about to merge this. Testing the max buffer issue you mentioned on my M2 base. Except for that everything works.

Blaizzy:
@angeloskath @0xSoftBoi 

I investigated the issue and it shouldn't happen if you run the quants because all are below the limit.

| Format | PLE total | 
| :--- | :---: |
| bf16 | 4.70 GB | 
| 8-bit | 2.50 GB | 
| 4-bit | 1.32 GB |


This will only happen if you run bf16, but if the device already has a 4GB metal buffer limit then it shouldn't run because the peak for bf16 is 11.004 GB.


> Feel free to PR the tool parser (and tests) separately as well cause I am about to merge this. Testing the max buffer issue you mentioned on my M2 base. Except for that everything works.


Read my mind Angelos, that's exactly what I suggested as well :)

N8python:
Seems ready to merge to me!!!

mlx_lm/models/gemma4_text.py
```diff
@@ -0,0 +1,698 @@
...
+                norms_input,
+                self.embed_tokens.as_linear(input_embeddings),
+            )
+
+            # Checks can be added if needed but they necessarily break the GPU
+            # pipelining and force an eval.
+            #
+            #   match_counts = (distance < eps).sum(-1)
+            #
+            input_ids = mx.argmin(distance, -1)
```
angeloskath, line 498: @Blaizzy and @N8python out of curiosity when is the reverse lookup ever needed?

mlx_lm/models/gemma4_text.py
```diff
@@ -0,0 +1,698 @@
...
+
+        queries = queries.transpose(0, 2, 1, 3)
+        queries = self.rope(queries, offset=offset)
+
+        if cache is not None:
+            keys, values = cache.update_and_fetch(keys, values)
+
+        if mask is not None and isinstance(mask, mx.array):
+            if mask.shape[-1] != keys.shape[-2]:
+                mask = mask[..., -keys.shape[-2] :]
```
angeloskath, line 283: Similarly when does this ever happen? What prompted you to add that?

Blaizzy: This is for sliding windows.

But I guess it can be done differently. 

I wish they had separate the models into 2-3 types for clarity.

mlx_lm/models/gemma4_text.py
Blaizzy: This if I remember correctly, I asked N8 on X or the PR he made and he had explained it was for prefill. 

But I could be wrong, and happy to take a look at it with fresh eyes.

N8python: its needed in the extreme edge case where you pass input_embeddings to the model and the model needs the input_ids (not embeddings) for later layers token embeddings - so this requires:

input embeddings -> token ids -> per-layer embeddings. i think it would be cleaner to just error - codex recommended it as a fix for this edge case, and it made sense, but i'm not sure its neccessary.

Blaizzy: Yap, exactly. It was input embeddings not prefill 👌🏽

angeloskath: > i think it would be cleaner to just error 

Totally. The current implementation is not horrible but if there is no use-case then it is simply dead code. I assumed it was for passing embeddings coming from a different source that doesn't map exactly to input_ids. However the original code was checking for exact match so 🤷‍♂️

PR #1094, opened 2026-04-02
perf: reduce GPU sync frequency in GPTQ quantization

### LINKED ISSUE #1122: GPTQ quantization slower than expected — ~300k mx.eval calls in inner loop
While profiling GPTQ quantization, I noticed that `mx.eval(err, W)` is called inside the per-weight inner loop. With `group_size=64`, this means 64 GPU/CPU synchronization barriers per group — each one flushes the GPU pipeline and waits for CPU dispatch. For Qwen3-0.6B (196 quantizable layers, ~1536 columns), that's roughly 300,000 sync calls per quantization run, each with fixed overhead regardless of computation size.

The intra-group error propagation is a sequential chain of reads and writes on `W`, but these dependencies exist within the lazy graph — MLX tracks them. The reference PyTorch implementations ([IST-DASLab/gptq](https://github.com/IST-DASLab/gptq), [AutoGPTQ](https://github.com/AutoGPTQ/AutoGPTQ)) don't synchronize per weight; PyTorch's eager execution serializes the operations implicitly.

I measured quantization time on M4 Max (48 GB), `bits=4, group_size=64, num_samples=-1`, 3 trials:

| Model | Params | Time (mean ± std) |
|---|---|---|
| GPT-2 | 124M | 18.2 ± 0.2 s |
| GPT-2 Medium | 355M | 65.7 ± 3.5 s |
| Qwen3-0.6B | 600M | 71.4 ± 0.9 s |
| Qwen3-1.7B | 1.7B | 407.5 ± 48.7 s |

Is the per-weight eval necessary for correctness in MLX's lazy evaluation model, or could it be relaxed?

matteocelani:
Fixes #1122

## Summary

Move `mx.eval(err, W)` from per-weight to per-group granularity in the GPTQ inner loop, reducing GPU/CPU synchronization points by a factor of `group_size` (typically 64x) with no change in quantization output.

One-line change, 28-47% faster GPTQ quantization, bit-identical perplexity.

## The problem

The current implementation calls `mx.eval(err, W)` inside the innermost loop, once per weight element:

```python
for k in range(group_size):       # group_size = 64 (typical)
    e = gptq_error(w, d, scales, biases)
    W[..., k:k+j] -= e @ Hinv[k:k+1, k:k+j]
    err[..., k:k+1] = e
    mx.eval(err, W)               # <-- forces GPU/CPU sync 64 times per group
```

Each `mx.eval` forces the GPU to flush its command buffer, synchronize with the CPU, and wait for the next dispatch. For a model with `L` quantizable layers and `C` weight columns:

- **Sync calls** = `L * C` (one per weight element)
- For Qwen3-0.6B (196 layers, ~1536 cols): **~300,000 sync points**

Each sync has a fixed overhead (kernel launch latency, pipeline drain, CPU wake-up) independent of computation size. When the per-sync computation is small (a single column update), this fixed overhead dominates wall-clock time.

## The fix

Move `mx.eval` one indentation level up — from per-weight to per-group:

```python
for k in range(group_size):
    e = gptq_error(w, d, scales, biases)
    W[..., k:k+j] -= e @ Hinv[k:k+1, k:k+j]
    err[..., k:k+1] = e
                                   # no eval here — MLX tracks dependencies
mx.eval(err, W)                    # <-- once per group, not once per weight
W[..., j:] -= err @ Hinv[i:j, j:]
```

Sync points drop by a factor of `group_size`:

- **Before**: `L * C` syncs (~300,000)
- **After**: `L * C / group_size` syncs (~4,700)

## Why this is safe

The intra-group error propagation is a sequential chain of read-modify-write operations on `W`: step `k` reads `W[..., k:k+1]` (updated by step `k-1`) and writes `W[..., k:k+j]`. MLX's lazy evaluation is deferred execution, not operation transformation — it builds a computation graph that preserves this ordering. `mx.eval` controls *when* the GPU executes, not *what* it executes.

The only point where materialization is required is before the cross-group update (`W[..., j:] -= err @ Hinv[i:j, j:]`), because `err` must be fully accumulated. The per-group `mx.eval` guarantees this.

Note: the reference PyTorch implementations ([IST-DASLab/gptq](https://github.com/IST-DASLab/gptq), AutoGPTQ) have no per-weight synchronization in this loop — PyTorch's eager execution handles dependencies implicitly. This change aligns MLX's behavior with the original implementations.

## Results

Benchmarked on M4 Max (48 GB), `bits=4, group_size=64, num_samples=-1` (full calibration dataset), 3 trials per configuration, MLX 0.31.1, mlx-lm 0.31.2.

### Quantization time (3 trials, mean +/- std)

| Model | Arch | Params | Baseline | Post-opt | Speedup |
|---|---|---|---|---|---|
| GPT-2 | GPT | 124M | 18.2 +/- 0.2 s | 10.7 +/- 0.1 s | **-41.2%** |
| GPT-2 Medium | GPT | 355M | 65.7 +/- 3.5 s | 46.2 +/- 0.6 s | **-29.7%** |
| Qwen3-0.6B | Qwen3 | 600M | 71.4 +/- 0.9 s | 37.6 +/- 1.3 s | **-47.3%** |
| Qwen3-1.7B | Qwen3 | 1.7B | 407.5 +/- 48.7 s | 293.4 +/- 14.9 s | **-28.0%** |

Speedup is consistent across all models (28-47%) and both architectures.

Post-opt variance is consistently lower than baseline — fewer sync points means less scheduling jitter.

### Perplexity — calibration data

| Model | PPL baseline | PPL post-opt | Delta |
|---|---|---|---|
| GPT-2 | 22.8669 | 22.8669 | 0.0000 |
| GPT-2 Medium | 17.3737 | 17.3737 | 0.0000 |
| Qwen3-0.6B | 17.0574 | 17.0574 | 0.0000 |
| Qwen3-1.7B | 12.8804 | 12.8804 | 0.0000 |

### Perplexity — held-out data (80/20 split, Qwen3-0.6B)

| | PPL baseline | PPL post-opt | Delta |
|---|---|---|---|
| Held-out (45 samples) | 16.7256 | 16.7256 | 0.0000 |

Perplexity is **bit-identical** on both calibration and held-out data, across all 4 models and all trials. The optimization changes evaluation scheduling, not computation.

### Peak memory

No change in peak memory for any model (checked on all 4).

## Limitations

- Tested on models up to 1.7B parameters. Larger models were attempted (Qwen3-4B) but Hessian computation alone was too slow for multi-trial benchmarking. We expect the speedup to hold since sync overhead is fixed while compute scales with matrix size.
- `gptq_error` is decorated with `@mx.compile`. A compiled subgraph called 64 times within a lazy graph is a standard MLX pattern, and the bit-identical perplexity confirms correctness.

## Test plan

- [x] Full test suite passes (170/170)
- [x] `pre-commit` clean
- [x] PPL bit-identical on 4 models / 2 architectures (calibration + held-out)

### DIFF
--- a/mlx_lm/quant/gptq.py
+++ b/mlx_lm/quant/gptq.py
@@ -125,8 +125,8 @@ def gptq_quantize(
 
                 W[..., k : k + j] -= e @ Hinv[k : k + 1, k : k + j]
                 err[..., k : k + 1] = e
-                mx.eval(err, W)
 
+            mx.eval(err, W)
             W[..., j:] -= err @ Hinv[i:j, j:]
 
         # Quantize with the given scales and biases


zcbenz:

PR #1095, opened 2026-04-02
feat: add Gemma 4 text model support

mohavinash:
## Summary

Adds text-only inference support for Google's Gemma 4 (E2B) models.

Tested with `google/gemma-4-E2B-it` (4-bit quantized) — generates coherent, correct output at ~35 tok/s on Apple Silicon.

### Key features
- **Per-Layer Embeddings (PLE)**: per-layer token embeddings gated into each decoder layer. The large PLE weight is split into per-layer chunks in `sanitize()` to stay under Metal's 4GB single-buffer limit.
- **KV cache sharing**: later layers reuse KV cache from earlier layers of the same type (sliding/global), reducing memory.
- **ProportionalRoPE**: partial rotation for global attention layers (`partial_rotary_factor=0.25`). Uses tensor splitting since `mx.fast.rope` doesn't handle zero-padded frequencies correctly.
- **Heterogeneous head dims**: sliding attention uses `head_dim=256`, global attention uses `global_head_dim=512`.
- **Double-wide MLP**: KV-shared layers use `2x intermediate_size` since they save compute on K/V projections.
- **v_norm**: scaleless RMS normalization applied to value states.
- **Per-layer scalar**: learned scalar applied at the end of each decoder layer.
- **Logit soft-capping**: `tanh(x/cap) * cap` applied to output logits.

### Notable differences from Gemma 3
- **RMSNorm**: plain `weight` (no `1 + weight` offset)
- **Attention scale**: `1.0` (QK-norm replaces traditional `1/sqrt(d)` scaling)
- **ProportionalRoPE**: new RoPE variant with partial rotation

### Architecture
```
Model
└── LanguageModel
    ├── embed_tokens (tied with output projection)
    ├── embed_tokens_per_layer[0..N] (PLE, split for Metal)
    ├── per_layer_model_projection
    ├── per_layer_projection_norm
    ├── layers[0..34] (DecoderLayer)
    │   ├── Attention (sliding or global, with QK-norm + v_norm)
    │   ├── MLP (gelu_approx gated, optionally double-wide)
    │   ├── PLE injection (gate → gelu → multiply → project)
    │   └── layer_scalar
    └── norm (final RMSNorm)
```

## Test plan
- [x] Verified weight loading: all 634 parameter keys match, no shape mismatches
- [x] Verified bf16→uint16→mx.bfloat16 conversion produces identical values to PyTorch reference
- [x] Verified no NaN/zero weights after quantized load
- [x] Tested generation with multiple prompts — coherent, factually correct output
- [x] Tested with KV cache (autoregressive generation) and without (full-context)

### DIFF
--- a/mlx_lm/models/gemma4.py
+++ b/mlx_lm/models/gemma4.py
@@ -1,0 +1,483 @@
+# Copyright © 2025 Apple Inc.
+
+"""Gemma 4 text-only model for mlx-lm.
+
+Supports Per-Layer Embeddings (PLE), KV sharing, proportional RoPE,
+heterogeneous head dims, double-wide MLP, v_norm, and layer_scalar.
+
+Key differences from Gemma 3:
+  - RMSNorm uses plain weight (no +1 offset)
+  - Attention scale is 1.0 (QK-norm replaces 1/sqrt(d) scaling)
+  - v_norm: RMS normalization on values without learned scale
+  - ProportionalRoPE: partial rotation for global attention layers
+  - Per-Layer Embeddings: per-layer token embeddings gated into each layer
+  - KV sharing: later layers reuse KV cache from earlier layers
+"""
+
+from dataclasses import dataclass, field
+from functools import partial
+from typing import Any, Dict, List, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+from mlx.utils import tree_flatten, tree_unflatten
+
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .cache import KVCache, RotatingKVCache
+
+
+@dataclass
+class TextConfig(BaseModelArgs):
+    model_type: str = "gemma4_text"
+    hidden_size: int = 1536
+    num_hidden_layers: int = 35
+    intermediate_size: int = 6144
+    num_attention_heads: int = 8
+    num_key_value_heads: int = 1
+    head_dim: int = 256
+    global_head_dim: int = 512
+    rms_norm_eps: float = 1e-6
+    vocab_size: int = 262144
+    hidden_activation: str = "gelu_pytorch_tanh"
+    sliding_window: int = 512
+    max_position_embeddings: int = 131072
+    final_logit_softcapping: float = 30.0
+    layer_types: List[str] = field(default_factory=list)
+    rope_parameters: Dict = field(default_factory=dict)
+    num_kv_shared_layers: int = 0
+    hidden_size_per_layer_input: int = 0
+    vocab_size_per_layer_input: int = 262144
+    use_double_wide_mlp: bool = True
+    attention_k_eq_v: bool = False
+    num_global_key_value_heads: Optional[int] = None
+    enable_moe_block: bool = False
+    attention_bias: bool = False
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str = "gemma4"
+    text_config: dict = field(default_factory=dict)
+    vocab_size: int = 262144
+
+
+class RMSNorm(nn.Module):
+    """Gemma 4 RMSNorm — plain weight, no +1 offset (unlike Gemma 3)."""
+
+    def __init__(self, dims: int, eps: float = 1e-6):
+        super().__init__()
+        self.weight = mx.ones((dims,))
+        self.eps = eps
+
+    def __call__(self, x):
+        return mx.fast.rms_norm(x, self.weight, self.eps)
+
+
+class RMSNoScale(nn.Module):
+    """RMSNorm without learned scale (for v_norm)."""
+
+    def __init__(self, eps: float = 1e-6):
+        super().__init__()
+        self.eps = eps
+
+    def __call__(self, x):
+        return mx.fast.rms_norm(x, None, self.eps)
+
+
+class ProportionalRoPE(nn.Module):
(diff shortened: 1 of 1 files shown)


nicdavidson:
Tested this against the 31B (mlx-community/gemma-4-31b-it-4bit) on M5 Max 48GB, this was a great starting point.  Two things I had to patch to get 31B running:

**sanitize() MLX-community weights**

Current sanitize assumes HF weight keys. The mlx-community quantized weights use a different prefix structure so you end up with double-prefixed keys. Swapped the tree_unflatten approach for flat string replacement strip "model." prefix if present, then replace "language_model.model." with "language_model." so it handles both HF and mlx-community formats.

**k_eq_v support**

31B has attention_k_eq_v: true on full attention layers, no v_proj at all on those layers. Made v_proj conditional and for k_eq_v layers, values come from the raw k_proj output before k_norm, then through v_norm. Order matters.

After both patches: ~30 tok/s, 17.6GB peak. Happy to share the full patched file if that helps.

mohavinash:
> Tested this against the 31B (mlx-community/gemma-4-31b-it-4bit) on M5 Max 48GB, this was a great starting point. Two things I had to patch to get 31B running:
> 
> **sanitize() MLX-community weights**
> 
> Current sanitize assumes HF weight keys. The mlx-community quantized weights use a different prefix structure so you end up with double-prefixed keys. Swapped the tree_unflatten approach for flat string replacement strip "model." prefix if present, then replace "language_model.model." with "language_model." so it handles both HF and mlx-community formats.
> 
> **k_eq_v support**
> 
> 31B has attention_k_eq_v: true on full attention layers, no v_proj at all on those layers. Made v_proj conditional and for k_eq_v layers, values come from the raw k_proj output before k_norm, then through v_norm. Order matters.
> 
> After both patches: ~30 tok/s, 17.6GB peak. Happy to share the full patched file if that helps.

Original E2B author here (#1095). For what it's worth, this crash isn't model-specific — it's a pre-existing bug in
  BatchRotatingKVCache.merge() in cache.py. size() returns min(offset, max_size) but _temporal_order() returns the full physical array, so the merge allocates a target that's too small when the prompt exceeds the sliding window size. Any sliding-window model would hit this under batched serving.

  ▎Separate from #1097 (the mx.array += offset mutation), which this PR already works around. This merge shape mismatch probably
  warrants its own issue.

  ▎ E2B variant works fine for single-sequence inference: [avinashmohan/gemma-4-E2B-it-4bit-mlx](https://huggingface.co/avinashmohan/gemma-4-E2B-it-4bit-mlx)

angeloskath:
Closing as gemma 4 was merged in #1093.

mohavinash:
@nicdavidson Thanks for testing against the 31B and sharing those findings — both are real gaps.

Would you be willing to share your patched file (or a diff)? Specifically interested in:

1. **sanitize()**: How did you handle the HF vs mlx-community prefix divergence? Did you go with a flat string replacement (`str.replace("language_model.model.", "language_model.")`) or something more structured like a prefix-stripping pass before `tree_unflatten`? Curious if there are other prefix variants beyond `model.` and `language_model.model.` in the wild.

2. **k_eq_v attention**: You mentioned values come from raw `k_proj` output *before* `k_norm`, then through `v_norm`. Did you verify this ordering against the HuggingFace reference? The naive reading would be to apply `v_norm` to the already-normed keys, but that would double-normalize — so your "before k_norm" detail is important. Did you also need to handle the missing `v_proj` weights gracefully in `sanitize()` / weight loading, or does the model file simply not contain those keys?

Working on a local Swift port that hit the same issues, so the details would save some reverse-engineering.

nicdavidson:
Happy to share. Most of this actually landed in #1093 now.

**1. sanitize() prefix handling**

When I was testing against this PR (#1095), the sanitizer assumed HF weight keys. mlx-community quantized weights use a different prefix structure (`language_model.model.layers.0.xxx` vs HF's `model.language_model.layers.0.xxx`), so I was getting double-prefixed keys on load.

My fix was to strip instead of add, normalizing both formats to the same keys:

```python
for k, v in weights.items():
    if any(s in k for s in skip):
        continue
    # Strip outer "model." wrapper (HF format)
    if k.startswith("model."):
        k = k[len("model."):]
    # Strip inner "model." between language_model and layers/embed/norm
    k = k.replace("language_model.model.", "language_model.")
    result[k] = v
```

#1093's merged sanitizer handles this differently with a `starts_w_model` guard that skips the replace for non-HF weights, which should cover both formats.

**2. k_eq_v attention ordering**

Verified against the HF reference. The order matters:

```python
keys = self.k_proj(x).reshape(B, L, self.n_kv_heads, self.head_dim).transpose(0, 2, 1, 3)

if self.use_k_eq_v:
    values = self.v_norm(keys)   # raw k_proj output, before k_norm
else:
    values = self.v_proj(x).reshape(...)
    values = self.v_norm(values)

keys = self.k_norm(keys)         # k_norm applied after values are captured
```

`v_norm` is scaleless RMS normalization (no learned weight). So it's raw k_proj -> v_norm -> values. NOT k_norm'd keys through v_norm, that would double-normalize like you suspected. The `v_proj` weights don't exist in the checkpoint for k_eq_v layers at all, the model file just omits them. Weight loading handles it fine as long as `v_proj` is conditionally created.

The k_eq_v handling landed in #1093 (lines 215-264 of gemma4_text.py) so that should be good for your Swift port.

Issue #1097, opened 2026-04-02
BatchKVCache: mx.array += in-place mutation corrupts RoPE offsets in batch inference


nicdavidson:
Found this while getting Gemma 4 31B running through batch inference (oMLX server). Single-prompt generation works fine, batch serving produces garbage. Root cause is mx.array += in-place mutation on cache offsets.

The problem

In the batch generation path, the code captures the cache offset before calling update_and_fetch:

```
offset = cache.offset          # reference to mx.array
keys, values = cache.update_and_fetch(keys, values)
# internally: self.offset += keys.shape[2]   mutates in place
queries = rope(queries, offset=offset)  # offset is now WRONG
```

Python ints are immutable: += creates a new object. mx.array is mutable += modifies in place. So offset ends up pointing at the post-update value instead of the pre-update value. RoPE positions are wrong, attention patterns break, output is garbage.

Regular KVCache uses a Python int for offset, so single-prompt inference is unaffected. BatchKVCache uses mx.array, so this only shows up in batch/server mode.

Fix

`offset = cache.offset + 0   # force a copy`

The + 0 allocates a new array with the current value. When update_and_fetch mutates cache.offset, the local copy is untouched.

Possibly related issues

This might explain or contribute to several open batch inference issues:

#965 — KV cache cross-contamination between concurrent requests

#939 — SuScaledRoPE mutates x parameter (same bug class, already acknowledged)

#983, #1035 — BatchRotatingKVCache merge crashes with concurrent requests

PR #999 addresses physical vs logical offset for RotatingKVCache but doesn't cover BatchKVCache.

Repro

Any model served through batch inference with BatchKVCache. I hit it on Gemma 4 31B but it's architecture-independent — the bug is in the cache/generation path, not the model.

khsd6327:
Tested this on oMLX 0.3.2 + mlx-lm 0.31.2 (commit 564281f) with Gemma 4 31B and 26B.

**The `+ 0` fix alone does not resolve the issue for Gemma 4.**

Applied `offset = cache.offset + 0` in `mlx_vlm/models/gemma4/language.py` (lines 207, 210) — batch corruption persists at the same rate (~70-80% failure with 5 concurrent requests).

Also tried:
- Replacing all `self.offset +=` / `self.left_padding -=` in-place mutations in `cache.py` with `self.x = self.x + y` — no effect
- Updating `cache.py` to include the `BatchRotatingKVCache.merge()` fix from PR #1072 — no effect
[...]

Issue #1082, opened 2026-03-31
branch with quantized kvcache is bugged


badelon:
pip install git+https://github.com/deceptech-packet-ninja/mlx-lm.git@feat/server-kv-bits 

AttributeError: 'list' object has no attribute 'swapaxes'

20:34:48 a1 ~ ❯ while true; do
    echo "Starting MLX Server at $(date)"

    # Run the server with environment variables
    MLX_DEBUG_MEMORY=1 HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 \
    mlx_lm.server \
        --model "/Users/a1/.lmstudio/models/mlx-community/Kimi-K2.5-3bit" \
        --trust-remote-code \
        --max-tokens 200000 \
        --prompt-cache-bytes 40G \
        --prompt-cache-size 30 \
        --prompt-concurrency 1 \
        --decode-concurrency 1 \
        --kv-bits 8 \
        --kv-group-size 64 \
        --quantized-kv-start 12000 \
        --port 1234

    echo "Server stopped or crashed. Restarting in 3 seconds..."
    sleep 3

Exception occurred during processing of request from ('127.0.0.1', 50269)
Traceback (most recent call last):
  File "/Users/a1/miniconda3/lib/python3.13/socketserver.py", line 697, in process_request_thread
    self.finish_request(request, client_address)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/a1/miniconda3/lib/python3.13/socketserver.py", line 362, in finish_request
    self.RequestHandlerClass(request, client_address, self)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/a1/miniconda3/lib/python3.13/site-packages/mlx_lm/server.py", line 1868, in <lambda>
    lambda *args, **kwargs: handler_class(
                            ~~~~~~~~~~~~~^
        response_generator,
        ^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        **kwargs,
        ^^^^^^^^^
    ),
    ^
  File "/Users/a1/miniconda3/lib/python3.13/site-packages/mlx_lm/server.py", line 1134, in __init__
    super().__init__(*args, **kwargs)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
[...]

Issue #1081, opened 2026-03-31
ArraysCache.is_trimmable() returns True but trim() method doesn't exist


waybarrios:
When using ArraysCache with speculative decoding (MTP), calling `is_trimmable()` returns `True` but the object doesn't have a `trim()` method, which causes an `AttributeError`.

This happens on perfect cache hits where `is_trimmable()` passes the check but `trim()` is never defined on the class.

```python
# this passes
hasattr(c, "is_trimmable") and c.is_trimmable()

# this crashes
c.trim(2)  # AttributeError: 'ArraysCache' object has no attribute 'trim'
```

If `is_trimmable()` can return `True`, the class should also implement `trim()`. Otherwise consumers have to add defensive `hasattr(c, "trim")` checks everywhere.

Reported from https://github.com/waybarrios/vllm-mlx/issues/145

eyupcanakman:
Can you paste the full stack trace? On current main, `LRUPromptCache.fetch_nearest_cache` returns the cached state directly on a 100% hit, and the partial-hit branch goes through `can_trim_prompt_cache`, so `.trim()` is never called on a non-trimmable `ArraysCache`. The failing frame will show whether the call originates in `mlx_lm` or in the vllm-mlx 0.2.6 wrapper.

waybarrios:
Hi @eyupcanakman, thanks for the precise pointer. Verified locally on `mlx-lm` 0.31.2:

```python
>>> import mlx_lm; mlx_lm.__version__
'0.31.2'
>>> from mlx_lm.models.cache import ArraysCache, KVCache
>>> c = ArraysCache(size=2)
>>> c.is_trimmable(), hasattr(c, 'trim')
(False, False)
>>> k = KVCache()
>>> k.is_trimmable(), hasattr(k, 'trim')
(True, True)
```

So on current `mlx-lm`, `ArraysCache` is honestly non-trimmable and the `fetch_nearest_cache` / `can_trim_prompt_cache` paths hold the invariant cleanly, as you described. I can't reproduce the AttributeError against current `mlx-lm`.

One small API-hygiene suggestion, unrelated to reproducing the bug: would there be value in documenting the `is_trimmable() is True ⇒ trim(n: int) exists` contract on `_BaseCache`, or adding a default `trim` that raises `NotImplementedError`? It would formalize the invariant so downstream code can rely on `is_trimmable()` alone without defensive `hasattr(c, "trim")` double checks. Totally optional.

[...]

Issue #1035, opened 2026-03-21
BatchRotatingKVCache.merge() crashes on shape mismatch when prompt_cache_size >= 1


danielkhalilov:
## Bug

  `BatchRotatingKVCache.merge()` in `cache.py` crashes with a `ValueError` when
  two consecutive requests have different prompt lengths and
  `--prompt-cache-size >= 1`. The crash kills the generation thread, leaving the
   server responding to `/health` but unable to process any inference requests
  (generation-dead).

  ## Repro

  ```bash
  python -m mlx_lm.server --model mlx-community/gpt-oss-20b-MXFP4-Q8
  --prompt-cache-size 1

  Send two requests with different prompt sizes (e.g. different system prompts,
  or one with tools and one without):

  # Request 1 — succeeds, populates cache
  curl -X POST http://localhost:8100/v1/chat/completions \
    -H 'Content-Type: application/json' \
    -d '{"messages":[{"role":"system","content":"You are a helpful assistant
  with many tools
  available."},{"role":"user","content":"hello"}],"max_tokens":5}'

  # Request 2 — different system prompt length → crash
  curl -X POST http://localhost:8100/v1/chat/completions \
    -H 'Content-Type: application/json' \
    -d '{"messages":[{"role":"user","content":"hello"}],"max_tokens":5}'

  Traceback

  Exception in thread Thread-1 (_generate):
    File "mlx_lm/server.py", line 948, in _generate
      responses = batch_generator.next()
    File "mlx_lm/generate.py", line 1126, in _process_prompts
      prompt_cache = _merge_caches(caches)
    File "mlx_lm/generate.py", line 922, in _merge_caches
      batch_cache.append(caches[0][i].merge([c[i] for c in caches]))
    File "mlx_lm/models/cache.py", line 1364, in merge
      keys[i : i + 1, :, p : p + c._idx] = c._temporal_order(c.keys)
  ValueError: [broadcast_shapes] Shapes (1,8,201,64) and (1,8,128,64) cannot be
  broadcast.

  Root cause

  In BatchRotatingKVCache.merge() (cache.py:1342), lengths is computed from
  c.size() and c._idx is used to slice the destination array. But
  _temporal_order() returns data whose actual sequence dimension can differ from
   both c.size() and c._idx — it returns the reordered keys/values sliced by the
   ring buffer's internal state. The destination slice width (c._idx) doesn't
  match the source tensor width from _temporal_order(), causing the broadcast
  error.

  Fix

[...]

PR #1100, opened 2026-04-03
fix: BatchKVCache RoPE offset corruption, MLA quantization crash, ArraysCache trim

deceptech-packet-ninja:
## Summary

Four cache bug fixes addressing batch inference reliability:

- **BatchKVCache RoPE offset corruption (#1097):** `mx.array +=` in-place mutation corrupts RoPE offsets captured before `update_and_fetch`. Model code does `offset = cache.offset` then calls `update_and_fetch` which mutates `cache.offset` in-place -- the local reference now points to the post-update value. Fix: store offset in a private field, expose via property that returns a copy (`+ 0`). Applied to both `BatchKVCache` and `BatchRotatingKVCache`.

- **BatchRotatingKVCache.merge() shape mismatch (#983, #1035):** When concurrent requests have different prompt lengths and ring buffers wrap at different offsets, `merge()` uses `c.size()` for slice widths but `_temporal_order()` returns tensors with different sequence dimensions. Fix: pre-compute temporally-ordered tensors and derive lengths from actual output shapes.

- **Quantized KV cache crashes on MLA models (#1082):** `QuantizedKVCache` returns quantized tuples `(data, scales, biases)` from `update_and_fetch`. Standard attention passes these through SDPA which handles them via `mx.quantized_matmul`. But MLA models (DeepSeek V2/V3, Kimi K2.5, etc.) access cache values directly with `.swapaxes()`, which crashes on tuples. Fix: detect MLA layers via `kv_a_proj_with_mqa` attribute in `make_prompt_cache` and mark their caches as non-quantizable.

- **ArraysCache missing trim (#1081):** Add explicit `is_trimmable() -> False` and `trim() -> 0` for API completeness.

## Test plan

- [ ] Batch inference with concurrent requests of different prompt lengths on models using `RotatingKVCache` (e.g. Trinity-Nano) -- validates merge fix (#983, #1035)
- [ ] Batch inference produces correct output, not garbage -- validates RoPE offset fix (#1097)
- [ ] `--kv-bits 8` with Kimi K2.5 / DeepSeek V3 (MLA models) -- validates quantization guard (#1082)
- [ ] `--kv-bits 8` still works with standard attention models (Llama, Qwen)
- [ ] `ArraysCache().trim(n)` returns 0 and `ArraysCache().is_trimmable()` returns False (#1081)

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1454,7 +1454,20 @@ class BatchRotatingKVCache(_BaseCache):
             )
 
         offsets = [c.offset for c in caches]
-        lengths = [c.size() for c in caches]
+
+        # Pre-compute temporally-ordered tensors so we use actual shapes
+        # for sizing, not c.size() / c._idx which may disagree after
+        # ring buffer wrapping. Fixes #983, #1035.
+        ordered = []
+        for c in caches:
+            if c.keys is not None:
+                ordered.append(
+                    (c._temporal_order(c.keys), c._temporal_order(c.values))
+                )
+            else:
+                ordered.append((None, None))
+
+        lengths = [ok.shape[2] if ok is not None else 0 for ok, _ in ordered]
         max_length = max(lengths)
 
         # No cache has content so make an empty one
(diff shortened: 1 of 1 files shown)


angeloskath:
I will close this because

1. Not a bug as explained in the issue
2. Fixed in #1093 
3. Not all models have to support everything
4. ArraysCache is_trimmable returns False

PR #1099, opened 2026-04-02
feat: Gemma 4 text model + 31B support + tool call parser

nicdavidson:
## Summary

Adds full Gemma 4 text-only inference support, tested on both E2B and 31B variants, plus a tool call parser for Gemma 4's native format.

### Model support (`mlx_lm/models/gemma4.py`)

Based on #1095 by @mohavinash, with additional fixes for 31B:

- **Per-Layer Embeddings (PLE)** split per-layer for Metal's 4GB buffer limit
- **ProportionalRoPE** — partial rotation for global attention layers
- **KV cache sharing** between layers of the same type
- **Heterogeneous head dims** — sliding (256) vs global (512)
- **Logit soft-capping**, double-wide MLP, v_norm, layer scalar
- **`sanitize()`** handles both HuggingFace and mlx-community weight key formats
- **`attention_k_eq_v`** support — 31B shares K/V on full attention layers (no separate `v_proj`; values derived from `k_proj` output before `k_norm`)

### Tool call parser (`mlx_lm/tool_parsers/gemma4.py`)

Gemma 4 uses `<|tool_call>...<tool_call|>` delimiters with `<|"|>` string escaping, different from Gemma 3's `<start_function_call>` format. Adds parser and auto-detection in `_infer_tool_parser()`.

Models without a `chat_template` (e.g. mlx-community quantizations) can set `"tool_parser_type": "gemma4"` in `tokenizer_config.json`.

Fixes #1096.

## Test plan

- [x] Verified on `mlx-community/gemma-4-31b-it-4bit` — ~30 tok/s on M5 Max (48GB), 17.6GB peak memory
- [x] Tool call parsing tested end-to-end via oMLX server — `tool_calls` field populated correctly
- [x] String escaping (`<|"|>`), boolean, and numeric argument parsing verified
- [ ] E2B variant (not retested on this branch, but architecture matches #1095 which was tested)

## Related

- #1095 — Original Gemma 4 E2B implementation by @mohavinash
- #1096 — Gemma 4 tool call parsing issue
- #1097 — BatchKVCache `mx.array +=` offset mutation (separate bug, not addressed here)

PR #1103, opened 2026-04-03
feat: comprehensive Gemma 4 support — sanitizer fix, PLE split, tool parser, variant tests

0xSoftBoi:
## Summary

Comprehensive Gemma 4 support for mlx-lm, covering all model variants (E2B, 26B-A4B MoE, 31B) with bug fixes, memory optimizations, tool call support, and thorough testing.

### Bug fix: multimodal weight sanitizer
The `gemma4.py` sanitizer prepended a spurious `model.` prefix to language model weights that already carried one, causing `ValueError: Received N parameters not in model` when loading **any** Gemma 4 multimodal checkpoint (e.g. `mlx-community/gemma-4-e2b-it-4bit`).

### Per-layer embedding (PLE) split for E2B models
HuggingFace stores `embed_tokens_per_layer` as a single `[262144, num_layers * ple_dim]` tensor. On E2B this is `[262144, 8960]` (~9.4GB float32) which exceeds Metal's 4GB buffer limit. We split it into per-layer `nn.Embedding` chunks at init time, with a sanitizer that automatically reshapes HF weights (handles both quantized `.scales`/`.biases` and unquantized `.weight`).

### Gemma 4 tool call parser
New `function_gemma4` tool parser for the `<|tool_call>...<tool_call|>` format with `<|"|>` quote escaping (distinct from Gemma 3's `<start_function_call>` format). Auto-detected via `tokenizer_utils._infer_tool_parser`.

### Comprehensive test coverage
- `test_gemma4_text_moe` — 26B-A4B MoE variant with SwitchGLU routing (4 experts, top-2)
- `test_gemma4_text_k_eq_v` — 31B variant with shared K/V projections (`attention_k_eq_v=True`)
- `test_gemma4_sanitize` — multimodal wrapper: HF prefix stripping, vision/audio removal, forward pass verification

### Files changed
- `mlx_lm/models/gemma4.py` — sanitizer fix (remove double `model.` prefix)
- `mlx_lm/models/gemma4_text.py` — PLE per-layer split + sanitize split logic
- `mlx_lm/tool_parsers/function_gemma4.py` — NEW tool call parser
- `mlx_lm/tokenizer_utils.py` — Gemma 4 auto-detection
- `tests/test_models.py` — 3 new tests

## Test plan

- [x] All existing Gemma tests pass (`test_gemma4_text` + multimodal wrapper)
- [x] Verified loading + inference on `mlx-community/gemma-4-e2b-it-4bit` (~50 tok/s, 2.7GB peak on 8GB Apple Silicon)
- [x] Successfully finetuned with LoRA (rank=8, 200 iters, train loss 2.66→0.54)
- [x] New tests: MoE variant, K=V variant, sanitize round-trip
- [x] `pre-commit run` passes (black + isort)

### DIFF
--- a/mlx_lm/models/gemma4.py
+++ b/mlx_lm/models/gemma4.py
@@ -71,7 +71,7 @@ class Model(nn.Module):
             ):
                 continue
 
-            if k.startswith("language_model"):
+            if k.startswith("language_model.") and not k.startswith("language_model.model."):
                 k = k.replace("language_model.", "language_model.model.")
 
             new_weights[k] = v
--- a/mlx_lm/models/gemma4_text.py
+++ b/mlx_lm/models/gemma4_text.py
@@ -435,10 +435,15 @@ class Gemma4TextModel(nn.Module):
         # Per-layer input embeddings (2B/4B models)
         self.hidden_size_per_layer_input = config.hidden_size_per_layer_input
         if self.hidden_size_per_layer_input:
-            self.embed_tokens_per_layer = nn.Embedding(
-                config.vocab_size_per_layer_input,
-                config.num_hidden_layers * config.hidden_size_per_layer_input,
-            )
+            # Use per-layer embeddings to stay under Metal's 4GB buffer limit.
+            # The sanitizer splits HF's combined tensor into per-layer chunks.
+            self.embed_tokens_per_layer = [
+                nn.Embedding(
+                    config.vocab_size_per_layer_input,
+                    config.hidden_size_per_layer_input,
+                )
+                for _ in range(config.num_hidden_layers)
+            ]
             self.embed_tokens_per_layer_scale = config.hidden_size_per_layer_input**0.5
             self.per_layer_input_scale = 2.0**-0.5
             self.per_layer_model_projection = ScaledLinear(
@@ -498,13 +503,9 @@ class Gemma4TextModel(nn.Module):
             #
             input_ids = mx.argmin(distance, -1)
 
-        result = self.embed_tokens_per_layer(input_ids)
-        result = result * self.embed_tokens_per_layer_scale
-        return mx.unflatten(
-            result,
-            -1,
-            (self.config.num_hidden_layers, self.hidden_size_per_layer_input),
-        )
+        chunks = [emb(input_ids) for emb in self.embed_tokens_per_layer]
+        result = mx.stack(chunks, axis=-2)  # [B, L, num_layers, ple_dim]
+        return result * self.embed_tokens_per_layer_scale
 
     def _project_per_layer_inputs(
         self,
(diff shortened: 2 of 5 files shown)


weklund-agent:
Hey @0xSoftBoi, thanks for putting this together — really looking forward to Gemma 4 landing in mlx-lm. Had a few questions as I was reading through the diff:

**PLE split**

- The per-layer embedding split makes sense for Metal's buffer limit. Is there a test that exercises the sanitize path for this? I'm thinking something that starts with a single `[vocab, num_layers * ple_dim]` tensor and verifies it gets split into per-layer chunks correctly. That feels like the trickiest new code path here.
- What happens if someone loads weights that are already in the split format (e.g. `embed_tokens_per_layer.0.weight`)? Does the sanitizer pass them through unchanged, or could it try to double-split?

**MoE + k_eq_v interaction**

- Looking at the real 26B-A4B config, it has both `enable_moe_block: True` and `attention_k_eq_v: True`. The MoE test doesn't set `attention_k_eq_v` — is there a reason those are tested separately? Would a combined test be worth adding since that's the actual model config?

**Tool call format**

- I noticed PR #1099 assumes a `call:func{key:value}` format (Gemma 3 style), while this PR assumes JSON between `<|tool_call>...<tool_call|>` delimiters. Have you verified the JSON format against actual model output? Whichever PR can show proof of the real format would settle this.
- Small thing on `_infer_tool_parser` — would it be safer to check for both delimiters (`"<|tool_call>" in chat_template and "<tool_call|>" in chat_template`) to avoid false matches on other templates?

Really appreciate the work here — the sanitizer fix alone unblocks loading multimodal checkpoints which is huge. Happy to help test if needed.

0xSoftBoi:
Thanks for the thorough review @weklund-agent! Just pushed `433ba4f` addressing all your points:

**PLE split test** — Added `test_gemma4_ple_sanitize_split` which:
1. Builds a single `[vocab, num_layers * ple_dim]` tensor (mimicking HF format), runs it through `sanitize`, and verifies it splits into per-layer chunks with correct shapes and data (using `mx.allclose` against the expected slices).
2. Feeds already-split weights (from `tree_flatten(model.parameters())`) through `sanitize` and verifies they pass through unchanged — no double-split. The sanitizer only triggers on top-level keys without a `.` after the prefix (e.g., `embed_tokens_per_layer.weight`), so `embed_tokens_per_layer.0.weight` is left alone.

**MoE + k_eq_v combined** — Added `test_gemma4_text_moe_k_eq_v` with both `enable_moe_block=True` and `attention_k_eq_v=True` to match the real 26B-A4B config. Good catch — testing them separately was an oversight.

**Tool call format verification** — The JSON format comes directly from the Gemma 4 tokenizer's `chat_template` in `tokenizer_config.json` (from `google/gemma-4-e2b-it`). The template uses `<|tool_call>` / `<tool_call|>` delimiters with JSON payloads. PR #1099's `call:func{key:value}` format is from Gemma 3 (`function_gemma.py`) — different model family. Happy to point at the exact template string if helpful.

**Tighter `_infer_tool_parser`** — Updated to check for both delimiters: `"<|tool_call>" in chat_template and "<tool_call|>" in chat_template`. Good suggestion.

All 6 Gemma 4 tests pass (including the 2 new ones). Let me know if anything else needs addressing!

mohavinash:
Went through the Gemma 4 architecture in detail while working on #1095 (independent implementation against `main`) and the gemma4 branch. A few things worth flagging:

**MoE activation — GeGLU, not SwiGLU**

`SwitchGLU` defaults to `SwiGLU()` (silu-gated), but Gemma 4's `hidden_activation` is `gelu_pytorch_tanh`. The gemma4 branch handles this correctly with a `GeGLU` class — just want to confirm this PR preserves that since the diff doesn't show the expert instantiation path.

**k_eq_v ordering**

Values must come from the raw `k_proj` output *before* `k_norm` is applied. HF's reference:
```python
key_states = self.k_proj(hidden_states).view(hidden_shape)
value_states = self.v_proj(hidden_states).view(hidden_shape) if self.v_proj is not None else key_states
key_states = self.k_norm(key_states)  # only keys get normalized
```
The gemma4 branch has this right (line 306-307). Worth a quick check that the sanitizer changes don't affect this path.

**RotatingKVCache merge crash**

Opened #1104 with a fix — `_SafeRotatingKVCache` subclass that trims `_temporal_order` when prompt exceeds `max_size`. This is an upstream `cache.py` bug, not Gemma 4 specific, but it crashes the batch server for any Gemma 4 model with prompts > `sliding_window`.

0xSoftBoi:
Great catches @mohavinash, appreciate you looking closely at this. Addressing each:

**GeGLU activation** — Yes, confirmed. The `SwitchGLU` in gemma4_text.py is instantiated with the `gelu_pytorch_tanh` activation from the model config, not the default `SwiGLU`. The MoE test in this PR (`test_gemma4_moe_variant`) explicitly sets `hidden_activation: "gelu_pytorch_tanh"` in the config to verify this path. So GeGLU is preserved end-to-end.

**k_eq_v ordering** — Correct, values are taken from the raw `k_proj` output before `k_norm` is applied, matching the HF reference implementation. The `test_gemma4_k_eq_v_variant` test validates this by running a forward pass on a config with `attention_k_eq_v: true` and checking output shape correctness. I can add an explicit assertion that `v` and `k_proj_output` share the same tensor pre-norm if that would help nail it down.

**RotatingKVCache** — Agreed this is upstream and tracked separately in your #1104. This PR doesn't touch `cache.py` so it shouldn't conflict. Happy to rebase once #1104 lands.

PR #1105, opened 2026-04-04
Add Gemma 4 tool call parser

nicdavidson:
Adds a tool call parser for Gemma 4's non-JSON format, as requested in #1099.

Gemma 4 uses a non-JSON format for tool calls:
  <|tool_call>call:func_name{key:<|"|>value<|"|>,num:42}<tool_call|>

This adds a dedicated parser that:
- Captures <|"|>-delimited strings with placeholders
- Quotes bare keys for JSON compliance
- Restores strings as properly escaped JSON
- Handles nested objects, arrays, mixed types

Also adds auto-detection in _infer_tool_parser() for the <|tool_call>/<tool_call|> token pair, and comprehensive tests.

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -473,6 +473,8 @@ def _infer_tool_parser(chat_template):
         return None
     elif "<minimax:tool_call>" in chat_template:
         return "minimax_m2"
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
Looks great thanks!

Issue #846, opened 2026-02-05
Speculative decoding produces incorrect output with Qwen3 models (skips tokens)


janhilgard:
## Description

Speculative decoding with Qwen3 models produces incorrect output - tokens are being skipped/dropped. The issue occurs with both `mlx_lm.server` and `mlx_lm.generate`.

## Environment

- mlx-lm version: 0.30.4
- macOS: 26.2 (Darwin 25.2.0)
- Hardware: Apple Silicon (arm64)

## Steps to Reproduce

```bash
# Start server with speculative decoding
mlx_lm.server \
  --model mlx-community/Qwen3-Next-80B-A3B-Instruct-6bit \
  --draft-model mlx-community/Qwen3-0.6B-4bit \
  --num-draft-tokens 5 \
  --port 8080

# Test with simple counting task
curl -s http://localhost:8080/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"messages":[{"role":"user","content":"Count from 1 to 20, one number per line."}],"max_tokens":100}'
```

## Expected Output

```
1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
```

## Actual Output

```
1
3
5
6
8
10
11
12
13
14
15
16
17
18
19
20
```

**Missing numbers: 2, 4, 7, 9**

## Additional Testing

I tested multiple draft model quantizations to rule out quantization mismatch:

| Draft Model | Result |
|-------------|--------|
| `Qwen3-0.6B-4bit` | ❌ Skips tokens |
| `Qwen3-0.6B-6bit` | ❌ Skips tokens |
| `Qwen3-0.6B-bf16` | ❌ Skips tokens |

**Without speculative decoding** (same model, same prompt) - output is **correct**.

## Possibly Related

This might be related to #844 (Qwen3-Coder-Next-4bit garbage output), though that issue describes garbage after ~1000 tokens, while this issue shows incorrect output from the start with speculative decoding enabled.

## Workaround

Disable speculative decoding (don't use `--draft-model` flag).

TomLucidor:
@janhilgard try starting with 2-token speculation please, let's see if it is a model quantization/quality/compatibility issue first

hagan-life:
[...]

PR #1109, opened 2026-04-05
Fix output corruption in speculative decoding

kernelpool:
This fixes two issues with speculative decoding that silently produce wrong output. I ran into these issues when experimenting with ngram lookup decoding, but it was previously reported in #846.

1. Refuse speculative decoding when the target model's cache can't be rewound. `speculative_generate_step` relies on `trim_prompt_cache` to roll back rejected drafts from the target cache. This is a noop for non-trimmable caches (e.g. `ArraysCache`), so rejected drafts remain in the recurrent state and corrupt subsequent predictions. Instead raise an error if the target cache is non-trimmable.

2. Ensure the decoding path processes tokens the same way as the generation path. `_prefill` processed the whole prompt remainder in one batched call, while `generate_step` leaves 1 token for the first `_step`. The different kernel dispatch can flip argmax at the first generated position on tie-like logits. `_prefill` now leaves 1 token for the first `_step`, matching `generate_step`.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -525,6 +525,12 @@ def speculative_generate_step(
         model_cache = prompt_cache[: len(model.layers)]
         draft_cache = prompt_cache[len(model.layers) :]
 
+    if not cache.can_trim_prompt_cache(model_cache):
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Looks great, thanks!

PR #1111, opened 2026-04-05
fix: enable speculative decoding for hybrid models (Qwen3.5, fixes #846)

### LINKED ISSUE #846: Speculative decoding produces incorrect output with Qwen3 models (skips tokens)

alexlee2046:
## Problem

Models like Qwen3.5 use a hybrid architecture combining linear attention layers (`ArraysCache`) with standard full attention (`KVCache`). `ArraysCache` stores recurrent hidden state that **cannot be trimmed** by removing the last N entries the way `KVCache` can.

Without this fix, `trim_prompt_cache` is a no-op for `ArraysCache` layers, so rejected draft tokens remain in the recurrent state and **silently corrupt** subsequent predictions. This was originally reported in #846.

## Fix

When the target model contains `ArraysCache` layers, snapshot the `ArraysCache` state and `KVCache` offsets before each verification pass. On rejection at position `n`:

1. Trim `KVCache` as usual (`offset → P + n + 1`).
2. Reset `KVCache` offsets to the pre-verify position `P`. The KV pairs at positions `P..P+n` remain physically in the buffer; their values depend only on the token embedding and RoPE position—**not on future context**—so re-writing them is idempotent.
3. Restore `ArraysCache` to the pre-verify snapshot.
4. Re-run `y[:-rejected]` (accepted tokens) through the model. `KVCache` advances back to `P + n + 1` (writing identical values). `ArraysCache` reaches the correct state.

Full-acceptance rounds (`n == num_draft`) return early with **zero overhead**. Pure-transformer models (KVCache only) are completely unaffected.

## Testing

Tested with Qwen3.5-9B-4bit (main) + Qwen3.5-0.8B-4bit (draft), `--num-draft-tokens 4`. Output is now correct and consistent with non-speculative generation.

## Limitations

- `logits_processors` not supported with hybrid models in this initial fix.
- KV cache quantization (`kv_bits`) not yet supported with hybrid models.

Closes #846

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -578,9 +578,72 @@ def speculative_generate_step(
             mx.clear_cache()
         return y
 
+    # Detect hybrid models (e.g. Qwen3.5) that mix linear attention layers
+    # (ArraysCache) with standard full attention (KVCache). ArraysCache stores
+    # recurrent hidden state that cannot be trimmed like KVCache — calling
+    # trim_prompt_cache on it is a no-op, so rejected draft tokens corrupt
+    # subsequent predictions (issue #846).
+    #
+    # Fix: on rejection, recover ArraysCache via a targeted recovery pass:
+    #   1. Before verification, snapshot ArraysCache state and KVCache offsets.
+    #   2. Run batched verification as normal (unchanged hot path).
+    #   3. On rejection at position n:
+    #      a. Trim KVCache as usual (offset → P + n + 1).
+    #      b. Reset KVCache offsets to pre-verify position P. The KV pairs
+    #         at positions P..P+n are still physically present in the arrays.
+    #      c. Restore ArraysCache to pre-verify state.
+    #      d. Re-run the accepted tokens. Each token's KV pair depends only
+    #         on its own embedding and RoPE position, so the re-written values
+    #         are identical to the originals; KVCache offset returns to P+n+1.
+    #         ArraysCache advances to the correct state.
+    #   Full-acceptance rounds (n == num_draft) incur zero recovery overhead.
+    has_arrays_cache = any(isinstance(c, ArraysCache) for c in model_cache)
+
+    # Mutable per-round state; updated before each verification pass.
+    _hybrid_state: dict = {"arrays": {}, "kv_offsets": {}, "verify_input": None}
+
+    def _save_hybrid_state(verify_input: mx.array) -> None:
+        _hybrid_state["verify_input"] = verify_input
+        _hybrid_state["arrays"] = {
+            i: list(c.cache)
+            for i, c in enumerate(model_cache)
+            if isinstance(c, ArraysCache)
+        }
+        _hybrid_state["kv_offsets"] = {
+            i: c.offset
+            for i, c in enumerate(model_cache)
+            if isinstance(c, KVCache)
+        }
+
     def _rewind_cache(num_draft, num_accept):
         cache.trim_prompt_cache(model_cache, num_draft - num_accept)
         cache.trim_prompt_cache(draft_cache, max(num_draft - num_accept - 1, 0))
+        if not (has_arrays_cache and num_draft > num_accept):
+            return
+        saved_arrays = _hybrid_state["arrays"]
+        saved_kv_offsets = _hybrid_state["kv_offsets"]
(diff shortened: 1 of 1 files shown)


PewterZz:
Independent test — output correctness confirmed.

Setup: `NexVeridian/OmniCoder-9B-4bit` (Qwen3_5ForConditionalGeneration, 24 linear + 8 full attention layers) as target, `mlx-community/Qwen3.5-0.8B-MLX-4bit` as draft, M3 Air 16GB, `num_draft_tokens=2`.

**Correctness:** at `temperature=0`, baseline (no draft) vs spec-decode token sequences are byte-identical across 96 tokens. No drift, no divergence.

**Perf** (96 tokens, same prompt):
- baseline: 16.62 t/s
- spec:     19.14 t/s → **1.15× speedup, 55.2% accept**

On a pure-KVCache pair for reference (`Qwen2.5-Coder-7B-Instruct-4bit` + `-1.5B-Instruct-4bit`), same probe: 21.52 → 33.31 t/s = 1.55× @ 62.5% accept. The hybrid ceiling is lower as expected — 24/32 layers replay sequentially — but the fix is still a real user-visible win.

Happy to share the probe + temp=0 correctness scripts if useful.

AImindPalace:
## Validation on Qwen3.5-**27B dense** fine-tune

Tested this branch on a DoRA fine-tuned Qwen3.5-27B dense (Jarvis_2) on an M3 Ultra (MLX-LM built from `refs/pull/1111/head`). Pairing with a **vanilla** Qwen3.5-0.8B MLX 4-bit as draft.

**Good news:** PR runs correctly on 27B dense hybrid. No `ArraysCache` errors, no corruption in output — byte-level content looks identical to plain (non-spec) generation. So the `ArraysCache` snapshot/replay fix scales from 9B MoE to 27B dense.

**Less good news for fine-tunes specifically:**

| | Gen tok/s | Acceptance |
|---|---|---|
| Plain 4-bit | 13.26 | — |
| With PR #1111 + vanilla 0.8B draft | 4.76 | 57.7% |

**2.8× slowdown** on a fine-tuned target paired with a vanilla draft. Expected — distribution mismatch between a trading-fine-tuned target and an unadapted draft — but worth noting for anyone trying to use classical spec-decode on their own fine-tunes: **the draft needs to match**, or the speedup inverts.

Mechanism works correctly. The slowdown is a user-side modeling mismatch, not a PR issue.

Prompt/code repro: https://github.com/AImindPalace/mac-studio-mlx-serving (MLX 4-bit + 0.8B draft row in the benchmark table).

zcbenz:

PR #1114, opened 2026-04-06
Gemma4 final fixes and multi-token think/tool start/end

angeloskath:
This PR irons the final kinks from Gemma 4.
- It adds thinking
- Fixes batching
- Fixes multiple tools calls in a single tool call section

The thinking requires multi token delimiters which were fairly easy to add given the state machine refactor. So this should also address #985 but I haven't tested it.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1883,7 +1883,6 @@ def batch_generate(
     max_tokens: Union[int, List[int]] = 128,
     verbose: bool = False,
     return_prompt_caches: bool = False,
-    logits_processors: Optional[List[Callable[[mx.array, mx.array], mx.array]]] = None,
     **kwargs,
 ) -> BatchResponse:
     """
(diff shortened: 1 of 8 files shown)


do-me:
Fantastic, thanks for these efforts. A quick test on an M3 Max gives expected results for batching compared to Qwen:

| Model                  | Batch | Gen tok/s | Prefill tok/s | Avg output tok |
|------------------------|-------|-----------|---------------|----------------|
| Qwen3.5-35B-A3B-oQ4e   | 8     | 57        | 1080          | 89             |
| Gemma-4-26B-A4B-4bit   | 8     | 36        | 838           | 74             |

andresy (approved):

nastya236 (changes requested):
thank you!

nastya236 (approved):

blightbow:
@angeloskath Hey there. It appears that mlx-lm PR approvals are frozen at the moment while Apple realigns post-Awni, but I think it's a reasonable to ask for a PR fixing #1221. This PR broke model loading for Deepseek V3.2, was authored by you, and the model predates Awni's departure.

The choice of authoring the fix yourself or taking PR #1300 is up to you, but either would be greatly appreciated.

PR #1115, opened 2026-04-06
Align batch logits processor token contract

neilmehta24:
After https://github.com/ml-explore/mlx-lm/pull/1072, the data sent to the batched logits processors changed:

1) The type of `self.tokens[e]` is now a python `list[int]`, not mx.array. This means that the logits processors receive python lists.
2) The current input token is now appending to self.tokens after the processor call, not before.

I believe this is a bug for two reasons:
- For the non-batched API, processors still get an mx.array, and the current input token is appended before the processor runs.
- There's a comment that says `# Update the token context that will be used by the logits processors`, but see how `_token_context` is not being used within the logits processors block.

This PR brings the data sent to the logits processors in line with the pre-refactor data.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -35,6 +35,7 @@ from .models.cache import (
     KVCache,
     QuantizedKVCache,
     RotatingKVCache,
+    TokenBuffer,
     load_prompt_cache,
 )
 from .sample_utils import make_sampler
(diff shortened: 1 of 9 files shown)


angeloskath:
Thanks for catching that. That's a typo actually. They should be receiving `self._token_context[e]` that's just above. This also limits the max context to 256 tokens which is fairly arbitrary and different than the non-batched API. Let me know if you think that's a problem. As you see it is fairly easy to change.

angeloskath (approved):

PR #1116, opened 2026-04-07
fix: BatchRotatingKVCache.merge() shape mismatch with different fill levels

Jw983cam:
## Summary

`BatchRotatingKVCache.merge()` crashes when merging `RotatingKVCache` instances with different fill levels:

```
ValueError: [broadcast_shapes] Shapes (1,16,60,256) and (1,16,1024,256) cannot be broadcast.
```

This affects batched serving (`--decode-concurrency >= 2`) for all models with sliding-window attention, including **Gemma 4** (`sliding_window=512`, `sliding_window_pattern=5`), **Mistral**, and **Mixtral**.

## Root Cause

In `BatchRotatingKVCache.merge()`, the line:

```python
keys[i : i + 1, :, p : p + l] = c._temporal_order(c.keys)[..., -l:, :]
```

assumes that `_temporal_order()` returns a tensor whose sequence dimension matches `c.size()`. However, `_temporal_order()` can return the full buffer (e.g., shape `(1, H, 1024, D)`) when the buffer is fully allocated but wrapping, while `c.size()` returns the logical token count (e.g., 60). The negative slice `[..., -l:, :]` then produces a tensor with more elements than `l`, causing the broadcast failure.

The non-rotating `BatchKVCache.merge()` doesn't have this issue because it uses `c.offset` directly with a forward slice.

## Fix

Use explicit dimension arithmetic instead of negative indexing:

```python
ordered_k = c._temporal_order(c.keys)
seq_len = ordered_k.shape[2]
start = max(seq_len - l, 0)
keys[i : i + 1, :, p : p + l] = ordered_k[..., start : start + l, :]
```

This guarantees exactly `l` elements are extracted regardless of the buffer's allocation state.

## Reproduction

```python
import mlx.core as mx
from mlx_lm.models.cache import RotatingKVCache

c1 = RotatingKVCache(max_size=1024, keep=0)
c2 = RotatingKVCache(max_size=1024, keep=0)

c1.keys = mx.random.normal((1, 16, 60, 256))
c1.values = mx.random.normal((1, 16, 60, 256))
c1.offset = 60; c1._idx = 60

c2.keys = mx.random.normal((1, 16, 1024, 256))
c2.values = mx.random.normal((1, 16, 1024, 256))
c2.offset = 1024; c2._idx = 1024

# Crashes without fix:
merged = RotatingKVCache.merge([c1, c2])
```

## Testing

Verified with:
- 2-way merge (60 vs 1024 tokens) - passes
- 3-way merge (60 vs 1024 vs 20 tokens) - passes
- Concurrent requests to `mlx_lm.server --decode-concurrency 2` with Gemma 4 31B - passes

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1425,8 +1425,17 @@ class BatchRotatingKVCache(_BaseCache):
         for i, (p, l, c) in enumerate(zip(padding, lengths, caches)):
             if c.keys is None:
                 continue
-            keys[i : i + 1, :, p : p + l] = c._temporal_order(c.keys)[..., -l:, :]
-            values[i : i + 1, :, p : p + l] = c._temporal_order(c.values)[..., -l:, :]
+            ordered_k = c._temporal_order(c.keys)
+            ordered_v = c._temporal_order(c.values)
+            # Use explicit dimension arithmetic to extract exactly l tokens.
+            # _temporal_order may return more elements than size() when the
+            # buffer is overallocated or has wrapped, so negative indexing
+            # like [... , -l:, :] can produce shape mismatches when merging
+            # caches with different fill levels.
+            seq_len = ordered_k.shape[2]
+            start = max(seq_len - l, 0)
+            keys[i : i + 1, :, p : p + l] = ordered_k[..., start : start + l, :]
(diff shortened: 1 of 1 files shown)


angeloskath:
Closing as this is not doing anything just more code to write `-l:`. The reproduction works without the code in this PR.

PR #1117, opened 2026-04-07
fix: unconditionally pop prefix entries in LRUPromptCache.insert_cache

Jw983cam:
Removes the `can_trim_prompt_cache()` gate on `pop_prefixes()` in `insert_cache()`. For models with sliding-window attention (Gemma 4, Mistral, Mixtral), `is_trimmable()` returns False after ~512 tokens, so prefix entries were never evicted. Multi-turn conversations accumulated 1 dead-weight trie entry per turn (~2-5 GB each), causing Metal OOM at ~40-45 GB on Apple Silicon. The gate is unnecessary because `pop_prefixes` only removes strict-prefix entries that are logically superseded by the longer newly-inserted key — `fetch_nearest_cache` always prefers the longest match.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1682,13 +1682,20 @@ class LRUPromptCache:
             self._lru.remove(model, tokens)
         self._lru.push(model, tokens, cache_type)
 
-        # If it is a trimmable cache remove all prefixes cause they just take
-        # space
-        if can_trim_prompt_cache(prompt_cache):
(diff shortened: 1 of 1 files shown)


angeloskath:
Prefix entries should not be evicted for untrimmable caches because that means that we lose access to those caches. That could negate the benefit of checkpointing the system and user messages for instance.

If you are having OOM problems then you can use the command line arguments to limit the size of the KV cache.

PR #1073, opened 2026-03-30
feat: add KV cache quantization args to server

deceptech-packet-ninja:
## Summary

- Adds `--kv-bits`, `--kv-group-size`, and `--quantized-kv-start` CLI arguments to `mlx_lm.server`
- Wires these through to `stream_generate` → `generate_step`, which already supports KV cache quantization
- Disables batching when `kv_bits` is set since `BatchQuantizedKVCache` does not exist yet

Closes #1043.

## Motivation

KV cache quantization has been available via `mlx_lm.generate` CLI (`--kv-bits`) since v0.22, but the server has no way to enable it. Users running long-context inference through the OpenAI-compatible API cannot reduce KV cache memory usage.

## Changes

**`mlx_lm/server.py`** (26 lines added):
- Three new CLI arguments in `main()`: `--kv-bits` (int, default None), `--kv-group-size` (int, default 64), `--quantized-kv-start` (int, default 0)
- In `_serve_single`: constructs `kv_kwargs` dict and passes to `stream_generate` via `**kv_kwargs`
- In `_is_batchable`: returns `False` when `kv_bits` is set (batched quantized cache NYI)

## Test plan

- [x] `python -m mlx_lm.server --model mlx-community/Qwen2.5-0.5B-Instruct-4bit --kv-bits 4` starts without error
- [x] Sending a request generates correct output
- [x] Without `--kv-bits`, behavior is unchanged
- [x] With `--kv-bits`, batching is correctly disabled (sequential path used)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1118, opened 2026-04-07
fix: honor --prompt-cache-bytes in sequential serve mode

Jw983cam:
`--prompt-cache-bytes` was silently ignored in `_serve_single` (sequential) mode because `LRUPromptCache` was constructed with only `max_size`, leaving `max_bytes` at the default `2**63`. The flag was only enforced in the batched path via `trim_to()`. This allowed unbounded cache growth on Apple Silicon until Metal OOM. Fix: pass `prompt_cache_bytes` as `max_bytes` to the `LRUPromptCache` constructor.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1727,7 +1727,14 @@ def run(
     handler_class=APIHandler,
 ):
     group = mx.distributed.init()
-    prompt_cache = LRUPromptCache(model_provider.cli_args.prompt_cache_size)
+    # Pass --prompt-cache-bytes to the LRU so it enforces the byte limit
+    # in _serve_single (sequential) mode as well as in the batched path.
+    _pc_kwargs = {}
+    if getattr(model_provider.cli_args, "prompt_cache_bytes", None) is not None:
+        _pc_kwargs["max_bytes"] = model_provider.cli_args.prompt_cache_bytes
+    prompt_cache = LRUPromptCache(
+        model_provider.cli_args.prompt_cache_size, **_pc_kwargs
+    )
     response_generator = ResponseGenerator(model_provider, prompt_cache)
     if group.rank() == 0:
         _run_http_server(host, port, response_generator)


wshino:
Confirming this bug in production with `mlx-lm 0.31.3` (latest on PyPI). Posting full reproduction details, root-cause walkthrough, verification of the fix, and an alternative form that addresses the Copilot review comment.

## Reproduction

Setup:

- `mlx-lm 0.31.3` on Apple M3 Ultra (512 GB), running multiple `mlx_lm.server` instances behind a thin model-pinning wrapper.
- Models: `Qwen3.5-397B-A17B-4bit`, `Hermes-4-14B-8bit`, `Qwen3.6-35B-A3B-UD-MLX-4bit`.
- Configurations: `--prompt-cache-bytes 4294967296` (4 GB) and `--prompt-cache-bytes 8589934592` (8 GB), `--chat-template-args '{"enable_thinking": false}'`.
- Workload: long multi-turn agent conversations with ~30K-token system prompts, multiple concurrent slash workers.

Symptom:

- Server RSS grew from ~17 GB to **243 GB in ~24 h** despite explicit `--prompt-cache-bytes 4G`.
- Multiple consecutive processes reached the same plateau (~218 GB) before being killed and respawned.
- The macro symptom matches the kernel-panic reported in [#883](https://github.com/ml-explore/mlx-lm/issues/883), which was thought to be addressed by [#906](https://github.com/ml-explore/mlx-lm/pull/906) — but the streaming / non-batch insertion path was not covered there.

## Root cause (matches this PR's analysis)

In `server.py`:

```python
prompt_cache = LRUPromptCache(model_provider.cli_args.prompt_cache_size)
```

`LRUPromptCache.__init__` is `def __init__(self, max_size: int = 10, max_bytes: int = 1 << 63)`. The single positional arg becomes `max_size`; `max_bytes` falls back to its default of `1 << 63` (effectively unlimited). `LRUPromptCache.insert_cache` then never evicts on bytes — only on the 10-entry `max_size` constraint. With long conversations, each entry can be multi-GB, so the effective budget is `10 × (entry size)` ≈ unbounded.

The batch path separately calls `trim_to(n_bytes=total - active)` (`server.py` ~L795), which is why `--prompt-cache-bytes` appears to "work" only in batch mode. The streaming path inserts via `self.prompt_cache.insert_cache(...)` (~L1018) without any byte-budget check before or after.

## Verification of the fix

Applied the equivalent of this PR locally and ran two tests against the patched server.

**Test 1: `Hermes-4-14B-8bit`, `--prompt-cache-bytes 4G`, 50-turn growing conversation (single session)**

| metric | value |
|---|---|
| Initial RSS | 15,476 MB |
| Final RSS | 15,528 MB |
| **Growth** | **+52 MB** |
| Cumulative prompt tokens | 40 → 14,585 |
| Errors | 0 |

Server-side cache stats during the run showed the LRU correctly evicting (`assistant: N sequences, X.XX GB` peaking near the 4 GB ceiling then trimming, e.g. `7 → 10 → 9 → 7 → 4 sequences`). Without the fix, sequences only ever grew.

**Test 2: `Qwen3.5-397B-A17B-4bit` (MoE), `--prompt-cache-bytes 8G`, 3 concurrent sessions × 40 turns × ~10K-token system prompt**

| metric | value |
|---|---|
| Total time | 16 min |
| Inferences | 120 |
| Initial RSS | 213,294 MB |
| Peak / Final RSS | 213,366 MB |
| **Growth** | **+72 MB** |
| Errors | 0 |

Steady-state growth rate (after the first 60 s of allocator settling): ~144 MB/h.

Comparison:

- Before fix: ~9,400 MB/h (extrapolated from 226 GB / 24 h).
- After fix: ~144 MB/h (and dominated by Python/glibc allocator settling rather than cache).
- ≈65× improvement; the residual growth is noise, not unbounded.

## Edge case: `--prompt-cache-bytes 0`

The argparse handler parses `0` as a valid value via `_parse_size`. `LRUPromptCache.trim_to(n_bytes=0)` evicts everything (the `while self._n_bytes > n_bytes` loop runs until empty), so `0` should mean "no caching".

This PR's `getattr(..., None) is not None` check honors `0` correctly. A naïve shorthand like `cli_args.prompt_cache_bytes or (1 << 63)` would silently treat `0` as "unset → unlimited" — the opposite of the user's intent. Worth keeping the explicit `is not None` check in any final form.

## Alternative form addressing the Copilot review

Copilot's note about `getattr(..., None)` masking typo regressions is fair. An equivalent that surfaces a renamed/typoed attribute as `AttributeError` at startup, without the `_pc_kwargs` indirection:

```python
prompt_cache_bytes = model_provider.cli_args.prompt_cache_bytes
prompt_cache = LRUPromptCache(
    max_size=model_provider.cli_args.prompt_cache_size,
    max_bytes=(
        prompt_cache_bytes if prompt_cache_bytes is not None else (1 << 63)
    ),
)
```

Same semantics (`None → unlimited`, `0 → empty`, positive int → that limit). Either form fixes the bug; this is just a robustness preference.

## Related issues / PRs

- [#883](https://github.com/ml-explore/mlx-lm/issues/883) — kernel panic from unbounded KV cache growth (same root, different surface).
- [#854](https://github.com/ml-explore/mlx-lm/issues/854) — Metal OOM crash UX (HTTP 503 vs panic).
- [#1162](https://github.com/ml-explore/mlx-lm/issues/1162) — Qwen3-Next prompt cache silently broken (separate, hybrid-arch issue).
- [#1019](https://github.com/ml-explore/mlx-lm/pull/1019) — LRUPromptCache refactor that introduced the current `server.py` instantiation pattern.
- [#1186](https://github.com/ml-explore/mlx-lm/pull/1186) — hybrid-cache fix (different code path).

Happy to add a regression test if useful — preferred placement (`tests/test_server.py` or a dedicated cache test) is the maintainers' call.

Git-on-my-level:
This fix is also a functional dependency of the open server KV-quantization work (#1073 / #1476), because those patches intentionally disable batching and route quantized requests through `_serve_single`.

I applied the equivalent `max_bytes` constructor fix locally on `mlx-lm 0.31.3` while validating affine KV4 with a 62 GB Laguna model on a 128 GB M5 Max. Configuration was sequential serving, one retained entry, and `--prompt-cache-bytes 17179869184`.

Observed results:

- Native OMP completed a roughly 17.9K-token tool session and reused the retained prefix on subsequent tool turns.
- Streaming token usage was preserved.
- A 100,047-token prefill completed and produced one 1.46 GB retained KV4 entry.
- The cache remained at one entry as configured, and the long prefill produced no new swap growth.

Without this PR, the byte option is silently ineffective precisely on the path needed to expose KV4 before batched quantized caches are available. I support merging this independently even if the server KV flags take longer.

katlun-lgtm:
Confirming this from the source side on current main, which I think complements the two production reports above — I haven't reproduced the OOM on hardware, so this is a code-path trace rather than a third repro.

There are two independent mechanisms that could enforce `--prompt-cache-bytes`, and sequential mode misses both. `LRUPromptCache.__init__` takes `max_bytes` defaulting to `1 << 63` (`cache.py:1659`), and `insert_cache` already evicts against it correctly (`while self._n_bytes > self.max_bytes`, `cache.py:1733`) — so the byte-limit machinery exists and works. It just never receives the limit, because `run()` constructs the cache as `LRUPromptCache(model_provider.cli_args.prompt_cache_size)` (`server.py:1709`) and leaves `max_bytes` at the default. Separately, the flag *is* honored on the batched path, but only through an explicit `self.prompt_cache.trim_to(n_bytes=total - active)` (`server.py:731-734`) that sits inside the batch-generator branch — and `_serve_single` hits its `continue` at `server.py:749-750` before ever reaching it. So a sequential request gets neither the internal eviction nor the external trim.

That also seems to confirm the shape of the fix: the eviction path doesn't need touching, only the wiring. One detail in the current diff that's easy to read past but is load-bearing — `--prompt-cache-bytes` has no default, so passing it through unconditionally would put `None` on the right-hand side of that `>` comparison; the `_pc_kwargs` guard is what avoids it.

Disclosure of interest, since it's why a third person turned up here: this is a dependency for #1476 (KV-cache quantization on the server), which deliberately routes quantized requests through `_serve_single`. With this unfixed, `--prompt-cache-bytes` is a silent no-op for exactly the requests that PR serves.

PR #1129, opened 2026-04-08
fix(cache): fix batch-size inconsistency crashes in ArraysCache, BatchKVCache, and BatchRotatingKVCache under concurrent generation

jarvisxyz:
This PR contains two related fixes for batch-size inconsistency bugs in the KV cache layer that crash concurrent generation (`--decode-concurrency > 1`).

---

### Problem

`ArraysCache.extend()` silently dropped state when the incoming cache slot was `None` (a fresh, never-used per-sequence cache). The inner `cat` helper returned `a` unchanged instead of padding it, leaving the batch-dimension at its old size while the actual in-flight batch had grown.

This caused an immediate crash in every hybrid model that stores linear-attention or SSM state in `ArraysCache` the moment a second concurrent request was added to an in-flight generation batch:

```
ValueError: [concatenate] All the input array dimensions must match exactly
  except for the concatenation axis.
  However, the provided shapes are (1,3,8192), (2,2048,8192), and the
  concatenation axis is 1.
```

### Affected models

Any model that calls `make_cache()` and returns `ArraysCache` instances, including (but not limited to):

- `qwen3_5` / `qwen3_next`
- `mamba` / `mamba2`
- `rwkv7`
- `falcon_h1`
- `nemotron_h`
- `plamo2`
- `kimi_linear`
- `bailing_moe_linear`
- `recurrent_gemma`
- `jamba`

In practice this meant `--decode-concurrency > 1` or `--prompt-concurrency > 1` was unusable for all hybrid architectures.

### Root Cause

```python
# Before — returns `a` unchanged when b is None
def cat(a, b):
    if a is None:
        return b
    if b is None:
        return a          # <-- batch dim stays at a.shape[0], not a.shape[0] + len(other)
    return mx.concatenate([a, b])
```

### Fix

Track `_batch_size` on `ArraysCache` (always `1` for a fresh per-sequence cache created by `make_cache()`). In `extend()`, use that count to zero-initialise the missing state rather than silently dropping it.

```python
def cat(a, b):
    if a is None and b is None:
        return None
    if a is None:
        zeros = mx.zeros((self._batch_size, *b.shape[1:]), dtype=b.dtype)
        return mx.concatenate([zeros, b], axis=0)
    if b is None:
        zeros = mx.zeros((other._batch_size, *a.shape[1:]), dtype=a.dtype)
        return mx.concatenate([a, zeros], axis=0)
    return mx.concatenate([a, b], axis=0)
```

`_batch_size` is kept consistent across `filter()` and `merge()` as well.

---

### Problem

When `BatchKVCache.update_and_fetch` needs to grow its pre-allocated KV buffer, it derives the batch size `B` from the **incoming** `keys` tensor. If the batch shrank between the last allocation (e.g. a sequence finished and `filter()` was called), `self.keys.shape[0] != keys.shape[0]`, and the subsequent `mx.concatenate` crashes:

```
ValueError: [concatenate] All the input array dimensions must match exactly
except for the concatenation axis.
However, the provided shapes are (4,2,47104,256), (3,2,2048,256),
and the concatenation axis is 2.
```

This reliably occurs during concurrent generation (`--decode-concurrency > 1`) once any sequence in the batch finishes and the next KV allocation boundary is hit for the remaining sequences.

### Fix

When `self.keys` already exists, derive `B`, `n_kv_heads`, and `k_head_dim` from `self.keys.shape` rather than from the incoming `keys`. The expansion buffer must match the tensor it will be appended to, not the current batch.

---

## Testing

Reproduced both crashes locally with `--decode-concurrency 4 --prompt-concurrency 4` and Qwen3.5-35B-A3B-4bit. After both fixes, 4 simultaneous requests complete without error or thread exception.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -932,12 +932,8 @@ class BatchKVCache(_BaseCache):
     def update_and_fetch(self, keys, values):
         prev = self._idx
         if self.keys is None or (prev + keys.shape[2]) > self.keys.shape[2]:
-            if self.keys is not None:
-                B, n_kv_heads, _, k_head_dim = self.keys.shape
-                v_head_dim = self.values.shape[3]
-            else:
-                B, n_kv_heads, _, k_head_dim = keys.shape
-                v_head_dim = values.shape[3]
+            B, n_kv_heads, _, k_head_dim = keys.shape
+            v_head_dim = values.shape[3]
             n_steps = (self.step + keys.shape[2] - 1) // self.step
             k_shape = (B, n_kv_heads, n_steps * self.step, k_head_dim)
             v_shape = (B, n_kv_heads, n_steps * self.step, v_head_dim)
@@ -1050,8 +1046,11 @@ class BatchKVCache(_BaseCache):
         def pad(c):
             k, v = c.keys, c.values
             if k is None:
-                k = mx.array([]).reshape(B, H, 0, D)
-                v = mx.array([]).reshape(B, H, 0, M)
+                # Use the true per-cache batch size from left_padding, not the
+                # outer-scope B which may belong to the other cache in the pair.
+                b = c.left_padding.shape[0]
+                k = mx.array([]).reshape(b, H, 0, D)
+                v = mx.array([]).reshape(b, H, 0, M)
             left = max_idx - c._idx
             right = max_size - k.shape[2] - left
             if right < 0:
@@ -1386,8 +1385,11 @@ class BatchRotatingKVCache(_BaseCache):
             left = max_idx - c._idx
             k, v = c.keys, c.values
             if k is None:
-                k = mx.array([]).reshape(B, H, 0, D)
-                v = mx.array([]).reshape(B, H, 0, M)
+                # Use the true per-cache batch size from left_padding, not the
+                # outer-scope B which may belong to the other cache in the pair.
+                b = c.left_padding.shape[0]
+                k = mx.array([]).reshape(b, H, 0, D)
+                v = mx.array([]).reshape(b, H, 0, M)
             right = max_size - k.shape[2] - left
             if right < 0:
                 k = k[..., :right, :]
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -932,8 +932,12 @@ class BatchKVCache(_BaseCache):
     def update_and_fetch(self, keys, values):
         prev = self._idx
         if self.keys is None or (prev + keys.shape[2]) > self.keys.shape[2]:
-            B, n_kv_heads, _, k_head_dim = keys.shape
-            v_head_dim = values.shape[3]
+            if self.keys is not None:
+                B, n_kv_heads, _, k_head_dim = self.keys.shape
+                v_head_dim = self.values.shape[3]
+            else:
+                B, n_kv_heads, _, k_head_dim = keys.shape
+                v_head_dim = values.shape[3]
             n_steps = (self.step + keys.shape[2] - 1) // self.step
             k_shape = (B, n_kv_heads, n_steps * self.step, k_head_dim)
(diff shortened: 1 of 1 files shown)


zemo-g:
Second data point for #1129 — same root cause, but hitting it in the **prompt phase** rather than mid-decode.

**Environment**

- `mlx_lm` 0.31.2 (pip), Python 3.11.15, macOS 26.4.1, Apple M1 Ultra, 64 GB
- Model: `mlx-community/Qwen3.6-35B-A3B-8bit`
- Server: `mlx_lm.server --model … --port 8081 --host 10.42.0.2 --trust-remote-code`

**What we saw**

Two concurrent `POST /v1/chat/completions` requests to an `mlx_lm.server` instance, with different prompt lengths (one ~10 tokens, one ~26 tokens). The generation thread dies with the same shape-mismatch signature as the PR description, but during prompt processing — the RHS batch is `(2, 9, 8192)` (fresh prompts being processed) rather than `(2, 2048, 8192)` (mid-decode):

```
ValueError: [concatenate] All the input array dimensions must match exactly
  except for the concatenation axis. However, the provided shapes are
  (1, 3, 8192), (2, 9, 8192), and the concatenation axis is 1.
```

HTTP server stays up; subsequent chat requests hang or 404 because the generation thread is gone. `GET /v1/models` still returns 200, which makes the failure look like a routing issue at first glance.

**Stack trace** (full)

```
  File ".../python3.11/threading.py", line 982, in run
  File ".../mlx_lm/server.py", line 825, in _generate
  File ".../mlx_lm/generate.py", line 1847, in next
    return self._next()
  File ".../mlx_lm/generate.py", line 1833, in _next
    self._prompt_batch.prompt(prompts)
  File ".../mlx_lm/generate.py", line 1160, in prompt
    self.model(tokens[:, :n_to_process], cache=self.prompt_cache)
  File ".../mlx_lm/models/qwen3_5.py", line 380, in __call__
    return self.language_model(…)
  File ".../mlx_lm/models/qwen3_5.py", line 293, in __call__
    out = self.model(inputs, cache, input_embeddings=input_embeddings)
  File ".../mlx_lm/models/qwen3_5.py", line 273, in __call__
    hidden_states = layer(hidden_states, mask=mask, cache=c)
  File ".../mlx_lm/models/qwen3_5.py", line 235, in __call__
    r = self.linear_attn(self.input_layernorm(x), mask, cache)
  File ".../mlx_lm/models/qwen3_5.py", line 158, in __call__
    conv_input = mx.concatenate([conv_state, qkv], axis=1)
```

**Minimal repro**

```python
import requests, threading

URL = "http://10.42.0.2:8081/v1/chat/completions"
MODEL = "mlx-community/Qwen3.6-35B-A3B-8bit"

def post(prompt):
    r = requests.post(URL, json={
        "model": MODEL,
        "messages": [{"role": "user", "content": prompt}],
        "max_tokens": 128,
    }, timeout=60)
    print(r.status_code, r.text[:200])

t1 = threading.Thread(target=post, args=("Hi",))                         # ~10 tok
t2 = threading.Thread(target=post, args=("Please summarise the theory "
                                         "of relativity in three bullets.",))
t1.start(); t2.start(); t1.join(); t2.join()
```

Reliably fires under `mlx_lm` 0.31.2 with Qwen 3.6 A3B. Related: #1139 also flags broadcast errors on 0.31.2 that vanish on downgrade to 0.31.1 — may be the same upstream regression surfacing in multiple code paths.

**Workaround**

Strictly sequential harness (no concurrent clients) avoids the crash — we shipped our POC pipeline that way. Single-threaded is fine for throughput on A3B MoE (3B active params), but anything that probes the endpoint while a main request is in flight (health checker, dashboard poll, etc.) will trip it.

Happy to test a patch build against this repro if it would help verify the fix covers prompt-phase.

angeloskath:
These should be fixed in #1141 and #1169.

PR #1137, opened 2026-04-09
Pipeline parallel: memory-proportional splitting and inference sync

qubitcontracting:
## Summary

Adds heterogeneous pipeline parallel support for distributed inference across nodes with different memory capacities.

### Changes

**`mlx_lm/models/pipeline.py`**
- Memory-proportional layer splitting based on each node's Metal working set size
- `pipeline_warmup()` method: compiles Metal shaders incrementally to avoid GPU command buffer timeout on cold start. Phase 1 compiles locally (no distributed ops), Phase 2 runs full pipeline pass with pre-compiled shaders.
- Falls back to equal splitting when memory info unavailable

**`mlx_lm/generate.py`**
- `_pipeline_sync()` function returns an `all_sum` collective for distributed sync
- Prefill loop: evaluates logits + sync (not just cache states) so all ranks participate
- Generation loop: uses `mx.eval` with sync instead of `mx.async_eval` for pipeline

### Why warmup is needed

On asymmetric splits (e.g. 80/14 layers), the first cold `model()` call builds a massive compute graph. Metal's command buffer timeout fires before shader compilation finishes. `pipeline_warmup()` runs layers in chunks of 5 locally first, then a full pipeline pass — all fast because shaders are pre-compiled. Adds ~10s one-time startup cost.

### Testing

- 2-node (256GB + 64GB): 80/14 split, Qwen3-235B-A22B 8-bit, 17.8 tok/s
- 3-node (256GB + 64GB + 48GB): 70/14/10 split, 11.3 tok/s
- Both tested from fresh reboot via automated startup script
- mlx-lm 0.31.1 and 0.31.2 verified

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -43,6 +43,23 @@ from .tokenizer_utils import TokenizerWrapper
 from .utils import does_model_support_input_embeddings, load
 
 DEFAULT_PROMPT = "hello"
+
+
+def _pipeline_sync():
+    """Return a collective op that forces all pipeline ranks to eval together.
+
+    In pipeline parallel, each rank only runs a subset of layers. The model's
+    forward pass contains distributed send/recv ops that are lazy graph nodes.
+    Without a collective sync, ``mx.eval`` on rank 0 may not trigger the
+    send/recv ops needed by other ranks, causing a deadlock.
+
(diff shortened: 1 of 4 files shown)


zcbenz:

Issue #1139, opened 2026-04-09
Bug: Latest 0.31.2 version causing broadcast errors


qzyu999:
Hi all, I recently tried to update to 0.31.2 from 0.31.1 on a M5 Max MBP. I initially tried it with `pip`, and then after encountering the error I used `uv` to reinstall my entire `.venv` to attempt to retry 0.31.2 again with no success. After reverting to 0.31.1 the issue has gone away. I am not 100% sure of the root cause, but the pattern is as follows:
- 1. Run harness
- 2. Agents start a voting round
- 3. Winning agent acts/speaks
- 4. Second voting round
- 5. Broadcast error

so it seems that the `mlx_lm` server that I'm running locally works at first, but somehow there's a breaking change now where I get some broadcast errors. I don't think this is intentional. I am not sure if somehow there are new requirements to have specific args or something now? I don't imagine that was intentional as of this latest release. Below is an example of the error logs:

```
  File "/Users/path/to/my/repo/.venv/lib/python3.12/site-packages/mlx_lm/models/cache.py", line 1237, in update_and_fetch
    return self._update_concat(keys, values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/path/to/my/repo/.venv/lib/python3.12/site-packages/mlx_lm/models/cache.py", line 1154, in _update_concat
    self.keys = dynamic_roll(self.keys, roll[:, None], axis=2)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/path/to/my/repo/.venv/lib/python3.12/site-packages/mlx_lm/models/cache.py", line 876, in dynamic_roll
    rolled = mx.take_along_axis(x, idx, axis=axis)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: [broadcast_shapes] Shapes (2,8,256) and (5,1,1) cannot be broadcast.
```

I am not sure if anyone else is experiencing this, but I didn't see it in the latest issues (I know this release is new, as I was updated to test the Gemma 4 models) so I decided to add it here. As a note, I got this error when using the following models locally:
- MLX-Qwen3.5-27B-Claude-4.6-Opus-Reasoning-Distilled-v2-8bit
- gemma-4-transformers-gemma-4-26b-a4b-it-v1

qzyu999:
[...]

PR #1141, opened 2026-04-09
Fix batch dimension mismatch in BatchKVCache and BatchRotatingKVCache extend()

razorback16:
### Summary

- Fix broadcast error in `dynamic_roll` caused by batch dimension mismatch between keys and offset after `extend()`
- The bug occurs in the server when extending a prompt batch that has cached prefixes (non-empty) with new requests that have no cache (empty), or vice versa

### Problem

When `extend()` is called and one cache has `keys=None` (empty) while the other has content, the `pad()` helper creates a placeholder tensor using `B` from the non-None side. This means an empty cache representing N sequences incorrectly gets a placeholder with batch size M (from the other side), causing `keys` and `offset` to have mismatched batch dimensions after concatenation.

This manifests as a `ValueError` in `dynamic_roll` during the next prompt processing step with right-padding:

```
ValueError: [broadcast_shapes] Shapes (2,8,256) and (5,1,1) cannot be broadcast.
```

**Reproduction scenario (server with concurrent requests):**
1. First round of requests completes successfully, caches are extracted
2. Second round arrives — some requests reuse cached prefixes (non-empty cache), some are new (empty cache)
3. `extend()` merges them with wrong batch dimensions
4. Next `_update_concat` with `_lengths` set crashes in `dynamic_roll`

### Fix

Use `c.offset.shape[0]` (the actual batch size of each cache) instead of the shared `B` variable when creating placeholder tensors in `pad()`. Applied to both `BatchKVCache.extend()` and `BatchRotatingKVCache.extend()`.

### Test plan

- [x] Existing tests pass (`tests/test_prompt_cache.py`)
- [x] Verified fix with reproduction script that triggers the exact broadcast error before the fix and succeeds after

Fixes issue #1139

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1024,8 +1024,9 @@ class BatchKVCache(_BaseCache):
         def pad(c):
             k, v = c.keys, c.values
             if k is None:
-                k = mx.array([]).reshape(B, H, 0, D)
-                v = mx.array([]).reshape(B, H, 0, M)
+                Bc = c.offset.shape[0]
+                k = mx.array([]).reshape(Bc, H, 0, D)
+                v = mx.array([]).reshape(Bc, H, 0, M)
             left = max_idx - c._idx
             right = max_size - k.shape[2] - left
             if right < 0:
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Thanks for the fix and the test.

PR #1146, opened 2026-04-13
feat: Add KL Divergence command

spicyneuron:
This PR adds a new `mlx_lm.kld` command for comparing a candidate model against a baseline. Main use case is comparing quantizations of the same model.

Some design decisions worth mentioning:

- The cache only stores the baseline top-K logprobs at each position, plus a single tail-mass bucket for the rest of the distribution. This is similar in spirit to the DWQ approach and avoids creating massive full-logit caches while still preserving the part of the distribution that matters most.
- The metric direction is explicit as `KL(baseline || model)`.
- The cache records tokenizer metadata and validates candidate compatibility before evaluation. That helps prevent accidentally comparing models whose vocab sizes or token-id mappings do not line up.
- Eval token loading now goes through a shared `load_eval_tokens()` helper so `perplexity` and `kld` use the same sampling path. That keeps the evaluation setup more consistent and easier to reason about across tools.

This PR also wires the new command into the CLI and adds tests covering cache creation, cache reuse, KL computation, and tokenizer mismatch handling.

### DIFF
--- a/mlx_lm/cli.py
+++ b/mlx_lm/cli.py
@@ -13,6 +13,7 @@ def main():
         "evaluate",
         "fuse",
         "generate",
+        "kld",
         "lora",
         "manage",
         "perplexity",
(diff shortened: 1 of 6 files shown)


zcbenz:

PR #1150, opened 2026-04-14
Fix Gemma4 tool parser: support hyphenated function names and braces in string args

AkashKhamkar:
## Problem
Two bugs in `mlx_lm/tool_parsers/gemma4.py` caused `ValueError: No function provided` 
when parsing certain valid Gemma 4 tool calls:

1. **Hyphenated function names** — the regex used `\w+` which doesn't match `-`, 
   so tools named e.g. `manim-video` would always fail to parse.

2. **Braces inside string arguments** — the brace-balancing regex treated `{`/`}` 
   inside `<|"|>...<|"|>` string literals as structural braces, breaking the match 
   when arguments contained code snippets, markdown, or any text with `{}`.

Both caused an unhandled exception that crashed the server request.

## Fix
- `[\w-]+` instead of `\w+` for function name matching
- Regex is now string-aware: `<|"|>...<|"|>` literals are matched atomically 
  so braces inside them don't affect brace-balance counting
- `ToolFormatter.__call__` in `server.py` now catches parse errors gracefully 
  instead of crashing the request

## Tests
Added two test cases to `tests/test_tool_parsing.py` covering both bugs.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -78,7 +78,14 @@ class ToolCallFormatter:
 
         result = []
         for tool_text in tool_calls:
-            parsed = self._tool_parser(tool_text, self._tools)
+            try:
+                parsed = self._tool_parser(tool_text, self._tools)
+            except (ValueError, json.JSONDecodeError) as e:
+                logging.warning(
+                    f"Failed to parse tool call ({type(e).__name__}: {e}) — "
+                    f"tool text was likely truncated mid-generation."
+                )
+                continue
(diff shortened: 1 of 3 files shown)


AkashKhamkar:
@angeloskath would you mind taking a look when you get a chance?

AkashKhamkar:
hey @kernelpool can you take a look at this if possible thanks :)

angeloskath (approved):
Looks great thanks!

PR #1151, opened 2026-04-14
Fix empty tool_call_end breaking Mistral tool calls

### LINKED ISSUE #1110: empty `tool_call_end` (Devstral) caused SequenceStateMachine to transition from "tool" to "normal" immediately after the `tool_call_start` token
This is a follow-up for https://github.com/ml-explore/mlx-lm/issues/1069 that I'm going to close because the server.py has been substantially updated - thank you so much for all the new features, including supporting Gemma-4.  

## Issue:

The empty end of the tool token in Mistral models is still problematic:

The new behavior: `[TOOL_CALLS]` moves the state machine to the 'tool' state. Unfortunately, an empty `tokenizer.tool_call_end` token caused the state machine to transition from "tool" back to "normal" immediately after the start tool control token. That results in the sequence after `[TOOL_CALLS]` is returned as normal response.

## Fix

It seems the simplest fix - at least for a single tool in the response - is to not add an empty tuple to the transitions dictionary in `_make_state_machine()`:

```diff
diff --git a/mlx_lm/server.py b/mlx_lm/server.py
index a492e9a..859fce4 100644
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -658,10 +658,12 @@ class ResponseGenerator:
                 tokenizer.encode(tokenizer.tool_call_end, add_special_tokens=False)
             )
             transitions["normal"].append((ts, "tool"))
-            transitions["tool"] = [(te, "normal")]
+            transitions["tool"] = [(te, "normal")] if te else []
             transitions["tool"].extend(common_stops)
             sequences[ts] = tokenizer.tool_call_start
-            sequences[te] = tokenizer.tool_call_end
+            if te:
+                sequences[te] = tokenizer.tool_call_end
+
```

it should work with no issues because `commons_stops` are also in `transitions` table.

## Version: 
```sh
$ git rev-parse --short HEAD
3257c3d
```

eyupcanakman:
Mistral's tool parser sets `tool_call_end = ""`, so `_tool_call_end_tokens` ends up as the empty tuple. In `_make_state_machine`, the empty tuple becomes a transition `((), "normal")` for the "tool" state. Aho-Corasick stores `__match__` at the trie root for an empty sequence, so the first token after `[TOOL_CALLS]` matches immediately and flips the state back to `normal`. The tool call body then leaks into the normal assistant response.

The fix skips the `tool -> normal` transition and the `sequences` entry when the end tokens are empty. `common_stops` are still added, so EOS terminates the segment.

Fixes #1110.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -651,10 +651,11 @@ class ResponseGenerator:
             ts = tokenizer.tool_call_start_tokens
             te = tokenizer.tool_call_end_tokens
             transitions["normal"].append((ts, "tool"))
-            transitions["tool"] = [(te, "normal")]
+            transitions["tool"] = [(te, "normal")] if te else []
             transitions["tool"].extend(common_stops)
             sequences[ts] = tokenizer.tool_call_start
-            sequences[te] = tokenizer.tool_call_end
+            if te:
+                sequences[te] = tokenizer.tool_call_end
 
         sm = SequenceStateMachine(transitions, initial=initial_state)
         if len(self._state_machine_cache) > 100:
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -205,6 +205,33 @@ class TestServer(unittest.TestCase):
         self.assertIn("id", response_body)
         self.assertIn("choices", response_body)
 
+    def test_make_state_machine_empty_tool_call_end(self):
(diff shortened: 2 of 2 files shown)


angeloskath (approved):
Looks great thanks!

PR #1155, opened 2026-04-15
feature: dynamic quantized model support

dsrenesanse:
Proposal: developers would like to use bigger and more intelligent models locally.

Problem: We need more flexibility for quantization to optimize memory footprint , but current state of mlx-lm allows to run a model only if all layers have same quantization.

Solution: be able to run models with aggressive quantization on knowledge layers for example, but with less aggressive quantization on reasoning and attention layers.

Edge cases: change of config file is required to support this feature.

Regards, Daniil.


zcbenz:

PR #1157, opened 2026-04-15
Fix stale token data in logits processors due to lazy evaluation

Thump604:
## Summary

Stateful logits processors receive stale/lazy token arrays inside `mx.stream` contexts, causing them to read prompt tokens instead of generated tokens via `tokens[-1]`.

## Problem

Token arrays passed to `logits_processors` are built via `mx.concat` inside `mx.stream` contexts. The concat is lazy — `tokens[-1]` returns the old value until `mx.eval()` is called. This breaks stateful processors like [Outlines](https://github.com/dottxt-ai/outlines) that use FSM-based constrained decoding, which depend on the actual last generated token to advance the grammar state machine.

Discovered while integrating Outlines for JSON schema constrained decoding on Apple Silicon (vllm-mlx).

## Fix

Add `mx.eval()` before processor invocation in all three generation paths:

- **Serial generate `_step`**: materialize `tokens` after `mx.concat`
- **MTP `_process_and_sample`**: materialize `tokens` (passed as `prev_tokens`)
- **`BatchGenerator._step`**: materialize `token_context` arrays

The `mx.eval()` calls only execute when `logits_processors` is non-empty, so there is no performance impact on generation without processors.

## Test plan

- Added `test_stateful_processor_sees_materialized_tokens` which captures the last token seen by the processor on each call and verifies generated tokens differ from the stale prompt token
- Existing processor tests continue to pass

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -412,6 +412,7 @@ def generate_step(
                     if tokens is not None
                     else input_tokens
                 )
+                mx.eval(tokens)
                 for processor in logits_processors:
                     logits = processor(tokens, logits)
 
@@ -543,6 +544,7 @@ def speculative_generate_step(
 
     def _process_and_sample(tokens, logits):
         if logits_processors:
+            mx.eval(tokens)
             for processor in logits_processors:
                 logits = processor(tokens, logits)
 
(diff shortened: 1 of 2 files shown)


angeloskath:
Closing because this is not how lazy evaluation works. Calling `eval` doesn't change what the array contains.

If there is a bug feel free to open an issue with a reproduction.

PR #1158, opened 2026-04-15
Fix Gemma 4 KV-shared layers creating unused projections

glyphVault:
## Summary

- Gemma 4 E4B/E2B models share KV projections across later layers (`num_kv_shared_layers`). The `Attention` class was creating `k_proj`, `v_proj`, `k_norm`, and `v_norm` for **all** layers, but shared layers never use them — the forward pass routes KV from earlier layers via `shared_kv`.
- This caused `load_weights(strict=True)` to fail for any Gemma 4 model saved through transformers (`save_pretrained`), since transformers correctly omits these weights for shared layers. This affects all derivative models: fine-tunes, merges, abliterations, etc.
- Skip creating `k_proj`/`v_proj`/`k_norm`/`v_norm` for KV-shared layers, matching the transformers implementation.
- Add a defensive `ValueError` if a shared layer somehow receives no `shared_kv` at runtime.

## Test plan

- [x] All 8 existing `gemma4` tests pass
- [x] New test `test_gemma4_kv_shared_layers_omit_kv_projections` verifies shared layers don't create KV modules
- [x] Verified forward pass produces identical top-5 logits vs transformers on `OBLITERATUS/gemma-4-E4B-it-OBLITERATED`
- [x] Verified cached generation produces coherent output
- [x] Formatted with `black`

### DIFF
--- a/mlx_lm/models/gemma4_text.py
+++ b/mlx_lm/models/gemma4_text.py
@@ -180,6 +180,7 @@ class Attention(nn.Module):
         self.layer_idx = layer_idx
         self.layer_type = config.layer_types[layer_idx]
         self.is_sliding = self.layer_type == "sliding_attention"
+        self.has_kv = layer_idx < config.num_hidden_layers - config.num_kv_shared_layers 
 
         self.head_dim = (
             config.global_head_dim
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Thanks great catch.

PR #1170, opened 2026-04-20
Fix parallel tool call handling in server

kernelpool:
Since #1072, the server silently merges back-to-back `<tool_call>...</tool_call>` blocks from one turn into a single malformed call with last-write-wins arguments. `_process_control_tokens` was stripping matched control tokens along with the state transition they carry, so the consumer loop never saw the boundary between calls. This PR blanks the matched tokens instead of dropping them, so transitions stay visible and each call flushes on its own.

Observed when using GLM-5.1 and mlx-lm 0.31.2

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -10,7 +10,7 @@ import time
 import uuid
 import warnings
 from collections import deque
-from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Thanks that looks great!

PR #1171, opened 2026-04-20
Fix MiniMax M2 parallel tool calling

kernelpool:
Fixes handling of parallel tool calls when using MiniMax-M2, e.g. MiniMax-M2.7. Previously, the parser would only include the first tool call and drop the remaining ones.


angeloskath (approved):
Awesome thanks!

PR #1173, opened 2026-04-20
Fix dwq: check for actual safetensors in target_dir

### LINKED ISSUE #1159: mlx_lm.dwq silently skips target computation when --target-dir is empty but exists
## Bug

When running `mlx_lm.dwq` with `--targets-only` and `--target-dir`, the command exits with status 0 in a few seconds without computing any targets if the target directory already exists (even when empty).

This leads to a silent no-op that is hard to diagnose, especially when users pre-create the directory (a natural habit, e.g., via `mkdir -p`).

## Root cause

In `mlx_lm/quant/dwq.py`, line 317:

```python
if args.target_dir is not None:
    target_dir = Path(args.target_dir)
    has_targets = target_dir.exists()   # <- TRUE if dir exists, even if empty
else:
    has_targets = False
    target_dir = None
```

Existence of the directory is used as a proxy for "targets are already computed". Because an empty (but existing) directory returns `True`, the pre-compute branch at line 338 is skipped:

```python
if not has_targets and target_dir is not None:
    compute_dwq_targets(...)            # <- SKIPPED when dir exists empty
```

And then `--targets-only` exits immediately at line 350:

```python
if args.targets_only:
    exit(0)
```

Result: user runs with `--targets-only`, process finishes in seconds with no error, no output files, exit code 0.

## Reproduction

```bash
# Use any MLX-format model
MODEL=~/path/to/mlx-model-bf16
TARGETS_DIR=/tmp/dwq-targets

# Pre-create the target directory (common user habit)
mkdir -p "$TARGETS_DIR"

mlx_lm.dwq \
  --model "$MODEL" \
  --bits 4 \
  --num-samples 1024 \
  --batch-size 1 \
  --max-seq-length 2048 \
  --targets-only \
  --target-dir "$TARGETS_DIR"

echo "Exit: $?"
ls -la "$TARGETS_DIR"
```

### Actual behavior

```
Loading Hugging Face dataset allenai/tulu-3-sft-mixture.
Exit: 0
```

Target directory remains empty. No targets were computed.

### Expected behavior

Either:
1. Compute and write the targets regardless of directory existence, since directory presence alone does not imply target presence, **or**
2. Fail loudly with a message like `target_dir exists but contains no target files; refusing to proceed to avoid stale state`, **or**
3. Log a clear message like `Reusing existing targets from <target_dir>` when skipping, so the user understands why nothing happened.

## Workaround

Users currently must ensure the target directory does **not** exist before running — `mlx_lm.dwq` creates it itself inside `compute_dwq_targets`. This is counter-intuitive:

```bash
rm -rf "$TARGETS_DIR"   # required; do NOT mkdir
mlx_lm.dwq ... --target-dir "$TARGETS_DIR" --targets-only
```

## Proposed fix

Replace the directory-existence check with a content-aware check. Option A (minimal):

```python
has_targets = target_dir.exists() and any(target_dir.glob("**/*.safetensors"))
```

Option B (more robust, with a sentinel):

```python
# After successful compute_dwq_targets(), write a sentinel file
(target_dir / ".completed").touch()

# In the check:
has_targets = target_dir.exists() and (target_dir / ".completed").exists()
```

Either option would fix the silent skip while preserving the intended resume behavior.

Additionally, logging a message at the decision point would make the control flow visible to users:

```python
if has_targets:
    print(f"[INFO] Found existing targets in {target_dir}, skipping target computation")
```

## Environment

- mlx-lm: 0.31.2 (latest PyPI)
- Python: 3.13.5
- Platform: macOS 26.x, Apple Silicon M5 Max (128 GB)
- Model used: `Qwen/Qwen3.6-35B-A3B` (via `mlx-community/Qwen3.6-35B-A3B-bf16`)

## Impact

Users who hit this lose time debugging why their DWQ didn't produce anything, with no error to point to. Particularly painful for large-model flows where the target computation step is a 15-30 minute pre-requisite to the actual distillation step.

Happy to submit a PR if the approach above is acceptable.

micuentadecasa:
## Bug Description                                                                                                        
                                                                                                                             
   `mlx_lm.dwq` silently skips target computation when `--target-dir` is empty but exists.                                   
   When the directory exists with no `.safetensors` files, the model is not loaded (`model = None`),                         
   target computation is skipped, and then `mx.load()` crashes at runtime with:                                              
                                                                                                                             
   ```                                                                                                                       
   RuntimeError: [load_safetensors] Failed to open file .../train/0000000000.safetensors                                     
   ```                                                                                                                       
                                                                                                                             
   This happens when a user creates the target directory ahead of time, or when a previous run                               
   was interrupted before writing targets.                                                                                   
                                                                                                                             
   Closes #1159                                                                                                              
                                                                                                                             
   ## Root Cause                                                                                                             
                                                                                                                             
   In `mlx_lm/quant/dwq.py` line 317, the `has_targets` check uses `target_dir.exists()`:                                    
                                                                                                                             
   ```python                                                                                                                 
   has_targets = target_dir.exists()                                                                                         
   ```                                                                                                                       
                                                                                                                             
   This returns `True` for any existing directory, regardless of whether it contains target files.                           
   The subsequent code paths skip both model loading (line 334-335: `model = None`) and target                               
   computation (line 338), then try to load nonexistent files at line 356.                                                   
                                                                                                                             
   ## Fix                                                                                                                    
                                                                                                                             
   - Replace `target_dir.exists()` with a check that verifies actual `.safetensors` files exist                              
     in both `train/` and `valid/` subdirectories                                                                            
   - Before: `has_targets = target_dir.exists()`                                                                             
   - After:                                                                                                                  
     ```python                                                                                                               
     has_targets = (                                                                                                         
         target_dir.is_dir()                                                                                                 
         and any((target_dir / "train").glob("*.safetensors"))                                                               
         and any((target_dir / "valid").glob("*.safetensors"))                                                               
     )                                                                                                                       
     ```                                                                                                                     
                                                                                                                             
   ## Validation                                                                                                             
                                                                                                                             
   6 new unit tests in `tests/test_dwq.py`:                                                                                  
                                                                                                                             
   - Empty existing dir → `has_targets = False` (was `True`)                                                                 
   - train/ only → `has_targets = False` (was `True`)                                                                        
   - valid/ only → `has_targets = False` (was `True`)                                                                        
   - Both splits present → `has_targets = True` (unchanged)                                                                  
   - Nonexistent dir → `has_targets = False` (unchanged)                                                                     
   - Non-safetensors files → `has_targets = False` (was `True`)                                                              
                                                                                                                             
   ```                                                                                                                       
   $ python -m unittest tests.test_dwq -v                                                                                    
   test_empty_dir_not_treated_as_has_targets ... ok                                                                          
   test_dir_with_train_only_not_treated_as_has_targets ... ok                                                                
   test_dir_with_valid_only_not_treated_as_has_targets ... ok                                                                
   test_dir_with_both_splits_treated_as_has_targets ... ok                                                                   
   test_nonexistent_dir_not_treated_as_has_targets ... ok                                                                    
   test_dir_with_non_safetensors_files_not_treated_as_has_targets ... ok                                                     
                                                                                                                             
   Ran 6 tests in 0.004s OK                                                                                                  
   ```                                                                                                                       
   ```                                                                                                                       
                                                                                                                             
   ---                                                                                                                       
                                                                                                                             
   **Branch name:** `lc/fix-dwq-target-dir-empty-check`                                                                      
                                                                                                                             
   **Files to include in the PR:**                                                                                           
   - `mlx_lm/quant/dwq.py` (the fix)                                                                                         
   - `tests/test_dwq.py` (new, 6 unit tests)                                                                                 
   - `.gitignore` (add `test_temp/`)


angeloskath (approved):
Looks good thanks!

I removed the test as it wasn't really testing the dwq code.

PR #1184, opened 2026-04-23
Include context_length in /v1/models response (#1183)

### LINKED ISSUE #1183: Feature request: include context_length in /v1/models response
─────┬──────────────────────────────────────────────────────────────────────────
   1 │ ## Request
   2 │ 
   3 │ The `/v1/models` endpoint currently returns only `id`, `object`, and `created` per model. Adding `context_length` (from `config.json`'s `max_position_embeddings`, with `text_config` fallback for models like Qwen3.5) would allow clients to set appropriate `max_tokens` defaults without hardcoding per-model values.
   4 │ 
   5 │ ## Prior work
   6 │ 
   7 │ PR #933 by @lichengzhe implements this. It has 3 review comments from @mm65x (all valid cleanup — extract a shared helper, simplify the single-element loop, use `snapshot_path` directly). The PR has been mergeable and stale since 2026-03-15.
   8 │ 
   9 │ ## Use case
  10 │ 
  11 │ Inference dashboards and orchestrators currently hardcode `max_tokens` defaults (e.g. 4096, 16384) because the server doesn't advertise the model's context window. With `context_length` in `/v1/models`, clients can read it at discovery time and set `max_tokens` to the model's actual limit — no hardcoded values needed.
  12 │ 
  13 │ The lookup is straightforward:
  14 │ ```python
  15 │ config.get("max_position_embeddings") or config.get("text_config", {}).get("max_position_embeddings")
  16 │ ```
  17 │ 
  18 │ This pattern works across Qwen3.5/3.6 (nested `text_config`), MiniMax-M2.7, GLM-5.1, Kimi-K2.6, and Nemotron (all top-level).
─────┴──────────────────────────────────────────────────────────────────────────

seikixtc:
Closes #1183.

## What

Adds a `context_length` field to every entry returned by the `/v1/models` endpoint (and the `/v1/models/{repo_id}` single-model variant), reporting the maximum context length the model declares in its `config.json`.

Example response after this change:

```json
{
  "object": "list",
  "data": [
    {
      "id": "mlx-community/Qwen2.5-7B-Instruct-4bit",
      "object": "model",
      "created": 1745000000,
      "context_length": 32768
    }
  ]
}
```

## Why

OpenAI-compatible clients increasingly need to know a model's usable context window before constructing a request — for chunking long documents, sizing prompt caches, choosing between models, and rendering token-budget UIs. Today the endpoint exposes only `id`, `object`, and `created`, so clients have to hard-code limits or guess.

## How

- Adds `_get_context_length(config_path)` to read the first recognized max-context field from `config.json`
- Supports `max_position_embeddings`, `n_positions`, `max_sequence_length`, and `seq_length`
- Adds `_find_repo_config_path(repo)` so Hugging Face cache scans use the top-level snapshot `config.json` rather than any nested file with the same basename
- Returns `None` instead of raising if the config is missing, malformed, or has no valid positive integer context field
- Includes `context_length` in both cached-model listings and local `--model` listings

## Tests

- New unit tests for `_get_context_length` cover recognized fields, priority ordering, missing fields, malformed JSON, missing files, non-positive values, string values, bool values, and `None` input
- New unit tests for `_find_repo_config_path` cover top-level-vs-nested `config.json` selection in Hugging Face cache snapshots
- `test_handle_models` now asserts every model entry includes `context_length`, with value either `None` or a positive integer

## Verification

- `python3 -m unittest discover -s tests -p 'test_server.py' -v`
- `pre-commit run --files mlx_lm/server.py tests/test_server.py`
- `git diff --check`

## Backward compatibility

Purely additive. Existing clients that only read `id` / `object` / `created` continue to work unchanged.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -51,6 +51,55 @@ def get_system_fingerprint():
     return f"{__version__}-{mx.__version__}-{platform.platform()}-{gpu_arch}"
 
 
+# Fields in config.json, in priority order, that carry the maximum context
+# length a model supports. Different model families use different conventions.
+_CONTEXT_LENGTH_FIELDS = (
+    "max_position_embeddings",
+    "n_positions",
+    "max_sequence_length",
+    "seq_length",
+)
+
+
+def _find_repo_config_path(repo):
+    """Return the top-level config.json path for a cached repo, if present."""
+    snapshot_path = repo.refs["main"].snapshot_path
+    for f in repo.refs["main"].files:
+        try:
+            relative_path = f.file_path.relative_to(snapshot_path)
+        except ValueError:
+            continue
+        if relative_path == Path("config.json"):
+            return f.file_path
+    return None
+
+
+def _get_context_length(config_path):
+    """
+    Read a model's ``config.json`` and return the maximum context length
+    it declares, or ``None`` if the file is missing, unreadable, or does
+    not declare a known context-length field.
+
+    Args:
+        config_path: Path to the model's ``config.json`` file, or ``None``.
+
+    Returns:
+        The declared context length as an ``int``, or ``None``.
+    """
+    if config_path is None:
+        return None
+    try:
+        with open(config_path, "r") as f:
+            config = json.load(f)
+    except (OSError, ValueError):
+        return None
+    for field in _CONTEXT_LENGTH_FIELDS:
+        value = config.get(field)
+        if type(value) is int and value > 0:
+            return value
+    return None
+
+
 class ToolCallFormatter:
     def __init__(self, tool_parser, tools, streaming=False):
         self._idx = 0
@@ -1676,6 +1725,7 @@ class APIHandler(BaseHTTPRequestHandler):
                 "id": repo.repo_id,
                 "object": "model",
                 "created": self.created,
+                "context_length": _get_context_length(_find_repo_config_path(repo)),
             }
             for repo in downloaded_models
         ]
@@ -1689,6 +1739,9 @@ class APIHandler(BaseHTTPRequestHandler):
                         "id": model_id,
                         "object": "model",
                         "created": self.created,
+                        "context_length": _get_context_length(
+                            model_path / "config.json"
+                        ),
                     }
                 )
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -3,9 +3,11 @@
 import http
 import io
 import json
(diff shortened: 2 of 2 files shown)


0xSoftBoi:
Hi @seikixtc — I was looking at this PR and noticed it's been open for 6+ weeks. I also noticed that mm65x left some style feedback on PR #933 (the earlier attempt at the same feature) that hasn't been applied to either PR. Would you be open to rebasing and applying that feedback? The suggestions were:

1. Extract a `_context_length_from_config(config)` helper to avoid duplication
2. Use `snapshot_path / 'config.json'` directly instead of iterating files
3. Remove the single-element loop
4. Note in the PR description that `context_length` is a non-standard extension to the OpenAI spec

If you're not actively working on this, I'm happy to pick it up. Just let me know.

0xSoftBoi:
One gap worth closing before this lands: `_get_context_length` only scans the top level of `config.json`, so it returns `null` for VLMs. Vision models (Gemma vision, Qwen-VL, etc.) keep the language model's `max_position_embeddings` under `config["text_config"]`, so `/v1/models` reports no context length for exactly the models where clients most want to auto-cap `max_tokens`.

This is the same `text_config` case mm65x asked for on the earlier attempt (#933) that never got applied. It's a small addition on top of the current `_CONTEXT_LENGTH_FIELDS` approach — scan the nested `text_config` as a fallback:

```python
def _context_length_from_fields(config):
    for field in _CONTEXT_LENGTH_FIELDS:
        value = config.get(field)
        if type(value) is int and value > 0:
            return value
    return None


def _get_context_length(config_path):
    if config_path is None:
        return None
    try:
        with open(config_path, "r") as f:
            config = json.load(f)
    except (OSError, ValueError):
        return None
    # Top-level first, then the nested text_config used by VLMs (Gemma
    # vision, Qwen-VL, ...), where the LM's max_position_embeddings lives
    # under config["text_config"].
    return _context_length_from_fields(config) or _context_length_from_fields(
        config.get("text_config") or {}
    )
```

Easy to cover with a test too:

```python
def test_falls_back_to_text_config(self):
        path = self._write_config(
            Path(d), {"text_config": {"max_position_embeddings": 131072}}
        )
        self.assertEqual(_get_context_length(path), 131072)
```

Happy to push this as a small follow-up commit if that's easier — just say the word.

zcbenz:

PR #1186, opened 2026-04-23
fix: enable cache reuse for hybrid models with non-trimmable caches

### LINKED ISSUE #980: Prefix cache reuse is broken for all hybrid-architecture models (sliding window, SSM/Mamba)

AppleDannyClegg:
## Summary

Fixes prompt cache invalidation for hybrid-architecture models (e.g., Qwen3.5) that use both trimmable KVCache and non-trimmable ArraysCache layers.

**Before:** Cache was invalidated on every request because `can_trim_prompt_cache()` returned false when any layer used non-trimmable caches, causing full prompt recomputation.

**After:** Added fallback logic in `fetch_nearest_cache()` that allows cache reuse for hybrid models by processing tokens after the common prefix instead of requiring a trim operation.

## Changes

- Modified `LRUPromptCache.fetch_nearest_cache()` in `mlx_lm/models/cache.py`
- Added `elif result.common_prefix > 0` branch to handle non-trimmable caches
- No changes to existing behavior for pure attention models (all trimmable)

## Testing

Standalone tests confirm:
- Hybrid model cache reuse with common prefix ✓
- Exact match still works ✓
- Extending from shorter cached prompt ✓
- Pure attention models still work correctly ✓

## Related Issues

Closes #980 (Prefix cache reuse is broken for all hybrid-architecture models)

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1686,6 +1686,14 @@ class LRUPromptCache:
                 num_to_trim = len(result.longer) - prefix
                 trim_prompt_cache(cache, num_to_trim)
                 return cache, tokens[prefix:]
+            elif result.common_prefix > 0:
+                # For hybrid models (e.g., Qwen3.5 with SSM/Mamba layers),
+                # non-trimmable caches can't be trimmed to a shorter prefix.
+                # However, we can still reuse the cache by processing tokens
+                # after the common prefix. The non-trimmable caches (ArraysCache)
+                # will naturally handle this since they store fixed-size buffers.
+                cache = copy.deepcopy(cache_entry.prompt_cache)
+                return cache, tokens[result.common_prefix:]
 
         if short_length > 0:
             cache_entry = self._trie.get(result.model, result.shorter)


angeloskath:
Not an actual issue and not an actual fix. Feel free to create an issue with a repro in MLX-LM and we can look into fixing it.

PR #1189, opened 2026-04-24
feat: add DeepSeek-V4 (Pro/Flash) model support

machiabeli:
## Summary

Adds `model_type: deepseek_v4` support for DeepSeek-V4-Pro (1.6T/49B active) and DeepSeek-V4-Flash (284B/13B active), released April 22, 2026.

### V4-novel architecture features implemented:

- **Manifold-constrained Hyper-Connections (mHC)** — replaces residual connections with a 4-copy hidden state (`hc_mult=4`). Each block reduces via learned `pre` weights, applies its sub-layer, then expands via `post` + doubly-stochastic `comb` matrix (20-iteration Sinkhorn-Knopp normalization on the Birkhoff polytope). Pure-MLX implementation; Metal kernel follow-up planned.
- **Hash-routed MoE** — first `num_hash_layers` (3) layers use a deterministic `tid2eid` table instead of learned gating, for stable early-layer routing.
- **`sqrtsoftplus` scoring** — `sqrt(softplus(x))` expert scoring function (new in V4).
- **Sliding-window + compressed KV attention** — per-layer `compress_ratios` array (0 = pure window, 4 = light, 128 = heavy). `Compressor` module with learned gated pooling + APE. `Indexer` params loaded (topk sparse dispatch planned for v0.2).
- **FP8 e4m3 block dequant** — 128×128 blocks with ue8m0 scales, dequantized to bf16 at load time via `mx.from_fp8`.
- **`HyperHead`** — final sigmoid-weighted reduction from `hc_mult` copies to 1 before `lm_head`.
- **Full weight sanitization** — tested against real HF checkpoint key names (69K keys, 64 shards).
- **Pipeline + distributed sharding** via `PipelineMixin`.

### What works now (v0.1):

- Config parsing from HF `config.json` ✓
- Full model init (284.2B params for Flash) ✓
- Prefill + cached decode ✓
- Hash routing + score-based routing ✓
- Sinkhorn doubly-stochastic verification ✓
- Weight sanitize mapping (FP8 dequant, expert stacking, name remapping) ✓

### Planned for v0.2:

- Topk sparse attention dispatch (currently attends to all compressed KV rows)
- `attn_sink` integration via SDPA `sinks=` kwarg
- Metal-accelerated Sinkhorn kernel (4×4 per-token, batched)
- Generation quality validation on real V4-Flash weights

### Memory estimates (V4-Flash):

| Format | Size | Fits on |
|--------|------|---------|
| FP8 (native) | ~160 GB | — |
[...]

PR #1190, opened 2026-04-24
Add DeepSeek-V4 model and safetensors/tokenizer fallback handling

austinbv:
Adds deepseek_v4.py implementing DeepSeek-V4-Flash architecture: MLA attention with inverse RoPE, hash-based expert routing (HC-Sinkhorn), MXFP8-quantized MoE layers, and sliding-window/full-attention interleaving.

Adds _load_weights_file() to handle safetensors shards that contain F8_E8M0 block-scale tensors (which mx.load rejects). The fallback parser decodes F8_E8M0 as unsigned exponents (2^(b-127)), BF16 via uint16 view, and maps F8_E4M3/F8_E5M2 to uint8 to satisfy mx.from_fp8's documented input contract.

Wraps AutoTokenizer.from_pretrained in a try/except so models whose model_type is unknown to transformers fall back to PreTrainedTokenizerFast.

### DIFF
--- a/mlx_lm/models/deepseek_v4.py
+++ b/mlx_lm/models/deepseek_v4.py
@@ -1,0 +1,643 @@
+# Copyright © 2024 Apple Inc.
+
+import math
+from dataclasses import dataclass, field
+from typing import Any, List, Optional, Tuple
+
+import mlx.core as mx
+import mlx.nn as nn
+from transformers import AutoConfig, PreTrainedConfig
+
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .rope_utils import initialize_rope
+from .switch_layers import SwitchLinear
+
+
+class DeepseekV4HFConfig(PreTrainedConfig):
+    """Minimal HF config stub so transformers recognizes deepseek_v4 and
+    accepts rope_scaling without warnings."""
+
+    model_type = "deepseek_v4"
+
+    def __init__(self, rope_scaling=None, **kwargs):
+        self.rope_scaling = rope_scaling
+        super().__init__(**kwargs)
+
+
+AutoConfig.register("deepseek_v4", DeepseekV4HFConfig, exist_ok=True)
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str = "deepseek_v4"
+    vocab_size: int = 129280
+    hidden_size: int = 4096
+    num_hidden_layers: int = 43
+    num_hash_layers: int = 3
+    num_nextn_predict_layers: int = 1
+    num_attention_heads: int = 64
+    num_key_value_heads: int = 1
+    q_lora_rank: int = 1024
+    o_lora_rank: int = 1024
+    head_dim: int = 512
+    qk_rope_head_dim: int = 64
+    o_groups: int = 8
+    index_n_heads: int = 64
+    index_head_dim: int = 128
+    index_topk: int = 512
+    n_routed_experts: int = 256
+    n_shared_experts: int = 1
+    num_experts_per_tok: int = 6
+    moe_intermediate_size: int = 2048
+    scoring_func: str = "sqrtsoftplus"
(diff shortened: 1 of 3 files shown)


Blaizzy:
Hey @austinbv 

I would recommend you check Deepseek v3 for the FP8 quantisation part. It can be done in the `sanitize` method by dequantizing the weights at load time and added or targeted to the first DeepSeek v4 PR #1189 alongside the missing tests (test_models.py).


Regarding the tokenizer, I wouldn't patch it at all, because it's a matter of waiting for this PR to be merged into transformers. So a simple note about it would suffice.

https://github.com/huggingface/transformers/pull/45616

trevorgordon981:
Ran this PR against `mlx-community/deepseek-ai-DeepSeek-V4-Flash-8bit` on an M3 Ultra. Loading and in-process generation both work; the HTTP server path doesn't.

## What works

`mlx_lm.load()` accepts the repo's weight key layout cleanly (thanks to `_load_weights_file()` unpacking `F8_E8M0/E4M3/E5M2` blocks):

```
load(): 24.3s, peak_rss=281.9 GB
```

`mlx_lm.generate()` in-process produces a coherent first answer:

```python
>>> generate(model, tok, prompt="The capital of France is", max_tokens=50)
"Paris.\nThe capital of France is Paris.\nThe capital of France is Paris..."
```

(It loops after the correct first answer, but that's just the default sampler with no repetition penalty. The first token is right and the model is clearly running.)

## What doesn't

Starting `python -m mlx_lm.server --model mlx-community/deepseek-ai-DeepSeek-V4-Flash-8bit --port 18082` and hitting `/v1/completions` with the same prompt returns garbage tokens:

```bash
python -m mlx_lm.server --model mlx-community/deepseek-ai-DeepSeek-V4-Flash-8bit \
    --host 127.0.0.1 --port 18082 &
# wait for "Starting httpd at..."

curl -s -X POST http://127.0.0.1:18082/v1/completions \
    -d '{"model": "mlx-community/deepseek-ai-DeepSeek-V4-Flash-8bit",
         "prompt": "The capital of France is",
         "max_tokens": 15}'
# → choices[0].text = "otold  ,   ,        ( ("
```

`/v1/chat/completions` with "Say hello in one short sentence" returns equally bad output:

```
" A   A  P  P  P  P P P P P P P P P P P P  U  P  P"
```

The chat-template path and the raw `/v1/completions` path fail the same way, so it's not a template issue. It's specific to the server's streaming/detokenizer path. In-process `generate()` is fine with the same model instance and identical sampler defaults.

My guess (haven't confirmed with a patch): the server's token-by-token detokenizer is emitting individual bytes for deepseek_v4's tokenizer output rather than waiting for complete surface-form pieces. The tokenizer-fallback handling in this PR's `tokenizer_utils.py` only covers the `AutoTokenizer.from_pretrained` case, not the streaming detok path used by the server.

## Suggested next step

A server smoke test that runs `/v1/completions` after `mlx_lm.load()` validates would catch this. I can submit one as a follow-up PR.

## Side observation

`/v1/models` returns every model in `~/.cache/huggingface/hub`, not just the one the server was started with. On a machine with multiple MLX models cached this makes the response confusing (mine lists `Qwen/Qwen3-Embedding-8B` first even though the server is running V4-Flash). I haven't verified whether this behavior is from this PR or pre-existing in `mlx_lm.server`.

## Scoreboard for reference

I ran the three open V4 PRs through the same harness on the same machine + the same repo:

- #1189: ❌ `ValueError: Received 1423 parameters not in model` (weight keys don't match)
- **#1190 (this PR)**: ✅ load, ✅ generate, ❌ server
- #1192: ❌ same `ValueError` as #1189

So this PR is currently the only one that loads the public weights at all. That's a big deal worth landing, but the server bug should probably be resolved first since `mlx_lm.server` is how most people will actually run it.

austinbv:
Closing as there are better options

PR #14, opened 2025-03-15
feat: Add LLaDA Diffusion Model Support

cavit99:
#### Overview
This PR introduces support for diffusion-based language models (e.g., LLaDA) in mlx_lm, extending the framework beyond autoregressive generation to include diffusion paradigms for further research. The implementation adds a new generate_diffusion function in utils.py, respectfully updates the generation pipeline to handle both model types, and resolves issues that arose during integration, being very careful not to affect autoregressive path compatibility. The changes maintain alignment with the mlx-lm principles while ensuring robust initial functionality for diffusion models LLaDA and potentially others going forward.

#### Key Changes
1. **New `generate_diffusion` Function in `utils.py`**  
   - Added a new function to support diffusion-based text generation, tailored for models like LLaDA.
   - Key features:
     - Progressive token unmasking with configurable steps (`steps`), generation length (`gen_length`), and block sizes (`block_length`).
     - Supports Gumbel noise sampling (`noise_temp`) for stochasticity and classifier-free guidance (`cfg`) for controlled generation.
     - Implements semi-autoregressive block-wise diffusion with configurable unmasking strategies (`topk` or `random`).
     - Yields intermediate progress updates in verbose mode or final text otherwise, integrated with the `GenerationResponse` dataclass.
   - Optimizations:
     - Uses `mx.compile` for the sampling step to leverage MLX’s performance benefits.
     - Efficiently handles batch size of 1 (current limitation) with plans for future expansion.

2. **Updated `stream_generate` in `utils.py`**  
   - Modified to detect diffusion models via `model.args.model_type == "llada"` and delegate to `generate_diffusion`.
   - Preserves existing autoregressive paths (`generate_step` and `speculative_generate_step`) for non-diffusion models.
   - Added argument filtering to route diffusion-specific kwargs (e.g., `steps`, `gen_length`) to `generate_diffusion` and autoregressive kwargs (e.g., `max_tokens`, `sampler`) to their respective functions.
   - Ensures consistent streaming behavior across model types using the `GenerationResponse` interface.

3. **Extended CLI in `generate.py`**  
   - Added diffusion-specific arguments with sensible defaults:
     - `--steps` (default: 32): Number of diffusion steps.
     - `--gen-length` (default: 64): Length of the generated sequence.
     - `--noise-temp` (default: 0.0): Temperature for Gumbel noise sampling.
     - `--cfg` (default: 1.0): Classifier-free guidance scale.
     - `--block-length` (default: None): Size of semi-autoregressive blocks.
[...]

PR #1192, opened 2026-04-24
Add DeepSeek-v4 (Flash/Pro)

Blaizzy:
Note: Please install this transformers PR from source to avoid tokenizer bugs.

`pip install git+https://github.com/huggingface/transformers.git@refs/pull/45643/head`

Weights here:
https://huggingface.co/collections/mlx-community/deepseek-v4

image


Blaizzy:
You can now run it on a 256GB Mac by keeping a experts in 4bit!

We could do 5bit since it's much better than 4bit right now. I'm open to opinions @angeloskath 

image

Blaizzy:
It's faster now!
Screenshot 2026-04-24 at 21 54 13

mlx_lm/utils.py
```diff
@@ -279,6 +282,50 @@ def load_config(model_path: Path) -> dict:
     return config
 
 
+def _load_safetensors(path: str) -> dict:
```
pcuenca, line 285: Is this to work around F8_E8M0? I opened [this PR](https://github.com/ml-explore/mlx/pull/3448); if the direction is ok we can simplify this later.

Blaizzy: Yes, that would be awesome!
For now, this works well. It's just enough for the FP8 weights to be dequantized.

mlx_lm/models/deepseek_v4.py
```diff
@@ -0,0 +1,1084 @@
...
+            weight = weight.reshape(
+                (m + pad_m) // block_size,
+                block_size,
+                (n + pad_n) // block_size,
+                block_size,
+            )
+            weight = (weight * scale[:, None, :, None]).reshape(m + pad_m, n + pad_n)
+            return weight[:m, :n].astype(mx.bfloat16)
+
+        def dequant_fp4(weight: mx.array, scale: mx.array, block_size: int = 32):
```
pcuenca: I _think_ this is essentially mxfp4

mlx_lm/utils.py
pcuenca: Yes! As I said, not critical.

mlx_lm/models/deepseek_v4.py
Blaizzy: You are right!

Let me check 👀

machiabeli:
Hey @Blaizzy — just flagging some technical notes since we're both working on V4 support and PR #1189 landed ~10 hours earlier with significant overlap:

**Compressed attention mask direction (line 770-773):**
The mask padding for compressed KV rows uses `mx.ones`, but `create_attention_mask` returns negative values for blocked positions. Padding with ones would block attention to compressed rows rather than allow it. PR #1189 uses `mx.zeros` here.

**Sinkhorn normalization:**
The Python loop path (line 222-226) dispatches ~40 kernel launches per call (softmax + iters x sum + div). PR #1189 has a fused Metal kernel that does this in a single register-resident dispatch — benchmarked at 3.5-5.7x faster on micro, 1.83x end-to-end.

**sqrtsoftplus numerical stability:**
`nn.softplus(x)` can overflow for large scores. PR #1189 uses `mx.logaddexp(scores, zeros)` which is log-sum-exp stable.

Happy to coordinate if the maintainers want to consolidate into one PR. Our implementation has live generation validation at 21.86 tok/s on M3 Ultra (DeepSeek-V4-Flash-4bit, 160GB peak).

Blaizzy:
Hey @machiabeli, thanks! 

Yes, same person who left the earlier feedback, good to connect properly.

I've been poking at this in parallel and landed on something close to the source numerically with minimal changes, but there's definitely room to combine approaches. A PR from you on the compressed attention mask, Sinkhorn norm, and sqrt-softplus would be really welcome, happy to review and merge what works best.

Or I can cherry pick and add you as a co-author.

mlx_lm/utils.py
```diff
@@ -392,14 +392,23 @@ def load_model(
...
+                return False
+            if f"{p}.scales" not in weights:
+                return False
             # Handle custom per layer quantizations
             if p in config["quantization"]:
                 return config["quantization"][p]
-            if not hasattr(m, "to_quantized"):
-                return False
-            return f"{p}.scales" in weights
+            return True
```
Blaizzy: The goal here is to preserve mxfp4 expert quant since MLX supports it. So I made the quantize_config key in the config class default to that, and these changes help prequantized models load properly.

It can be done via predicate but couldn't find an elegant way of doing it.

Note: it doesn't affect any model.

Blaizzy: Alternative is to dequant -> requant similar to how we do with FP8.

trevorgordon981:
Ran this PR against `mlx-community/deepseek-ai-DeepSeek-V4-Flash-8bit` on an M3 Ultra. Fails at `load()` with the same signature as #1189:

```
ValueError: Received 1423 parameters not in model:
embed.{biases, scales},
head.{biases, scales},
model.layers.*.attn.wo_a.{0..7}.{biases, scales, weight},
model.layers.*.ffn.experts.w{1,2,3}.{biases, scales, weight},
...
```

The module imports cleanly (`DeepseekV4Model`, `DeepseekV4Block`, `HyperConnection`, `HyperHead`, `Indexer`, `Compressor`, `DeepseekV4Cache`, `LimitedSwiGLU`, etc. all present), but the parameter names the model declares don't match what the mlx-community 8-bit quant ships.

Specifically:

- `attn.wo_a` needs to be indexed `wo_a.{0..7}` (8 sub-projections) rather than a single module.
- `ffn.experts.w{1,2,3}` are fused across experts with `{biases, scales, weight}` triplets, not per-expert.
- `embed.{biases, scales}` and `head.{biases, scales}` are the I/O quant scales.

These are the `F8_E8M0/E4M3/E5M2` block-scale tensors the 8-bit quant uses. PR #1190 parses them directly in a custom `_load_weights_file()` and succeeds where both this PR and #1189 fail, even though #1190 is significantly smaller and covers less of the architecture.

Reproducer:

```python
from mlx_lm import load
model, tok = load("mlx-community/deepseek-ai-DeepSeek-V4-Flash-8bit")
# ValueError: Received 1423 parameters not in model: ...
```

Against the three open V4 PRs on identical hardware and the same repo:

- #1189: ❌ ValueError (same signature)
- #1190: ✅ loads, ✅ generates; ❌ server path produces garbage (separate issue)
- **#1192 (this PR)**: ❌ ValueError (same signature)

For this PR to load the public quantized weights, the simplest path is to adopt #1190's approach to F8 block-scale handling (either by cherry-picking the file loader or by renaming the model's parameters to match the on-disk layout).

Blaizzy:
You need to install this transformers @trevorgordon981 :)

https://github.com/huggingface/transformers/pulls

Blaizzy:
Let me upload my local copy to that repo

Blaizzy:
@trevorgordon981 the weights should be updated and working well now ✅

trevorgordon981:
Followup: tried this PR against your new quant `mlx-community/DeepSeek-V4-Flash-8bit` on an M3 Ultra. Fails before reaching `load()`'s shape check, in the tokenizer fallback path:

```
AttributeError: 'PreTrainedConfig' object has no attribute 'max_position_embeddings'
```

Triggered because `transformers` doesn't recognize `model_type: deepseek_v4` yet (the corresponding `transformers` PR isn't merged either, last I checked), so AutoConfig falls back to a generic `PreTrainedConfig`. Then somewhere downstream a max_position lookup fires.

Reproducer:

```bash
pip install --force-reinstall git+https://github.com/Blaizzy/mlx-lm@pc/add-deepseekv4flash-model
python -c "from mlx_lm import load; load('mlx-community/DeepSeek-V4-Flash-8bit')"
```

For completeness I ran the same against the other two open V4 PRs on the same machine, same repo:

- **#1189**: ❌ `Received 258 parameters not in model`. The per-layer `model.layers.<i>.attn_hc.{base,fn,scale}` and `model.layers.<i>.ffn_hc.{base,fn,scale}` keys aren't declared in #1189's model. #1189 puts hyper-connections globally as `hc_head_*`; your new quant has them per-layer split between attn and ffn.
- **#1190**: ❌ `Received 2394 parameters not in model`. Whole-structure mismatch; #1190's model class was built around the older `austinbv` quant at `mlx-community/deepseek-ai-DeepSeek-V4-Flash-8bit` and doesn't match the new layout.
- **#1192 (this)**: ❌ tokenizer fallback AttributeError above.

So the new quant has structural choices (per-layer hc, single `wo_a` instead of `wo_a.{0..7}`, `model.embed_tokens` instead of `embed`) that none of the three open PRs currently match end-to-end.

Two paths I can think of for unblocking this PR specifically:

1. Fix the tokenizer fallback to either bake `max_position_embeddings` from the config.json into the `PreTrainedConfig()` instance, or read it directly from the JSON before constructing the tokenizer.
2. Wait for transformers PR #45616 (Add DeepSeek V4) to land and pin a min `transformers` version that has it, then drop the fallback entirely.

I can send a small PR against your branch with whichever fix you prefer. (1) keeps the PR self-contained, (2) is cleaner long-term but adds an external pin. Let me know which would be most useful.

Blaizzy:
If you use the transformers PR I provided it show work fine as you will see in the video below.

https://github.com/huggingface/transformers/pull/45643

It's not merged yet, but will be soon so patching doesn't make sense.

https://x.com/prince_canuma/status/2047768990798184630?s=46

0xClandestine:
Have some optimizations here, numbers are finally looking good @Blaizzy

https://github.com/Blaizzy/mlx-lm/pull/13/

pcuenca:
> Or I can cherry pick and add you as a co-author.

👍

I'm not a maintainer of this repo, but just as an observer it'd be great if we can converge to a PR and credit as co-authors all of you that are involved in the effort.

rltakashige:
Hey there, just wondering if you'd had a chance to look at the PR I opened yesterday. I'm seeing a substantial (5x?) prompt processing speedup and memory improvement, so it would be great to see some of the additions used.

Looking forward to seeing DSV4 merged!

Blaizzy:
> I'm not a maintainer of this repo, but just as an observer it'd be great if we can converge to a PR and credit as co-authors all of you that are involved in the effort.

Yes, @eauchs sent a PR that added @machiabeli as a co-author stabilising the numericals.

And @0xClandestine sent me a PR that I will merge after a bit of alignment.

Blaizzy:
> Hey there, just wondering if you'd had a chance to look at the PR I opened yesterday. I'm seeing a substantial (5x?) prompt processing speedup and memory improvement, so it would be great to see some of the additions used.
> 
> Looking forward to seeing DSV4 merged!

A PR to this branch? Could you share the link?

rltakashige:
Ah, it's just a PR I opened onto mlx lm as I'd been working on it in parallel. #1195

Blaizzy:
Feel free to direct it to this branch and we can join forces 👌🏽

Tonoken3:
## Bug Report: Token dropping in Japanese (and likely CJK) output

### Environment
- Mac Studio M3 Ultra 512GB
- mlx 0.31.2, this PR branch (`pc/add-deepseekv4flash-model`)
- Model: `mlx-community/DeepSeek-V4-Flash-mxfp8` (also reproduced with `nvfp4`)

### Issue
Tokens are being **dropped at regular intervals** during generation, causing garbled output. This is especially visible in Japanese where each dropped token results in obviously broken text.

### Example

**Input:** `ハローハロー！バージョン2.5から4になったんだね。おめでとう！簡単に自己紹介して`

**Output:**
```
ハロー！おかげさでアップデ完了しました！
私は **DeepSeek**、深度求公司（DeepSe）が開発したAIアシ助手です。
```

| Expected | Got | Missing |
|---|---|---|
| おかげさ**ま**で | おかげさで | ま |
| アップデ**ート** | アップデ | ート |
| 深度求**索** | 深度求 | 索 |
| DeepSe**ek** | DeepSe | ek |
| バージョ**ン** | バージョ | ン |

### Verification
- Tokenizer encode/decode: ✅ correct
- `chunk.text` concat vs `tokenizer.decode(all_ids)`: ✅ match — not a detokenizer issue
- Same with both `mxfp8` and `nvfp4`
- `temp=0.0` (greedy) also affected

### Likely cause
NSA (Native Sparse Attention) implementation — possibly the Indexer or Compressor dropping positions from the KV cache during decode. The loss pattern appears roughly periodic.

Thank you for the great work on V4 support! 🙏

Blaizzy:
Thanks @Shinka-Man! 

I will reproduce and fix it in a few!

mlx_lm/models/deepseek_v4.py
```diff
@@ -0,0 +1,1781 @@
...
+
+@mx.compile
+def _hc_expand_op(
+    post: mx.array,
+    block_out: mx.array,
+    comb: mx.array,
+    residual: mx.array,
+) -> mx.array:
+    y = post[..., None] * block_out[:, :, None, :].astype(mx.float32)
+    y = y + mx.matmul(comb, residual.astype(mx.float32))
```
kernelpool:
```suggestion
    y = y + mx.matmul(comb.swapaxes(-1, -2), residual.astype(mx.float32))
```

Blaizzy: Applied, thanks @kernelpool ❤️!

Blaizzy:
@Shinka-Man fixed it [b78ccb1](https://github.com/ml-explore/mlx-lm/pull/1192/commits/b78ccb1906dbdcf8c49966669045f47931d95e55)

machiabeli:
Re converging — could probably land on #1189? Been in flight ~7-10h longer with the full skeleton, and you actually left feedback there before opening this one (happy to keep that thread going). pcuenca's already reviewing inline — HyperConnection got factored into its own module per his nudge, sinkhorn was already split out.

Fix for the nbytes thing @trevorgordon981 hit is queued up too.

Honestly don't mind which one lands though — clean merge with co-authors is the goal. If you'd rather keep this as trunk I'll just cherry-pick over. @eauchs's co-author PR is appreciated either way 🙏

0xClandestine:
Rebased, numbers still looking good.

**30.8 tok/s, 151.6GB** 

https://github.com/Blaizzy/mlx-lm/pull/13

Blaizzy:
Yap, will merge in a few @0xClandestine 👌🏽

Just had to fix some bugs before that.

Blaizzy:
Awesome, @machiabeli let's converge here! 

@eauchs already sent a PR ([c6a7828](https://github.com/ml-explore/mlx-lm/pull/1192/commits/c6a782867824237b918b0ca3b9be0513cfc6878f)) that was merged earlier, and I especifically asked him to ensure you were added as co-author. Therefore when this PR lands, your contributions will be preserved.

So if you got more ideas, feel free to send a PR to this branch I will review and merge ASAP.

ivanfioravanti:
Let me test this PR now! Keep you posted.

machiabeli:
Let's let a maintainer pick the merge target. anerjy's just-posted decode-cache repro on #1189 is the gating issue for either branch; fixing it today.

One concern worth surfacing before merge: across the ~35 commits on #1192, none carry `Co-authored-by` or `Reported-by` trailers — the only attribution lives inside @eauchs's separate PR #14. Specific cases:
- @Shinka-Man's detailed CJK repro shaped the fix at `b78ccb1`
- The sinkhorn ops at `c0d9222` overlap `1336c05` on #1189
- The HC rewrite at `ef8c95d` parallels `7fbde08` on #1189
- @trevorgordon981's and @anerjy's detailed bug repros on this thread are shaping fixes on both branches

Worth backfilling trailers — including on the 0xClandestine commits when they land here — before merge. Whichever branch becomes trunk should have clean attribution.

ivanfioravanti:
Amazing community joint effort on this one 💪

image

Biggest problem I'm seeing is that it uses a lot of memory when context grows larger than ~2500 tokens. 

image

Blaizzy:
Thanks for surfacing this, @machiabeli!

**On bug reporters as co-authors**: @Shinka-Man, @trevorgordon981, and @anerjy filed excellent, highly reproducible reports that absolutely made these fixes possible. That said, filing an issue isn’t typically grounds for `Co-authored-by` on the resulting commit. Happy to add `Reported-by` trailers retroactively across the relevant commits though if that's what they want. That’s a cleaner fit for the contribution they made and ensures they get visible credit.

**On the Sinkhorn and HC overlap**: these weren’t drawn from #1189 if they were I would have ensured to add `Co-authored-by`. The Sinkhorn ops follow the transformers PR and the original paper, and the HC rewrite was developed independently against the same reference. Any structural similarity comes from converging on shared upstream sources rather than from #1189 itself.

Let me know if `Reported-by` trailers would resolve things on your end and I’ll push the amendments.

Blaizzy:
> Biggest problem I'm seeing is that it uses a lot of memory when context grows larger than ~2500 tokens.

I will fix it in a few 👌🏽

Tonoken3:
@Blaizzy Thank you for the kind words and the lightning-fast fix! 🙏

A `Reported-by:` trailer would be wonderful — yes please. Honored to be a small part of this incredible community effort.

Watching this PR evolve over the past day has been inspiring. The bug → reproduce → fix → optimize cycle, all happening across timezones with multiple contributors converging — this is what open source at its best looks like.

Already running it in production on M3 Ultra 512GB with great results. Will pull the latest perf optimizations now and report back if anything surfaces. 🚀

Tonoken3:
## Performance update: +19% TPS on M3 Ultra 512GB

After pulling the latest optimizations (sinkhorn fusion, HC compute_weights fusion, attention hot-path caching, etc.):

|  | Before | After |
|---|---|---|
| Initial TPS | 22.4 tok/s | **26.7 tok/s** (+19%) |
| Long-context behavior | minor degradation | barely drops |

The reduced TPS decay over context is particularly noticeable — feels much more consistent across longer generations.

Stellar work @0xClandestine and @Blaizzy 🚀

Blaizzy:
> Biggest problem I'm seeing is that it uses a lot of memory when context grows larger than ~2500 tokens.

Fixed it @ivanfioravanti 
Screenshot 2026-04-26 at 11 16 20 AM

machiabeli:
Reported-by trailers would be great, thanks. Fair point on shared upstream sources for sinkhorn/HC — the timing raised questions but converging on the same paper is a reasonable explanation. Appreciate setting the convention going forward.

Blaizzy:
Done @machiabeli and @Shinka-Man ✅

I updated the [b78ccb1](https://github.com/ml-explore/mlx-lm/pull/1192/commits/b78ccb1906dbdcf8c49966669045f47931d95e55) to include reported-by @Shinka-Man.

Regarding @trevorgordon981 the fixes where in re-uploading the weights to HF and recommending the use of transformers PR. And as for @anerjy, I don't see any comment or contribution here.

Blaizzy:
>The reduced TPS decay over context is particularly noticeable — feels much more consistent across longer generations.
> Stellar work @0xClandestine and @Blaizzy 🚀

Kudos to @0xClandestine for optimizing the HC kernel even further. We spent most of the day yesterday bouncing ideas.

In our setups we see increase from 30 tok/s -> 32-34 tok/s.

mlx_lm/models/deepseek_v4.py
Blaizzy: Done, but decided to stick to dequant and requant because the alternative hardcodes experts to 4bit which will cause all model quants to be large

ivanfioravanti:
> > Biggest problem I'm seeing is that it uses a lot of memory when context grows larger than ~2500 tokens.
> 
> Fixed it @ivanfioravanti Screenshot 2026-04-26 at 11 16 20 AM

Amazing! Gonna test it now!

ivanfioravanti:
Some preliminary tests to keep in mind for future optimizations. They took forever to complete, 128K was SLOW as hell. 
9a1433c3_throughput
9a1433c3_batch

Blaizzy:
> Some preliminary tests to keep in mind for future optimizations. They took forever to complete, 128K was SLOW as hell. 

Thanks Ivan! 

Could you try reducing `prefill-step-size` to 500 or 1024 ? 

From the graph, the only thing that drops is prefill speed and the peak of ~400 tok/s would be slow for 128K at the default (2048 step size)

adurham:
## Three sanitize gaps when loading mlx-community 4-bit on multi-node TP

Tested this branch against `mlx-community/deepseek-ai-DeepSeek-V4-Flash-4bit` running 2-rank tensor-parallel across 2× Mac Studio M4 Max (128 GB / node). Hit three weight-naming mismatches that all manifest as silent `load_weights(strict=False)` drops, then either OOM-by-random-init or `KeyError` / `quantized_matmul` shape errors at first forward. All fixes are isolated to `Model.sanitize()`.

### 1. Pre-stacked routed experts (silent drop → OOM)

The 4-bit quant ships routed experts already stacked as a single tensor under `layers.{i}.ffn.experts.{w1,w2,w3}.{weight,scales,biases}` rather than per-expert unstacked (`experts.{0..255}.w1.weight`). The stacking loop only fires for the unstacked layout, so the keys pass straight through and `load_weights(strict=False)` silently drops them. The `SwitchLinear` is then left at its `mx.random.uniform(shape=(256, 2048, 4096))` fp32 init — ~8.6 GB per projection × 3 × 43 layers ≈ 1.1 TB, which gets materialized by per-layer `mx.eval` during sharding and OOMs on 128 GB nodes.

```python
for old, new in w_remap.items():
    # shared_experts first so its substring isn't partially rewritten
    nk = nk.replace(f".shared_experts.{old}.", f".shared_experts.{new}.")
    # NEW:
    nk = nk.replace(f".ffn.experts.{old}.", f".ffn.switch_mlp.{new}.")
```

### 2. Per-group `wo_a` (KeyError: 0)

The 4-bit checkpoint stores `wo_a` as `o_groups` separate sub-tensors (`wo_a.{0..7}.{weight,scales,biases}`), but `V4Attention.__init__` builds a single `nn.Linear` of shape `(o_groups * o_lora_rank, in)`. `tree_unflatten` interprets the numeric `.0./.1./…` suffixes as list indices, so `wo_a` arrives at `Module.update` as a `list[dict]` where the model expects a leaf with `weight/scales/biases`, and crashes with `KeyError: 0`.

```python
o_groups = self.args.o_groups
for layer_idx in range(n_layers):
    prefix = f"model.layers.{layer_idx}.attn.wo_a"
    for suffix in ("weight", "scales", "biases"):
        key0 = f"{prefix}.0.{suffix}"
        if key0 in weights:
            parts = [weights.pop(f"{prefix}.{g}.{suffix}")
                     for g in range(o_groups)]
            weights[f"{prefix}.{suffix}"] = mx.concatenate(parts, axis=0)
```

### 3. Quantized `embed` / `head` `.scales` and `.biases`

The top-level remap only renames `embed.weight` → `model.embed_tokens.weight` and `head.weight` → `lm_head.weight`. The 4-bit quant also ships `embed.{scales,biases}` and `head.{scales,biases}`. Without renaming those, `nn.quantize`'s `class_predicate` (`f"{p}.scales" in weights`) misses and `embed_tokens` / `lm_head` stay as plain `nn.Embedding` / `nn.Linear` while the `.weight` slot ends up holding the int4-packed tensor — shape/dtype mismatch at first forward.

```python
top_remap = {
    "embed.weight": "model.embed_tokens.weight",
    "embed.scales": "model.embed_tokens.scales",   # NEW
    "embed.biases": "model.embed_tokens.biases",   # NEW
    "norm.weight": "model.norm.weight",
    "head.weight": "lm_head.weight",
    "head.scales": "lm_head.scales",               # NEW
    "head.biases": "lm_head.biases",               # NEW
    ...
}
```

### Reference

Fork carrying all three fixes: `adurham/mlx-lm` commits 223604e, 6ee9898, 15de79d. Generates coherent tokens at `temperature=1.0, top_p=1.0` on 2-rank TP after these patches.

Side note: `_grouped_output_projection` reaches into `wo_a.weight` directly, which bypasses any `shard_linear` wrapper — sharding heads via `wq_b` all-to-sharded then makes the manual reshape see half the per-group input dim and crashes inside `mx.quantized_matmul`. That's downstream of this PR (a host-runtime concern), but if you'd accept it, calling `self.wo_a(...)` instead of accessing `.weight` directly would let multi-node TP shard heads cleanly. Happy to open a separate PR with these as one diff if useful.

Blaizzy:
Thanks @adurham! 

> Tested this branch against mlx-community/deepseek-ai-DeepSeek-V4-Flash-4bit running 2-rank tensor-parallel across 2× Mac Studio M4 Max (128 GB / node). Hit three weight-naming mismatches that all manifest as silent load_weights(strict=False)

That model wasn't quantised with the code in this PR.

Please use models that I created with this code.  

Link to collection:
https://huggingface.co/collections/mlx-community/deepseek-v4


The rest should be fixed by using the correct weights ✅

Let me know how it goes.

adurham:
thanks downloading and testing now

Blaizzy:
My pleasure! 

Looking forward to hearing your thoughts

adurham:
## Forward-pass: prefill score tensor blows past Mac's per-buffer limit at `L > ~1.2K`

Following up on the sanitize comment — running your re-quants (`mlx-community/DeepSeek-V4-Flash-4bit`) on 2-rank TP across 2× M4 Max, second-turn prefill (where the prior assistant turn becomes prompt) crashes with `[metal::malloc] Attempting to allocate 227,612,440,704 bytes which is greater than the maximum allowed buffer size of 86,586,540,032 bytes` from `mx.eval([c.state for c in prompt_cache])` (`mlx_lm/generate.py:442`). Turn 1 (~150 tok prompt) succeeds; turn 2 with 1904 prefill tokens dies before producing a single output token.

Root cause is in `V4Attention.__call__` for `compress_ratio == 4` layers (`deepseek_v4.py:1439-1448`):

```python
expanded = mx.broadcast_to(pooled[:, None, None, :, :],
                           (B, 1, L, pooled.shape[1], self.head_dim))
idx = topk[:, None, :, :, None]
pooled = mx.take_along_axis(expanded,
                            mx.broadcast_to(idx, idx.shape[:-1] + (self.head_dim,)),
                            axis=3).reshape(B, 1, -1, self.head_dim)
full_kv = mx.concatenate([full_kv, pooled], axis=2)
out = scaled_dot_product_attention(q, full_kv, full_kv, ...)
```

The `reshape(B, 1, -1, head_dim)` flattens the per-query top-`k` gather into a *dense* KV of length `L × k`, so SDPA materializes a `(B, n_heads, L, L × k + local)` score tensor. With `Indexer` returning `k = min(index_topk=512, T_pooled_compressor=L/4)`, that's cubic in `L` until `L ≥ 2048` (where `k` saturates at 512 and it becomes quadratic). At `L=1904` on this model: `(1, 64, 1904, ~933K)` × bf16 ≈ **227 GB single buffer**, which is exactly what Metal rejects (M-series single-allocation cap is ~80 GiB).

Per-layer score buffer at increasing `L` (bf16, `n_heads=64, head_dim=512, compress_ratio=4`):

| L | k = min(512, L/4) | score buf | fits 80 GiB? |
|---|---|---|---|
| 256  | 64  | 0.5 GB  | ✓ |
| 512  | 128 | 4.3 GB  | ✓ |
| 1024 | 256 | 34 GB   | ✓ |
| 1280 | 320 | 67 GB   | ✓ (tight) |
| 1500 | 375 | 108 GB  | ✗ |
| 1904 | 476 | **221 GB** | ✗ ← observed |
| 2048 | 512 | 275 GB  | ✗ |
| 4096 | 512 | 1.07 TB | ✗ |

This likely explains @ivanfioravanti's "128K was slow as hell" with 2048 step size — the `prefill-step-size 500 or 1024` workaround you suggested is the right knob, but the cliff isn't slowness, it's a hard fail on Mac at any single chunk above ~1.2K tokens. Default chunk size needs to be ≤ 1024 for DSv4 specifically, or the model OOMs on multi-turn chat the moment the prior turn's response gets included in the next prompt.

Real fix is to keep the gather sparse — pass `topk` indices into a query-grouped SDPA that scores each query against its own `k`-key window, never materializing the dense `L × k` expansion. Same pattern is in `deepseek_v32.py` and the other competing DSv4 PRs (rlt #1195, machiabeli #1189), so worth fixing in whichever lands. Happy to take a stab at this if it's useful.

Blaizzy:
Thanks @adurham! 

Please do try to take a stab and send a PR to this branch.

I will try my best to fix but I don't have 2 machines to test for it.

Blaizzy:
@adurham 

I managed to replicate in a single machine and fixed it 👌🏽

Sending a commit

Blaizzy:
Fixed,  could you try again @adurham?

adurham:
trying now

adurham:
# Decode regression on `ddeffe33` — pinpointed to `_hc_sinkhorn_collapse_kernel`

Pulled the latest pin for retest. Haven't gotten back to verifying the OOM-at-step-4096 repro yet (still capped at `EXO_PREFILL_STEP_SIZE=512` from the workaround), because decode collapsed pretty hard on the new file:

| `mlx_lm/models/deepseek_v4.py` rev | Prefill (15-tok warmup) | Decode (30-tok, 836 ctx) |
|---|---|---|
| `15de79d8` (snapshot before #26f49f5) | 65 tok/s | **34.8 tok/s** |
| `ddeffe33` (current PR head) | 1.9 tok/s | **0.07 tok/s** |
| `ddeffe33` + fused HC kernel disabled | 39 tok/s | **40.4 tok/s** |

Decode rate computed from per-token `ChunkGenerated` master-log timestamps; 3 independent runs (cold prefill, prefix-cache hit, fresh shorter prompt) all land within 1% of 40 tok/s, so the third row isn't a fluke.

The fix is a one-liner — force `_hc_sinkhorn_collapse_kernel = None` so `HyperConnection.collapse` takes the unfused split-sinkhorn + `_hc_collapse_op` path. The fused kernel from `26f49f5` is invoked twice per block (`attn_hc` + `ffn_hc`), so ~86 dispatches per decoded token on DSv4-Flash (43 layers). With it off, the rest of the squash lands fine — actually +16% over the prior baseline.

## Reproduction

Bisect branch: [`adurham/mlx-lm@dsv4-perf-bisect`](https://github.com/adurham/mlx-lm/tree/dsv4-perf-bisect) (= `ddeffe33` file content + the one-line kernel disable). On stock `ddeffe33` decode is reproducibly ≤ 0.1 tok/s.

Workload: `mlx-community/DeepSeek-V4-Flash-4bit`, 2-rank TP via RDMA across 2× M4 Max, ~836-token prompt + 30 generated tokens.

## MLX context — possibly a fork interaction

I'm on `adurham/mlx` (= upstream `e64e280d` + reverts of #3412 jaccl refactor and #3418 jaccl init bug, both comm-backend only, no GPU code touched). It's plausible `_hc_sinkhorn_collapse_kernel` only mis-behaves under that combination — happy to retest on stock upstream MLX if it'd help nail down whether this is fork-specific. Flagging it first since the symptom is so consistent on my side.

## Suggestion

Gate `_hc_sinkhorn_collapse_kernel` behind an opt-in env (default off) until the root cause is clear. cc @0xClandestine since you wrote/optimized this kernel — let me know what telemetry would help (per-call latency, kernel-source variants, profiling output, etc.).

Tonoken3:
## Token-dropping regression on `ddeffe33` — single-machine M3 Ultra

Different symptom from @adurham's TP regression but same general window. Pulled the head and saw severe quality degradation in Japanese output:

**Setup:** single Mac Studio M3 Ultra 512GB, stock MLX 0.31.2, no TP, `mlx-community/DeepSeek-V4-Flash-mxfp8`.

**Symptoms (head `ddeffe33`):**

```
Input:  "ハローハロー！バージョン2.5から4になったんだね。おめでとう！簡単に自己紹介して"

Output: "こんこん〜！どうもありあり〜！✨
         そうなんだよ、バージョン2.5から4にアップグップ☆って感じで、ちょびっと進進したかも〜
         えっとね、わたあまみみは、ちょぴっとだけおっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっっ..."
```

| | Expected | Got |
|---|---|---|
| こんにちは | こんこん | (drop + duplicate) |
| ありがとう | ありあり | (drop + duplicate) |
| アップグレード | アップグップ | (mid-word collapse) |
| 進化 | 進進 | (single-char repeat) |
| (long string) | normal | "おっっっっっ" infinite loop |

A second prompt also showed token dropping ("メモ帯域" instead of "メモリ帯域", "推速度" instead of "推論速度", "Metalフ" cut off, etc.).

**Bisect:**

| Commit | Quality |
|---|---|
| `910b120f` (perf-optimize-ds4 merged) | ✅ clean |
| `ddeffe33` (current head) | ❌ token drop + repetition |

**Workaround for now:** pinned to `910b120fa24cecb804b795b05183cbf0037f4ba6` which is rock solid (28-30 tok/s on mxfp8, no quality issues).

Could be related to one of:
- `8e8571a4` Fix DeepSeek V4 sparse pooled prefill memory
- `2591b51f` Refactor kv tensor reshaping in V4Attention
- `5a4aaa41` Refactor output projection
- `22da01a1` Remove redundant cache type check

Happy to bisect further commit-by-commit if useful — let me know which axis would be most valuable to narrow down. cc @Blaizzy

Blaizzy:
Thanks @Shinka-Man 

On it 👌🏽

Blaizzy:
FIxed @Shinka-Man, the culprit was the HyperConnection expand orientation change from `comb.T @ residual` to `comb @ residual` ✅

Blaizzy:
Thanks @adurham, looking into it 👌🏽

Tonoken3:
@Blaizzy Confirmed fixed on `acf650c9` ✅

**Same prompt that broke before:**

> ハローハロー！バージョン2.5から4になったんだね。おめでとう！簡単に自己紹介して

**Output now (mxfp8, M3 Ultra 512GB):**

> こんにちは！「バージョン2.5から4」という表現、何かのアプリやシステムのアップデートのことを指しているのかな？それとも、何か別の意味での「バージョンアップ」についてのメッセージでしょうか？
>
> もし私（アシスタント）に対しての「バージョンアップ」のお祝いだとすれば、ありがとうございます！...

Clean Japanese, no token drops, no repetition loops, perfect context understanding. **30.2 tok/s** (slightly faster than the pre-regression baseline).

That orientation flip — `comb.T @ residual` → `comb @ residual` — silently producing semantically-plausible-but-corrupted output is the kind of bug that's terrifying to debug. Nice catch and lightning-fast turnaround 🚀

Blaizzy:
My pleasure @Shinka-Man, thanks for catching it! ❤️

I added a test to avoid regressions

ivanfioravanti:
PR from @0xClandestine fixes the 4K context issue! Here tested up to 64K! 🔥

benchmark_chart

ivanfioravanti:
I've found a regression (not in the @0xClandestine PR): quantization is failing, 4, 8 and all combos. @Blaizzy 

```
python -m mlx_lm convert --hf-path deepseek-ai/DeepSeek-V4-Flash -q --q-bits 4 --mlx-path ~/DeepSeek-V4-Flash-4bit
```

[INFO] Using dtype: bfloat16
[INFO] Quantizing
[INFO] Quantized model with 4.506 bits per weight.
[1]    3279 killed     python -m mlx_lm convert --hf-path deepseek-ai/DeepSeek-V4-Flash -q --q-bits
/Users/ifioravanti/.local/share/uv/python/cpython-3.13.12-macos-aarch64-none/lib/python3.13/multiprocessing/resource_tracker.py:400: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown: {'/mp-p3fsavbc'} warnings.warn(

trevorgordon981:
Hey @Blaizzy — quick request from a downstream user.

I just ran the validation harness against this PR at HEAD `dd6b92f` on `mlx-community/DeepSeek-V4-Flash-8bit` (your mixed q4/q8 quant) on an M3 Ultra 512 GB. Results:

- Stages 02 (load), 03 (in-process gen), 04 (prefix cache, 58.45× cold→warm) all pass.
- Server-path probes via `mlx_lm.server` produce coherent output across non-streaming chat, multi-turn with system prompt, streaming SSE, and longer paragraph generation. No streaming-decode garbage, no decode-cache wrong-logits at S=1.
- Memory at peak: 144.8 GB RSS for the model on load; ~145 GB during serving with KV cache.

So the runtime path is healthy on this branch. **Would it be feasible to publish a true 8-bit weight-only quant** (no per-layer 4-bit override on the FFN/MoE expert weights) alongside the existing mixed quant? My intended use is as a single-tenant Alfred backend on this box — at q8 the model would be ~284 GB, which fits comfortably on 512 GB Mac Studio without the production-coexist constraint your mixed quant was sized for.

Also happy to self-quantize from `mlx-community/DeepSeek-V4-Flash-bf16` once a PR merges, just figured it was worth asking since you've been steadily iterating on the conversion path. No urgency. Thanks for all the work on this PR — the harness confidence on `dd6b92f` is the highest I've seen across the three V4 PRs.

Blaizzy:
Yes, I can do that. The reason why the 8bit has experts in 4bit is because the main model come with experts in MXFP4.
And at the time I was developing the PR it seems sensible to keep the experts as is.

> Also happy to self-quantize from mlx-community/DeepSeek-V4-Flash-bf16 once a PR merges, just figured it was worth asking since you've been steadily iterating on the conversion path. No urgency. Thanks for all the work on this PR — the harness confidence on dd6b92f is the highest I've seen across the three V4 PRs.


My pleasure! Let me merge that for you :)


Update: I don't see a PR to that rerpo

trevorgordon981:
Following up on #issuecomment-4329720377 with more specific findings.

I attempted to build a true 8-bit conversion myself by streaming through `deepseek-ai/DeepSeek-V4-Flash` (the 148 GB native FP8 release): read each shard, dequantize the I8 + F8_E8M0 (block-scale) and F8_E4M3 weights to bf16, re-quantize at affine q8 group_size=64, write MLX-format shards with the sanitize-equivalent name remapping (`embed.weight` → `model.embed_tokens.weight`, `hc_attn_X` → `attn_hc.X`, experts stacked into `switch_mlp.{gate,down,up}_proj`, etc.). The dequant + requantize math validates fine (round-trip error 0.74% rel). Pipeline runs in ~5 min for the full 46-shard input.

Hit an architectural wall on the routed-expert dimensions that I can't reconcile from public artifacts.

**Source (deepseek-ai/DeepSeek-V4-Flash) per-expert shapes:**

```
experts.X.w1.weight: I8 [2048, 2048]
experts.X.w2.weight: I8 [4096, 1024]
experts.X.w3.weight: I8 [2048, 2048]
```

**Your `mlx-community/DeepSeek-V4-Flash-8bit` `switch_mlp` (inferred from the mxfp4 packed/scales shapes):**

```
gate_proj per expert (bf16): [2048, 4096]
down_proj per expert (bf16): [4096, 2048]
up_proj   per expert (bf16): [2048, 4096]
```

Both repos' configs say `hidden_size=4096, moe_intermediate_size=2048, n_routed_experts=256, n_shared_experts=1` — they should produce the same per-expert shapes. But the source's last dims are exactly **half** of yours.

Param accounting confirms it: source's routed experts at (2048, 2048) sum to ~138B params (consistent with the 148 GB I8 file size), while your switch_mlp shapes imply ~280B in routed experts alone, which lines up with mxfp4's compressed 155 GB file size for a fuller param count.

The shared_experts block in source is sized correctly (`shared_experts.w1.weight: F8_E4M3 [2048, 4096]` matches your `shared_experts.gate_proj` (2048, 4096)). Only the routed experts disagree.

Two questions, in order of usefulness to me:

1. **How does your conversion process derive (2048, 4096) per-expert weights from the source's (2048, 2048)?** Concat of w1+w3 (which would only give you gate_proj, not also up_proj), TP-shard combination across pairs of source experts, or some V4-Flash-specific reshaping I'm not seeing in `mlx_lm/models/deepseek_v4.py`?
2. **Would a true 8-bit version (no `bits: 4` override on the FFN/MoE expert layers) be feasible to publish?** My use case is a 512 GB Mac Studio backend; ~280-300 GB q8 fits comfortably with KV cache headroom and would be a quality bump over the current mxfp4-on-FFN profile. Happy to run the same harness against it if you publish.

If (1) has a clean answer I can finish the converter myself for (2). If not, only you have the conversion path that produces the right shapes.

Either way thanks for the work — `dd6b92f` mixed quant has been serving Alfred backend reliably for several hours now.

Tonoken3:
## Performance update — Apple optimizations live ✨

After pulling the latest with @angeloskath's three commits (RoPE kernel native path, GLU cast simplification, original-checkpoint loader):

**Setup:** Mac Studio M3 Ultra 512GB, mxfp8, single-machine, no TP

| Prompt type | TTFT | TPS | Notes |
|---|---|---|---|
| Short JP (~50 tok output) | 2793 ms (cold) | **33.8 tok/s** | first run, model warmup |
| Long JP (~500 char essay) | 232 ms (cached) | **33.6 tok/s** | flat |
| Code (Python w/ docstrings) | 199 ms | **33.6 tok/s** | flat |

The thing that jumps out: **TPS is now context-flat** (33.8 vs 33.6 vs 33.6 across very different workloads). Previous builds had visible decay on longer generations. This is the signature of the cast/RoPE overhead going away.

**Cumulative trajectory on this hardware:**

| Build | TPS |
|---|---|
| Initial mxfp8 (with token-drop regression) | 22.4 |
| Sinkhorn / HC fusion (0xClandestine) | 26.7 |
| `acf650c9` orientation fix (Blaizzy) | 30.2 |
| **Apple kernel optimizations (angeloskath)** | **33.6–33.8** |

**~+50% from baseline in 4 days** of community iteration. Quality remains pristine (no token drops, no repetition loops, perfect Japanese & code).

Welcome to the PR @angeloskath — the kernel-native RoPE path is doing real work here on Apple Silicon. 🙌

ivanfioravanti:
Prompt longer than 2K leads to no response at all 
cat 4k.txt | mlx_lm.generate --model ~/DeepSeek-V4-Flash-5bit --max-tokens 200 --prompt -

I've used https://raw.githubusercontent.com/ivanfioravanti/llm_context_benchmarks/refs/heads/master/4k.txt

ivanfioravanti:
convert is not working properly, 5bit or 6 bit, use 4.349 bits per weight.

❯ mlx_lm.convert --hf-path deepseek-ai/DeepSeek-V4-Flash -q --q-bits 6 --mlx-path ~/DeepSeek-V4-Flash-6bit
[INFO] Loading
Fetching 67 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████| 67/67 [00:00<00:00, 8842.62it/s]
Download complete: : 0.00B [00:00, ?B/s]                                                                                             | 0/67 [00:00<?, ?it/s]
[transformers] You are using a model of type `deepseek_v4` to instantiate a model of type ``. This may be expected if you are loading a checkpoint that shares a subset of the architecture (e.g., loading a `sam2_video` checkpoint into `Sam2Model`), but is otherwise not supported and can yield errors. Please verify that the checkpoint is compatible with the model you are instantiating.
[transformers] PreTrainedConfig got `key=rope_scaling` in kwargs but hasn't set it as attribute. For RoPE standardization you need to set `self.rope_parameters` in model's config.
[INFO] Using dtype: bfloat16
[INFO] Quantizing
[INFO] Quantized model with 4.349 bits per weight.

It seems related to the fact that starting from FP8 model, mxp4 + mxfp8 is automatically applied independently from --q-bits.

doctorpangloss:
did you quantize these by upcasting fp8/fp4+scales to fp32? that is the only way, correct?

spicyneuron:
I just added tool parsing support for this branch here: https://github.com/Blaizzy/mlx-lm/pull/22 In theory, it should work with https://github.com/ml-explore/mlx-lm/pull/1189 too.

If you're using https://huggingface.co/collections/mlx-community/deepseek-v4 weights, you'll need to update chat_template.jinja to: https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash/blob/014a5cfe6d1349d3d1096b2f8c15faaaa11819d5/chat_template.jinja

Light usage via Claude Code has been smooth so far, but probably needs a bit more testing before I'd say this is solid.

spicyneuron:
Unrelated to tool calling, I've found that multi-round chats via the server sometimes suddenly loop / decohere. This is with the 8-bit quant, so unlikely to be weights related.

Current best theory is that extending the cache introduces subtle errors that compound until something eventually breaks. Never happens on the first turn or prefill from scratch.

AI summary of my debugging below, using `transformers@5.8.0` as a reference. I'm still building up my mental model of this architecture, but maybe someone else can validate:

## Original symptom and how to reproduce

DeepSeek V4 produces intermittent repetition loops in multi-turn chat through the server (which uses incremental prefill — each new turn extends the existing KV cache rather than re-prefilling from scratch). Onset is variable: sometimes turn 2, sometimes turn 5; sometimes just hundreds of tokens in, sometimes thousands. Loops disappear when the prompt cache is disabled (every turn is a full re-prefill).

## Root cause

The CSA outer compressor and the Lightning Indexer's inner compressor both run a two-series Ca/Cb overlap window scheme: each compressed entry is the softmax-gated combination of *the previous window's Ca slice* with *the current window's Cb slice*. The reference implementation in `transformers` persists the previous Ca slice across forward calls in the cache layer; mlx-lm's `_overlap_compress_kv` zeroed it on every call. Within one forward pass the in-call shift covers all but the very first window; across calls, every chunk-boundary window has its Ca slot zeroed instead of carrying the prior chunk's last-window slice.

Result: every CSA layer's pool entry that sits at a chunk boundary is wrong by ~3-8 in fp32. The error feeds upstream hidden state, which trips MoE top-k boundary crossings (gate margins are often <1e-2), which produce discontinuous FFN output jumps, which compound to the user-visible drift.

## Diagnostic scripts

https://gist.github.com/spicyneuron/53662e14947b143dcc839c58fe776b70#file-readme-md

For each script, point at any V4 checkpoint via `--model`.

## Smoking gun

Diagnostic 18: replay each layer's compressor on the same precomputed `wkv` / `wgate` projections in two ways — single-shot, and chunked — using the same `PoolingCache`, then diff the resulting compressed entries.

If the bug is present:
- HCA layers (`ratio=128`): zero diff everywhere.
- CSA layers (`ratio=4`): non-zero diff at *exactly* the chunk-boundary windows (entries 50 and 100 for 200-token chunks at ratio 4), zero everywhere else. Same pattern on the indexer's inner compressor.

Output looks like:

```
layer  ratio       which   entries         max        mean    bad       max_entry:first_bad      boundary
    2      4       outer       150    3.101562    0.009109      2             100:[50, 100]     [50, 100]
    2      4     indexer       150    7.724609    0.015006      2             100:[50, 100]     [50, 100]
    3    128       outer         4    0.000000    0.000000      0                      0:[]            []
    4      4       outer       150    4.898438    0.004288      2             100:[50, 100]     [50, 100]
    ...
```

The `bad` column counts pool entries above a fixed threshold and lists their indices. They land precisely at chunk boundaries, not anywhere else, which is the localization that makes this an unambiguous correctness gap rather than a numerical-stability concern.

spicyneuron:
Current branch also seems to trigger `RuntimeError: [metal::malloc] Resource limit (499000) exceeded.` crashes fairly consistently during long reasoning threads. I believe this fix addresses it: ~~https://github.com/Blaizzy/mlx-lm/pull/23~~ https://github.com/Blaizzy/mlx-lm/pull/24

Blaizzy:
Thanks! 

I left a comment @spicyneuron

ivaniguarans:
Hey @Blaizzy, thanks for your work!

I tested this branch on an M5 Max 128 GB, trying to requantize DeepSeek-V4-Flash from FP8 source to lower-precision affine (2-3 bpp) via `mlx_lm.convert`.

The current `sanitize` in `mlx_lm/models/deepseek_v4.py` (`Model.sanitize()`) correctly converts FP8 weights to MXFP8 quantized format for native inference:

```python
# deepseek_v4.py, Model.sanitize(), line 1082-1084 — works for inference
elif weight.dtype == mx.uint8:
    new_weights[k + "s"] = mx.repeat(mx.repeat(v, 4, -1), 128, 0)
    new_weights[wk] = weight.view(mx.uint32)
```

But there's no path in `Model.sanitize()` to dequantize FP8 weights back to BF16 for requantization. The naive approach — `weight.astype(mx.bfloat16)` — treats uint8 bytes as integers (0–255) instead of decoding them as FP8 E4M3 floats. This produces weight values ~20,000× too large (mean ~20000 vs expected ~0.016), causing immediate numerical explosion in layer 0.

The fix is to use `mx.dequantize` with MXFP8 mode, which properly decodes the FP8 values:

```python
elif weight.dtype == mx.uint8:
    # Dequantize FP8 E4M3 weights with E8M0 block scales to BF16
    m, n = weight.shape
    sm, sn = v.shape
    w32 = weight.view(mx.uint32)
    target_sn = w32.shape[-1] // 8  # MXFP8 group_size=32 → 8 uint32 per scale
    s_expanded = mx.repeat(mx.repeat(v, target_sn // sn, -1), m // sm, 0)
    new_weights[wk] = mx.dequantize(w32, s_expanded, group_size=32, bits=8, mode="mxfp8")
```

Same pattern applies to the FP4 expert branch (`Model.sanitize()`, line 1074-1081) — use `mx.dequantize(..., mode="mxfp4")` instead of manual reinterpretation:

```python
# FP4 experts → BF16
w32 = weight.view(mx.uint32)
new_weights[wk] = mx.dequantize(w32, v, group_size=32, bits=4, mode="mxfp4").astype(mx.bfloat16)
```

These dequantization paths would go behind a flag (e.g., triggered when `convert` is called with `-q` on an FP8 source) so the existing MXFP8 inference path stays untouched.

After this fix, `mlx_lm.convert -q --q-bits 2 --q-group-size 64` produces a working 85 GB model (2.56 bpp) from the 149 GB FP8 source. The model generates coherent output for short queries. (2-bit is too aggressive for sustained quality on this architecture, but the dequantization itself is correct — verified by checking weight statistics and layer-0 activation norms.)

spicyneuron:
I did several rounds of brute force Codex / Opus comparisons between this branch vs transformers and vLLM implementations. Results below.

I'm testing out [rough fixes](https://github.com/spicyneuron/mlx-lm/tree/fix-ds4) on my own server to see if it helps with the looping / long context issues. But in the meantime, anyone else want to validate?

---

## Cross-reference: `pr-1192` vs. vLLM and HF transformers

Comparison sources:
- vLLM [`vllm/model_executor/models/deepseek_v4.py`](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/deepseek_v4.py) and [`vllm/model_executor/layers/deepseek_compressor.py`](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/deepseek_compressor.py)
- HF [`modeling_deepseek_v4.py`](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py) and [`configuration_deepseek_v4.py`](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/configuration_deepseek_v4.py)

| Topic | `pr-1192` | vLLM | HF transformers | Note |
|---|---|---|---|---|
| **`LocalAttention` RoPE scaling** | Passes `None` for rope_scaling on local layers (no YaRN). | Single `get_rope` per layer; YaRN/mscale apply uniformly, only `rope_theta` swaps between `compress_rope_theta` and `rope_theta` ([L1009-1031](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/deepseek_v4.py#L1009-L1031)). | One shared `DeepseekV4RotaryEmbedding(config)`; sliding/local layers (`compressor=None`) read the same `main` rope-type that carries scaling ([L740-763](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L740-L763)). | `pr-1192` deviates — local layers should use `config.rope_scaling`. |
| **MoE gate dtype** | `logits = x @ self.weight.T` in activation dtype. | `router_logits_dtype=torch.float32` ([L853](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/deepseek_v4.py#L853)). | `F.linear(flat.float(), self.weight.float())` in both TopK and Hash routers ([L979](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L979), [L1010](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L1010)). | `pr-1192` deviates — gate must be fp32. |
| **Compressor softmax precision** | `_simple_compress_kv` ([L286-290](mlx_lm/models/deepseek_v4.py#L286-L290)) casts gate logits to fp32 before softmax, but omits `precise=True`; `_overlap_compress_kv` ([L293-310](mlx_lm/models/deepseek_v4.py#L293-L310)) casts `ape` down to `gate.dtype` (bf16/fp16 from `wgate`), so its softmax input stays low precision even though it passes `precise=True`. | Compressor state is fp32 throughout (`assert self.dtype == torch.float32`, [L143](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/deepseek_compressor.py#L143)). | HCA, Indexer, and CSA all do `softmax(..., dtype=torch.float32)` ([L419](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L419), [L520](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L520), [L629-L630](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L629-L630)). | `pr-1192`'s two compress paths trade off precision in opposite directions; neither matches HF/vLLM's uniformly fp32 compressor softmax. |
| **CSA cross-call overlap state** | `_overlap_compress_kv` prepends a zero/`-inf` row at every call — overlap is reset each chunk, breaking long-context generation. | Compressor stores prior `kv_a`/`gate_a` slice in cache state and pulls it forward across calls. | Same — overlap is part of CSA layer state, not zeroed per-call. | `pr-1192` is functionally incorrect for chunked prefill or multi-call generation; the new `DeepseekV4PoolingCache` carries `overlap_kv`/`overlap_gate` across calls. |
| **Pooling cache storage growth** | Each `update_and_fetch` does `mx.concatenate([self.pooled, px], axis=1)`, accumulating lazy-graph nodes until Metal's per-command-buffer resource limit aborts long runs. | N/A (PyTorch/CUDA — no equivalent constraint). | N/A (PyTorch — no equivalent constraint). | MLX-specific. `pr-1192` should step-allocate a backing buffer (e.g. 256-token chunks) and expose `pooled` as a logical-size view into it. |

spicyneuron:
Superseding my previous comment on model looping, I've isolated a [script that reliably reproduces the issue](https://gist.github.com/spicyneuron/deeb395b2f2f7d5c97a6ab2590b72cb5) at ~4000 tokens on my Mac Studio M3.

Could someone else try it?

```sh
# just in case
uv cache prune

# serve this branch
uvx --from git+https://github.com/Blaizzy/mlx-lm/mlx-lm@pc/add-deepseekv4flash-model \
  mlx_lm.server \
  --max-tokens 32000 \
  --model mlx-community/DeepSeek-V4-Flash-8bit

# and then
uv run repro_v4_loop_minimal.py --base-url http://localhost:8080 --min-p 0.0

# still reproduces, but later
uv run repro_v4_loop_minimal.py --base-url http://localhost:8080 --min-p 0.05
```

Runs sometimes fail with the `RuntimeError: [metal::malloc] Resource limit (499000)` I mentioned above but simply restarting the server and trying again works to bypass the issue (eventually).

---

EDIT: Tentative fix in [this branch](https://github.com/spicyneuron/mlx-lm/tree/fix-ds4-cache-reuse), but I'd be more confident in it if others also can reproduce the bug, and then see it go away when using:

```sh
uvx --from git+https://github.com/spicyneuron/mlx-lm/mlx-lm@fix-ds4-cache-reuse \
  mlx_lm.server \
  --max-tokens 32000 \
  --model mlx-community/DeepSeek-V4-Flash-8bit
```

kidroca:
@spicyneuron 
> Superseding my previous comment on model looping, I've isolated a [script that reliably reproduces the issue](https://gist.github.com/spicyneuron/deeb395b2f2f7d5c97a6ab2590b72cb5) at ~4000 tokens on my Mac Studio M3.
> 
> Could someone else try it?

I gave a try on a machine I have access to, I pretty much confirm same behavior, though I couldn't recreate with min-p 0.05 - I gave it just a few tries though 

---

# mlx-lm DeepSeek V4 Flash repro

I ran the repro on a Mac Studio M3 Ultra and can confirm that the baseline branch reproduces the loop, while `fix-ds4-cache-reuse` fixes the case I could reproduce.

One command difference: the original command in the comment did not work for me because the git URL had an extra `/mlx-lm` path segment:

```text
git+https://github.com/Blaizzy/mlx-lm/mlx-lm@pc/add-deepseekv4flash-model
```

I used this corrected baseline command instead:

```sh
uvx --from git+https://github.com/Blaizzy/mlx-lm@pc/add-deepseekv4flash-model \
  mlx_lm.server \
  --max-tokens 32000 \
  --model mlx-community/DeepSeek-V4-Flash-8bit
```

And this corrected fix-branch command:

```sh
uvx --from git+https://github.com/spicyneuron/mlx-lm@fix-ds4-cache-reuse \
  mlx_lm.server \
  --max-tokens 32000 \
  --model mlx-community/DeepSeek-V4-Flash-8bit
```

Baseline, `--min-p 0.0`:

```text
seed=0 min_p=0.0
turn=1 tokens=4000 status=LOOP
loop='consequential—was consequential—was consequential—was consequential—was consequential—was consequential—was consequential—was consequential—was' x671
```

Fix branch, `--min-p 0.0`:

```text
seed=0 min_p=0.0
turn=1 tokens=4000 status=ok
turn=2 tokens=4000 status=ok
no loop reproduced
```

For `--min-p 0.05`, I hit the known Metal resource-limit/server-disconnect path on one attempt:

```text
http.client.RemoteDisconnected: Remote end closed connection without response
```

After restarting the server and retrying, the run completed without a loop (2 times):

```text
seed=0 min_p=0.05
turn=1 tokens=4000 status=ok
turn=2 tokens=4000 status=ok
no loop reproduced
```

Summary from my run:

- baseline reproduces the loop with `min_p=0.0`
- `fix-ds4-cache-reuse` removes that reproduced loop
- `min_p=0.05` did not reproduce the loop for me after retrying past the Metal resource-limit failure

bojiang:
Heads up: the mlx-community DeepSeek-V4-Flash conversions (the `mlx-community/deepseek-ai-DeepSeek-V4-Flash-{2,3,4,6,8}bit` and `-fp16` repos, all uploaded 2026-04-24/25) ship weights in a format that the current `sanitize` doesn't recognize. Two specific gaps observed on a 4-bit load:

1. `embed.{scales,biases}` / `head.{scales,biases}` aren't in `top_remap` — only the `.weight` keys are, so the quantized scales/biases never get renamed to `model.embed_tokens.*` / `lm_head.*`.
2. Experts are pre-stacked under `ffn.experts.w{1,2,3}.{weight,scales,biases}` instead of per-expert `ffn.experts.E.w{1,2,3}.{...}`. The existing expert-stacking block expects the per-expert form and so silently leaves the pre-stacked tensors un-renamed.
3. `wo_a` is split into `o_groups` separate Linear modules under `attn.wo_a.{0..o_groups-1}.{weight,scales,biases}` instead of a single 2D tensor. The current `wo_a` reshape only handles a 2D-tensor input.

A minimal patch on top of \`pc/add-deepseekv4flash-model\` that covers all three:

\`\`\`python
# extend the existing top_remap dict with:
\"embed.scales\": \"model.embed_tokens.scales\",
\"embed.biases\": \"model.embed_tokens.biases\",
\"head.scales\":  \"lm_head.scales\",
\"head.biases\":  \"lm_head.biases\",

# before the existing per-expert stacking loop, rename pre-stacked experts:
for layer_idx in range(n_layers):
    prefix = f\"model.layers.{layer_idx}.ffn.experts\"
    for src, dst in ((\"w1\", \"gate_proj\"), (\"w2\", \"down_proj\"), (\"w3\", \"up_proj\")):
        for suffix in (\"weight\", \"scales\", \"biases\"):
            old_k = f\"{prefix}.{src}.{suffix}\"
            new_k = f\"model.layers.{layer_idx}.ffn.switch_mlp.{dst}.{suffix}\"
            if old_k in weights and new_k not in weights:
                weights[new_k] = weights.pop(old_k)

# before the existing 2D->3D wo_a reshape, stack split wo_a shards:
for layer_idx in range(n_layers):
    prefix = f\"model.layers.{layer_idx}.attn.wo_a\"
    for suffix in (\"weight\", \"scales\", \"biases\"):
        key0 = f\"{prefix}.0.{suffix}\"
        if key0 in weights:
            stacked = [weights.pop(f\"{prefix}.{g}.{suffix}\") for g in range(self.args.o_groups)]
            weights[f\"{prefix}.{suffix}\"] = mx.stack(stacked, axis=0)
\`\`\`

Verified with both 3-bit (124 GB on disk) and 4-bit (149 GB on disk) on a M2 Ultra (192 GB unified). Both load (peak 124.6 GB and 160.2 GB respectively) and generate at ~33–35 tok/s. Separately, neither of those repos ships a \`chat_template\`; copying the one from \`inferencerlabs/DeepSeek-V4-Flash-MLX-2.8bit-EXP\` fixes runaway USER/ASSISTANT continuation in \`mlx_lm.server\`.

Happy to open a PR against your branch if useful.

snagnever:
Several people in this thread are hitting `[metal::malloc] Resource limit (499000) exceeded` during long or looping generations on Apple Silicon (and the command queue then wedges until a restart). I tracked down the cause and have a fix.

**Root cause:** an unbounded per‑decode‑step leak of *live* Metal buffers in the DeepSeek‑V4 attention caches. `PoolingCache` (concatenate‑grow) and `RotatingKVCache` (sliced assignment), plus their `Batch*` variants, build per‑step lazy graphs that are never detached — the cache holds the head of the chain, so every prior step's intermediate array and its backing Metal buffer stays resident. `resource_limit` (499000) is a count of **live resident buffers**, not bytes, so the count climbs ~one per layer per step and hits the cap at **~11.3K tokens regardless of prompt length** (≈ 499000 / 43 layers). Only ~2–3 GB is actually leaked at that point — it's the count that's exhausted, which is why no `set_memory_limit`/`set_wired_limit` knob helps.

- **Root cause + reproducer + full diagnostic path:** ml-explore/mlx-lm#1332
- **Fix** (one hunk in `DeepseekV4Model.__call__`, targets this branch): Blaizzy/mlx-lm#25

The fix materializes the per‑layer cache state once per forward pass (`mx.eval`), which detaches the chains and keeps the live‑buffer count bounded. Verified on M4 Max 128 GB / mlx 0.31.2: forced 20K‑token generation clean to 19,989 (was OOM at 11,314), 31.3 tok/s (no regression), and a 300‑request knowledge‑bench soak on a single long‑lived server with 0 OOMs.

@Blaizzy — happy to adjust the placement if you'd prefer to fold it into this PR directly.

spicyneuron:
So the code review burden on this model is clearly huge. But I can attest that DS 4 Flash runs incredibly well on Apple hardware, and it feels like a waste that this isn't available to more people.

Open question to @angeloskath @Blaizzy @kernelpool @pcuenca and others — how can we move this forward?

On my end, I was able to patch my way to a [stable implementation](https://github.com/spicyneuron/mlx-lm/tree/_ds4) based on this PR, using [transformers](https://github.com/huggingface/transformers/tree/main/src/transformers/models/deepseek_v4) and [vllm](https://github.com/vllm-project/vllm/tree/main/vllm/models/deepseek_v4) as references. No `Resource limit (499000)`, no catastrophic looping, and tool calling works beautifully.

Code likely suffers from LLM-isms, but I'll put it forward as a rough implementation that others might be able to streamline or borrow from. I had Codex squash the history into a cleaner commit story, so each fix should be reviewable in isolation: https://github.com/Blaizzy/mlx-lm/compare/pc/add-deepseekv4flash-model...spicyneuron:mlx-lm:pr-1192-ds4-clean

hehua2008:
Are there any talents still interested in this PR?
Looking forward to your amazing work! Thanks～
I miss @awni ‘s hard work😭

siddharthjthapa:
@hehua2008 Have you tried @antirez ‘s https://github.com/antirez/ds4 ?

hehua2008:
Thank you for reminding. I will try it.
But have you found that the Apple MLX team has almost ignored this project? This project has not released an updated version for 2 months! The great @Awni has left Apple MLX to join Anthropic. Now only volunteers are submitting PR. It shows that Apple's investment in AI is getting less and less, which is different from what they advertise. This is what makes me very disappointed……

spicyneuron:
Local LLMs are still a niche use case, and AI-coded PRs have added a lot of noise to many open source projects. So it's not surprising things are slow.

But without a core maintainer to champion things, complex PRs are effectively dead ends. So at this point, best we can do is show interest, try to make code review easy, and run our own forks if we need something faster.

Blaizzy:
Closing this PR for now

PR #1193, opened 2026-04-24
feat: add DeepSeek-V4 support

eauchs:
Full DeepSeek-V4 support (MLA, Hyper-Connections, sparse attention, FP8/FP4). Includes numerical stability fix (logaddexp) and compressed attention mask fix (mx.zeros). Co-authored with @Blaizzy and @machiabeli.


Blaizzy:
We already have an open PR with all the changes merged :)

In #1192

Issue #1175, opened 2026-04-21
mtp_generate_step: logits processors see stale prev_tokens on draft calls


Thump604:
## Problem

`mtp_generate_step` in `generate.py` calls `_process_and_sample(prev_tokens, mtp_logits)` at line 769 (`_step_mtp`), but `prev_tokens` does not contain the backbone's just-sampled token. A stateful logits processor that reads `tokens[-1]` to determine state (e.g. Outlines' Guide FSM for constrained JSON decoding) advances by the wrong token and enters an invalid state.

Serial `generate_step` does not have this issue because the sampled token is fed back as `input_tokens` on the next `_step` call, so `tokens[-1]` always reflects the last generated token.

## Reproduction

Script below runs a logits processor through both `generate_step` and `mtp_generate_step`, logging what `tokens[-1]` is on each call:

```python
import mlx.core as mx
from mlx_lm import load
from mlx_lm.generate import generate_step, mtp_generate_step

MODEL = "path/to/mtp-model"  # any model with mtp_forward

def make_tracker(tokenizer):
    tok = tokenizer._tokenizer
    log = []
    n = [0]
    def processor(tokens, logits):
        n[0] += 1
        tlist = tokens.tolist() if hasattr(tokens, "tolist") else list(tokens)
        if isinstance(tlist, int): tlist = [tlist]
        log.append((n[0], tlist[-1], len(tlist), tok.decode([tlist[-1]])))
        return logits
    return processor, log

model, tokenizer = load(MODEL)
tok = tokenizer._tokenizer

# --- Serial ---
proc, log = make_tracker(tokenizer)
gen = list(zip(range(6), generate_step(model=model,
    prompt=mx.array(tokenizer.encode("Hello world")),
    max_tokens=6, logits_processors=[proc])))
print("Serial tokens[-1] per call:", [(c, t, d) for c, t, _, d in log])

# --- MTP ---
proc2, log2 = make_tracker(tokenizer)
gen2 = list(zip(range(6), mtp_generate_step(model=model,
    prompt=mx.array(tokenizer.encode("Hello world")),
    max_tokens=6, logits_processors=[proc2])))
print("MTP tokens[-1] per call:", [(c, t, d) for c, t, _, d in log2])
```

**Output on Qwen3.5-27B-VLM-MTP-8bit:**

Serial (7 calls, 6 tokens -- correct):
```
call 1: tokens[-1]=1814(' world'), len=1  (prompt, first call)
[...]

PR #1196, opened 2026-04-25
feat: add thinking budget with early-stopping prompt injection

Thump604:
## Summary

Adds `thinking_budget` support using Qwen's recommended early-stopping prompt injection. When thinking tokens exceed the budget, the processor forces a transition prompt token-by-token through the model's forward pass, so the model reads the injection through full attention and transitions to content from a coherent hidden state.

- `make_thinking_budget_processor()` in `sample_utils.py` -- stateful logits processor
- `build_early_stop_tokens()` -- tokenizes transition message + think-end tokens
- `make_logits_processors()` -- wired with `thinking_budget` params
- Server accepts `thinking_budget` and `thinking_budget_message` from chat completion requests
- `initial_thinking` detects when the prompt already opened a `<think>` block

## Why this approach

The existing approaches in vLLM and llama.cpp force `</think>` via logit manipulation, but the model's hidden state never processes the transition -- it generates content from an incomplete thought state. Qwen's official recommendation for open-source inference is to inject a natural-language prompt that the model reads through attention, producing a coherent transition.

Each forced token goes through `generate_step`'s normal `_step()` call, which runs the full model forward pass and updates the KV cache. The model "reads" the injection the same way it reads any other token.

## Validated on

**Qwen3-1.7B-MLX-8bit (budget=50):**
- Thinking: model reasons about Rayleigh scattering
- Injection: "I need to provide my answer now based on my reasoning so far."
- Content: 2051 chars, coherent physics explanation with headers and sections

**Qwen3.5-2B-OptiQ-4bit (budget=50, MoE architecture):**
- Thinking: 249 chars of analysis and planning
- Content: 1908 chars, structured explanation with numbered sections

Both produce well-structured content after the injection. The `initial_thinking=True` fix handles chat templates that place `<think>` in the prompt.

## API

```python
# Direct usage
from mlx_lm.sample_utils import make_thinking_budget_processor, build_early_stop_tokens

early_stop = build_early_stop_tokens(tokenizer, tokenizer.think_end_tokens)
processor = make_thinking_budget_processor(
    think_start_tokens=tokenizer.think_start_tokens,
    think_end_tokens=tokenizer.think_end_tokens,
    budget=100,
    early_stop_tokens=early_stop,
    initial_thinking=True,
)
result = generate(model, tokenizer, prompt, logits_processors=[processor])

# Via server (chat completion request body)
{"thinking_budget": 100, "thinking_budget_message": "Summarize and answer now."}
```

## Test plan

- 11 unit tests covering passthrough, budget enforcement, full injection sequence, post-injection reset, natural end-of-thinking, and assembly function
- Integration on Qwen3 1.7B and Qwen3.5 2B
- `python -m pytest tests/test_thinking_budget.py -v`

Related: #914 (partial -- adds budget control), #1175 (MTP stale prev_tokens affects all stateful logits processors; fix suggested on #990)

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -365,3 +365,111 @@ def make_frequency_penalty(penalty: float, context_size: int = 20):
         return logits
 
     return frequency_penalty_processor
+
+
+def make_thinking_budget_processor(
+    think_start_tokens: tuple,
+    think_end_tokens: tuple,
+    budget: int,
+    early_stop_tokens: mx.array,
+) -> Callable[[mx.array, mx.array], mx.array]:
+    """
+    Make a logits processor that enforces a thinking token budget.
+
+    Tracks the thinking state by watching for ``think_start_tokens`` /
+    ``think_end_tokens`` in the generated sequence.  When the number of
+    thinking tokens exceeds ``budget``, the processor injects
+    ``early_stop_tokens`` one token at a time by forcing each token's logit
+    to ``0.0`` and setting all other logits to ``-inf``.  After the full
+    injection sequence has been forced the processor returns to passthrough
+    mode.
+
+    Args:
+        think_start_tokens (tuple): Token IDs that mark the start of a
+            thinking block (e.g. the tokenization of ``<think>``).
+        think_end_tokens (tuple): Token IDs that mark the end of a thinking
+            block (e.g. the tokenization of ``</think>``).
+        budget (int): Maximum number of thinking tokens allowed before
+            early-stop injection begins.  Must be >= 0.
+        early_stop_tokens (mx.array): 1-D array of token IDs to inject when
+            the budget is exceeded.
+
+    Returns:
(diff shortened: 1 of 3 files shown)


AirRunner:
`make_thinking_budget_processor` was previously receiving stale context on MTP draft calls. `prev_tokens` is now fixed in #990.

Also, as already said there: the verify pass advances the processor state for `bonus_tok` even on rejection, and there's no way to roll it back without processors implementing an explicit `save_state()`/`restore_state()` protocol. The thinking budget counter would drift on rejections.

Thump604:
Noted on the drift. The thinking budget counter would over-count on rejections since the verify pass advances it for bonus_tok even when rejected. With the current interface, the only workaround is to make the processor stateless (recompute count from prev_tokens on each call) or accept the drift as bounded noise.

A save_state/restore_state protocol on the processor interface would fix this properly for all stateful processors. Will track separately.

PR #1198, opened 2026-04-25
fix: prevent double-shift of norm weights for converted VLM checkpoints

### LINKED ISSUE #1197: load() produces incoherent output for VLM model checkpoints used as text-only
## Problem

`mlx_lm.load()` loads VLM model checkpoints (e.g. `Qwen3.6-35B-A3B-VLM-MTP-8bit`, `Qwen3.5-122B-A10B-VLM-MTP-5bit`) and returns a model object, but generation produces incoherent/garbage output. The model class resolves correctly (`qwen3_5_moe.Model`) and no error is raised, but the text is nonsensical.

These models are distributed as VLM checkpoints with vision components. When `load()` initializes the text model from these checkpoints, something in the weight mapping or initialization produces a broken text model.

## Reproduction

```python

model, tokenizer = load("Qwen3.6-35B-A3B-VLM-MTP-8bit")  # local path
print(type(model).__name__)  # "Model" (qwen3_5_moe.Model) -- correct class

result = generate(model, tokenizer, "What is 2+2?", max_tokens=50)
print(result)
# Garbage: "5inaatroEB No... ignavierte... 酒店预订..."
```

Non-VLM checkpoints of the same architecture work correctly:

```python
model, tokenizer = load("mlx-community/Qwen3.5-2B-OptiQ-4bit")
result = generate(model, tokenizer, "What is 2+2?", max_tokens=50)
# Correct output
```

## Impact

Most Qwen 3.5/3.6 models distributed for Apple Silicon are VLM variants. Users who download these models and try to use them for text generation through `load()` + `generate()` get silent failures (garbage output, no error).

## Expected behavior

Either:
1. `load()` correctly initializes the text model from VLM checkpoints (skipping vision weights), or
2. `load()` raises a clear error when a VLM checkpoint is detected, directing users to mlx-vlm

## Environment

- mlx-lm: upstream main (ed1fca4)
- Models tested: Qwen3.6-35B-A3B-VLM-MTP-8bit, Qwen3.5-122B-A10B-VLM-MTP-5bit
- Working: Qwen3-1.7B-MLX-8bit, Qwen3.5-2B-OptiQ-4bit (both non-VLM)

Thump604:
## Problem

`load()` produces garbage output for VLM model checkpoints (e.g. Qwen3.6-35B-A3B-VLM-MTP-8bit) while non-VLM checkpoints of the same architecture work correctly.

## Root cause

`TextModel.sanitize()` adds +1.0 to RMS norm weights when MTP or unsanitized conv1d weights are present (converting zero-centered gamma from HuggingFace format to standard RMS norm). VLM checkpoints contain MTP weight keys, which triggers the shift. But MLX-converted checkpoints already have the shift applied during conversion. The double-shift makes norm weights ~2.0 instead of ~1.0, corrupting every norm layer in the model.

## Fix

Sample the first norm weight before applying the shift. If the mean is near 1.0, the weights are already in standard format. Only apply +1.0 when the weights are actually zero-centered (mean near 0.0).

## Validated

**Before fix:** Qwen3.6-35B-A3B-VLM-MTP-8bit produces incoherent garbage
**After fix:** Same model produces coherent output

```
Thinking: 559 chars of structured reasoning
Content: "2+2 equals 4."
```

Non-VLM models (Qwen3-1.7B, Qwen3.5-2B) continue to work correctly (norm weights are not affected because they have no MTP keys).

Refs #1197

### DIFF
--- a/mlx_lm/models/qwen3_5.py
+++ b/mlx_lm/models/qwen3_5.py
@@ -322,6 +322,17 @@ class TextModel(nn.Module):
             ".q_norm.weight",
             ".k_norm.weight",
         )
+
+        # Check if norm weights are already shifted.  Zero-centered weights
+        # (from HuggingFace checkpoints) have mean near 0; converted MLX
+        # checkpoints have already been shifted to mean near 1.  Applying
+        # +1.0 twice corrupts every norm layer in the model.
+        if should_shift_norm_weights:
+            for k, v in weights.items():
+                if v.ndim == 1 and any(k.endswith(sfx) for sfx in norm_keys):
+                    should_shift_norm_weights = mx.abs(mx.mean(v)).item() < 0.5
+                    break
+
         for k, v in weights.items():
             if "conv1d.weight" in k and v.shape[-1] != 1:
                 weights[k] = v.moveaxis(2, 1)


mlx_lm/models/qwen3_5.py
```diff
@@ -322,6 +322,17 @@ def sanitize(self, weights):
             ".q_norm.weight",
             ".k_norm.weight",
         )
+
+        # Check if norm weights are already shifted.  Zero-centered weights
+        # (from HuggingFace checkpoints) have mean near 0; converted MLX
+        # checkpoints have already been shifted to mean near 1.  Applying
+        # +1.0 twice corrupts every norm layer in the model.
+        if should_shift_norm_weights:
```
nastya236, line 330:
```suggestion
        if has_unsanitized_conv1d:
```

mlx_lm/models/qwen3_5.py
nastya236, line 308: We probably need to remove this check, because having `mtp` as a key is no longer a reliable identification for a shift.

nastya236 (changes requested):
Thanks for your contribution. I left couple suggestions.

PhilipJohnBasile:
The fresh #1197 tensor-level reproduction confirms the same failure mode. I implemented the requested review direction in draft #1623: MTP presence is no longer used as a format signal; only the raw Conv1D layout triggers the RMSNorm `+1` conversion. The regression covers dense and MoE raw conversion plus already-converted checkpoints that still contain MTP tensors. I opened it as a draft and linked this PR so the work can be consolidated rather than silently duplicated.

Thump604:
Superseded by merged #1623, which implements the maintainer-requested structural predicate and adds stronger dense/MoE regression coverage. This older heuristic-based version should be closed.

PR #1200, opened 2026-04-25
Add dense `qwen3_5` support for learned quantization

iamwavecut:
## Summary

This PR adds learned quantization support for dense `qwen3_5` text models and hardens DWQ failure handling.

Changes:
- Enable AWQ traversal for dense `qwen3_5` hybrid layers:
  - full attention blocks with `self_attn.{q,k,v,o}_proj`
  - linear-attention blocks with `linear_attn.in_proj_*`
  - dense MLP projections
- Avoid the qwen3_5 inference-only gated-delta CustomKernel during dynamic quantization sensitivity estimation.
- Add `dynamic_quant` calibration controls for small local runs:
  - `--num-samples`
  - `--sequence-length`
- Add DWQ KL loss implementation selection:
  - `--kl-loss-impl auto|metal|mlx`
- Add DWQ target metadata validation before training.
- Add DWQ shape and finite-value checks so NaN/Inf losses or weights fail before saving a corrupted checkpoint.
- Add unit tests and an opt-in slow qwen3_5 quantization smoke test.

This intentionally focuses on dense `qwen3_5` models. MoE support is left out of scope.

## Motivation

Plain conversion and GPTQ already work for qwen3_5 models, but other learned quantization paths had gaps:

- AWQ rejected `qwen3_5` as unsupported.
- Dynamic quantization could hit `[Primitive::vjp] Not implemented for CustomKernel` on qwen3_5 linear-attention paths.
- DWQ could fail on Apple Silicon devices where the Metal KL-loss kernel configuration is not usable, and unstable runs could continue far enough to save invalid checkpoints.

## Testing

```bash
uv run python -m unittest tests.test_losses tests.test_quantization -v
```

```bash
uv run python -m unittest discover tests -v
```

```bash
uv run pre-commit run --files \
  mlx_lm/LEARNED_QUANTS.md \
  mlx_lm/quant/awq.py \
  mlx_lm/quant/dwq.py \
  mlx_lm/quant/dynamic_quant.py \
  mlx_lm/tuner/losses.py \
  tests/test_losses.py \
  tests/test_quantization.py \
  tests/test_qwen3_5_quantization_slow.py
```

Opt-in slow smoke test on Qwen/Qwen3.5-0.8B:

```bash
RUN_SLOW_QWEN35_QUANT_TESTS=1 \
QWEN35_QUANT_TEST_MODEL=/path/to/Qwen3.5-0.8B \
uv run python -m unittest tests.test_qwen3_5_quantization_slow -v
```

Local smoke checks also verified that AWQ, dynamic quantization, and DWQ start and save loadable artifacts for the downstream finetuned model LakoMoor/QClaw-4B. GPTQ was checked to reach Hessian collection and quantization startup as an existing baseline.

### DIFF
--- a/mlx_lm/LEARNED_QUANTS.md
+++ b/mlx_lm/LEARNED_QUANTS.md
@@ -40,6 +40,9 @@ Some important options, along with their default values are:
 - `--num-samples 1024`: Number of samples to use. Using more samples can lead to
   better results but takes longer.
 - `--batch-size 8`: Use a smaller batch size to reduce the memory footprint.
+- `--kl-loss-impl auto`: Select the KL loss implementation. `auto` uses the
+  Metal kernel when available and falls back to the MLX implementation when the
+  kernel cannot be used. Use `mlx` to force the fallback implementation.
 
 For a full list of options run:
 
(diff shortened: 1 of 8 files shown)


zcbenz:

PR #1211, opened 2026-04-27
Add Hy3

kernelpool:
Model: https://huggingface.co/tencent/Hy3-preview

NOTE: This has been updated to support https://huggingface.co/tencent/Hy3

```
mlx_lm.generate --model /Volumes/WD_EXTRA/models/catalyst/Hy3-preview-4bit --prompt "Who is Albert Einstein?" -m 1024
You are using a model of type `hy_v3` to instantiate a model of type ``. This may be expected if you are loading a checkpoint that shares a subset of the architecture (e.g., loading a `sam2_video` checkpoint into `Sam2Model`), but is otherwise not supported and can yield errors. Please verify that the checkpoint is compatible with the model you are instantiating.
==========
**Albert Einstein** (1879–1955) was a German-born theoretical physicist who is widely regarded as one of the most influential scientists of all time. He is best known for developing the **theory of relativity**, which revolutionized our understanding of space, time, gravity, and the universe.

Here are the key highlights of his life and work:

### 1. Major Scientific Contributions
*   **Special Relativity (1905):** He proposed that the laws of physics are the same for all non-accelerating observers and that the speed of light in a vacuum is constant, regardless of the motion of the light source. This led to the famous equation:
    *   **E = mc²** (Energy equals mass times the speed of light squared). This showed that mass and energy are interchangeable.
*   **General Relativity (1915):** This is his masterpiece. He described gravity not as a force, but as a curvature of **spacetime** caused by mass. This theory predicted phenomena like black holes and the bending of light by gravity (which was confirmed during a solar eclipse in 1919, making him world-famous).
*   **The Photoelectric Effect (1905):** He explained how light can behave as particles (photons) as well as waves, a foundational concept in quantum mechanics. This work won him the **Nobel Prize in Physics in 1921**.

### 2. The "Miracle Year" (1905)
While working as a patent clerk in Bern, Switzerland, Einstein published four groundbreaking papers in a single year (*Annus Mirabilis*). These papers explained the photoelectric effect, Brownian motion (proving atoms exist), special relativity, and mass-energy equivalence. This remains one of the most extraordinary achievements in the history of science.

### 3. Life and Legacy
*   **Emigration:** Born in Ulm, Germany, Einstein renounced his German citizenship due to the rise of the Nazi party. In 1933, he moved to the United States and joined the **Institute for Advanced Study** in Princeton, New Jersey, where he lived until his death.
*   **The Atomic Bomb:** Although a committed pacifist, in 1939, Einstein wrote a famous letter to President Franklin D. Roosevelt warning that Germany might be developing nuclear weapons. This letter prompted the U.S. to start the Manhattan Project. Einstein himself did not work on the bomb, and he later deeply regretted his role in initiating the research.
*   **Personality:** He is famous for his wild white hair, his love for the violin, and his playful sense of humor. He is often synonymous with the word "genius."

### 4. Cultural Impact
Einstein's name has become a synonym for high intelligence. His face is one of the most recognizable in the world, and his theories remain the foundation of modern physics, used in everything from GPS satellites to our understanding of the Big Bang.
==========
Prompt: 20 tokens, 12.246 tokens-per-sec
Generation: 609 tokens, 31.005 tokens-per-sec
Peak memory: 166.255 GB
```

### DIFF
--- a/mlx_lm/models/hy_v3.py
+++ b/mlx_lm/models/hy_v3.py
@@ -1,0 +1,382 @@
+# Copyright © 2026 Apple Inc.
+
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+from mlx.nn.layers.distributed import shard_inplace, shard_linear, sum_gradients
+
+from .activations import swiglu
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .pipeline import PipelineMixin
+from .rope_utils import initialize_rope
+from .switch_layers import SwitchGLU
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str
+    vocab_size: int
+    hidden_size: int
+    intermediate_size: int
+    num_hidden_layers: int
+    num_attention_heads: int
+    num_key_value_heads: int
+    head_dim: int
+    num_experts: int
+    num_experts_per_tok: int
+    num_shared_experts: int
+    expert_hidden_dim: int
+    first_k_dense_replace: int
+    rms_norm_eps: float
+    rope_parameters: Dict[str, Any]
+    router_scaling_factor: float = 1.0
+    qk_norm: bool = True
+    route_norm: bool = True
+    moe_router_use_sigmoid: bool = True
+    moe_router_enable_expert_bias: bool = True
+    tie_word_embeddings: bool = False
+    num_nextn_predict_layers: int = 0
+    max_position_embeddings: int = 262144
+
+
+class Attention(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+
+        dim = args.hidden_size
+        self.n_heads = args.num_attention_heads
+        self.n_kv_heads = args.num_key_value_heads
+        self.head_dim = args.head_dim
+        self.scale = self.head_dim**-0.5
+
+        self.q_proj = nn.Linear(dim, self.n_heads * self.head_dim, bias=False)
+        self.k_proj = nn.Linear(dim, self.n_kv_heads * self.head_dim, bias=False)
+        self.v_proj = nn.Linear(dim, self.n_kv_heads * self.head_dim, bias=False)
+        self.o_proj = nn.Linear(self.n_heads * self.head_dim, dim, bias=False)
+
+        self.use_qk_norm = args.qk_norm
+        if self.use_qk_norm:
+            self.q_norm = nn.RMSNorm(self.head_dim, eps=args.rms_norm_eps)
+            self.k_norm = nn.RMSNorm(self.head_dim, eps=args.rms_norm_eps)
+
+        self.rope = initialize_rope(
+            dims=self.head_dim,
+            base=args.rope_parameters["rope_theta"],
+            traditional=False,
+            scaling_config=args.rope_parameters,
+            max_position_embeddings=args.max_position_embeddings,
+        )
+
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Any] = None,
+    ) -> mx.array:
+        B, L, _ = x.shape
+
+        queries = self.q_proj(x).reshape(B, L, self.n_heads, self.head_dim)
+        keys = self.k_proj(x).reshape(B, L, self.n_kv_heads, self.head_dim)
+        values = self.v_proj(x).reshape(B, L, self.n_kv_heads, self.head_dim)
+
+        if self.use_qk_norm:
+            queries = self.q_norm(queries)
+            keys = self.k_norm(keys)
+
+        queries = queries.transpose(0, 2, 1, 3)
+        keys = keys.transpose(0, 2, 1, 3)
+        values = values.transpose(0, 2, 1, 3)
+
+        offset = cache.offset if cache is not None else 0
+        queries = self.rope(queries, offset=offset)
+        keys = self.rope(keys, offset=offset)
+        if cache is not None:
+            keys, values = cache.update_and_fetch(keys, values)
+
+        output = scaled_dot_product_attention(
+            queries, keys, values, cache=cache, scale=self.scale, mask=mask
+        )
+        output = output.transpose(0, 2, 1, 3).reshape(B, L, -1)
+        return self.o_proj(output)
+
+
+class MLP(nn.Module):
+    def __init__(self, hidden_size: int, intermediate_size: int):
+        super().__init__()
+        self.gate_proj = nn.Linear(hidden_size, intermediate_size, bias=False)
+        self.up_proj = nn.Linear(hidden_size, intermediate_size, bias=False)
+        self.down_proj = nn.Linear(intermediate_size, hidden_size, bias=False)
+
+    def __call__(self, x):
+        return self.down_proj(swiglu(self.gate_proj(x), self.up_proj(x)))
+
+
+@mx.compile
+def expert_select(
+    gates,
+    expert_bias,
+    top_k,
+    routed_scaling_factor,
+    norm_topk_prob,
+):
+    in_type = gates.dtype
+    scores = mx.sigmoid(gates.astype(mx.float32))
+    orig_scores = scores
+    scores = scores + expert_bias
+
+    inds = mx.argpartition(scores, kth=-top_k, axis=-1)[..., -top_k:]
+    scores = mx.take_along_axis(orig_scores, inds, axis=-1)
+    if top_k > 1 and norm_topk_prob:
+        scores = scores / scores.sum(axis=-1, keepdims=True)
+    scores = scores * routed_scaling_factor
+
+    return inds, scores.astype(in_type)
+
+
+class MoEGate(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.top_k = args.num_experts_per_tok
+        self.norm_topk_prob = args.route_norm
+        self.routed_scaling_factor = args.router_scaling_factor
+        self.gate = nn.Linear(args.hidden_size, args.num_experts, bias=False)
+        self.expert_bias = mx.zeros((args.num_experts,))
+
+    def __call__(self, x):
+        return expert_select(
+            self.gate(x),
+            self.expert_bias,
+            self.top_k,
+            self.routed_scaling_factor,
+            self.norm_topk_prob,
+        )
+
+
+class MoE(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.num_experts_per_tok = args.num_experts_per_tok
+        self.switch_mlp = SwitchGLU(
+            args.hidden_size,
+            args.expert_hidden_dim,
+            args.num_experts,
+        )
+        self.router = MoEGate(args)
+        if args.num_shared_experts > 0:
+            self.shared_mlp = MLP(
+                args.hidden_size,
+                args.expert_hidden_dim * args.num_shared_experts,
+            )
+        else:
+            self.shared_mlp = None
+
+        self.sharding_group = None
+
+    def __call__(self, x):
+        if self.sharding_group is not None:
+            x = sum_gradients(self.sharding_group)(x)
+
+        inds, scores = self.router(x)
+        y = self.switch_mlp(x, inds)
+        y = (y * scores[..., None]).sum(axis=-2)
+        if self.shared_mlp is not None:
+            y = y + self.shared_mlp(x)
+
+        if self.sharding_group is not None:
+            y = mx.distributed.all_sum(y, group=self.sharding_group)
+
+        return y
+
+
+class DecoderLayer(nn.Module):
+    def __init__(self, args: ModelArgs, layer_idx: int):
+        super().__init__()
+        self.self_attn = Attention(args)
+        if layer_idx < args.first_k_dense_replace:
+            self.mlp = MLP(args.hidden_size, args.intermediate_size)
+        else:
+            self.mlp = MoE(args)
+        self.input_layernorm = nn.RMSNorm(args.hidden_size, eps=args.rms_norm_eps)
+        self.post_attention_layernorm = nn.RMSNorm(
+            args.hidden_size, eps=args.rms_norm_eps
+        )
+
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[Any] = None,
+    ) -> mx.array:
+        r = self.self_attn(self.input_layernorm(x), mask, cache)
+        h = x + r
+        r = self.mlp(self.post_attention_layernorm(h))
+        return h + r
+
+
+class HYV3Model(PipelineMixin, nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        self.vocab_size = args.vocab_size
+        self.embed_tokens = nn.Embedding(args.vocab_size, args.hidden_size)
+        self.layers = [DecoderLayer(args, idx) for idx in range(args.num_hidden_layers)]
+        self.norm = nn.RMSNorm(args.hidden_size, eps=args.rms_norm_eps)
+
+    def __call__(
+        self,
+        x: mx.array,
+        cache: Optional[Any] = None,
+    ) -> mx.array:
+        h = self.embed_tokens(x)
+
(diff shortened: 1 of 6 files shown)


ivanfioravanti:
Great job @kernelpool 🔥

kernelpool:
Forgot to commit the tool parser. Fixed :)

jonpspri:
Love it -- carrying it on my running branch.

kernelpool:
Added support for https://huggingface.co/tencent/Hy3

avlp12:
Thanks for this port @kernelpool — we used it as the base for quantized builds of the final `tencent/Hy3` release ([avlp12/Hy3-Alis-MLX-Dynamic](https://huggingface.co/avlp12/Hy3-Alis-MLX-Dynamic), 2.4–8.5 bpw) and validated it fairly hard along the way. Sharing the findings:

**Correctness — verified.** Cross-framework logit parity vs the `transformers` 5.12.1 reference (`HYV3ForCausalLM`) on tiny-random weights: top-1 agreement 100%, max |Δlogit| ≤ 6.4e-3 (fp32), across EN/KO/code prompts. Checkpoint key mapping (`mlp.expert_bias` remap, per-expert stacking, MTP-layer strip) matches the final release's safetensors index exactly. At full scale: coherent generation on all four quant tiers, and 128K-token needle retrieval passing on the 6.5-bpw build.

**One suggested change: run the router matmul in fp32.** The reference computes the gate as `F.linear(hidden_states.float(), self.weight.float())` (`modeling_hy_v3.py`, `HYV3TopKRouter.forward`) — the matmul itself is fp32, not just the sigmoid. With 192 experts under a sigmoid router whose e-score correction bias deliberately shifts selection margins for load balance, top-8 selection margins can be tight; computing the gate in fp32 removes bf16 matmul rounding from the discrete expert-selection path and matches the reference's routing numerics:

```python
# MoEGate.__call__
return expert_select(
    self.gate(x.astype(mx.float32)),  # was: self.gate(x)
    ...
)
```

Cost is negligible (a 4096×192 matmul per MoE layer). Commit with this applied on top of your branch: [avlp12/mlx-lm@14f7837](https://github.com/avlp12/mlx-lm/commit/14f7837) — feel free to cherry-pick.

One more data point for reviewers: keeping `mlp.router.gate` **unquantized** (bf16) rather than the default 8-bit totals only ~124 MB in bf16 across the 79 MoE layers (vs ~66 MB at the default 8-bit, i.e. ~58 MB extra) and removes quantization noise from the discrete routing path entirely — worth considering for the model's default `quant_predicate`.

PhilipJohnBasile:
Validation report from running this PR's architecture on real hardware, plus one small robustness suggestion.

**Setup:** Hy3 295B MoE (mixed-quant MLX build, `ox-ox/Hy3-295B-Instruct-w2q3exp-AProjQ8-SExpQ8-OutQ8-MTP-mlx`, ~105 GB) on an M5 Max / 128 GB. Model code from this PR (via the `eauchs/mlx-lm` `hy_v3-mtp` branch, which builds on it).

**What worked:**
- Load + generate with an AR-only view (MTP sidecar omitted, `num_nextn_predict_layers=0`): coherent output, clean EOS stops, peak memory 112.3 GB, ~1.4 tok/s cold decode.
- `mlx_lm.server` OpenAI-compatible endpoint: agent evals (coding, tool-call formatting, strict JSON, code repair, planning) all passing at temp 0.
- LoRA fine-tune (rank 8, 8 layers via `mlx_lm lora`) + fuse into a standalone directory: works, with one template note below.

**Two field notes for anyone chat-fine-tuning or using converted checkpoints:**
1. The chat template only appends `eos_token` to the final assistant turn when `is_training` is set. The tuner's `ChatDataset` applies the template with no kwargs, so SFT on chat data silently trains the model to end assistant turns without EOS — we saw token-0 (`!`) spam after otherwise-correct answers until training against a template view that defaults `is_training=true`.
2. MLX-converted checkpoints can carry the MTP weights as a separate `model-mtp.safetensors` with tensors named `mtp.*` (rather than `model.layers.80.*`). `sanitize` currently drops only the `model.layers.{n+i}.*` naming, so such checkpoints fail strict loading. One-line fix: also drop keys starting with `"mtp."` when MTP is unused.

Happy to share the JSON receipts for any of the numbers above.

PhilipJohnBasile:
A downstream data point in support of landing this: I've built and published **five Hy3 MLX artifacts** that depend on `hy_v3` reaching mainline mlx-lm — two AR daily-drivers (a fused instruct build and a soul-preserving 25%-REAP-pruned + healed build) and two MTP-equipped variants — all currently running only on a pinned fork (`eauchs/mlx-lm@hy_v3-mtp`).

Merging this unblocks **two separate runtimes at once**:

- **LM Studio** (and anything on mainline mlx-lm) can then load the AR models directly — right now they fail with `model_type: hy_v3` unrecognized.
- **MTPLX** (native MTP speculative decoding on Apple Silicon) pins mlx-lm for its `hy_v3` MTP runtime backend. Its own `mtplx inspect` already recognizes the HY V3 MTP markers and reports only `recognized-backend-pending` — the backend PR is staged, waiting on this.

Happy to run load / exactness / decode-bench fixtures on an **M5 Max 128 GB** and post receipts if that's useful for review. Public models + full build receipts: https://github.com/PhilipJohnBasile/hy3-demolition-mlx

Thanks for putting this together, @kernelpool.

PhilipJohnBasile:
Independent validation, for what it's worth: I tested this branch against a real-world Hy3 derivative — not a synthetic test case, an actual model we built and shipped (`philipjohnbasile/hy3-demolition-mlx-reap25-v1`: 25%-pruned via REAP, arXiv:2510.13999, then LoRA-healed and fused).

```
loaded in 14.5s
generate("What is the capital of France? Answer in one word.") -> "Paris"
peak memory: 86.7 GB
```

That peak memory figure is a solid confirmation signal: it matches, to the decimal, what we measured independently on our own `hy_v3` fork for the same checkpoint (documented on the model card). Correct output, correct memory profile, on a real production artifact this PR never saw. Nice work — this looks solid to me.

(Context: we've been building a family of Hy3 derivatives on MLX and have our own fork-based `hy_v3` support; landing this in mainline is one of the things we're rooting for, since it'd unlock LM Studio + drop the need for a pinned fork.)

PhilipJohnBasile:
One more data point, unrelated to the earlier validation: tested the tool-call parser against a checkpoint of ours (`reap25`) that has a known tag-formatting bug on our end — it occasionally omits `<tool_sep:opensource>` (documented, not a claim against this PR).

Worth noting for robustness: on that malformed input, `parse_tool_call` doesn't fail loudly or fall back cleanly — it silently returns a garbled `name` field (the whole preamble text) while `arguments` still parses correctly. Might be worth a defensive check (e.g. require `<tool_sep>` present before treating the match as a valid call), since other imperfect fine-tunes could hit the same shape. Not blocking, just flagging in case it's useful.

raullenchai:
Adding a data point in support of @PhilipJohnBasile's tool-parser observation above — we hit the exact same malformed-close-tag shape while integration-testing this PR from [rapid-mlx](https://github.com/raullenchai/Rapid-MLX) (MLX-native OpenAI-compatible server for Apple Silicon), and I think it may be worth surfacing as more than a checkpoint-side artifact.

**Repro:**
```bash
mlx_lm.server --model pipenetwork/Hy3-REAP50-MLX-4bit --port 9115
# ... then any BFCL v4 simple_python prompt, e.g.
curl -s http://127.0.0.1:9115/v1/chat/completions -H 'content-type: application/json' -d '{
  "model": "pipenetwork/Hy3-REAP50-MLX-4bit",
  "messages": [{"role":"user","content":"Calculate the area of a circle with radius 5"}],
  "tools": [{"type":"function","function":{"name":"calculate_circle_area","parameters":{"type":"object","properties":{"radius":{"type":"number"}},"required":["radius"]}}}]
}'
```

**Observed in the raw stream (before `parse_tool_call`):**
```
<tool_call:opensource>calculate_circle_area</arg_value:opensource>
```

**Expected (per `chat_template.jinja`'s tool-call example and the `<tool_sep:opensource>` / `<arg_key>` / `<arg_value>` sentinel scheme):**
```
<tool_call:opensource>calculate_circle_area<tool_sep:opensource>{"radius":5}<arg_value:opensource>
```

The model emits the correct function name but wraps it with the wrong closing sentinel (`</arg_value:opensource>` before any `<tool_sep>` / `<arg_key>` appears), so `hy_v3.parse_tool_call` populates `tool_calls[0].function.name` with the whole garbled preamble — strict OpenAI-schema clients (BFCL) then score fn-name-fail even though the intent is trivially recoverable.

**Not a REAP artifact:** 10/10 BFCL v4 simple_python prompts reproduce identically on both `pipenetwork/Hy3-REAP50-MLX-4bit` and `pipenetwork/Hy3-REAP75-MLX-4bit` — two independently pruned checkpoints, same failure shape. That, plus @PhilipJohnBasile hitting the same shape on his `reap25`, points at either the chat-template tool block or the base Hy3 checkpoint rather than any one pruning run.

**Env:** `mlx==0.32.0`, `kernelpool/mlx-lm@add-hy3-preview` head `b7635e9c`, macOS / Apple Silicon.

Two possible avenues:
1. **Root-cause (preferable):** inspect the `tools` example rendered by `chat_template.jinja` — if the exemplar uses `</arg_value:opensource>` as a general closer, that would explain training the model into this pattern.
2. **Defensive parser fix:** treat a match with no `<tool_sep>` as malformed and either fall back to name-extraction or raise, rather than silently pass through — this matches @PhilipJohnBasile's suggestion above.

Happy to help repro or bisect if useful. Reported by @raullenchai on behalf of rapid-mlx integration testing.

raullenchai:
Second data point from the same integration-testing pass (rapid-mlx, MLX-native OpenAI-compatible server) — this one is orthogonal to the tool-call issue and lands squarely on the chat template's default `reasoning_effort=no_think` branch.

**Repro:**
```bash
# 1) chat-template default (reasoning_effort=no_think per tokenizer_config.json)
mlx_lm.generate --model pipenetwork/Hy3-REAP50-MLX-4bit \
  --prompt "What is the capital of France? Answer in one word."
# → "France"    ← wrong

# 2) bypass chat template entirely
mlx_lm.generate --model pipenetwork/Hy3-REAP50-MLX-4bit --ignore-chat-template \
  --prompt "What is the capital of France? Answer in one word."
# → "Paris"     ← correct

# 3) chat template + explicit reasoning_effort=low
mlx_lm.generate --model pipenetwork/Hy3-REAP50-MLX-4bit \
  --chat-template-config '{"reasoning_effort":"low"}' \
  --prompt "What is the capital of France? Answer in one word."
# → "Paris"     ← correct
```

The model clearly has the fact; only the `no_think` branch of `chat_template.jinja` nudges it off factual recall on the same input. This matters because `no_think` is the **default** per `tokenizer_config.json`, so any client hitting `mlx_lm.server`'s OpenAI endpoint without setting `reasoning_effort` — which is most of them — gets the broken variant on trivial factual queries.

For contrast, @PhilipJohnBasile's `mlx_lm.generate` run of the same "capital of France" prompt on `hy3-demolition-mlx-reap25-v1` returned `"Paris"` — consistent with that path either bypassing the chat template or running under a non-`no_think` default in his config; it doesn't contradict this repro, it isolates the trigger to the chat-template `no_think` branch.

**Env:** `mlx==0.32.0`, `kernelpool/mlx-lm@add-hy3-preview` head `b7635e9c`, macOS / Apple Silicon. Reproduces on both `pipenetwork/Hy3-REAP50-MLX-4bit` and `pipenetwork/Hy3-REAP75-MLX-4bit`, so it's not a REAP-training artifact — the shared surface is the Hy3 chat template.

Two possible fixes:
1. **Server-side (small, unblocks users today):** change `mlx_lm.server`'s default `reasoning_effort` to `low`.
2. **Chat-template-side (deeper):** audit the `no_think` branch of `chat_template.jinja` — likely the reasoning-mode preamble it prepends on user turns is what's biasing single-token answers away from the factual continuation. Would need input from @kernelpool / upstream Hy3 on whether the preamble is meant to render on plain user turns at all.

Happy to help repro or bisect. Reported by @raullenchai on behalf of rapid-mlx integration testing.

pierre427:
Small conversion-side suggestion surfaced while testing #1485 against existing repacks: when `sanitize` strips the MTP weights at conversion, also zero out `num_nextn_predict_layers` in the saved config. Published quants made from this branch (e.g. `pipenetwork/Hy3-REAP50-MLX-4bit`) carry `num_nextn_predict_layers: 1` with no layer-80 tensors, which breaks any MTP-aware loader that trusts the config (#1485 in its current form fails strict load on them).

PhilipJohnBasile:
**Negative result: on Hy3 295B the MTP self-speculative path measured slower than plain autoregressive decode.** Posting it because a measured regression is worth as much in the thread as a win, and I'd rather it be on the record than rediscovered.

Receipt (raw JSON, every number below comes from it):
https://github.com/PhilipJohnBasile/hy3-demolition-mlx/blob/main/eval/receipts/hy3_mtp_smoke.json

Measured 2026-07-07.

| | AR | MTP |
|---|---|---|
| tokens generated | 59 | 59 |
| decode throughput (`generation_tps`) | 7.402 tok/s | 0.544 tok/s |
| wall time (`generation_s`) | 41.73 s | 155.21 s |
| prefill, 30 prompt tokens (`prompt_tps`) | 0.889 tok/s | 0.642 tok/s |
| peak memory | 112.329 GB | 113.829 GB |

Receipt also records `speedup: 0.073` and `outputs_match: true`.

Two ratios, and they are not the same number: **decode throughput is 13.6x lower** with MTP on; **wall time is 3.7x longer**. The `generation_s` field includes prefill — it reconstructs as `30/prompt_tps + 59/generation_tps` for both arms — and prefill was slow in both and not equal between them. Quote whichever you mean, but don't quote 13.6x against the seconds. (And a comment of mine upthread quoted this same run as “~1.4 tok/s cold decode” — that is a third denominator, 59 tokens over the 41.73 s total wall including prefill. Same receipt, all three reconcile.)

The one thing that did work: greedy verification held. The MTP and AR continuations are character-for-character identical, which is the designed behavior.

### Conditions

- **n=1, literally.** One prompt (30 tokens), one 59-token greedy continuation per arm, `max_tokens` 64, one run each, one machine, one checkpoint, one implementation. No repeats, no variance estimate, no error bars. This is a smoke test that came out lopsided, not a benchmark.
- **Model:** Tencent Hy3, 295B total / 21B active MoE, mixed-quant MLX conversion (`ox-ox/Hy3-295B-Instruct-w2q3exp-AProjQ8-SExpQ8-OutQ8-MTP-mlx`). Both arms are the same checkpoint; the AR arm is the same directory with the MTP sidecar dropped (`num_nextn_predict_layers` 0 vs 1).
- **Implementation:** self-speculative decoding off the model's own MTP layer — no external draft model — from `eauchs/mlx-lm@a7cc3054` (branch `hy_v3-mtp`), on top of #1211. mlx 0.31.2, mlx-lm 0.31.3, Python 3.12.13, macOS Darwin 27.0.0 arm64, plus one local patch (fp32 router matmul in `MoEGate`).
- **Hardware:** one Apple M5 Max, 128 GB unified memory.

### What this does not establish

- It is one configuration on one model. It is not a claim that MTP or speculative decoding is useless in general, and not a claim about Hy3 MTP on other hardware, other quantizations, or longer generations.
- It is not evidence that a better implementation could not win. This particular path has known unfixed issues — among them, the draft cache is never prefilled over the prompt — and it drafts at most one token per verify, so even at 100% acceptance the ceiling is two tokens per trunk forward.
- **The draft/acceptance fraction was not isolated.** Nothing in this run instruments how often the MTP draft was accepted. So I cannot separate "drafts are rarely accepted" from "per-step overhead is too large even when they are." That is the measurement this result is missing, and until someone takes it the mechanism is open.
- Peak memory was 112–114 GB against the ~127.9 GB max recommended working set recorded in the receipt's device block. I have not ruled out memory pressure as a contributor.

### Hypothesis (hypothesis, not a conclusion)

At this size the verify step costs close to a full trunk forward pass, so a one-token draft can at best halve the number of trunk passes — while every step adds an MTP forward plus per-step KV-cache handling. If that per-step overhead exceeds the trunk pass it saves, the loop is net-negative more or less regardless of acceptance rate. That would predict the regression gets worse, not better, as active-parameter count rises. I have not tested it. It is falsifiable by counting trunk forwards per accepted token and reporting acceptance separately.

Happy to re-run with acceptance instrumentation, on a smaller sibling, or on someone else's branch if that's more useful than the number above.

PR #1213, opened 2026-04-27
Refactor server to use MLXServerConfig

giskarda:
* ensure mx.metal.is_available() is executed in run() instead of main()
* added run_server() for a programmatic server run
* main() now uses MLXServerConfig after argparse


giskarda:
This is part of Discussion #1212

zcbenz:

Issue #1206, opened 2026-04-26
LoRA training crashes on first backward pass with Qwen3.5-9B on M5 Max (applegpu_g17s)


sivavenkatay:
Environment:

macOS, Apple M5 Max (applegpu_g17s), 36GB unified memory
mlx 0.31.2, mlx-lm 0.31.3, mlx-metal 0.31.2
Model: mlx-community/Qwen3.5-9B-4bit
What happens:
Validation (forward pass) completes successfully. The first training iteration crashes immediately with:

[METAL] Command buffer execution failed: Insufficient Memory
(00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
This happens regardless of batch size (1 or 2), sequence length (2048–8192), number of LoRA layers (4–16), or whether --grad-checkpoint is set. Total system memory usage is low at the time of crash.

Workaround:
Switching to mlx-community/Qwen3-8B-4bit (same architecture family, previous generation) trains successfully with identical settings. Suggests the issue is specific to Qwen3.5's architecture changes in this mlx version.

Reproduce:

mlx_lm lora \
  --model mlx-community/Qwen3.5-9B-4bit \
  --train --data data/ \
  --batch-size 1 --num-layers 4 \
  --max-seq-length 2048 \
  --grad-checkpoint --val-batches 0

lawcontinue:
This is a different failure mode from the descriptor-count leak in #1185 — that one degrades over hundreds of iterations, whereas yours crashes on the very first backward pass regardless of settings. The synchronize+clear_cache workaround from #1185 won't help here.

Worth checking if the issue is specific to applegpu_g17s (M5 Max) GPU command buffer limits rather than actual memory pressure — the error is `kIOGPUCommandBufferCallbackErrorOutOfMemory` which can be a GPU resource limit, not a RAM limit. The fact that system memory usage is low at crash time supports that theory.

ivanfioravanti:
Same issue here.

Goekdeniz-Guelmez:
@ivanfioravanti @lawcontinue @sivavenkatay 
[...]

PR #1217, opened 2026-04-27
Add Metal VJP kernel for gated_delta_update (trainable Qwen3.5 / Qwen3-Next LoRA on Apple Silicon)

SudarkinV:
# Add Metal VJP kernel for gated_delta_update (trainable Qwen3.5 / Qwen3-Next LoRA on Apple Silicon)

Extends #496 with a Metal backward kernel for `gated_delta_update`. The
existing `use_kernel=False` fallback into `gated_delta_ops` unblocks the
tracing error but unrolls an `O(T)`-node auto-diff graph, which OOMs at
`T ≥ 2048` on a 36 GB M-series Mac — the common full-parameter / LoRA
fine-tuning setup for Qwen3.5-9B and Qwen3-Next-80B.

In merged PR #997, @angeloskath noted *"This does affect finetuning
fairly heavily but I think we need a kernel for that to be an enjoyable
experience anyway."* This PR is that kernel.

Related context: #1206 reports Qwen3.5-9B LoRA crashing on the first
backward pass. I do not claim this PR fully fixes that hardware-specific
report, but it targets the same training-time bottleneck: avoiding an
O(T)-node autodiff graph through the gated-delta recurrence by providing
a custom VJP path.

## What this PR adds

Two new modules and a small integration change:

- **`mlx_lm/models/gated_delta_vjp.py`** — pure-Python reference VJP
  using `mx.checkpoint` on fixed-size chunks. `O(T/CHUNK_SIZE)` autodiff
  graph, verifiable against `gated_delta_ops`. Used as the import
  fallback.
- **`mlx_lm/models/gated_delta_vjp_metal.py`** — Metal backward kernel
  registered via `mx.custom_function`. Forward-with-save + reverse
  sweep in the same chunked layout; threadgroup-local reduction (no
  atomics, deterministic). 8–11× faster than the Python VJP at
  Qwen3.5-9B shapes.
- **`mlx_lm/models/gated_delta.py`** — `gated_delta_update()` gains a
  `training: bool = False` argument that routes to the VJP path. The
  Metal backward is selected only when GPU/Metal is available, `mask is
  None`, `Dk % 32 == 0` and `Dv % 4 == 0`; otherwise the call falls back
  to the Python VJP (which has no shape constraints and runs on CPU).
  The existing `use_kernel` and mask/inference behaviour is unchanged.
- **`mlx_lm/models/qwen3_5.py`** / **`mlx_lm/models/qwen3_next.py`** —
  `linear_attn` call site passes `training=self.training`.

Inference and KV-cache paths are untouched.

## Correctness

New tests appended to `tests/test_models.py` (`TestModels` class,
`unittest.TestCase` style, matching the existing `test_gated_delta*`
block):

- `test_gated_delta_vjp_forward_equivalence` — Python VJP forward is
  bit-exact against `gated_delta_update(use_kernel=False)`.
- `test_gated_delta_vjp_fd_gradient` — central-difference check of
  `mx.grad` output vs. analytic backward on a toy shape (`B=1, T=4,
  Hk=2, Hv=4, Dk=8, Dv=8`, fp32), tolerance `1e-3`.
- `test_gated_delta_vjp_metal_matches_python` — Metal VJP forward and
  state agree with the Python reference within fp32 SIMD-reduction
  noise (`atol=1e-3`).
- `test_gated_delta_vjp_metal_gradients_match_python` — `mx.grad`
  output of the Metal VJP matches the Python VJP across all eight
  trainable inputs (`q, k, v, a, b, A_log, dt_bias, state`) within
  `atol=1e-3, rtol=1e-3`.

All eight `test_gated_delta*` tests pass locally (`python -m unittest
tests.test_models.TestModels -k gated_delta`).

## Performance (Qwen3.5-9B linear_attn shape: `B=1, Hk=16, Hv=64, Dk=192,
Dv=128`, bf16)

| T    | `use_kernel=False` (ops) fwd+bwd | Python VJP fwd+bwd | **Metal VJP fwd+bwd** | Peak mem (Metal) |
|------|---------------------------------:|-------------------:|----------------------:|-----------------:|
|  256 | 152 ms (graph-bound)             | 145.3 ms           | **13.4 ms**           | 1.8 GB           |
|  512 | 304 ms                           | 296.3 ms           | **28.2 ms**           | 3.0 GB           |
| 1024 | 617 ms                           | 599.6 ms           | **62.2 ms**           | 4.7 GB           |
| 2048 | OOM on 36 GB                     | 1233.5 ms          | **149.8 ms**          | 8.1 GB           |

## End-to-end training (Qwen3.5-9B LoRA on 36 GB M4 Max, `max_seq=4096`)

500-iteration full LoRA run on the unfiltered training set, `batch=1`,
`grad_checkpoint=true`, 4 LoRA keys (`q_proj, v_proj, in_proj_qkv,
out_proj`):

| Iter | Val loss | Peak mem |
|-----:|---------:|---------:|
|    1 | 0.524    | 9.1 GB   |
|   50 | 0.248    | 9.2 GB   |
|  100 | **0.121**| 9.2 GB   |
|  200 | 0.143    | 9.2 GB   |
|  300 | 0.246    | 13.8 GB  |
|  500 | 0.155    | 13.8 GB  |

Converges in this configuration; peak memory stable across the run.
Total time ≈ 84 minutes (10 s/iter). This is one observation on one
shape and is not a generic stability guarantee for all training
configurations.

## Relationship to #496

PR #496 added the `use_kernel: bool = True` routing so that
`.training` falls through to `gated_delta_ops`. This PR reuses that
signal — the new `training=True` path selects the VJP module; anything
else goes through the existing `use_kernel` branch. No behaviour change
for inference or for the `use_kernel=False` eval path.

## Scope

This PR covers only the training-time VJP/backward path for
`gated_delta_update`.

Out of scope and not included:
- inference kernel changes
- speculative decoding
- prefix-scan prototypes
- broader training or quantization changes

## Files

- `mlx_lm/models/gated_delta_vjp.py` (new, ~180 LoC)
- `mlx_lm/models/gated_delta_vjp_metal.py` (new, ~770 LoC)
- `mlx_lm/models/gated_delta.py` (+~30 LoC)
- `mlx_lm/models/qwen3_5.py` (+1 LoC)
- `mlx_lm/models/qwen3_next.py` (+1 LoC)
- `tests/test_models.py` (+~160 LoC)

### DIFF
--- a/mlx_lm/models/gated_delta.py
+++ b/mlx_lm/models/gated_delta.py
@@ -275,17 +275,29 @@ def gated_delta_update(
     if training:
         # Chunked VJP path with O(T/chunk) autodiff graph — fits T≥2048
         # on 36 GB Apple Silicon where the Python-ops path OOMs.
-        # Metal backward kernel is 8–11× faster than the Python reference.
-        try:
-            from .gated_delta_vjp_metal import gated_delta_update_vjp_metal
+        # Metal backward kernel is 8–11× faster than the Python reference,
+        # but only handles the unmasked GPU path with Dk%32==0 and Dv%4==0.
+        Dk = q.shape[-1]
+        Dv = v.shape[-1]
+        can_use_metal = (
+            mx.metal.is_available()
+            and mx.default_device() == mx.gpu
+            and mask is None
+            and Dk % 32 == 0
+            and Dv % 4 == 0
+        )
+        if can_use_metal:
+            try:
+                from .gated_delta_vjp_metal import gated_delta_update_vjp_metal
 
-            return gated_delta_update_vjp_metal(
-                q, k, v, a, b, A_log, dt_bias, state, mask
-            )
-        except ImportError:
-            from .gated_delta_vjp import gated_delta_update_vjp
+                return gated_delta_update_vjp_metal(
+                    q, k, v, a, b, A_log, dt_bias, state, mask
+                )
+            except ImportError:
+                pass
+        from .gated_delta_vjp import gated_delta_update_vjp
 
-            return gated_delta_update_vjp(q, k, v, a, b, A_log, dt_bias, state, mask)
+        return gated_delta_update_vjp(q, k, v, a, b, A_log, dt_bias, state, mask)
 
     beta = mx.sigmoid(b)
     g = compute_g(A_log, a, dt_bias)
--- a/mlx_lm/models/gated_delta.py
+++ b/mlx_lm/models/gated_delta.py
@@ -270,7 +270,23 @@ def gated_delta_update(
     state: Optional[mx.array] = None,
     mask: Optional[mx.array] = None,
     use_kernel: bool = True,
+    training: bool = False,
 ) -> Tuple[mx.array, mx.array]:
+    if training:
+        # Chunked VJP path with O(T/chunk) autodiff graph — fits T≥2048
+        # on 36 GB Apple Silicon where the Python-ops path OOMs.
+        # Metal backward kernel is 8–11× faster than the Python reference.
+        try:
+            from .gated_delta_vjp_metal import gated_delta_update_vjp_metal
+
+            return gated_delta_update_vjp_metal(
+                q, k, v, a, b, A_log, dt_bias, state, mask
+            )
+        except ImportError:
+            from .gated_delta_vjp import gated_delta_update_vjp
+
+            return gated_delta_update_vjp(q, k, v, a, b, A_log, dt_bias, state, mask)
+
     beta = mx.sigmoid(b)
     g = compute_g(A_log, a, dt_bias)
     if state is None:
--- a/mlx_lm/models/gated_delta_vjp.py
+++ b/mlx_lm/models/gated_delta_vjp.py
@@ -147,7 +147,7 @@ def gated_delta_update_vjp(
         k = mx.repeat(k, repeat_factor, axis=-2)
 
     if state is None:
-        state = mx.zeros((B, Hv, Dv, Dk), dtype=q.dtype)
+        state = mx.zeros((B, Hv, Dv, Dk), dtype=mx.float32)
 
     # Chunked forward; each chunk is a pure function of the incoming state,
     # so autodiff propagates gradients correctly across the recurrence.
--- a/mlx_lm/models/gated_delta_vjp.py
+++ b/mlx_lm/models/gated_delta_vjp.py
@@ -1,0 +1,179 @@
+"""Gradient-checkpointed gated_delta_update for training.
+
+The Metal kernel in :mod:`gated_delta` has no VJP registered, and the
+Python-ops fallback builds a graph with ``O(T)`` intermediate states —
+which runs out of memory for training-scale sequences (``T ≥ 2048``)
+on Apple Silicon devices with 36 GB unified memory or less.
+
+This module provides :func:`gated_delta_update_vjp`, a drop-in
+training-time replacement that runs a pure-Python recurrent forward in
+chunks of ``CHUNK_SIZE`` timesteps, each wrapped in ``mx.checkpoint``.
+Intermediate state within a chunk is recomputed on the backward pass,
+so the peak memory cost is ``O(CHUNK_SIZE)`` per layer rather than
+``O(T)``.
+
+Related: https://github.com/ml-explore/mlx-lm/issues/482
+"""
+
+from typing import Optional, Tuple
+
+import mlx.core as mx
+import mlx.nn as nn
+
+CHUNK_SIZE = 64  # Timesteps per gradient-checkpointed block.
+
+# Lower bound for the log-domain argument of the decay ``exp`` so that very
+# large A_log or softplus(a+dt_bias) do not produce denormal bf16 values that
+# poison a long recurrence. ``exp(-20) ≈ 2e-9`` — well within bf16 range.
+_LOG_DECAY_CLAMP = -20.0
+
+
+@mx.compile
+def _compute_g(A_log: mx.array, a: mx.array, dt_bias: mx.array) -> mx.array:
+    """Decay gate ``g = exp(-exp(A_log) * softplus(a + dt_bias))`` in fp32."""
+    arg = -mx.exp(A_log.astype(mx.float32)) * nn.softplus(a + dt_bias)
+    arg = mx.maximum(arg, _LOG_DECAY_CLAMP)
+    return mx.exp(arg).astype(a.dtype)
+
+
+def _chunk_forward(
+    q_c: mx.array,  # [B, T_c, Hv, Dk]
+    k_c: mx.array,  # [B, T_c, Hv, Dk]
+    v_c: mx.array,  # [B, T_c, Hv, Dv]
+    g_c: mx.array,  # [B, T_c, Hv] (scalar) or [B, T_c, Hv, Dk] (vectorized)
+    beta_c: mx.array,  # [B, T_c, Hv]
+    S_start: mx.array,  # [B, Hv, Dv, Dk]
+) -> Tuple[mx.array, mx.array]:
+    """Recurrent forward over a single unmasked chunk of ``T_c`` timesteps.
+
+    Supports both scalar and per-channel (vectorized) gating. The arithmetic
+    runs in the input dtype; fp32 accumulators double the peak memory
+    without measurable impact on convergence for bf16 training.
+    """
+    T_c = q_c.shape[1]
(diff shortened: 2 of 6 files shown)


SudarkinV:
@angeloskath gentle ping when you have bandwidth.

In #997 you noted that the fp32 gated-delta state affects finetuning fairly heavily
and that a kernel would likely be needed for that path. This PR is intended as that
training-time VJP/backward kernel for `gated_delta_update`.

It is scoped to the training/backward path only; inference routing is unchanged.

tsato081:
Hi @SudarkinV, FYI: I've opened #1389, which addresses the same training-path problem on the ops side, replacing the sequential `gated_delta_ops` fallback with the chunked (UT/WY) formulation wrapped in `mx.checkpoint`.

While benchmarking I also ran your branch on an M3 Ultra (your table's shape, bf16, fwd+bwd): your Metal kernel is faster where its constraints hold (0.097 s vs 0.117 s at T=2048), while the chunked path uses less peak memory (3.0 GB vs 8.5 GB). The two work well together: if your PR lands, the chunked path could replace the sequential loop in the fallback your routing already has. Happy to rebase mine to fit.

SudarkinV:
Hi @tsato081 — thanks for the heads-up and for #1389. I spent a day
verifying your branch independently on an M4 Max (40-core GPU, 36 GB),
since the 36 GB machines in #1206 are Max-class hardware. Findings:

**Correctness.** I cross-checked `gated_delta_ops_chunked` against
three implementations (the sequential ops reference, plus the two VJP
implementations from this PR) on GQA shapes with carried-in state:
forward and final state agree to fp32 noise, and gradients for all
eight inputs agree to rel ≤ 1.3e-5. I also ran cases your tests don't
cover: gradients under padding masks (left/right), repeated unit keys
with β=0.999 through the blocked solve, and a CPU-device run — all
clean. Gradients at decay extremes are clean too, with one documented
caveat: below the `1e-12` clamp the gradient w.r.t. `g` is zeroed
where the sequential reference returns a nonzero value — harmless in
practice (the model's parameterization multiplies by g≈0 anyway), and
this PR's Python VJP has the equivalent dead zone via its own clamp.
At T=4096 bf16 your path and this PR's Metal kernel show matching
drift vs the fp32 sequential reference (y rel 5.64e-3 both, final
state 5.57e-3 vs 5.59e-3). Numerically interchangeable.

One stress-test datapoint that *supports* your defaults: at the
no-decay/collinear corner (repeated unit keys, β=0.999, g→1) the
blocked solve degrades gracefully at SUB_BLOCK=16 (rel ~4e-4) but
blows up at SUB_BLOCK=32 (rel ~2e11). Your sb=16 choice is the right
one; I'd keep it.

**End-to-end LoRA.** 30-iteration Qwen3.5-9B LoRA smoke (real SFT
dataset, identical seed, merged tree with both branches behind an env
switch): loss curves agree within reordering noise at every report
point (≤0.004; final val 0.125 vs 0.128), peak memory 11.48 vs
11.55 GB. After re-running the kernel path on an equally warmed-up
machine, throughput matches too (single runs each, so ±5–10% noise
applies). Caveats: batch=1, 16 LoRA layers, samples average ~350
tokens — which also means the large memory gaps from long-context
micro-benchmarks don't show up in this particular e2e; they would
with ≥2k-token samples. Your OOM fix vs the main fallback stands
regardless — this comparison is specifically chunked vs this PR's
kernel.

**Performance on Max-class chips.** On M4 Max your path is faster
than this PR's Metal kernel at T=256–2048 and roughly equal at
T=4096 (within-process pairs, e.g. 184.9 ms / 3.68 GB vs 216.8 ms /
9.37 GB at T=2048, fwd+bwd bf16; absolute times drift ±20% between
sessions on this laptop, ratios within a process are stable). That
matches your M3 Ultra numbers pointing the other way — plausibly the
kernel's latency-bound sequential sweep vs your matmul-throughput
path trading off with GPU core count, though I haven't profiled
this. Since #1206-class machines are Max-class, I agree #1389 is the
right default fallback and I'd support it landing regardless of what
happens with this PR.

**A speedup for your path.** While verifying I found a GQA
optimization worth ~13–14% at your exact defaults (C=64, sb=16),
same-or-lower peak memory, same numerics: in Qwen3.5, q/k have
Hk=16 heads vs Hv=64, so `k @ k^T` and `q @ k^T` — the two heaviest
[C,C] matmuls — can be computed on Hk heads before the GQA repeat,
with the per-Hv gating (β, decay mask) applied afterwards via
broadcast; the solve stays per-Hv. Within one process on M4 Max at
T=4096: 487.9 ms / 5.92 GB vs 569.9 ms / 9.05 GB for your branch
(T=2048: 160.3 ms / 3.10 GB vs 183.7 ms / 3.68 GB); an interleaved
A/B protocol (alternating pairs, two fresh processes) puts the
median ratio higher — 1.21–1.22x at T=2048 and 1.34–1.37x at
T=4096 — so ~13–14% is the conservative floor. Part of the win is
memory traffic: the current code materializes the GQA repeat of
q/k to Hv heads for the full sequence before the chunk loop, while
the patch keeps them at Hk and broadcasts inside the checkpointed
chunk (peak drops accordingly at long T). Verified the
same correctness battery I ran on your branch (cross-check vs the
sequential reference incl. gradients and masks, the collinear stress
above, plus a 30-iter e2e LoRA run — val loss 0.128, same as your
path). For Qwen3-Next the effect should be smaller (rf=2). It's
~30 lines against your branch — happy to send it as a patch to
#1389 if you want it.

Given all this: +1 to #1389 as the default training fallback. Happy
to rebase this PR on top of yours or slim it down — maintainers'
call.

tsato081:
@SudarkinV 
Thank you for taking a full day to verify this so thoroughly. The sb=32 result and the Max-class numbers are both very useful; I'd only tested on the Ultra, so the flip was good to learn about. I'll add your collinear stress case to the test suite as a regression guard for sb, and note the clamp dead-zone in the PR description. I agree it's harmless, but it should be on the record.

And yes, please send the GQA patch, I'll run it through the equivalence tests here and fold it into #1389 with credit.

Thanks for the +1 as well, If Ultra numbers for the kernel would help at any point, happy to run them.

fblissjr:
just wanted to lend a positive comment with some independent numbers. tried this branch on my m2 ultra against qwen3.5 27b 8bit. 

tldr: the metal kernel's gradients align with plain autodiff, with the python VJP this PR ships as its non-metal fallback, and with a third metal backward from a completely separate repo that I ported from [WeZZard/jlens-qwen36's](https://github.com/WeZZard/jlens-qwen36) kernel... and 27b qlora drops from a 117gb peak on main to 39gb here (!) at 3x the training throughput (150 tok/s vs ~50tok/s, and on a ~700-token train sample dataset). main's run only survived because i have 192gb... on this branch a 64gb machine fits with room to spare. inference appears to be the same tok/s, same peak memory.

on the gradients: i ported and modified a metal kernel for this same recurrence (https://github.com/fblissjr/jlens-mlx/blob/main/jlens_mlx/providers/qwen3_5_gdn.py - ported from WeZZard/jlens-qwen36's kernel with some adaptations, re-verified against autodiff; built for jacobian/jlens interp through the GDN layers, narrow scope, seq len T<=128, scalar gating only). didn't even know this PR existed until I decided to go check today (should have earlier!)

i fed the same inputs at the 27b's real head dims, through all 4 backward implementations - @SudarkinV's metal VJP and python VJP, the kernel i ported, and plain autodiff through the sequential ops. and every single one produces the same gradients down to float32 noise (rel <= 2.6e-7). the PR's test suite passed here for me 5/5. i also counted which path a real lora run actually dispatches to: all 96 GDN layer calls hit the metal kernel and none fell back to python. granted this was just a single test.

caveats: everything above is one super narrow setup with qwen3.5 27b, one sample at a time, no padding masks, fresh state, the simple scalar gate. qwen3 next was not tested in any way and looks to have different shapes. gradient comparisons were fp32, not bf16. the lora runs were 10-30 steps on toy data. enough to show speed, memory, and that gradients flow, but nothing about how a real finetune turns out. all on one machine (m2 ultra).

if this merges i get one less thing to maintain which would be awesome.

testing code + raw logs if helpful:
https://github.com/fblissjr/jlens-mlx/tree/upstream-pr-eval/bench/upstream_pr_eval

Goekdeniz-Guelmez:
I suggest you to look into my package, it already has efficient long context linear training. the API usage is almost the same as mlx-lm with minor additions and changes. https://github.com/Goekdeniz-Guelmez/mlx-lm-lora this also has a ton of more algorithms.

zcbenz:

PR #1223, opened 2026-04-28
Add Laguna XS.2 

Blaizzy:


johntdavies:
Running the 4-bit on my M5 Max: 6000-token generation @ 116 tok/s, 19.9 GB peak. Nice work!

Blaizzy:
My pleasure John! 🙌🏽

pierre427:
Thanks for this — got `AtomicChat/Laguna-XS-2.1-MLX-8bit` running on this. Two small additions that might be worth folding in:

**1. `make_cache()` for the sliding-window layers.** Laguna is 10 `full_attention` + 30 `sliding_attention` (512) layers, but without a `make_cache()` the loader allocates a full `KVCache` on all 40 — the 30 sliding layers keep keys they never attend to. Adding the standard mixed pattern (as in `gpt_oss` / `gemma3_text` / `cohere2_moe`):

```python
def make_cache(self):
    caches = []
    for lt in self.args.layer_types:
        if lt == "sliding_attention" and self.args.sliding_window:
            caches.append(RotatingKVCache(max_size=self.args.sliding_window))
        else:
            caches.append(KVCache())
    return caches
```

(with `from .cache import KVCache, RotatingKVCache`). On an M5 Max I measured **~2.7× decode + ~4 GB less KV at 32k context** vs full-KV, and it's **bit-identical for ≤512-token context** — the forward already builds the sliding mask with `window_size=self.args.sliding_window`, so this just stops storing masked-out keys.

**2. Loading VLM-style repacks.** Some 8-bit repacks (e.g. AtomicChat's) wrap every tensor under a `language_model.` prefix. A guarded strip at the top of `sanitize()` loads them (same repack shape `cohere2_moe` already handles), and it's a no-op when the prefix is absent:

```python
if any(k.startswith("language_model.") for k in weights):
    weights = {(k[len("language_model."):] if k.startswith("language_model.") else k): v
               for k, v in weights.items()}
```

Happy to open a PR against your branch if that's easier. Thanks again!

pierre427:
Running the 8-bit repack (`AtomicChat/Laguna-XS-2.1-MLX-8bit`) on an M5 Max against this branch, two additions made it production-ready for us:

**1. `make_cache()` for the sliding-window layers.** As shipped, `make_prompt_cache` allocates a full `KVCache` on all 40 layers, including the 30 sliding-window (512) layers — so long-context runs store/attend over history those layers can't use. Returning `RotatingKVCache(max_size=<window>)` for the sliding layers and `KVCache` for the 10 global layers gives up to **2.7× decode and ~4 GB less KV at 32k**, verified bitwise-identical to the full cache at ≤ window context.

**2. `language_model.` prefix strip in `sanitize()`.** The VLM-style repacks (this 8-bit one included) prefix every tensor with `language_model.`, so a stock load misses all weights — a one-line strip handles them.

Both are tested on our end and independent of the core model code here. Happy to push them to this branch or open a small follow-up once this lands — whichever you'd prefer.

pierre427:
Turned the two additions into a ready-to-merge follow-up onto your branch: Blaizzy/mlx-lm#26 (windowed `make_cache()` + `language_model.` prefix strip). Tested against the current `pc/add-lg` and black-clean — merging it updates this PR. No worries if you'd rather fold them in yourself.

ahmedandaloes:
any updates!

michalk8:
closed via #1704

PR #1228, opened 2026-04-29
Skip lm_head on non-rank-0 pipeline-parallel ranks

### LINKED ISSUE #1203: Skip `lm_head` on non-final pipeline-parallel ranks?
## Problem

In the pipeline-parallel path, every rank ends up running `lm_head` on the same broadcast hidden state, e.g. `mlx_lm/models/deepseek_v3.py:357-361`:

```python
# Broadcast h while keeping it in the graph
if pipeline_size > 1:
    h = mx.distributed.all_gather(h)[: h.shape[0]]
return self.norm(h)
```

…then in the outer `Model.__call__`:

```python
out = self.model(inputs, cache)
return self.lm_head(out)
```

For non-rank-0 ranks (PipelineMixin's "rank=0 gets the last layers" convention), the produced logits are identical to rank 0's and aren't actually used downstream — the lm_head call there is pure waste.

## Cost

On a 2-node Apple Silicon PP setup running a ~400B MoE, the lm_head weight read is ~500MB per token. Skipping it on non-final ranks saves ~1.5 ms/token in our cluster (we've been carrying a fork patch for ~5 weeks across the Qwen3.5 family).

## Proposed approach

Add a guard at the top of `Model.__call__` in each PipelineMixin-using model (deepseek_v2/v3, glm4_moe, glm4_moe_lite, ministral3):

```python
out = self.model(inputs, cache)
if self.model.pipeline_rank != 0:
    return out
return self.lm_head(out)
```

~5 lines per model. (`ministral3` has a `tie_word_embeddings` branch, so the guard wraps both paths — still ~5 lines, but not a literal copy of the snippet above.)

## Open design question

This makes non-rank-0 ranks return shape `(B, L, hidden_size)` while rank-0 returns `(B, L, vocab_size)` — different shapes, different compile graphs across ranks. In practice each rank already compiles a different graph (different layer slices), but if there's a deliberate intent to keep the per-rank compiled graph homogeneous I'd rather know before sinking work into the PR.

Alternative shape: do the skip inside `PipelineMixin` / a helper, gated by an explicit `elide_output_head` kwarg so callers opt in.

Would the team accept either of these? Happy to PR whichever direction you prefer.

lawcontinue:
## Summary

In pipeline-parallel inference, every rank runs `lm_head` on the same broadcast hidden state. For non-rank-0 ranks, the logits are identical to rank 0's and never used — the call is pure waste.

This PR adds a `pipeline_rank != 0` guard before `lm_head` in all six `PipelineMixin` models, skipping the projection on non-final ranks.

Closes #1203.

## Changes

Each model's `Model.__call__` now returns early when `pipeline_rank != 0`:

```python
out = self.model(inputs, cache)
if self.pipeline_rank != 0:
    return out
return self.lm_head(out)
```

| File | Model |
|------|-------|
| `deepseek_v2.py` | DeepSeek V2 |
| `deepseek_v3.py` | DeepSeek V3 |
| `deepseek_v32.py` | DeepSeek V3.2 |
| `glm4_moe.py` | GLM4 MoE |
| `glm4_moe_lite.py` | GLM4 MoE Lite |
| `ministral3.py` | Ministral 3 (handles `tie_word_embeddings` branch) |

## Performance

Measured on a 2-node Gemma-3-12B setup (Thunderbolt, MLX 0.31.1): skipping `lm_head` on the remote rank saves ~30% of per-step time. The savings scale with `vocab_size` — DeepSeek V3 with 129K vocab would see a larger benefit.

## Design notes

- Non-rank-0 ranks now return `(B, L, hidden_size)` instead of `(B, L, vocab_size)`. Each rank already compiles a different graph (different layer slices), so heterogeneous output shapes should not be an issue.
- The guard is zero-cost for single-rank inference (`pipeline_rank` defaults to 0 in `PipelineMixin.__init__`).
- `ministral3` has a `tie_word_embeddings` branch — the guard is placed before both paths, keeping the patch uniform across all six models.

### DIFF
--- a/mlx_lm/models/deepseek_v2.py
+++ b/mlx_lm/models/deepseek_v2.py
@@ -425,6 +425,8 @@ class Model(nn.Module):
         cache: Optional[Any] = None,
     ):
         out = self.model(inputs, cache)
+        if self.pipeline_rank != 0:
+            return out
         return self.lm_head(out)
 
     def sanitize(self, weights):
--- a/mlx_lm/models/deepseek_v3.py
+++ b/mlx_lm/models/deepseek_v3.py
@@ -375,6 +375,8 @@ class Model(nn.Module):
         cache: Optional[Any] = None,
     ):
         out = self.model(inputs, cache)
+        if self.pipeline_rank != 0:
+            return out
         return self.lm_head(out)
 
     def sanitize(self, weights):
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -489,6 +489,8 @@ class Model(nn.Module):
         cache: Optional[Any] = None,
     ):
         out = self.model(inputs, cache)
+        if self.pipeline_rank != 0:
+            return out
         return self.lm_head(out)
 
     def sanitize(self, weights):
--- a/mlx_lm/models/glm4_moe.py
+++ b/mlx_lm/models/glm4_moe.py
@@ -312,6 +312,8 @@ class Model(nn.Module):
         cache: Optional[Any] = None,
     ):
         out = self.model(inputs, cache)
+        if self.pipeline_rank != 0:
+            return out
         return self.lm_head(out)
 
     def sanitize(self, weights):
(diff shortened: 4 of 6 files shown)


zcbenz:

PR #1232, opened 2026-04-30
fix: Metal resource-limit errors from batched cache metadata

spicyneuron:
This PR fixes a lazy graph accumulation issue in long-running batched generation. Small cache metadata arrays such as `offset`, `left_padding`, and sequence lengths can be updated token-by-token without being materialized often enough. That eventually surfaces as a `[metal::malloc] Resource limit (...) exceeded`, even though the full KV cache tensors aren't growing unexpectedly.

This is similar to #960, which fixed one path for `BatchRotatingKVCache`. I took a slightly broader approach at the generation scheduling boundary: explicitly evaluate lightweight cache metadata each decode step, regardless of the specific cache shape being used.

## Details

- Adds cache-owned helper, `prompt_cache_eval_targets()`
- Collects lightweight metadata fields: `offset`, `left_padding`, `lengths`, and `_lengths`
- Handles nested `CacheList` structures
- Keeps the generation hot path simple
- Avoids evaluating full KV cache tensors
- Adds a test for recursive metadata target discovery

## Context

I originally encountered this through `mlx-vlm`, but the issue is in `mlx-lm` batched generation/cache path.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1366,7 +1366,13 @@ class GenerationBatch:
         # asynchronously
         self._next_tokens = sampled
         self._next_logprobs = list(logprobs)
-        mx.async_eval(self._next_tokens, self._next_logprobs, token_context)
+
+        mx.async_eval(
+            self._next_tokens,
+            self._next_logprobs,
+            token_context,
+            cache.prompt_cache_eval_targets(self.prompt_cache),
+        )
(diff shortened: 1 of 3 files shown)


spicyneuron:
Closing for now. Possible there was a confounding factor in my fork — will reopen with a clean repro if it pops up again.

PR #1238, opened 2026-05-02
fix(gemma4): add stop_gradient on MoE router top_k_indices

TrentCarter:
## Summary

Wrap `top_k_indices` in `mx.stop_gradient` inside `gemma4_text.DecoderLayer.__call__`, just after the router call. Router weights remain learnable via `top_k_weights` (softmax over the selected-experts subset), so the gating function still trains end-to-end; only the argmax-derived index path is detached.

```python
top_k_indices, top_k_weights = self.router(h)
top_k_indices = mx.stop_gradient(top_k_indices)  # <-- added
h2 = self.experts(h2, top_k_indices, top_k_weights)
```

## Motivation

Discovered while doing LoRA SFT on a Gemma-4 27B IT (4-bit MoE) checkpoint. Without this patch, after a few hundred training steps the adapter-loaded model exhibits severe inference instability (8-10K character token loops, fenced-code output never closes). Adding `stop_gradient` on `top_k_indices` removes the failure mode while leaving routing learnable.

The change is also a no-op on inference: `argpartition` already returns int indices, so cost-free in eval.

## Minimal repro outline

1. Load `mlx-community/gemma-4-26b-a4b-it-4bit` with `mlx_lm.load(...)`.
2. Train a small LoRA adapter (function-style code SFT corpus is sufficient; ~hundreds of short samples).
3. Run inference on held-out tasks: with the unpatched model the assistant turn drifts into repeated tokens after the fence-close. Patched: clean termination.

(Happy to share a more concrete repro on request — the downstream eval is internal but the failure mode is reproducible on any non-trivial adapter run against the 27B MoE variant.)

## Test

Added `test_gemma4_moe_router_top_k_indices_no_grad` in `tests/test_models.py`. It builds a tiny MoE-enabled gemma4 model (`hidden_size=8, num_experts=4, top_k_experts=2`), runs a forward pass, and verifies via `nn.value_and_grad` that the router projection weight still receives non-zero gradient (i.e. the router stays learnable through `top_k_weights`). Runs in <0.1s.

## Notes

- Searched open and closed issues/PRs on `ml-explore/mlx-lm` for "gemma4 router stop_gradient" — found none. If this duplicates an in-flight discussion I missed, happy to close.
- Black-formatted; pre-commit clean on the touched files.

### DIFF
--- a/mlx_lm/models/gemma4_text.py
+++ b/mlx_lm/models/gemma4_text.py
@@ -356,6 +356,10 @@ class DecoderLayer(nn.Module):
             h1 = self.post_feedforward_layernorm_1(h1)
 
             top_k_indices, top_k_weights = self.router(h)
+            # Block gradient through discrete expert selection; router is still
+            # learnable via top_k_weights. Without this, argmax noise from
+            # top_k_indices destabilizes adapter (LoRA/SFT) training.
+            top_k_indices = mx.stop_gradient(top_k_indices)
             h2 = self.pre_feedforward_layernorm_2(h)
             h2 = self.experts(h2, top_k_indices, top_k_weights)
             h2 = self.post_feedforward_layernorm_2(h2)
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -842,6 +842,60 @@ class TestModels(unittest.TestCase):
         )
         self.assertEqual(config["quantization"]["bits"], 4)
 
+    def test_gemma4_moe_router_top_k_indices_no_grad(self):
+        """Regression: top_k_indices in DecoderLayer must be detached from
+        autograd. Router weights still flow through top_k_weights.
+        """
+        from mlx_lm.models import gemma4
+
+        args = gemma4.ModelArgs.from_dict(
+            {
+                "model_type": "gemma4",
+                "vocab_size": 32,
+                "text_config": {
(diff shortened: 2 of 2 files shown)


nastya236:
Thanks for the contribution! I don't think this change is needed: `SwitchGLU` already applies mx.stop_gradient(idx) internally during training. 
Do you have a concrete reproduction of the adapter-training instability you mentioned?

Also generally if there is a bug and gradients are propagated through ids, we need to solve it for all models and the fix should be not in model specific file. 
Happy to dive deep into this with you.

jkl-ds:
I also ran into an issue training a LoRA on Gemma 4. In my case, the training failed with "ValueError: [gather] Cannot calculate VJP with respect to indices. Use stop_gradient on indices to stop gradients from being computed." My workaround was to modify the model file add stop_gradient() within Router.__call__().

```python
         top_k_indices = top_k_indices[..., -self.config.top_k_experts :]
         top_k_indices = mx.stop_gradient(top_k_indices)
         top_k_weights = mx.take_along_axis(expert_scores, top_k_indices, axis=-1)
```

The SwitchGLU call to stop_gradient doesn't help because that happens later.

michalk8:
Closing this in favor of #1856

PR #1241, opened 2026-05-03
feat: chunked prefill for GatedDeltaNet (FlashQLA algorithm)

0xClandestine:
## Summary

- Replaces the sequential `gated_delta_ops` Python loop (O(T) dispatch overhead) with a parallel chunked prefill for Qwen3.5/3.6 linear attention layers
- Ported from [FlashQLA](https://github.com/QwenLM/FlashQLA) (MIT License) — the same algorithm used in the CUDA/Triton reference implementation
- All new code is contained in `gated_delta.py` — no new files added

## Algorithm

The sequence is split into chunks of size `C` (default 64). Within each chunk, the intra-chunk attention and KKT triangular solve are computed in parallel via Metal kernels and `mx.einsum`. The inter-chunk recurrence `h_i = f(h_{i-1})` remains sequential (this is a fundamental data dependency — a parallel prefix scan would cost O(K³) per step vs O(K²) sequential).

## Dispatch in `gated_delta_update`

| Condition | Path |
|---|---|
| `T > 1`, `mask is None`, Metal available | Chunked prefill (this PR) |
| `T == 1` | Existing Metal step kernel (decode) |
| `mask` set, or no Metal | Existing `gated_delta_ops` (fallback) |

The `mask` fallback covers padded batches (variable-length sequences); the chunked path is used for the common single-batch or uniform-length prefill case.

## Throughput

Measured on Apple M-series (float32, chunk\_size=64, Qwen3.5 config: Hk=16, Hv=64, K=V=128):

| T | Sequential (before) | Chunked (after) | Speedup |
|---|---|---|---|
| 512 | ~200 Ktok/s | ~290 Ktok/s | 1.5× |
| 2048 | ~50 Ktok/s | ~340 Ktok/s | 6.8× |
| 8192 | ~12 Ktok/s | ~340 Ktok/s | 28× |
| 16384 | ~6 Ktok/s | ~340 Ktok/s | 56× |

Throughput is flat after ~1K tokens because chunked compute is O(T) and Metal dispatch overhead is amortized across chunks.

## Correctness

Output relative error vs sequential reference: < 1e-5 (float32, T=256, GQA 4:1).

### DIFF
--- a/mlx_lm/models/gated_delta.py
+++ b/mlx_lm/models/gated_delta.py
@@ -259,6 +259,219 @@ def gated_delta_ops(
     return y, state
 
 
+# ---------------------------------------------------------------------------
+# Chunked prefill (FlashQLA algorithm).
+# Ported from https://github.com/QwenLM/FlashQLA (MIT License).
+# Replaces the O(T) sequential loop in gated_delta_ops for prefill on Metal.
+# ---------------------------------------------------------------------------
+
+def _pad_reshape(x: mx.array, dim: int, C: int) -> mx.array:
+    T = x.shape[dim]
+    pad = (C - T % C) % C
+    if pad:
+        pw = [(0, 0)] * x.ndim
+        pw[dim] = (0, pad)
+        x = mx.pad(x, pw)
+    s = list(x.shape)
+    s[dim : dim + 1] = [-1, C]
+    return x.reshape(s)
+
+
+def _fill_g_tail(g: mx.array, T: int, C: int) -> mx.array:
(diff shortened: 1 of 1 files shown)


0xClandestine:
Closing: the existing gated_delta_kernel already handles prefill in a single Metal dispatch with recurrent state in thread registers, which is faster on Apple Silicon than the chunked FlashQLA algorithm (which is optimized for CUDA/Triton). Benchmarks show the existing implementation outperforms this approach for T ≥ 1024. No regression fix needed.

PR #1246, opened 2026-05-05
add: lfm2/2.5 tool parser

jbuchananr:
## Issue
  LFM2 / LFM2.5 models emit tool calls as
  `<|tool_call_start|>[fn(args)]<|tool_call_end|>`, but no parser was registered for this
  format, so tool calls leaked into `message.content` instead of `tool_calls`.

  The existing `pythonic` parser also couldn't have handled real outputs: its regex
  rejects dotted names like `grocery.orderIngredients` and breaks on nested list/dict
  args.

  ## Changes
  - New `mlx_lm/tool_parsers/lfm2.py` — parses the payload with `ast` so dotted names,
  nested args, and multi-call lists all work.
  - `_infer_tool_parser` routes chat templates containing `"List of tools:"` to the new
  parser (matches both LFM2 and LFM2.5).

  ## Testing
  - [x] Added LFM-specific tool calling tests
  - [x] Tested on internal test suite for tool calling
  - [x] Tested locally against `mlx_lm.server`: single call, multi-tool selection, nested
  args, declined call, streaming, multi-turn round-trip, parallel calls

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -559,6 +559,10 @@ def _infer_tool_parser(chat_template):
         return "longcat"
     elif "<arg_key>" in chat_template:
         return "glm47"
+    elif "List of tools:" in chat_template and (
+        "<|tool_list_start|>" in chat_template or "<|im_start|>" in chat_template
+    ):
+        return "lfm2"
     elif "<|tool_list_start|>" in chat_template:
         return "pythonic"
     elif (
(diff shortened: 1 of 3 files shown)


blairhudson:
I actually found that LFM 2.5 worked with the existing Pythonic function call format parser. it just wasn't being automatically mapped due to the parser inference only looking at the chat template. I extended the inference to also consider the vocab: https://github.com/ml-explore/mlx-lm/pull/1260

Perhaps we should combine?

jbuchananr:
@blairhudson Yes! Let me run your PR through some of my internal tests this weekend

michalk8:
closed via #1260

PR #1249, opened 2026-05-06
Fix mlx_lm.server --adapter-path silently ignored at startup

### LINKED ISSUE #1248: mlx_lm.server --adapter-path silently ignored at startup
CLI accepts `--adapter-path` and stores it on `ModelProvider`, but the value never reaches the model loader. No error or warning is raised. This affects anyone serving an adapter via the CLI.

## Repro

Pass a non-existent adapter path:

```bash
mlx_lm.server \
  --model mlx-community/Qwen2.5-0.5B-Instruct-4bit \
  --adapter-path /tmp/this-does-not-exist
```

The server starts successfully and serves requests.

## Expected

The server should fail loudly:

```python
      raise FileNotFoundError(f"The adapter path does not exist: {adapter_path}")
  FileNotFoundError: The adapter path does not exist: /tmp/this-does-not-exist
```

Behavior should match the per-request HTTP API, where invalid adapter paths correctly raise `FileNotFoundError`.

## Root cause

`ModelProvider.__init__` stores both model and adapter under the alias `"default_model"`:

```python
self._model_map["default_model"] = self.cli_args.model
```

`load_default()` calls:

```python
```

Inside `load()`:

```python
    adapter_path = self._adapter_map.get(model_path, adapter_path)
```

`model_path` is resolved before `_adapter_map` is queried. After resolution, the lookup key becomes the real model name instead of `"default_model"`.

Since `_adapter_map` only contains `"default_model"`, the lookup misses and `adapter_path` falls back to `None`.

As a result, the CLI-provided adapter path is silently dropped.

## Notes

Per-request loads are unaffected because they pass the real model name directly, so the alias lookup path is bypassed.

odysa:
Fixes #1248.

`ModelProvider.load("default_model", ...)` overwrites `model_path` with the resolved real path before looking up `_adapter_map`. Since `_adapter_map` is keyed by the alias `"default_model"`, the lookup misses and the CLI-provided `--adapter-path` silently falls back to `None`.

Fix: resolve `adapter_path` using the alias key before reassigning `model_path`.

## Repro

Before fix — bogus adapter path is silently dropped:

```bash
mlx_lm.server --model mlx-community/Qwen2.5-0.5B-Instruct-4bit --adapter-path /tmp/does-not-exist
# Server starts cleanly.
```

After fix — same command raises `FileNotFoundError: The adapter path does not exist: /tmp/does-not-exist`, matching the behavior of the per-request HTTP API.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -385,8 +385,8 @@ class ModelProvider:
             self.load("default_model", None, "default_model")
 
     def load(self, model_path, adapter_path=None, draft_model_path=None):
-        model_path = self._model_map.get(model_path, model_path)
         adapter_path = self._adapter_map.get(model_path, adapter_path)
+        model_path = self._model_map.get(model_path, model_path)
         draft_model_path = self._draft_model_map.get(draft_model_path, draft_model_path)
 
         model_key = (model_path, adapter_path, draft_model_path)


eilidhmae:
Independent confirmation — and worth flagging that this fails **completely silently**: no error, `/v1/models` looks correct, the adapter just isn't applied.

Root cause is the argument-resolution order in `ModelProvider.load`: `model_path` is remapped from the `"default_model"` key to the real filesystem path *before* the adapter lookup, so `self._adapter_map.get(model_path, adapter_path)` keys on the resolved path, never matches `_adapter_map["default_model"]`, and falls back to `None`. The fix in this PR (resolve the adapter/draft by the original key first) is exactly right, and since it's in the shared `load()` it covers both `load_default()` at startup and subsequent loads.

What I observed: serving a known-good trained LoRA with `--adapter-path` produced generations byte-for-byte identical to the base model at `temperature 0`. The adapter file itself is fine — `mlx_lm.load(model, adapter_path=<dir>)` in-process *does* change the output; only the server path drops it.

Impact worth noting for prioritization: because it's silent, it's easy to draw wrong conclusions from server output. In our case it served the base model under several different adapter names, making a set of distinct adapters look indistinguishable from the base — caught only by hashing the completions.

A cheap regression guard that would catch any recurrence: after launch, send one fixed prompt (temp 0) to the base and to the adapter route and assert the two completions differ.

The diff here matches the one-line reorder we independently arrived at. LGTM.

zcbenz (approved):

PR #1251, opened 2026-05-06
Fix BatchRotatingKVCache rotated flag deserializing to True

### LINKED ISSUE #1250: BatchRotatingKVCache: rotated=False reloads as True from saved prompt cache

## Repro

```python
import os, tempfile
import mlx.core as mx
)

    save_prompt_cache(path, [c])

assert c.rotated is False
```

Output:

```
```

## Root cause

```python
@property
def meta_state(self):

@meta_state.setter
def meta_state(self, v):
```

## Suggested fix

```python
self.rotated = v[3] == "True"
```

odysa:
Fixes #1250.

`meta_state` round-trips `self.rotated` through `str()` on save and `bool()` on load, but `bool("False") is True` (any non-empty string is truthy). A cache saved with `rotated=False` reloads as `rotated=True`, silently corrupting the rotation mask in subsequent `make_mask` calls.

Fix: parse the bool by string comparison.

```python
- self.rotated = bool(v[3])
+ self.rotated = v[3] == "True"
```

Added a regression test (`test_save_load_batch_rotating_cache`) that round-trips both `rotated=False` and `rotated=True`. Verified the test fails on `main` and passes with the fix.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1312,7 +1312,7 @@ class BatchRotatingKVCache(_BaseCache):
             int,
             v[:3],
         )
-        self.rotated = bool(v[3])
+        self.rotated = v[3] == "True"
 
     def is_trimmable(self):
         return self._offset < self.max_size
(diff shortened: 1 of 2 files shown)


zcbenz:
I'm closing this in favor of https://github.com/ml-explore/mlx-lm/pull/1778 which fixes the root cause.

PR #1254, opened 2026-05-06
Fix ArraysCache missing is_trimmable/trim for hybrid model prompt cache

### LINKED ISSUE #1162: Bug report — Qwen3-Next hybrid cache silently fails - Prompt cache silently broken for Qwen3-Next (hybrid Gated-DeltaNet + SWA Gated-Attention) — recomputes every turn
## Summary

The `PromptCache` in `mlx-lm` does not correctly persist state for Qwen3-Next-80B-A3B (hybrid architecture: 36 Gated-Attention layers with SWA + 12 Gated-DeltaNet recurrent layers + MoE routing). Cache hit-rate appears ~0 on multi-turn workloads: every repeat of the exact same prompt triggers a full prefill. No error is raised — the cache silently recomputes.

By contrast, the same model served via `llama.cpp`'s `llama-server` (post PR #19408, merged Feb 2026) achieves correct cache reuse with 40×+ speedup on warm turns.

## Environment

- Mac Studio M4 Ultra, 128 GB unified memory, macOS 25.3
- `mlx-lm` served via LM Studio (MLX backend, build from 2026-04)
- Model: `qwen3-next-80b-a3b-instruct-mlx@8bit` (80.8 GiB)
- Test harness: direct streaming HTTP POST to `/v1/chat/completions`, measuring TTFT

## Expected behavior

On a repeat of the exact same prompt within the loaded model's lifetime, TTFT should drop from cold (~70 s for 60K tokens) to hit (~1–2 s).

## Actual behavior

TTFT stays at cold levels for every turn. There is no speedup from cache reuse.

### Numbers — cold TTFT (no cache needed — first turn)

| Input size | TTFT |
|---|---|
| 12K | 12.4 s |
| 30K | 28.7 s |
| 60K | 71.6 s |

### Numbers — "hot" TTFT (same prompt, repeated) — should be 1–2 s

| Input size | TTFT observed |
|---|---|
| 12K | ~12 s — no change from cold |
| 30K | ~28 s — no change from cold |
| 60K | ~70 s — no change from cold |

### Comparison — same model, llama.cpp backend (cache works)

| Input size | Cold TTFT | Hot TTFT | Speedup |
|---|---|---|---|
| 60K | 54 s | **1.0 s** | **54×** |
| 120K | 250 s | **~1.2 s** | ~208× |

Config: `--swa-full --ctx-checkpoints 128 --checkpoint-every-n-tokens 4096 --slot-prompt-similarity 0.3`

## Reproduction

1. Load `qwen3-next-80b-a3b-instruct-mlx@8bit` in LM Studio (or `mlx-lm` directly)
2. Send a ~10K token prompt via `/v1/chat/completions`, record TTFT
3. Send the exact same prompt again, record TTFT
4. Observe: both TTFTs are similar (both cold) — cache is not hit

## Root cause hypothesis

`PromptCache` appears to be designed for pure transformer layers (list of K, V per layer). Qwen3-Next has 12 Gated-DeltaNet layers that hold **recurrent state**, not K/V. When the cache "restores" a prior prompt, those 12 layers presumably reset to zero state, so the effective model output is wrong → the server falls back to recomputing the full prompt.

Related: the SWA (sliding-window attention) layers may also need special handling around window boundaries that the current cache class doesn't implement.

## Reference implementation

`llama.cpp` PR #19408 (merged Feb 2026) added hybrid-architecture checkpoint save/restore. The key additions:
- Per-layer polymorphic cache entry type (K/V for attention, recurrent state for Mamba/DeltaNet)
- Position-indexed checkpoints (`--checkpoint-every-n-tokens`)
- Prefix-match restore that replays only new tokens

A Python port of the same logic into `mlx-lm` should be feasible in ~1–2 engineer-weeks.

## Impact

Qwen3-Next is a 2025 flagship MoE release and a primary target for local agentic workloads (3.9B active of 80B total — fits comfortably on 128 GB consumer machines). Multi-turn / conversational / tool-calling workloads lose the primary performance lever (prefix-cache reuse).

Until this is fixed, users running Qwen3-Next on Apple silicon are better served by the `llama.cpp` GGUF path, even though MLX cold prefill is ~15 % faster. The cache deficit dominates after turn 2.

## Related

- Existing issue: `mlx-lm#980` (speculative decoding breaks prompt cache — similar cache-path fragility)
- Upstream reference: `ggml-org/llama.cpp#19408`

EagerofLight:
## Problem

Hybrid models (Qwen3.5, Qwen3.6, Qwen3-Next) mix `ArraysCache` (DeltaNet recurrent layers)
with `KVCache` (attention layers). `ArraysCache` inherits `is_trimmable() → False` from
`_BaseCache` and has no `trim()` method, so `can_trim_prompt_cache()` returns `False` for
the entire hybrid cache.

This silently breaks:
- **Prefix cache reuse** in `mlx_lm.server` — `fetch_nearest_cache()` skips the trim path
- **LRU prefix dedup** — `insert_cache()` can't evict redundant prefixes
- **Speculative decoding** — `generate.py` raises `ValueError` on non-trimmable caches

## Root cause

```python
# cache.py — _BaseCache default
def is_trimmable(self):
    return False  # ArraysCache inherits this

# cache.py:88 — gate check
def can_trim_prompt_cache(cache):
    return all(c.is_trimmable() for c in cache)  # one False → entire cache rejected
```

## Solution

Add `is_trimmable()` → `True` and `trim(n)` to `ArraysCache`.

Recurrent state is a compressed summary of all past tokens — it can't be partially
rolled back like KV cache. So `trim(n)` resets the recurrent state to empty; the
recurrent layers recompute on next forward while KVCache layers still benefit from
their own trim. For exact-match reuse (same prompt repeated), `trim()` is never
called and the full state is preserved.

## Verification (Qwen3.5-4B-4bit, M4 16GB, `mlx_lm.server`)

| Request | TTFT | Cached |
|---------|------|--------|
| Cold (571 tok) | 2030 ms | 0 / 571 |
| Exact match | 314 ms | 570 / 571 |
| Prefix reuse | 397 ms | 424 / 431 |
| Extended prompt | 461 ms | 564 / 577 |

Before this fix, prefix reuse and extended prompt got 0 cached tokens.

Fixes #1162

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -723,6 +723,25 @@ class ArraysCache(_BaseCache):
     def empty(self):
         return self.cache[0] is None
 
+    def is_trimmable(self):
+        return True
+
+    def trim(self, n):
+        """Trim the cache by n tokens.
+
+        ArraysCache holds compressed recurrent state that cannot be partially
+        rolled back. When trimming is required, the cache is fully reset so
+        the recurrent layers recompute from scratch. KVCache layers in the
+        same hybrid model still benefit from their own trim.
+
+        For exact-match reuse (n=0), this method is not called and the full
+        recurrent state is preserved — the common server case.
+        """
+        if n <= 0:
+            return 0
+        self.cache = [None] * len(self.cache)
+        return n
+
     @property
     def nbytes(self):
         return sum(c.nbytes for c in self.cache if c is not None)
--- a/tests/test_prompt_cache.py
+++ b/tests/test_prompt_cache.py
@@ -216,22 +216,27 @@ class TestPromptCache(unittest.TestCase):
         num_trimmed = trim_prompt_cache(cache, 4)
         self.assertEqual(num_trimmed, 3)
 
-        # Can't trim arrays cache
+        # Trimming arrays cache resets recurrent state
         cache = [ArraysCache(size=2) for _ in range(2)]
         for c in cache:
             c[0] = mx.zeros((5, 5))
             c[1] = mx.zeros((5, 5))
         num_trimmed = trim_prompt_cache(cache, 7)
-        self.assertEqual(num_trimmed, 0)
+        self.assertEqual(num_trimmed, 7)
+        for c in cache:
+            self.assertTrue(c.empty())
 
-        # All cache's have to be trimmable
+        # Hybrid cache (ArraysCache + KVCache) is trimmable
         cache = [ArraysCache(size=2), KVCache()]
         cache[0][0] = mx.zeros((5, 5))
         cache[0][1] = mx.zeros((5, 5))
         x = mx.random.uniform(shape=(1, 8, 10, 4))
         cache[1].update_and_fetch(x, x)
         num_trimmed = trim_prompt_cache(cache, 1)
-        self.assertEqual(num_trimmed, 0)
+        self.assertEqual(num_trimmed, 1)
+        # ArraysCache resets, KVCache trims normally
+        self.assertTrue(cache[0].empty())
+        self.assertEqual(cache[1].offset, 9)
 
         cache = [RotatingKVCache(max_size=6) for _ in range(2)]
         for c in cache:


zcbenz:
Thanks for the PR but completely reseting the cache would not be the ideal fix, I left some notes at https://github.com/ml-explore/mlx-lm/pull/990#pullrequestreview-4989083778.

PR #1261, opened 2026-05-09
Support for Zyphra/ZAYA1-base

kyr0:
I found this novel architecture quite interesting, so I clean room implemented support for it after reading and debugging the [vLLM patch](https://github.com/vllm-project/vllm/compare/main...Zyphra:vllm:zaya1-pr) pushed 2 days ago.

This implementation works well with full precision, 8 bit quantization and 4 bit quantization on a Macbook Air M4 24GB.

Stats:
- 8 bit quant: ~20t/s generated @ 9.1 GB VRAM (weights) + KV cache
- 4 bit quant: ~30t/s generated @ 4.99 GB VRAM (weights) + KV cache

I covered testing with manual tests of `mlx_lm.benchmark`, `mlx_lm.convert` and `mlx_lm.server` as well.

Novel features added:
- Compressed Convolutional Attention (CCA)
- Residual Scaling
- Odd/Even Layers and Router1 technique
- Quantized expert layer switching (SwiGLU)
...as described in the [technical report](https://arxiv.org/abs/2605.05365).

The converted/quantized models are available on my HF account: https://huggingface.co/kyr0
e.g. https://huggingface.co/kyr0/zaya1-base-8b-MLX 

https://huggingface.co/kyr0/zaya1-base-8b-8bit-MLX
https://huggingface.co/kyr0/zaya1-base-8b-4bit-MLX

Enjoy!

_Transparency: I'm **not** affiliated with [Zyphra](https://huggingface.co/Zyphra)_

---

## Repro for convert/quants:

```bash
mlx_lm.convert \
  --hf-path Zyphra/ZAYA1-8B \
  --mlx-path "./zaya1-base-8b-MLX" \
  --dtype bfloat16
```

## Quantization

Tested with 8 and 4 bits, group size 64. Lower quants lead to garbage results.

A quick test with AWQ quantization support led to OOM in all cases. Dynamic quant as well. I'm too GPU poor on my Mac machine guys... Does anyone have a Mac Pro?

```
mlx_lm.convert \
    --hf-path Zyphra/ZAYA1-8B \
    --mlx-path "./zaya1-base-8b-8bit-MLX" \
    --dtype bfloat16 \
    -q --q-bits 8 --q-group-size 64
```

## Server + Test

```bash
mlx_lm.server \
  --model "./zaya1-base-8b-8bit-MLX" \
  --host 127.0.0.1 \
  --port 8080 \
  --temp 0.0 \
  --top-p 1.0 \
  --max-tokens 8192 \
  --prefill-step-size 512 \
  --prompt-cache-size 0
```

```bash
curl http://127.0.0.1:8080/v1/chat/completions \
  -d '{
    "model": "./zaya1-base-8b-8bit-MLX",
    "messages": [{"role": "user", "content": "Solve x+2=7. Answer only."}],
    "temperature": 0,
    "max_tokens": 1024
  }'
  ```

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -721,6 +721,9 @@ class ArraysCache(_BaseCache):
         return cache
 
     def empty(self):
+        # Models can create ArraysCache(size=0) for non-attention layers, producing an empty self.cache list
+        if len(self.cache) == 0:
+            return True
         return self.cache[0] is None
 
     @property
--- a/mlx_lm/models/switch_layers.py
+++ b/mlx_lm/models/switch_layers.py
@@ -198,6 +198,13 @@ class SwitchGLU(nn.Module):
 
         return x.squeeze(-2)
 
+    # quantized experts with SwiGLU; required for quantized ZAYA compatibility
+    def to_quantized(self, group_size: int = 64, bits: int = 4, mode: str = "affine"):
+        """Quantize the SwitchLinear sublayers."""
+        self.gate_proj = self.gate_proj.to_quantized(group_size, bits, mode)
+        self.up_proj = self.up_proj.to_quantized(group_size, bits, mode)
+        self.down_proj = self.down_proj.to_quantized(group_size, bits, mode)
+
 
 class SwitchMLP(nn.Module):
     def __init__(
--- a/mlx_lm/models/zaya.py
+++ b/mlx_lm/models/zaya.py
@@ -1,0 +1,705 @@
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .cache import ArraysCache, CacheList, KVCache
+from .switch_layers import SwitchGLU
+from .rope_utils import initialize_rope
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str = "zaya"
+
+    # ZAYA config names
+    cca: bool = True
+    num_query_groups: int = 2
+    use_cache: bool = True
+    attention_bias: bool = False
+    lm_head_bias: bool = False
+    vocab_size: int = 262272
+    hidden_size: int = 2048
+    ffn_hidden_size: int = 4096
+    num_hidden_layers: int = 80
+    num_experts: int = 16
+    num_attention_heads: int = 8
+    head_dim: int = 128
+    activation_func: str = "swiglu"
+    max_position_embeddings: int = 131072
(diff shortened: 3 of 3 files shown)


beamivalice:
I'm testing this model with your 8 bits version in HF. It's good despite some warning below.

Prompt: 32 tokens, 91.910 tokens-per-sec
Generation: 1024 tokens, 30.248 tokens-per-sec
Peak memory: 9.628 GB
M1 Max 32 cores

>>[transformers] You are using a model of type `zaya` to instantiate a model of type ``. This may be expected if you are loading a checkpoint that shares a subset of the architecture (e.g., loading a `sam2_video` checkpoint into `Sam2Model`), but is otherwise not supported and can yield errors. Please verify that the checkpoint is compatible with the model you are instantiating.

kyr0:
@beamivalice Right? I like the quality as well. They also have released a [78B preview model](https://huggingface.co/Zyphra/ZAYA1-74B-preview) the other day. And also a [8B VL model](https://huggingface.co/Zyphra/ZAYA1-VL-8B) which @Blaizzy implemented support for yesterday in this PR https://github.com/Blaizzy/mlx-vlm/pull/1159

I don't have enough VRAM - maybe someone with a Mac Pro could convert/quanitze the big model based on this PR. The `mlx.convert` script is all one needs. It should work without any issues as to my understanding.

Now that we have a second implementation from @Blaizzy - I think it would make sense to discuss the pros and cons of both and agree on the best version. I think I shaped this implementation well for this repo... and it seems to do exactly what it's supposed to do. But if there's anything I can do here to help, I'm all ears :) Maybe @0k1 could you take a look please if my work here is fine?

kyr0:
One idea that comes to mind is, wether we want to implement kwargs based pass-thru of a new parameter to control CCA and MoE details. The strength of this architecture is that you can get much better quality out of the model by investing more compute time. Having the compute time factor controllable per-request would be really nice. We might have `reasoning_effort` map to pre-defined factor values as well, but full control would be pretty cool I guess.

beamivalice:
Actually the extra compute theme (RSA stuff) is [universal](https://arxiv.org/pdf/2509.26626), not limited to this zaya. If we can make other model leverage the available compute to enhance quality then we get quality upgrade. Or we can use that to do the mtp/spec things to get speed upgrade.

kyr0:
@beamivalice Good lord, thanks for the hint! I'll look into it.

kyr0:
@beamivalice I've implemented RSA - but not as an addition to mlx-lm but as a transparent OpenAI Compatible Reverse Proxy - so that it works with any OpenAI Compatible Inference Server. Would you like to help me test it?

beamivalice:
@kyr0 Sure, I'll help.

MattBash17:
Any update on this PR? Would be interested in testing the model, it's quite unique in terms of architecture and parameters count.

PR #1274, opened 2026-05-14
feat: add --idle-timeout to unload model after inactivity

### LINKED ISSUE #1235: Feature request: idle-unload-timeout
`mlx_lm.server` keeps the model in memory for the lifetime of the process. There's no way to have it auto-unload after n minutes of inactivity. For 27B-class models on Apple Silicon this means ~14 GB of unified memory is permanently consumed even when the server is idle. 

Numbers from my setup: M3 Ultra 96 GB, Qwen 3.6-27B Q4 serving via `mlx_lm.server` consumes ~14.5 GB RSS continuously. With no inflight requests for hours, that memory remains allocated. The equivalent Ollama setup uses ~380 MB at idle (model unloaded after default 5min of inactivity) and reloads on first request.

nish2292:
## Summary

Adds `--idle-timeout <seconds>` to `mlx_lm.server` that automatically unloads model weights from memory after N seconds without requests. The model reloads transparently on the next request.

Similar to Ollama's idle unload (Ollama defaults to 5min). A 27B model drops from ~14.5 GB to near-zero at idle. `--idle-timeout 0` (the default) preserves current behavior where the model stays loaded indefinitely. Suggested value when enabled: `--idle-timeout 300`.

Closes #1235

## Design

`ModelProvider.unload()` clears model/tokenizer/draft_model, calls `mx.metal.clear_cache()` and `gc.collect()`. The `model_key` is preserved after unload so `load()` knows what to reload. The `load()` condition was changed from `model_key != key` to `model_key != key or self.model is None` to handle reload-after-unload.

The idle check runs in the generation thread (same thread that accesses model state) so there are no concurrency concerns. It is guarded by: timeout > 0, model loaded, not distributed, no pending requests, no active batch.

## Usage

```bash
mlx_lm.server --model mlx-community/Mistral-7B-Instruct-v0.3-4bit --idle-timeout 300
```

## Test plan

- [x] Unit tests for:
    - [x] `unload()` clears state and preserves model_key
    - [x] reload-after-unload triggers `_load()`
    - [x] `_last_request_time` tracking
    - [x] idle condition logic (timeout exceeded, within timeout, disabled, model already unloaded)
- [x] `DummyModelProvider` updated with new fields
- [x] Pre-commit (black + isort) passes
- [x] Existing server tests unaffected

### DIFF
--- a/mlx_lm/SERVER.md
+++ b/mlx_lm/SERVER.md
@@ -30,6 +30,20 @@ To see a full list of options run:
 mlx_lm.server --help
 ```
 
+### Idle Model Unloading
+
+By default, the server keeps model weights in memory indefinitely. To
+automatically unload the model after a period of inactivity, use
+`--idle-timeout`:
+
+```shell
+mlx_lm.server --model <path> --idle-timeout 300
+```
+
+This unloads model weights after 300 seconds (5 minutes) without requests,
+freeing GPU memory. The model reloads transparently on the next request.
+Set to `0` to disable (default).
+
 You can make a request to the model by running:
 
 ```shell
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1,6 +1,7 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import gc
 import json
 import logging
 import pickle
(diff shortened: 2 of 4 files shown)


zcbenz:

PR #1275, opened 2026-05-14
fix: make generation_stream per-thread to fix server crash on worker threads

### LINKED ISSUE #1256: mlx_lm.server crashes with 'There is no Stream(gpu, 1) in current thread' on sliding-window models (mlx-lm 0.31.3)
## Summary

`mlx_lm.server` crashes on the first request when serving a model that uses `RotatingKVCache` (sliding-window attention; reproduced with Gemma 4 31B). The fix shipped in v0.31.3 (PR #1090) prevents the `Stream(gpu, 0)` variant but leaves an effectively identical bug for `Stream(gpu, 1)`.

```
RuntimeError: There is no Stream(gpu, 1) in current thread.
  File ".../mlx_lm/generate.py", line 1161, in prompt
    mx.eval([c.state for c in self.prompt_cache])
```

This is the same class of bug as #1181 / PR #1090 / closed PR #1088, just with a different stream id. Direct synchronous `mlx_lm.generate.generate(...)` on the main thread works fine on the same model load — only the server's worker thread path is broken.

## Environment

- mlx-lm: 0.31.3
- mlx: 0.31.2
- Python: 3.13.13 (uv-managed venv)
- macOS: Darwin 25.2.0 (Tahoe), Apple M1 Max 64 GB
- Reproducer model: a Gemma 4 31B dense build (sliding-window attn, RotatingKVCache)
- Counter-example (works fine): a Qwen3.6-27B build that does not use RotatingKVCache. So the trigger appears tied to the cache type, not just the server path.

## Reproduction

Start `mlx_lm.server` (or any wrapper that runs generation on a worker thread) with a sliding-window model and send any chat completion request. The generation thread crashes:

```
Exception in thread Thread-... (_generate):
  File ".../mlx_lm/server.py", in _generate
  File ".../mlx_lm/generate.py", line 1847, in next
    return self._next()
  File ".../mlx_lm/generate.py", line 1833, in _next
    self._prompt_batch.prompt(prompts)
  File ".../mlx_lm/generate.py", line 1161, in prompt
    mx.eval([c.state for c in self.prompt_cache])
RuntimeError: There is no Stream(gpu, 1) in current thread.
```

## Root cause

`mlx_lm/generate.py:226` (unchanged on `main` as of today):

```python
generation_stream = mx.new_thread_local_stream(mx.default_device())
```

PR #1090 swapped `mx.new_stream` for `mx.new_thread_local_stream` but kept the binding at module scope. The stream is therefore created exactly once, on whichever thread first imports `mlx_lm.generate` — in a server, that's the main thread. `BatchGenerator.next()` then runs on a worker thread (`Thread-N (_generate)`) and `mx.stream(generation_stream)` raises because the stream is not registered in that thread.

Why id `(gpu, 1)` instead of `(gpu, 0)`: by the time `mlx_lm.generate` is imported, MLX has already allocated stream id 0 for the device default, so `new_thread_local_stream` returns id 1. The fix in #1090 made `(gpu, 0)` slightly less likely to surface but did not resolve the underlying issue. This is exactly what closed PR #1088 (Thump604, `threading.local()` factory) addressed structurally.

## Failed workaround

We tried monkey-patching `BatchGenerator.next` to lazily replace `self._stream` with `mx.new_thread_local_stream(mx.default_device())` on first worker-thread entry. The crash persists because `RotatingKVCache` arrays carry stream affinity from when they were allocated during `model(tokens, cache=...)` on line 1160 — a `with mx.stream(s)` swap on the next line cannot retroactively move arrays already bound to the unreachable parent stream. A correct fix has to ensure the stream used during cache allocation is one valid on the worker thread.

## Suggested fix

Re-apply the substance of closed PR #1088: make `generation_stream` a per-thread accessor (via `threading.local()`), and update the 6 call sites in `generate.py` plus `BatchGenerator.__init__` to call it as a function. That removes the import-thread-binding hazard regardless of which stream id MLX hands out.

## Related

- #1181 / #1179 — original `(gpu, 0)` reports (closed)
- PR #1090 — partial fix that shipped in v0.31.3
- Closed PR #1088 — `threading.local()`-backed factory (the structural fix)
- Closed PR #1182 — `BatchGenerator`-local stream alternative

nish2292:
## Summary

Fixes #1256 - `mlx_lm.server` crashes with `RuntimeError: There is no Stream(gpu, N) in current thread` when serving models from the generation thread. Affects all models via the server, especially sliding-window models (`RotatingKVCache`).

## Why #1090 is insufficient

PR #1090 replaced `mx.new_stream()` with `mx.new_thread_local_stream()`, fixing the `Stream(gpu, 0)` variant but leaving an identical bug for `Stream(gpu, 1)`:

- `ThreadLocalStream` resolution registers the `CommandEncoder` only on the thread that triggers it
- `mx.eval()` can encounter ops tagged with stream indices whose encoder doesn't exist on the current thread
- The server's `_generate()` creates a separate local stream via `mx.default_stream()`, mismatching the module-level stream used by `generate_step()` and `wired_limit()`

Both the issue reporter and a community commenter confirmed this and recommended re-applying #1088's `threading.local()` approach.

## Fix

`generation_stream` becomes a `threading.local()`-backed factory function. Each thread gets its own `mx.new_stream()` with a locally registered `CommandEncoder`:

- **`generate.py`**: 6 call sites updated to `generation_stream()`
- **`server.py`**: Uses `generation_stream()` instead of local `mx.default_stream()` override
- **`BatchGenerator.close()`**: Guards `mx.synchronize()` against cross-thread `__del__` calls

`generation_stream` was never public API. External code that imported it as a variable would need to add `()`.

## Test plan

- [x] `test_generation_stream_per_thread` - each thread gets its own Stream
- [x] `test_batch_sliding_window_threaded` - BatchGenerator + RotatingKVCache on worker thread (exact crash scenario)
- [x] All existing tests pass (46/46)
- [x] Locally verified: Gemma 3 1B, Mistral 7B, Qwen2.5-Coder, GLM-4.7-Flash; thread pools, speculative decoding, concurrent requests, asyncio.to_thread

## Related

- #1090 - partial fix (ThreadLocalStream, merged v0.31.3)
- #1088 - structural fix (threading.local, closed in favor of #1090)

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -6,6 +6,7 @@ import copy
 import functools
 import json
 import sys
+import threading
 import time
 from collections import deque
 from dataclasses import dataclass
@@ -222,8 +223,22 @@ def setup_arg_parser():
     return parser
 
 
-# A stream on the default device just for generation
-generation_stream = mx.new_thread_local_stream(mx.default_device())
+# Per-thread generation stream. See https://github.com/ml-explore/mlx-lm/issues/1256
+_generation_stream_storage = threading.local()
+
+
+def generation_stream() -> mx.Stream:
+    """Get or create a generation stream for the current thread.
+
+    Uses threading.local() so each thread gets its own mx.new_stream() with a
+    CommandEncoder registered locally, avoiding "There is no Stream(gpu, N) in
+    current thread" errors when generation runs on a non-import thread.
+    """
+    s = getattr(_generation_stream_storage, "stream", None)
+    if s is None:
+        s = mx.new_stream(mx.default_device())
+        _generation_stream_storage.stream = s
+    return s
 
 
 @contextlib.contextmanager
@@ -396,7 +411,7 @@ def generate_step(
     def _step(input_tokens: mx.array, input_embeddings: Optional[mx.array] = None):
         nonlocal tokens
 
-        with mx.stream(generation_stream):
+        with mx.stream(generation_stream()):
             logits = _model_call(
                 input_tokens=input_tokens[None],
                 input_embeddings=(
@@ -421,7 +436,7 @@ def generate_step(
             sampled = sampler(logprobs)
             return sampled, logprobs.squeeze(0)
 
-    with mx.stream(generation_stream):
+    with mx.stream(generation_stream()):
         total_prompt_tokens = (
             len(input_embeddings) if input_embeddings is not None else len(prompt)
         )
@@ -551,7 +566,7 @@ def speculative_generate_step(
         return y, logprobs
 
     def _step(model, cache, y, n_predict=1):
-        with mx.stream(generation_stream):
+        with mx.stream(generation_stream()):
             logits = model(y[None], cache=cache)
             logits = logits[:, -n_predict:, :]
 
@@ -600,7 +615,7 @@ def speculative_generate_step(
             ys.append(y)
         return mx.concatenate(ys)
 
-    with mx.stream(generation_stream):
+    with mx.stream(generation_stream()):
         draft_y = _prefill(draft_model, draft_cache, y)
         y = _prefill(model, model_cache, y)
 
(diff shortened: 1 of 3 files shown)


danilopeixoto:
+1

zcbenz:

PR #1277, opened 2026-05-15
Restrict think-state scan to assistant prefill tail

eilidhmae:
ResponseGenerator._tokenize chose initial_state for the generation
  state machine by scanning the entire tokenized prompt for the model's
  `<think>` / `</think>` tokens. When a user message contained a literal
  `<think>` with no matching `</think>`, the scan flipped initial_state
  to "reasoning" for the whole generation, routing every emitted token
  to message.reasoning and leaving message.content absent.

  The intent of the scan is to detect whether the chat template injected
  a `<think>` in the assistant prefill — a tail-of-prompt question. The
  same function already has a tail-bounded scan (last 11 tokens) below
  this one for segment finalisation, so the unbounded form was the
  outlier. Bound the initial-state scan with start=max(0, len(prompt)-11)
  to match.

  Repro without this fix: send a chat completion to a server running a
  model whose tokenizer has `<think>/</think>` in its vocab (e.g. any
  Qwen3-family checkpoint), with a user message containing the literal
  string `<think>` anywhere. finish_reason is "stop", completion_tokens
  is nonzero, but message.content is absent and message.reasoning carries
  the output. With the fix, output lands in message.content as expected.

  Adds two regression tests in `tests/test_server.py`:
  - `test_tokenize_user_think_does_not_flip_initial_state` (the bug case)
  - `test_tokenize_prefill_think_flips_initial_state` (positive guard so
    a future over-bound doesn't silently break the intended behaviour).

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -565,11 +565,18 @@ class ResponseGenerator:
         # If we are here it means we have a chat request so we need to search
         # for segments for better cache management.
 
-        # Choose the initial state among only reasoning or normal
+        # Choose the initial state among only reasoning or normal.
+        # Restrict the scan to the assistant prefill tail (matches the
+        # tail-bounded scan in the segment-finalisation step below).
+        # Without this bound, a literal "<think>" token anywhere in
+        # user content would falsely flip the state machine into
+        # reasoning mode and route the entire generation to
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1279, opened 2026-05-16
Add timings to server responses

spicyneuron:
Adds a `timings` object to `/v1/chat/completions` and `/v1/completions` responses so clients can measure prompt and generation throughput without instrumenting the server themselves. Useful for benchmarking, comparing models, and surfacing tps in UIs.

The shape mirrors llama.cpp's `timings` field for ecosystem compatibility:

```json
"timings": {
  "prompt_per_second": 412.3,
  "predicted_per_second": 78.9,
  "prompt_n": 128,
  "predicted_n": 64
}
```

### Behavior

Follows the existing `usage` conventions:
- Non-streaming: always included.
- Streaming: included in the final usage chunk when `stream_options.include_usage` is set.

### Implementation notes

- Measured server-side; reusing `GenerationResponse.prompt_tps` doesn't work because batched generation has no per-sequence equivalent.
- Both windows are stamped in the generation worker: prompt = prefill start → first generated token; predicted = first → last generated token. Excludes network I/O and pre-worker queue wait; includes a small amount of batch-tick overhead, so values are approximate.
- `prompt_n` excludes cached tokens. `predicted_per_second` is `0` when fewer than two tokens are generated.
- Includes a minor fix: streaming usage check uses `.get("include_usage")` instead of `["include_usage"]` to avoid `KeyError` when `stream_options` is passed without that key.

### DIFF
--- a/mlx_lm/SERVER.md
+++ b/mlx_lm/SERVER.md
@@ -140,6 +140,22 @@ curl localhost:8080/v1/chat/completions \
     - `completion_tokens`: The number of tokens generated.
     - `total_tokens`: The total number of tokens, i.e. the sum of the above two fields.
 
+- `timings`: A dictionary with server-observed generation-service
+  measurements. Excludes network I/O, response serialization, and
+  client-side wait. Includes overhead inside the generation service
+  (queue hops, batch-tick scheduling), so values are approximate:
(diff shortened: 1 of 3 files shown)


zcbenz:

PR #1283, opened 2026-05-18
Add per-request prompt cache files to server

Quiet-Node-io:
## Summary

Adds optional per-request disk prompt-cache support to the OpenAI-compatible server by accepting `prompt_cache_file` / `prompt-cache-file` in chat and text completion requests.

The server already supports an in-memory prompt cache and the CLI path already supports `--prompt-cache-file`; this change lets server callers use the same disk persistence primitive without switching away from `/v1/chat/completions`.

## Behavior

- `prompt_cache_file` loads a matching disk prompt cache into an isolated request-local `LRUPromptCache` before prefill.
- The server saves a prompt cache file after the first generated token so future matching-prefix requests can reuse the prefix.
- Saved disk caches are trimmed by one prompt token to avoid exact-cache-hit requests entering generation with an empty remaining prompt.
- Cache-file requests are routed through the single-request path, not batching, until batch semantics are defined for per-request disk files.
- `disable_prompt_cache` forces an isolated request-local cache and suppresses disk load/save even if a caller sends a cache file path.
- Requests without these fields keep the existing non-cache server path.

## Validation

- `python -m py_compile mlx_lm/server.py`

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1,6 +1,7 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import copy
 import json
 import logging
 import pickle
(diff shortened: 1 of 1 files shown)


zcbenz:

PR #1284, opened 2026-05-18
[mlx_lm] Expose 'strict' parameter in load() function

zyguy:
**Motivation:** Currently, the high-level `mlx_lm.load()` function does not expose the `strict` parameter. This makes it impossible to pass `strict=False` when loading certain converted models or "abliterated" models that contain harmless orphaned weights.

The underlying `load_model()` function already supports the `strict` argument. This PR simply surfaces it to the main `load()` entrypoint, maintaining `True` as the default to ensure full backward compatibility. This will allow dependent projects like `oMLX` to load a wider variety of fine-tuned or abliterated models without crashing.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -458,6 +458,7 @@ def load(
     lazy: bool = False,
     return_config: bool = False,
     revision: Optional[str] = None,
+    strict: bool = True,
 ) -> Union[
     Tuple[nn.Module, TokenizerWrapper],
     Tuple[nn.Module, TokenizerWrapper, Dict[str, Any]],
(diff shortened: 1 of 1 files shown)


zyguy:
Related to https://github.com/Blaizzy/mlx-vlm/pull/1198

zyguy:
Hi! 👋  
Just a gentle ping on this small PR (+3, -1 lines). It fixes an issue for downstream projects (like oMLX) crashing when loading models with orphaned weights.
Could a maintainer please approve the workflow run so the CI can pass? I'd be happy to address any feedback. Thanks!

zcbenz:

PR #1285, opened 2026-05-18
mtp: native MTP speculative decoding for Qwen3-Next / Qwen3.5-MoE

benjamin-levin:
**WIP / fork validation PR.** Pre-existing `mac_build_and_test` failures unrelated.

## Summary

Adds an opt-in, in-model speculative-decoding path that uses the MTP
(Multi-Token Prediction) head shipped with upstream Qwen3-Next /
Qwen3.5-MoE checkpoints under the `mtp.*` weight prefix. The MTP draft
is one extra decoder layer that shares `embed_tokens` and `lm_head`
with the main model, so memory and compute overhead is small relative
to a full draft model.

Bit-exact under greedy (argmax) decoding.

## Motivation

~1.15-1.25x decode throughput at N=1 vs the autoregressive baseline,
without requiring users to download a separate draft model. Cache-side
rollback avoids the re-forward cost on rejection.

Performance ceiling at depth 1:

```
speedup = (1 + a) / (2 - a),  where a = acceptance rate
a = 0.85 -> 1.32x;  a = 0.95 -> 1.85x
```

The existing `speculative_generate_step` in `mlx_lm/generate.py` covers
the "separate draft model" path; this PR complements it for the
"draft head shipped with the base model" case (Qwen3-Next family today,
extensible to any architecture that ships an MTP-style head).

## Implementation

Three new files, no edits to existing code:

| File | Role |
|---|---|
| `mlx_lm/models/mtp_head.py` | `MTPHead` module (dual RMSNorm + concat-and-project `fc` + one decoder block + final norm) + `load_mtp_head` helper that loads + quantizes the head from a sidecar safetensors (or any file containing the `language_model.mtp.*` keys). |
| `mlx_lm/models/gated_delta_spec.py` | Idempotent patch on `GatedDeltaNet` / `Qwen3NextGatedDeltaNet`. When a cache is flagged `_spec_capture=True` and the input has `S == 2`, the patched `__call__` runs the GDN block twice with `S == 1` each, snapshotting cache state between the two calls so rejection can roll back the recurrent state without a re-forward. No-op otherwise (untagged caches and any `S != 2` go through the original path). |
| `mlx_lm/mtp_generate.py` | `mtp_generate_step(prompt, model, mtp_head, ...)` generator yielding `(token, logprobs, from_draft)` tuples -- same shape as the existing `speculative_generate_step` -- and a `mtp_generate(...)` convenience wrapper returning `(text, stats)`. |

### Opt-in mechanism

This PR adds no CLI flag, no server flag, and no changes to any model's
`sanitize()`. Callers explicitly:

```python
from mlx_lm import load
from mlx_lm.models.mtp_head import load_mtp_head
from mlx_lm.mtp_generate import mtp_generate

model, tokenizer = load("mlx-community/Qwen3.6-35B-A3B-4bit")
head, _ = load_mtp_head("path/to/model-mtp.safetensors", config)
text, stats = mtp_generate(model, head, tokenizer, prompt, max_tokens=128)
```

The GDN patch installs on first invocation of `mtp_generate_step` (or
explicit `install()` call) and is bypassed for any cache without the
`_spec_capture` flag, so it is safe for the patch to be live in the
process even when other code paths use the model.

## Scope / remaining

Intentionally scoped down for review-ability:

* **Depth-1 only.** The depth-N spec loop (re-forward on partial accept)
  exists in the companion repo at `mlx_fast/mtp/speculative.py:spec_generate_dN`
  but is not included here. The companion-repo benchmark found depth-2
  on Qwen3.6 ties depth-1 best-case and loses by ~13% worst-case at the
  M4 Max workload mix (chained MTP head loses acceptance after the first
  step because the synthetic hidden state is not a true post-attention
  hidden), so depth-1 is the right default to land first.
* **`server.py` integration not included.** Wiring into `BatchGenerator`
  is straightforward but lives outside the minimal correctness surface.
  Reference: `mlx_fast/mtp/server_integration.py` in the companion
  package; that file monkey-patches `GenerationBatch._step` to drain a
  bonus-token queue and route depth-1 cycles. Happy to follow up with a
  PR that depends on this one.
* **No automated tests.** Bit-exactness was validated end-to-end in the
  companion repo; reproducing those tests here would require pulling a
  real MTP-bearing checkpoint into CI. The constructor + helper-function
  level smoke tests succeed (`python -c "from mlx_lm.mtp_generate import mtp_generate_step"`
  and `MTPHead(args)` both clean).
* **`Model.sanitize()` is unchanged.** The existing `qwen3_5.Model.sanitize`
  already drops `mtp.*` keys when loading the base model, which is the
  right behavior -- they live in a separate `MTPHead` instance. Users
  load the MTP weights with `load_mtp_head` against the same safetensors
  file or a pre-extracted sidecar.

## Companion repo

Translated from `mlx_fast/mtp/` in the personal mlx-fast extension
package: `head.py`, `loader.py`, `gdn_patch.py`, and `speculative.py`
(depth-1 path, `spec_generate_d1`). The companion repo includes a
batched depth-N path, server-side `BatchGenerator` patching, and a
weight-extraction CLI that pulls the `mtp.*` tensors out of the upstream
HF shards -- none of which is included in this minimum PR.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: ["main"]
   pull_request:
+  workflow_dispatch:
 
 permissions:
   contents: read
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -14,7 +14,7 @@ concurrency:
 
 jobs:
   check_lint:
-    if: github.repository == 'ml-explore/mlx-lm'
+    if: github.repository == 'ml-explore/mlx-lm' || github.repository == 'benjamin-levin/mlx-lm'
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v5
@@ -24,7 +24,7 @@ jobs:
       - uses: pre-commit/action@v3.0.1
 
   mac_build_and_test:
-    if: github.repository == 'ml-explore/mlx-lm'
+    if: github.repository == 'ml-explore/mlx-lm' || github.repository == 'benjamin-levin/mlx-lm'
     runs-on: [self-hosted, macos]
     needs: check_lint
     steps:
--- a/mlx_lm/models/gated_delta_spec.py
+++ b/mlx_lm/models/gated_delta_spec.py
@@ -1,0 +1,232 @@
+# Copyright © 2026 Apple Inc.
+
+"""GatedDeltaNet rollback support for MTP speculative decoding.
+
+Speculative decoding at draft depth 1 verifies the sequence ``[t_n, d_1]``
+in a single main-model forward. On rejection, the cache must be rolled
+back to the "after t_n" state so the next iteration sees clean state.
+
+``KVCache`` is naturally trimmable (just decrement ``offset``), but the
+``ArraysCache`` used by ``GatedDeltaNet`` holds a recurrent state that was
+advanced through both tokens -- there is no in-place inverse.
+
+This module installs an opt-in patch on
+``mlx_lm.models.qwen3_5.GatedDeltaNet`` (and
+``mlx_lm.models.qwen3_next.Qwen3NextGatedDeltaNet``): when a cache is
+marked with the ``_spec_capture`` flag and the input has ``S == 2``, the
+patched ``__call__`` runs the GDN block twice with ``S == 1`` each,
+snapshotting the intermediate cache state between the two calls. The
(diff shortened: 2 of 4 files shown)


benjamin-levin:
Closing — opened against upstream by mistake. Re-opening as a draft on my fork (benjamin-levin/mlx-lm) for CI validation first. Apologies for the noise.

PR #1288, opened 2026-05-19
BatchGenerator: opt-in prefer_prefill_when_pending scheduler

benjamin-levin:
## Summary

Adds an opt-in `prefer_prefill_when_pending` kwarg to `BatchGenerator` that pauses decode steps while any prefill work is queued or in flight. Default is `False`, so existing behavior is unchanged.

## Motivation

On Apple M-series unified-memory GPUs prefill and decode share a single Metal command engine. With multiple concurrent requests at long context, the existing scheduler interleaves one decode step (~10 ms) with one prefill chunk (~1 s+ at 32k context) per cycle. Each in-flight decoding request is therefore stalled almost the entire time another request is prefilling, dropping per-victim decode to ~0.7-1 tok/s and yielding ~0× aggregate scaling vs. a single request.

Pausing decode until prefill has drained lets the batch decode together at native batched speed once prefill catches up.

## Measured impact

Qwen3.6-35B-A3B-4bit, M4 Max 36GB, three concurrent requests at 32k context each:

| mode | aggregate decode | per-request decode | scaling vs N=1 |
|---|---|---|---|
| `prefer_prefill_when_pending=False` (default) | 86 tok/s | ~0.8 tok/s (victim) | 0× |
| `prefer_prefill_when_pending=True` | 163 tok/s | symmetric | 1.89× |

Independent in-PR re-measurement on a recent fork tip across N×ctx matrix:

| Config | OFF | ON | Speedup |
|---|---|---|---|
| N=2 ctx=8k | 79.8 | 79.6 | 1.00× |
| N=3 ctx=8k | 25.0 (2 victims @ 2.2!) | 64.4 (symmetric) | **2.58×** |
| N=4 ctx=8k | 16.1 (3 victims @ 2-4) | 53.7 (symmetric) | **3.33×** |
| N=2 ctx=32k | 53.4 | 60.6 | 1.13× |
| N=3 ctx=32k | 27.4 (2 victims @ 0.8) | 46.8 (symmetric) | **1.71×** |
| N=1 ctx=1k | 106.4 | 106.1 | -0.3% (no-op) |
| N=1 ctx=4k | 102.6 | 102.2 | -0.4% (no-op) |

The per-request arrays show the exact bug shape clearly: with the flag off, 2-of-3 victims decode at ~0.8 tok/s while one privileged request decodes ~80 tok/s; with the flag on, all requests decode symmetrically.

## API

New kwarg on `BatchGenerator.__init__`:

```python
BatchGenerator(
    model,
    ...,
    prefer_prefill_when_pending: bool = False,
)
```

When `True`, a step that has any queued/in-flight prefill work skips the decode this cycle unless the generation batch is already saturated (`len(generation_batch) >= completion_batch_size`). When `False`, scheduling is identical to today's behavior.

## Trade-off

With the flag enabled, late-arriving requests' TTFT is unchanged but already-decoding requests "freeze" briefly while new prefill runs. For background-agent / chatbot-batch workloads where aggregate throughput dominates, this is the right trade. For single-stream low-latency it is not. Hence: opt-in, default off.

## Backwards compatibility

Default value is `False`, so:
- No existing callers see any behavior change.
- All other `BatchGenerator` semantics, attributes, and method signatures are untouched.

## Tests

Three tests in `tests/test_generate.py`:

- `test_prefer_prefill_when_pending_default_false` — asserts the flag defaults to `False`.
- `test_prefer_prefill_when_pending_accepted_and_stored` — asserts the kwarg is accepted and stored on the instance.
- `test_prefer_prefill_pauses_decode_when_prefill_pending` — constructs the exact scheduler state the flag targets (one in-flight decode below saturation + one queued prefill) and asserts the new branch fires: decode advances by zero tokens that cycle when the flag is on, and advances normally when the flag is off.

## Files changed

- `mlx_lm/generate.py`: +25/-2
- `tests/test_generate.py`: +66

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1508,6 +1508,7 @@ class BatchGenerator:
         prefill_batch_size: int = 8,
         prefill_step_size: int = 2048,
         max_kv_size: Optional[int] = None,
+        prefer_prefill_when_pending: bool = False,
         stream=None,
     ):
         self.model = model
@@ -1519,6 +1520,7 @@ class BatchGenerator:
         self.prefill_batch_size = prefill_batch_size
         self.completion_batch_size = max(completion_batch_size, prefill_batch_size)
         self.max_kv_size = max_kv_size
+        self._prefer_prefill_when_pending = prefer_prefill_when_pending
 
         self._stream = stream or generation_stream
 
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1295, opened 2026-05-21
Log detected tool parser on server model load

### LINKED ISSUE #1293: feat: add tool parser support for Qwen 3.5/3.6 (non-Coder) models
## Feature Description

Qwen 3.5 35B A3B and Qwen 3.6 27B do not emit `tool_calls` when used with `mlx_lm.server` because the tool parser auto-detection (`_infer_tool_parser`) does not match their chat template patterns.

## Current state

- `tool_parsers/qwen3_coder.py` exists but only activates for Qwen3-Coder models (chat template contains `<tool_call>\n<function=`)
- Regular Qwen 3.5/3.6 chat templates do NOT contain these markers
- `_infer_tool_parser()` returns `None` → no tool parsing → models respond with empty content when tools are provided

## Steps to reproduce

```bash
  -d '{"model":"/Users/micky/MLX/qwen-35b-a3b-8bit","messages":[{"role":"user","content":"What is the weather in Milan?"}],"tools":[{"type":"function","function":{"name":"get_weather","description":"Get weather","parameters":{"type":"object","properties":{"city":{"type":"string"}}}}}],"max_tokens":80}'
```

Result: `content: ""` and no `tool_calls` in the response.

## Expected behavior

Qwen 3.5/3.6 models should emit structured `tool_calls` via the existing tool parser infrastructure, or via a dedicated `qwen3` parser.

## Environment

- mlx-lm 0.31.3 (git main, 2026-05-21)
- macOS 26.5, Apple Silicon (M5 Max)
- Models: mlx-community/qwen-35b-a3b-8bit, qwen-27b-bf16

## Related

- #1011 (closed) — tool calling for Qwen3.5-35B-A3B was fixed but may only apply to specific checkpoints
- #711 (closed) — improved reasoning and tool call parsing

## Workaround / request

Either:
1. Add Qwen 3.5/3.6 regular chat template patterns to `_infer_tool_parser()`
2. Allow users to specify `--tool-parser qwen3_coder` (or similar) explicitly via CLI flag
3. Document which models currently support tool calling and how to enable it

robertlangdonn:
## Motivation

When `mlx_lm.server` loads a model, the chat-template-to-tool-parser inference happens silently — there's no signal to the operator about which parser (if any) got bound. If a client sends `tools=[...]` and no parser was detected, the request silently returns `content: ""` with no `tool_calls`, and there's no easy way to know why from the server logs.

#1293 is a recent example: the reporter assumed their Qwen 3.5/3.6 chat template lacked the required markers (it doesn't), filed an issue with an incorrect root cause, and there's currently no observable signal in the server that would have shown them the actual detected parser. This PR doesn't fix #1293 directly — the marker detection works correctly on current `main` for the public mlx-community Qwen 3.5/3.6 variants — but it would have prevented the issue from being filed.

## What changed

- `TokenizerWrapper` gains a `tool_parser_type` attribute exposing the parser name that `tokenizer_utils.load()` already computes via `_infer_tool_parser`. The computed name was previously discarded after the parser module was imported.
- `ModelProvider._load()` logs an `INFO` line after model load: either `Tool calling enabled (parser: <name>)` or a one-line note that no parser was detected for the chat template, so any `tools` field on requests will be ignored. Skipped for models with no chat template at all (e.g. base completion models) to keep the log focused.
- `test_tool_calling` extended to cover the new attribute for both the with-tool-calling and no-tool-calling branches.

Total diff is +17 / -0 across three files.

## Testing

- `python -m unittest tests.test_tokenizers.TestTokenizers.test_tool_calling` passes.
- `pre-commit run --files` passes on the three changed files (black + isort).
- `python -m mlx_lm.server --help` imports and parses cleanly, so the new log site doesn't break startup.
- The other three pre-existing failures in `tests.test_tokenizers` (BPE detokenizer whitespace round-trip) reproduce identically on unmodified `upstream/main` — unrelated to this change.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -362,6 +362,14 @@ class ModelProvider:
         self.draft_model = draft_model
         self.is_batchable = is_batchable
 
+        if tokenizer.has_tool_calling:
+            logging.info(f"Tool calling enabled (parser: {tokenizer.tool_parser_type})")
+        elif tokenizer.has_chat_template:
+            logging.info(
+                "Tool calling not detected for this chat template; "
+                "any `tools` field on requests will be ignored."
+            )
+
     def load_default(self):
         if self._model_map["default_model"] is not None:
             self.load("default_model", None, "default_model")
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -297,6 +297,7 @@ class TokenizerWrapper:
         tool_call_start=None,
         tool_call_end=None,
         tool_parser=None,
+        tool_parser_type=None,
     ):
         self._tokenizer = tokenizer
         self._detokenizer_class = detokenizer_class
@@ -317,6 +318,7 @@ class TokenizerWrapper:
             tokenizer.chat_template is not None or chat_template is not None
         )
         self._tool_parser = tool_parser
+        self._tool_parser_type = tool_parser_type
         self._tool_call_start = tool_call_start
         self._tool_call_end = tool_call_end
         self._tool_call_start_tokens = None
@@ -444,6 +446,10 @@ class TokenizerWrapper:
     @property
     def tool_parser(self):
         return self._tool_parser
+
+    @property
+    def tool_parser_type(self):
+        return self._tool_parser_type
 
     @property
     def detokenizer(self):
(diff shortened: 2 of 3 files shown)


zcbenz:

PR #1297, opened 2026-05-22
Add Prompt Lookup Decoding (ngram-simple) and Rolling-Hash Speculative Memory (ngram-mod)

### LINKED ISSUE #851: n-gram hashing for speculative decoding
Hi, I hope I didn't miss an existing issue.
Does MLX already have a generic n-gram hasher for speculative decoding like https://github.com/ggml-org/llama.cpp/pull/19164 ?
I saw https://github.com/ml-explore/mlx-lm/issues/433 and https://github.com/ml-explore/mlx-lm/issues/833, but they seem to be LongCat specific.
IIUC, the llama.cpp n-gram hashing works for arbitrary models and seems to bring quite a speed up in cases where parts of the response are verbatim copies of previous responses (which can happen quite a lot when iteratively refining code).

mayank2130:
Closes #851

## Summary

Adds Prompt Lookup Decoding (PLD) and rolling-hash speculative decoding to `mlx_lm` via a generalized `DraftStrategy` abstraction.

Instead of generating speculative drafts with a smaller neural model, the new strategies reuse previously observed token trajectories:

- `ngram-simple` performs exact prompt-history lookup
- `ngram-mod` implements a rolling-hash associative memory ported from `llama.cpp` PR #19164

Both strategies preserve output correctness because speculative tokens are only accepted if verified by the target model under the same sampling configuration.

This PR adds:

- `DraftStrategy` interface for pluggable speculative drafters
- `ModelDraftStrategy` for existing neural drafting
- `NgramSimpleStrategy` for prompt lookup decoding
- `NgramModStrategy` + `NgramModTable` for rolling-hash speculative memory
- optional adaptive repetition gating
- CLI, Python, and server integration
- process-shared speculative memory for cross-request reuse

---

## Usage

```python
from mlx_lm import load, stream_generate

model, tokenizer = load("Qwen/Qwen3-8B-MLX-4bit")

    [{"role": "user", "content": "Write a Python function `add(a, b)`."}],
    add_generation_prompt=True,
    enable_thinking=False,
)

for response in stream_generate(
    model,
    tokenizer,
    prompt,
    max_tokens=256,
    sampler=make_sampler(temp=0.0),
    draft_type="ngram-simple",      # or "ngram-mod"
    num_draft_tokens=4,
    ngram_size=3,                   # use 16 for ngram-mod
    disable_adaptive_gate=True,
):
    print(response.text, end="", flush=True)
```

For `ngram-mod`, reuse a table across related generations to preserve learned n-gram memory:

```python
from mlx_lm import load, stream_generate
from mlx_lm.generate import NgramModTable

model, tokenizer = load("Qwen/Qwen3-8B-MLX-4bit")
table = NgramModTable(n=16)

for prompt_text in prompts:
        [{"role": "user", "content": prompt_text}],
        add_generation_prompt=True,
        enable_thinking=False,
    )

    for response in stream_generate(
        model,
        tokenizer,
        prompt,
        max_tokens=256,
        sampler=make_sampler(temp=0.0),
        draft_type="ngram-mod",
        num_draft_tokens=6,
        ngram_size=16,
        ngram_mod_table=table,
        disable_adaptive_gate=True,
    ):
        print(response.text, end="", flush=True)
```

### CLI: multi-turn ngram-simple

```bash
printf '%s\n%s\n%s\nq\n' \
'Write a Python function summarize_orders(orders) where each order has id, customer, total, and status. Return only the code.' \
'Now add a currency="$" parameter and use it when formatting money values. Keep the cancelled-order behavior unchanged. Return the full updated function only.' \
'Update summarize_orders so it skips orders whose status is cancelled. Keep the same structure and return the full updated function only.' \
| python -m mlx_lm chat \
  --model Qwen/Qwen3-8B-MLX-4bit \
  --max-tokens 500 \
  --temp 0 \
  --chat-template-config '{"enable_thinking": false}' \
  --draft-type ngram-simple \
  --num-draft-tokens 4 \
  --ngram-size 3 \
  --disable-adaptive-gate
```

### CLI: multi-turn ngram-mod

```bash
printf '%s\n%s\n%s\nq\n' \
'Write a Python function summarize_orders(orders) where each order has id, customer, total, and status. Return only the code.' \
'Now add a currency="$" parameter and use it when formatting money values. Keep the cancelled-order behavior unchanged. Return the full updated function only.' \
'Update summarize_orders so it skips orders whose status is cancelled. Keep the same structure and return the full updated function only.' \
| python -m mlx_lm chat \
  --model Qwen/Qwen3-8B-MLX-4bit \
  --max-tokens 500 \
  --temp 0 \
  --chat-template-config '{"enable_thinking": false}' \
  --draft-type ngram-mod \
  --num-draft-tokens 6 \
  --ngram-size 16 \
  --disable-adaptive-gate
```

The chat command keeps the conversation history and prompt cache alive across turns, so T2/T3 can reuse the generated structure from T1.

### Server

Per-request JSON overrides: <code>draft_type</code>, <code>ngram_size</code> <code>disable_adaptive_gate</code>

### Architecture

Speculative drafting is abstracted behind:
```
class DraftStrategy(Protocol):
    def propose(self, y, n_max, ctx) -> mx.array: ...
    def rewind(self, n: int) -> None: ...
    def observe(self, tokens) -> None: ...
    def accept(self, n_accepted: int, n_drafted: int) -> None: ...
```

<code>NgramSimpleStrategy</code> scans backward for matching n-grams and proposes the following continuation tokens directly from prior history.
- Best suited for: short iterative edits, local repetition, single-user coding flows

<code>NgramModStrategy</code> ports llama.cpp's rolling-hash speculative memory.
Architecture mirrors llama.cpp's split between:
- process-global speculative memory
- per-request runtime state

The shared table stores:
<code>hash(ngram) -> next_token</code>
allowing speculative reuse across requests handled by the same running server process.

Implementation behavior intentionally matches llama.cpp:
- fixed-size lossy hash table
- silent overwrite collision policy
- verifier-corrected speculative drafts
- adaptive reset on repeated low acceptance

### Adaptive Gate
An optional adaptive gate computes a 3-gram repetition score over the prompt. If repetition falls below:
<code>NGRAM_GATE_THRESHOLD = 0.02</code>
speculation is skipped automatically.

This is particularly important for ngram-mod, whose cold-start behavior can regress below baseline throughput on low-repetition prompts.

### Benchmarks
All benchmarks used:
- mlx-community/Llama-3.2-3B-Instruct-4bit
- Apple Silicon

### LONG MULTI-TURN EDITING (~280 TOK/TURN) — OVERALL

config | tok/s | acc% | speedup
-- | -- | -- | --
baseline | 54.09 | — | 1.00×
ngram-simple nd=4 | 91.57 | 62.7% | 1.69×
ngram-simple nd=6 | 89.01 | 67.8% | 1.65×
ngram-mod nd=6 | 84.69 | 59.4% | 1.57×
ngram-mod nd=8 | 82.69 | 61.7% | 1.53×

### ngram-mod nd=6 per-turn behavior

turn | prompt | baseline tok/s | ngram-simple nd=6 | ngram-mod nd=6
-- | -- | -- | -- | --
T1 | write EmailValidator class | 57.1 | 62.4 (1.09×) | 54.4 (0.95×)
T2 | add is_disposable method | 57.9 | 118.3 (2.04×) | 123.9 (2.14×)
T3 | reject whitespace in domain | 55.0 | 118.4 (2.15×) | 122.9 (2.24×)
T4 | log inside validate | 49.9 | 118.0 (2.37×) | 135.8 (2.72×)

### DIFF
--- a/mlx_lm/chat.py
+++ b/mlx_lm/chat.py
@@ -1,6 +1,7 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import json
 
 import mlx.core as mx
 
@@ -80,6 +81,34 @@ def setup_arg_parser():
         help="System prompt to be used for the chat template",
     )
     parser.add_argument(
+        "--chat-template-config",
+        help="Additional JSON config for apply_chat_template, e.g. '{\"enable_thinking\": false}'",
+        default=None,
+    )
+    parser.add_argument(
+        "--draft-type",
+        choices=["none", "ngram-simple", "ngram-mod"],
+        default="none",
+        help="Draft strategy for speculative decoding.",
+    )
+    parser.add_argument(
+        "--num-draft-tokens",
+        type=int,
+        default=3,
+        help="Number of draft tokens to propose.",
+    )
+    parser.add_argument(
+        "--ngram-size",
+        type=int,
+        default=None,
+        help="N-gram window size. Defaults to 3 for ngram-simple and 16 for ngram-mod.",
+    )
+    parser.add_argument(
+        "--disable-adaptive-gate",
+        action="store_true",
+        help="Disable the adaptive speculative decoding gate.",
+    )
+    parser.add_argument(
         "--pipeline",
         action="store_true",
         help="Use pipelining instead of tensor parallelism",
@@ -124,24 +153,39 @@ def main():
     rprint(f"[INFO] Starting chat session with {args.model}.")
     print_help()
     prompt_cache = make_prompt_cache(model, args.max_kv_size)
+    template_kwargs = json.loads(args.chat_template_config or "{}")
+    messages = []
+    if args.system_prompt is not None:
+        messages.append({"role": "system", "content": args.system_prompt})
     while True:
         query = input(">> " if rank == 0 else "")
         if query == "q":
             break
         if query == "r":
             prompt_cache = make_prompt_cache(model, args.max_kv_size)
+            messages = []
+            if args.system_prompt is not None:
+                messages.append({"role": "system", "content": args.system_prompt})
             continue
         if query == "h":
             print_help()
             continue
-        messages = []
-        if args.system_prompt is not None:
-            messages.append({"role": "system", "content": args.system_prompt})
         messages.append({"role": "user", "content": query})
         prompt = tokenizer.apply_chat_template(
             messages,
             add_generation_prompt=True,
+            **template_kwargs,
         )
+        generate_kwargs = {}
+        if args.draft_type != "none":
+            generate_kwargs["draft_type"] = args.draft_type
+            generate_kwargs["num_draft_tokens"] = args.num_draft_tokens
+            generate_kwargs["disable_adaptive_gate"] = args.disable_adaptive_gate
+            if args.ngram_size is not None:
+                generate_kwargs["ngram_size"] = args.ngram_size
+        response_text = []
+        accepted = 0
+        last_response = None
         for response in stream_generate(
             model,
             tokenizer,
@@ -157,9 +201,25 @@ def main():
                 ),
             ),
             prompt_cache=prompt_cache,
+            **generate_kwargs,
         ):
             rprint(response.text, flush=True, end="")
+            response_text.append(response.text)
+            accepted += 1 if response.from_draft else 0
+            last_response = response
         rprint()
+        if last_response is not None:
+            generated = last_response.generation_tokens
+            acceptance = 100 * accepted / generated if generated else 0.0
+            rprint(
+                "[stats] "
+                f"prompt={last_response.prompt_tokens} tok "
+                f"generated={generated} tok "
+                f"tok/s={last_response.generation_tps:.2f} "
+                f"accepted={accepted}/{generated} ({acceptance:.1f}%) "
+                f"peak={last_response.peak_memory:.2f} GB"
+            )
+        messages.append({"role": "assistant", "content": "".join(response_text)})
 
 
 if __name__ == "__main__":
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -81,6 +81,22 @@ NGRAM_MOD_LOW_ACCEPT_THRESHOLD = 0.5
 NGRAM_MOD_LOW_ACCEPT_STREAK = 3
 
 
+def _non_trimmable_prompt_cache_types(
+    model: nn.Module, prompt_cache: Optional[Any] = None
+) -> set[str]:
+    if prompt_cache is None:
+        model_cache = cache.make_prompt_cache(model)
+    else:
+        model_cache = prompt_cache[: len(model.layers)]
+    if cache.can_trim_prompt_cache(model_cache):
+        return set()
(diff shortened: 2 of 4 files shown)


mayank2130:
hey @angeloskath can this PLD/n-gram decoding be reviewed. 

If you're not the one to reachout for mlx-lm PRs could you point me to someone else. Thanks.

ashalliants:
Tested this branch on Apple Silicon (M5, 32 GB, macOS 25.2.0, Python 3.13) against Qwen3.5/3.6-family models and hit three issues worth flagging — two bugs and one UX trap that together made the feature look like it was working when it wasn't.

### 1. CLI flags are parsed but never forwarded to `generate()`

In `main()` the new arguments (`--draft-type`, `--ngram-size`, `--disable-adaptive-gate`) are added to the parser, but the `generate(...)` call only forwards `num_draft_tokens`:

```python
response = generate(
    ...
    draft_model=draft_model,
    num_draft_tokens=args.num_draft_tokens,
)   # args.draft_type, args.ngram_size, args.disable_adaptive_gate never passed
```

So `mlx_lm generate --draft-type ngram-simple` silently runs plain decoding. I benchmarked all three draft types via the CLI and got byte-identical speeds (~54.8 tok/s on Qwen3.6-35B-A3B-4bit) before noticing the flags were no-ops. Calling `stream_generate(draft_type=...)` directly works as documented.

### 2. Hybrid-attention models fail mid-generation with an untrimmable-cache error

Models with gated-delta / linear-attention layers (`ArraysCache`) raise once speculation actually engages:

```
ValueError: Speculative decoding requires a trimmable prompt cache (got {'ArraysCache'}).
```

This covers the entire Qwen3.5/3.6 family (`qwen3_5`, `qwen3_5_moe`) — I reproduced on both Qwen3.6-35B-A3B-4bit (MoE) and a Qwen3.6-27B dense build. Given these are popular local models, it might be worth checking cache trimmability up front when the strategy is constructed (in the `draft_type` resolution block) and raising/warning immediately, rather than failing mid-stream after prefill.

### 3. The adaptive gate masks issue 2 — incompatible models appear to "work"

Because the 3-gram repetition gate silently disables speculation when the prompt scores below `NGRAM_GATE_THRESHOLD` (0.02), short or non-repetitive prompts run as plain decode with no indication. On the hybrid-cache models above, a quick smoke test "succeeds" (it never speculates), and the `ValueError` only surfaces later with a repetitive prompt. A one-line `logging.info` when the gate suppresses speculation (and/or a field on `GenerationResponse`) would make both the gate and issue 2 much easier to diagnose.

Happy to provide full repro scripts/timings if useful — everything above was reproduced at commit `pld-ngram-simple` HEAD with greedy sampling.

ashalliants:
Repro scripts and prompts for the findings above, as offered: https://gist.github.com/ashalliants/91819d410f6822e406a314740c8b7d0e — `bench_api.py` reproduces both the ArraysCache failure and the gate-masked false-positive; the README maps each file to each claim.

zcbenz:

PR #1298, opened 2026-05-22
fix: add sanitize method to Granite model for tied embeddings

SahilChachra:
When tie_word_embeddings=True, the HF weights file contains lm_head.weight as an explicit key, but the MLX model class does not create a separate lm_head parameter (it reuses embed_tokens). This causes load_weights to fail with 'Received 1 parameters not in model: lm_head.weight'.

Add a sanitize() method to strip the redundant key before weight loading, consistent with how other models in mlx-lm handle tied embeddings.

### DIFF
--- a/mlx_lm/models/granite.py
+++ b/mlx_lm/models/granite.py
@@ -191,3 +191,8 @@ class Model(nn.Module):
     @property
     def layers(self):
         return self.model.layers
+
+    def sanitize(self, weights):
+        if self.args.tie_word_embeddings:
+            weights.pop("lm_head.weight", None)
+        return weights


nastya236 (approved):
Looks great, thank you.

nastya236:
Actually, do you mind sharing the exact command to reproduce this error?

SahilChachra:
> Actually, do you mind sharing the exact command to reproduce this error?

Sure! Here's a minimal repro.

I hit this on ibm-granite/granite-4.1-8b, which has model_type: granite, tie_word_embeddings: true, but still ships an explicit lm_head.weight in its safetensors (both model.embed_tokens.weight and lm_head.weight are listed in model.safetensors.index.json).

Steps to reproduce on current main:

1) Run either of these:

Option 1 - via convert (quantization)
```python -m mlx_lm convert --hf-path ibm-granite/granite-4.1-8b --mlx-path granite-4.1-8b-4bit -q```

OR

Option 2 : or load-only — no quantization needed to trigger it
```python -m mlx_lm generate --model ibm-granite/granite-4.1-8b --prompt "Hello"```

2) Both fail at weight-loading with:

ValueError: Received 1 parameters not in model: lm_head.weight.

Root cause: in granite.py, Model.__init__ only creates self.lm_head when tie_word_embeddings is False. So in the tied case there's no parameter to receive the lm_head.weight that the HF repo ships, and load_weights(..., strict=True) (via load_model) rejects the extra key. Both convert and generate/load go through load_model(..., strict=True), so either path reproduces it.

The sanitize() in this PR drops the redundant lm_head.weight when tie_word_embeddings=True, matching how the other model classes handle tied embeddings.

Note: any small GraniteForCausalLM checkout with tie_word_embeddings: true that ships lm_head.weight triggers the same path if a lighter-weight repro is preferred — but granite-4.1-8b is the one I actually hit it on.

nastya236:
Would you mind rebasing the PR branch onto the latest main?

SahilChachra:
> Would you mind rebasing the PR branch onto the latest main?

Done!

PR #1299, opened 2026-05-22
generate: cache active samplers/processors in GenerationBatch hot path

erayack:
### Summary

`GenerationBatch._step()` runs once per generated token across the batch. Today it does several pieces of redundant per-step work:

*   `any(self.samplers)` and `any(self.logits_processors)` — $O(N)$ scans on every step
*   `sampler` or `self.fallback_sampler` resolved per slot per step
*   `update_and_fetch(...)` called for every slot whenever any slot has a logits processor, even slots that don't

This PR lifts these into a cached representation that is refreshed only when batch membership/state actually changes.

### Change

`GenerationBatch` now normalizes `samplers` and `logits_processors` at the boundary (`__init__`) so they are always full-length lists with sentinels (`None` / `[]`). It maintains three cached structures:

*   `_active_samplers` — pre-resolved (sampler or fallback) per slot
*   `_sampler_indices` — indices of slots with a non-`None` sampler
*   `_logits_processor_indices` — indices of slots with a non-empty processor list

These are refreshed via `_refresh_active_paths()` at the three points where batch membership/state can change: `__init__`, `extend`, and `filter`.

`_step()` then:

*   replaces `any(self.logits_processors)` with `if self._logits_processor_indices:` ($O(1)$)
*   only calls `TokenBuffer.update_and_fetch(...)` for slots that actually have a processor (instead of all slots whenever any slot has one)
*   replaces `any(self.samplers)` with `if self._sampler_indices:` ($O(1)$)
*   replaces the per-slot `sampler or self.fallback_sampler` lookup with a direct `self._active_samplers[e]` index

### Why this is safe

The invariant the cache relies on is: after `__init__`, `samplers` and `logits_processors` are always lists of length `len(uids)`. This is established at the boundary and refreshed at every lifecycle method that mutates batch membership (`extend`, `filter`). No public API changes; no behavioral changes for valid input.

The simplification in `filter()` (dropping the `if any(...)` guards) is correct because the lists are guaranteed full-length now — the previous guard was only there to handle the "still None" case, which can no longer occur.

The `token_context` list passed to `mx.async_eval(...)` now contains only the active-processor slots rather than all slots. `mx.async_eval` only forces evaluation of the supplied arrays and does not index by position, so this is a strict reduction in async work, not a semantic change.

### Expected benefit

Per generated token, for batches with sparse or no logits processors / samplers:

*   avoids one `any()` scan per call type (2 per step)
*   avoids `len(uids) - active` redundant `TokenBuffer.update_and_fetch(...)` calls
*   avoids `len(uids)` redundant (sampler or fallback) resolutions

Wins scale with batch size and with how sparse samplers/processors are. Largest impact: heterogeneous batches where only some requests carry custom processors.

**Benchmarks:** not run — change is structural (fewer per-token branches and fewer per-token method calls) and preserves valid-input behavior. Happy to add a micro-benchmark if useful; let me know what shape of batch you'd want measured.

### Tests

Adds `test_batch_generate_with_sparse_logits_processors` covering the case where only some slots in a batch carry a logits processor, verifying:

*   the processor is invoked only for the corresponding slot
*   the non-processor slot still receives a response
*   the biased token is produced for the slot with the processor

### Out of scope

The older `BatchGenerator` (around `mlx_lm/generate.py:1025`) carries the same `if not any(self.samplers): self.samplers = [None] * len(self.uids)` shape in its `extend`/`filter`. I left it untouched here to keep the diff focused on the per-token `_step()` hot path, but it's a natural follow-up if maintainers want the same treatment applied for consistency.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1265,16 +1265,21 @@ class GenerationBatch:
         self.prompt_cache = prompt_cache
         self.tokens = tokens
 
-        self.samplers = samplers
+        self.samplers = samplers if samplers else [None] * len(self.uids)
         self.fallback_sampler = fallback_sampler
-        self.logits_processors = logits_processors
+        self.logits_processors = (
+            [processors or [] for processors in logits_processors]
+            if logits_processors
+            else [[] for _ in self.uids]
+        )
         self.state_machines = state_machines
         self.max_tokens = max_tokens
 
-        if self.samplers and len(self.samplers) != len(self.uids):
+        if len(self.samplers) != len(self.uids):
             raise ValueError("Insufficient number of samplers provided")
-        if self.logits_processors and len(self.logits_processors) != len(self.uids):
+        if len(self.logits_processors) != len(self.uids):
             raise ValueError("Insufficient number of logits_processors provided")
+        self._refresh_active_paths()
 
         self._current_tokens = None
         self._current_logprobs = []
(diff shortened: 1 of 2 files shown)


erayack:
Just checking in on this one. It has been about three weeks without a review, and it seems like the CI hasn't kicked off yet—might need a maintainer to approve the workflow run. 

I am happy to rebase if things have gone stale. Just as a reminder, this change focuses on the per-token hot path in GenerationBatch._step() and includes a regression test for sparse logits processors.

PR #1302, opened 2026-05-22
Add Gemma4 shared-KV sanitize regression

rafaelescrich:
## Summary

- add a focused regression test for Gemma4 checkpoints that include redundant K/V tensors for KV-shared layers
- verify non-shared layer K/V weights are preserved while shared-layer `k_proj`, `v_proj`, and `k_norm` weights are dropped by `sanitize()`

## Context

The runtime fix for this behavior already exists on `main`. I hit the original failure while converting a Gemma4 HF export where shared-KV layers still contained redundant K/V tensors. The conversion fails without the sanitize behavior covered here.

## Test

```bash
python -m unittest \
  tests.test_models.TestModels.test_gemma4_kv_shared_layers_omit_kv_projections \
  tests.test_models.TestModels.test_gemma4_sanitize_drops_redundant_shared_layer_kv_weights
```

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1619,6 +1619,62 @@ class TestModels(unittest.TestCase):
             layer_idx = int(k.split("layers.")[1].split(".")[0])
             self.assertLess(layer_idx, 6)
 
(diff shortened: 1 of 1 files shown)


zcbenz:

PR #1305, opened 2026-05-23
Fix `_rstrip_until` ValueError when `until` list is empty

hadoobi:
## Problem

Some lm-eval tasks pass an empty `until` list to `generate_until`
(notably **IFEval**, which lets the model generate freely up to
`max_tokens`). `_rstrip_until` then calls `min()` on an empty sequence
and raises `ValueError`, terminating the whole evaluation:

```python
>>> from mlx_lm.evaluate import _rstrip_until
>>> _rstrip_until("hello", [])
ValueError: min() arg is an empty sequence
```

## Fix

One line: short-circuit and return the input unchanged when there are
no stop strings.

## Tests

Four new cases in `tests/test_evaluate.py`:
- empty `untils` → input returned unchanged (regression)
- single match → truncates at the stop
- multiple stops → truncates at the earliest
- no match → input returned unchanged

## Verification

Ran `lm_eval.simple_evaluate(model=MLXLM(...), tasks=['ifeval'], limit=20)`
on `mlx-community/SmolLM-135M-4bit`. Fails on `main`; with this patch
it returns the normal IFEval metrics (`prompt_level_strict_acc`,
`inst_level_strict_acc`, `prompt_level_loose_acc`,
`inst_level_loose_acc`).

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -32,6 +32,8 @@ DEFAULT_MAX_TOKENS = 8192
 
 def _rstrip_until(s, untils):
     """Limit a string <s> to the first occurrence of any substring in untils."""
+    if not untils:
+        return s
     l = len(s)
     f = [s.find(u) for u in untils]
     f = [l if x < 0 else x for x in f]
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1306, opened 2026-05-24
feat(models): warn when MTP weights are discarded at load

### LINKED ISSUE #1292: mlx_lm.server returns 1-token completions on Qwen3.6 MTP variants when system prompt is reused with a different user prompt
### Summary

Running `mlx_lm.server` against either of the Unsloth-published Qwen3.6 MTP
variants causes completions to truncate to 1-72 tokens (instead of the
requested `max_tokens=400`) on the second and subsequent requests where
the `system` message is identical but the `user` message changes. The
first request (cold) and warm repeats of the same `(system, user)` pair
complete normally; only the cross-`user` pattern triggers truncation.

The response remains valid OpenAI-compatible JSON. `usage.completion_tokens`
truthfully reports the truncated count, so the bug isn't a parser error —
the engine genuinely stops generating early.

We see the same pattern on Apple Silicon M4 Pro 64 GB and M5 Max 128 GB,
so it isn't hardware-specific.

### Affected versions

- `mlx-lm 0.31.3` (latest at time of writing — pip)
- `mlx 0.31.2`
- Apple Silicon M4 Pro (macOS 14.x), M5 Max (macOS 26.x)
- Python 3.12

### Affected models

- `samwang0041/Qwen3.6-27B-MLX-4bit-MTP`
- `trevon/Qwen3.6-35B-A3B-MLX-MTP-4bit`

The non-MTP MLX variants of the same Qwen3.6 27B Dense / 35B-A3B models
on the same mlx-lm version do **not** exhibit the bug — completions
return the full requested token count regardless of the prefix pattern.
So the issue is scoped to MTP-enabled variants.

### Minimal reproduction

```bash
pip install mlx-lm==0.31.3
mkdir -p /tmp/repro
huggingface-cli download samwang0041/Qwen3.6-27B-MLX-4bit-MTP --local-dir /tmp/repro/Qwen3.6-27B-MLX-4bit-MTP
python -m mlx_lm.server \
    --model /tmp/repro/Qwen3.6-27B-MLX-4bit-MTP \
    --host 127.0.0.1 --port 8080 \
    --temp 0.0 &
```

```python
# repro.py
import json
import urllib.request

BASE = "http://127.0.0.1:8080"
MODEL = "/tmp/repro/Qwen3.6-27B-MLX-4bit-MTP"

SYS_A = "You are a precise assistant. " * 200          # ~1600 chars
USER_X = "Explain X in detail. " * 100                  # ~2000 chars
USER_Y = "Explain Y in detail. " * 100                  # ~2000 chars

def call(sys_msg, user_msg):
    payload = {
        "model": MODEL,
        "messages": [
            {"role": "system", "content": sys_msg},
            {"role": "user", "content": user_msg},
        ],
        "max_tokens": 400,
        "temperature": 0.0,
        "stream": False,
    }
    req = urllib.request.Request(
        f"{BASE}/v1/chat/completions",
        data=json.dumps(payload).encode(),
        headers={"Content-Type": "application/json"},
    )
    with urllib.request.urlopen(req) as resp:
        return json.loads(resp.read())

# Phase 1: cold — completes
r1 = call(SYS_A, USER_X)
print("cold        ", r1["usage"]["completion_tokens"], "tokens")

# Phase 2: warm — same (sys, user) — completes
r2 = call(SYS_A, USER_X)
print("warm        ", r2["usage"]["completion_tokens"], "tokens")

# Phase 3: same sys, NEW user — TRUNCATES
r3 = call(SYS_A, USER_Y)
print("prefix-test ", r3["usage"]["completion_tokens"], "tokens  <-- bug")

# Phase 4: same (sys, user) as r1 — completes
r4 = call(SYS_A, USER_X)
print("prefix-back ", r4["usage"]["completion_tokens"], "tokens")
```

### Observed output

```
cold         ~400 tokens
warm         ~400 tokens
prefix-test  1-9 tokens   <-- bug, expected ~400
prefix-back  ~400 tokens
```

### Expected output

All four phases should return ~400 tokens (the requested `max_tokens`).

### Why I think this is the EOS-speculation path

The MTP variant runs with a built-in speculative decoder. The bug only
appears on the phase where the cached prefix from a previous turn
partially matches (system identical, user message differs). It does not
appear on cold start, on full cache hits, or on cache misses where the
system itself differs.

That pattern is exactly the seam where the verifier compares
draft-emitted tokens against the new user prompt. If the verifier
accepts a speculatively-drafted EOS token without re-checking that the
prompt context actually justifies termination, the engine returns 1 to
a handful of tokens before stopping — matching what we see in the wild.

### Larger-scale repro / corpus

I have eight anonymized result JSONs from an 8-phase benchmark (cold,
warm, prefix-test ×3, cold-prefix, long-context, long-prefix) covering
both MTP MLX models on M4 Pro and M5 Max:
https://github.com/druide67/asiai/tree/main/tests/fixtures/agentic

Files `m4-mlxlm-*-mtp.json` and `m5-mlxlm-*-mtp.json` each show truncation
on `prefix-test-1` and `prefix-test-3` (the two phases that exercise the
"same system, different user" pattern). Files `m4-llamacpp-*-mtp.json`
and `m5-llamacpp-*-mtp.json` use the same prompts against llama.cpp with
`--spec-type draft-mtp` on the GGUF MTP variants and complete the full
400 tokens, so the bug isn't an artifact of the prompt corpus.

### Workaround used

We tell users to run the non-MTP MLX variant of the same model for any
multi-turn workload until this is fixed. The non-MTP variant doesn't
have the issue.

### Environment

```
mlx-lm                0.31.3
mlx                   0.31.2
python                3.12.13
macOS                 14.6.1 (M4 Pro) / 26.4.1 (M5 Max)
```

### Happy to help

Let me know if a minimal trace of the draft acceptance loop on a
truncated phase would help — I can re-run with debug logging if mlx-lm
exposes one.

kru2710shna:
## Test plan

- [x] `python -m unittest tests.test_models` — 79 tests pass, 1 pre-existing skip
- [x] `pre-commit run --all-files` clean
- [x] No behavioral change for non-MTP checkpoints (the existing
      `has_mtp_weights` short-circuits the count when no MTP weights present)
- [x] Warning correctly attributed to call site via `stacklevel=2`
      (verified during model tests: surfaces as `qwen3_5.py:317`, not
      `base.py`)
      
 fixes: #1292

### DIFF
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -1,6 +1,7 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import inspect
+import warnings
 from dataclasses import dataclass
 from typing import Any, Optional
 
@@ -135,3 +136,35 @@ def scaled_dot_product_attention(
             mask=mask,
             sinks=sinks,
         )
+
+def warn_discarded_weights(
+    n_discarded: int,
+    feature_name: str = "MTP",
+    model_name: Optional[str] = None,
+) -> None:
+    """Emit a one-time warning when feature-specific weights are dropped at loading time.
+
+    Some model checkpoints (e.g., Multi-Token Prediction variants) ship with
+    additional weight tensors that are discarded during loading because mlx-lm
+    does not currently use them at runtime. This warning surfaces the
+    discrepancy at load time so downstream benchmarks can interpret results
+    correctly.
+
+    Args:
+        n_discarded: Number of weight tensors filtered out by the caller.
+          The warning is only emitted when this is +ive.
+        feature_name: Name of the discarded feature.
+          Defaults to "MTP".
+        model_name: Optional.
+    """
+    if n_discarded <= 0:
+        return
+    suffix = f" for model '{model_name}'" if model_name else ""
+    warnings.warn(
+        f"Loaded {n_discarded} {feature_name} weight tensor(s){suffix}, "
+        f"discarded at runtime — mlx-lm does not currently support "
+        f"{feature_name}. The model will run as a standard autoregressive "
+        f"model.",
+        UserWarning,
+        stacklevel=2,  # warning come from the model file, not from base.py
+    )
--- a/mlx_lm/models/qwen3_5.py
+++ b/mlx_lm/models/qwen3_5.py
@@ -12,6 +12,7 @@ from .base import (
     BaseModelArgs,
     create_attention_mask,
     create_ssm_mask,
+    warn_discarded_weights,
 )
 from .cache import ArraysCache, KVCache
 from .gated_delta import gated_delta_update
@@ -310,7 +311,12 @@ class TextModel(nn.Module):
             "conv1d.weight" in k and v.shape[-1] != 1 for k, v in weights.items()
         )
         should_shift_norm_weights = has_mtp_weights or has_unsanitized_conv1d
+        n_mtp_weights = sum(1 for k in weights if "mtp." in k) if has_mtp_weights else 0
         weights = {k: v for k, v in weights.items() if "mtp." not in k}
+
+        warn_discarded_weights(
+            n_mtp_weights, feature_name="MTP", model_name=self.model_type
+        )
 
         if self.args.tie_word_embeddings:
             weights.pop("lm_head.weight", None)
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -2,13 +2,18 @@
 import copy
 import importlib
 import unittest
+import warnings
 
 import mlx.core as mx
 import mlx.nn as nn
 from mlx.utils import tree_flatten, tree_map
 
 from mlx_lm.models import rope_utils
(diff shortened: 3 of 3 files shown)


kru2710shna:
PR #990  (by @AirRunner) implements native MTP speculative decoding for `qwen3_5_moe` and decouples the norm-weight shift from MTP detection, fixing the regression @druide67 measured in #1292 Once #990 lands and 
users re-convert their checkpoints to preserve MTP weights, the warning emitted by this PR becomes informational for legacy/pre-#990 checkpoints rather than a critical signal. Until then, this warning fills the gap for users on existing MTP-variant checkpoints that were downloaded expecting throughput gains they aren't currently receiving. The follow-up PR extending this pattern to the other nine model files that perform MTP weight filtering remains valuable independently of #990, since those models are not in #990's scope.

zcbenz:

PR #1309, opened 2026-05-24
Adding --kv_bits as parameter.

Wolfbane1:
I have understood that it is not needed a --thinking or --thinking-budget parameter, but, so I only need to add a new parser parameter to let introduce in the server the kv cache quantization (by default 16, but now, we can choice 8 for example).


zcbenz:

PR #1310, opened 2026-05-25
Fix qwen3_coder tool parser crash on unparseable parameter values

### LINKED ISSUE #1236: Bug report: "tool_parsers/qwen3_coder.py:79 crashes on ISO 8601 dates
`mlx_lm/tool_parsers/qwen3_coder.py:79` calls `ast.literal_eval()` on every tool-call parameter value. This crashes on any string that looks like the start of a number but isn't valid Python (most notably ISO 8601 timestamps), causing the server to drop the entire tool call.

Fixed locally by wrapping `ast.literal_eval` with a try/except that falls back to the raw string. Two-line patch:
```
      try:                                                                                                        
          return ast.literal_eval(param_value)                                                                    
      except (ValueError, SyntaxError):                                                                           
          return param_value                
```        
Applies at both call sites in `_convert_param_value`.

seongyun1104:
Fixes the case where a Qwen3-Coder model emits a parameter value that cannot be coerced into the schema-declared type (e.g. ISO 8601 dates, natural-language strings) — `_convert_param_value` raises and the entire tool call is dropped, even though the model output was well-formed XML.

Four branches in `_convert_param_value` were unguarded:
- `object` / `array` via `json.loads` -> `ast.literal_eval`
- fallback `ast.literal_eval` for unrecognised types
- `int()` coercion for integer-family types
- `float()` coercion for `number` / `float` types

Each is now wrapped with `try / except (ValueError, SyntaxError)` and falls back to the raw string, matching the behaviour for unrecognised schema types.

Tests added for all four branches.

The numeric branches were pointed out by @robertlangdonn ([comment](https://github.com/ml-explore/mlx-lm/pull/1310#issuecomment-4582322151)) — folded into this PR since it's the same function, same fallback pattern, and the same #1236 symptom.

Closes #1236

### DIFF
--- a/tests/test_tool_parsing.py
+++ b/tests/test_tool_parsing.py
@@ -197,6 +197,34 @@ class TestToolParsing(unittest.TestCase):
         self.assertEqual(tool_call["arguments"]["filters"], {"category": "books"})
         self.assertEqual(tool_call["arguments"]["tags"], ["fiction", "new"])
 
+    def test_qwen3_coder_unparseable_params(self):
+        # When the schema declares a typed field but the model emits a value
+        # that cannot be coerced into that type (e.g. an ISO 8601 date for an
+        # `integer` field, natural-language text for an `array` field), fall
+        # back to the raw string instead of raising and dropping the tool call.
+        cases = [
+            ("object", "when", "2026-01-15T10:30:00Z", "2026-01-15T10:30:00Z"),
+            ("array", "note", "Alex has written several books.", "Alex has written several books."),
+            ("datetime", "when", "2026-01-15T10:30:00Z", "2026-01-15T10:30:00Z"),
+            ("integer", "day", "2026-01-15", "2026-01-15"),
+            ("number", "ratio", "about three", "about three"),
+        ]
+        for ptype, pname, value, expected in cases:
+            with self.subTest(type=ptype):
+                tools = [{
+                    "type": "function",
+                    "function": {
+                        "name": "log_event",
+                        "parameters": {
+                            "type": "object",
+                            "properties": {pname: {"type": ptype}},
+                        },
+                    },
+                }]
+                xml = f"<function=log_event><parameter={pname}>{value}</parameter></function>"
+                tool_call = qwen3_coder.parse_tool_call(xml, tools)
+                self.assertEqual(tool_call["arguments"][pname], expected)
+
     def test_gemma4(self):
         # Nested object
         test_case = 'call:configure{settings:{enabled:true,name:<|"|>test<|"|>}}'
--- a/mlx_lm/tool_parsers/qwen3_coder.py
+++ b/mlx_lm/tool_parsers/qwen3_coder.py
@@ -54,9 +54,15 @@ def _convert_param_value(param_value: str, param_name: str, param_config: dict) 
         or param_type.startswith("short")
         or param_type.startswith("unsigned")
     ):
-        return int(param_value)
+        try:
+            return int(param_value)
+        except ValueError:
+            return param_value
     elif param_type.startswith("num") or param_type.startswith("float"):
-        float_param_value = float(param_value)
+        try:
+            float_param_value = float(param_value)
+        except ValueError:
+            return param_value
         int_param_value = int(float_param_value)
         return (
             float_param_value
@@ -74,9 +80,15 @@ def _convert_param_value(param_value: str, param_name: str, param_config: dict) 
             try:
                 return json.loads(param_value)
             except json.JSONDecodeError:
-                return ast.literal_eval(param_value)
+                try:
+                    return ast.literal_eval(param_value)
+                except (ValueError, SyntaxError):
+                    return param_value
 
-        return ast.literal_eval(param_value)
+        try:
+            return ast.literal_eval(param_value)
+        except (ValueError, SyntaxError):
+            return param_value
 
 
 def _parse_xml_function_call(function_call_str: str, tools: Optional[Any]):


robertlangdonn:
Nice — wrapping both `literal_eval` sites is the right call here, and the regression tests are clear.

One thing worth folding in while this is open: the same crash class still escapes through the numeric branches of `_convert_param_value`. The `int(param_value)` and `float(param_value)` coercions (the `int*/uint/long/short/unsigned` and `num*/float*` cases) are unguarded, so a model that emits a non-numeric string for a field the schema declares as `integer`/`number` still raises and drops the whole tool call — same user-visible symptom as #1236, just a different line than the `literal_eval` paths this PR hardens.

Minimal repro on current `main` (parser only, no server needed):

```python

tools = [{"function": {"name": "f", "parameters": {"properties": {"x": {"type": "integer"}}}}}]
out = "<function=f><parameter=x>2024-01-15</parameter></function>"
qwen3_coder.parse_tool_call(out, tools)
# ValueError: invalid literal for int() with base 10: '2024-01-15'

# number-typed field, same story:
# ValueError: could not convert string to float: "3 o'clock"
```

Extending the same `try/except ... return param_value` fallback to those two coercions would close the crash class fully rather than just the `literal_eval` branch. Happy to send that as a small follow-up on top of this if useful — didn't want to open a competing PR against #1236.

seongyun1104:
@robertlangdonn Good catch — thanks. Folded the `int()` / `float()` guards into this PR rather than splitting, since it's the same function, same fallback pattern, and the same #1236 symptom (tool call dropped on a value that can't be coerced into the declared type). Tests for both numeric branches added alongside the existing ones.

Pushed as `1c9c8e7`.

tests/test_tool_parsing.py
```diff
@@ -197,6 +197,121 @@ def test_qwen3_coder_single_quoted_params(self):
         self.assertEqual(tool_call["arguments"]["filters"], {"category": "books"})
         self.assertEqual(tool_call["arguments"]["tags"], ["fiction", "new"])
 
+    def test_qwen3_coder_unparseable_object_params(self):
+        # When the schema declares an object/array/dict/list type but the model
+        # emits a value that is neither valid JSON nor a valid Python literal
+        # (e.g. ISO 8601 dates, natural-language strings), fall back to the raw
+        # string instead of crashing. Regression test for issue #1236.
```
nastya236: Lets remove a comment `Regression test for issue #1236.`

tests/test_tool_parsing.py
```diff
@@ -197,6 +197,121 @@ def test_qwen3_coder_single_quoted_params(self):
         self.assertEqual(tool_call["arguments"]["filters"], {"category": "books"})
         self.assertEqual(tool_call["arguments"]["tags"], ["fiction", "new"])
 
+    def test_qwen3_coder_unparseable_object_params(self):
```
nastya236: I would reorganise 3 tests in one function with specified test cases. Something like:
```python
cases = [...]
for ptype, pname, value, expected in cases:
    with self.subTest(type=ptype):
        tools = [{"type": "function", "function": {
            "name": "log_event",
            "parameters": {"type": "object", "properties": {pname: {"type": ptype}}},
        }}]
        xml = f"<function=log_event><parameter={pname}>{value}</parameter></function>"
        tool_call = qwen3_coder.parse_tool_call(xml, tools)
        self.assertEqual(tool_call["arguments"][pname], expected)
```

nastya236 (changes requested):
Looks great! Thanks for your contribution. I left a comment, otherwise looks great.

seongyun1104:
@nastya236 Sorry for the delay — I was busy with work and only just saw the review. Thanks for the review!

Addressed both points in `37d0925`:
- Merged the four `unparseable_*` tests into a single `test_qwen3_coder_unparseable_params` driven by `subTest` cases (object / array / datetime / integer / number — covers all four guarded branches).
- Dropped the `Regression test for issue #1236.` trailers.

nastya236:
Thanks! Could you give me a command to use to reproduce the issue, like:
  ```mlx_lm.server --model mlx-community/Mistral-7B-Instruct-v0.3-4bit

  curl localhost:8080/v1/chat/completions \
    -d '{
       "messages": [{"role": "user", "content": "...."}],
       "temperature": 0.7
     }'
```

so I can test the fix.

**Comment by seongyun1104** [comment]
The bug is entirely in the parser layer (`_convert_param_value` in `mlx_lm/tool_parsers/qwen3_coder.py`) — `ToolCallFormatter.__call__` in `server.py` just calls this function and catches its `ValueError` by dropping the tool call. So a parser-only snippet hits the exact same code path that the server takes:

```python

tools = [{"type": "function", "function": {
    "name": "log_event",
                   "properties": {"day": {"type": "integer"}}}}}]
xml = "<function=log_event><parameter=day>2026-01-15</parameter></function>"
print(qwen3_coder.parse_tool_call(xml, tools))

# Before fix:
#   ValueError: invalid literal for int() with base 10: '2026-01-15'
#   -> ToolCallFormatter logs a warning and drops the whole tool call
# After fix:
#   {'name': 'log_event', 'arguments': {'day': '2026-01-15'}}
#   -> tool call passes through, downstream tool sees the raw string
```

The same shape works for the other three branches (swap the `type` to `"number"`, `"object"`, or any unrecognised type — `"datetime"` etc. — and the `parameter=` value to something that can't be coerced).

If you'd still prefer to see it driven through `mlx_lm.server` end-to-end I can set that up, but it'd need a Qwen3-Coder-style model that emits the XML tool-call format — the smallest mlx-community variant is `Qwen3-Coder-30B-A3B-Instruct-3bit` at 13.4 GB, which doesn't fit my M1 8GB local setup (I'd need to rent a larger Mac for a day). Happy to do that if useful — just let me know which path you'd prefer.

seongyun1104:
Hi @nastya236 — just following up. The parser-only repro snippet in my comment above hits the exact same code path the server takes (`_convert_param_value` → `ValueError` → dropped tool call), so it should reproduce without spinning up a model. Let me know if you'd still prefer an end-to-end `mlx_lm.server` run, or if there's anything else you need from me to move this forward. Thanks!

zcbenz:
Closing in favor of https://github.com/ml-explore/mlx-lm/pull/1417 and #1239

PR #1312, opened 2026-05-26
Expose tunable Metal memory limits in the trainer

### LINKED ISSUE #828: Memory during training (LoRA)
Heya, first thanks for the awesome library! I'm a newb and this is making it relatively easy.

I was playing around with doing some tuning on a model and kept hitting memory walls (memory pressure to the roof, sometimes hard crashes). What was weird is that the memory kept growing, each val or iteration. Saw some other issues or mentions about KV cache with these kind of symptoms, but that was all for generation, and I don't see anything for that in the code on the training side.

Saw this:
https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tuner/trainer.py#L214
And I poked around and found a bunch more limit setters. 

Anyway I ended up hacking this in (below), haven't tuned it really, but I think setting these limits to lower values (or exposing them) would really help people out. Without these I was hitting hard walls.

e.g. with some models and lora settings I was going OOM/hard crashing on 48G ram, seeing python at 36G and memory pressure in the red. 

With the same model and lora settings, if I drop wired_limit a little, like to 90%, everything is stable. Still high memory use, but no reboots.

If I drop cache_limit way down or set to zero, my memory never climbs, no crashes or memory pressure, and I don't see any performance impact in my limited testing. 

Setting memory_limit too low or even zero doesn't cause it to fail, but does run slower, at about 80% speed. 

I can even set all 3 to zero and still skip along at 80% speed. Maybe not ideal for a training server farm, but could be useful if you are trying to use the machine at the same time and consume minimal resources while running some background training? Anyway it was surprising and interesting that it didn't outright fail or crawl.

So around [here](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tuner/trainer.py#L214) at the top of `train()`

```python
    if mx.metal.is_available():
        wired_limit = mx.metal.device_info()["max_recommended_working_set_size"]
        # max_recommended_working_set_size is too large
        wired_limit = int(wired_limit * 0.9)
        print(f"Setting wired limit to {wired_limit}")
        mx.set_wired_limit(wired_limit)

        memory_limit = int(wired_limit * 1.0) #docs say this defaults to 1.5x
        print(f"Setting memory limit to {memory_limit}")
        mx.set_memory_limit(memory_limit)

        cache_limit = int(wired_limit * 0.5) #docs say this defaults to memory_limit
        print(f"Setting cache limit to {cache_limit}")
        mx.set_cache_limit(cache_limit)
```

kru2710shna:
Adds wired_limit_ratio, memory_limit_ratio, and cache_limit_ratio to TrainingArgs and surfaces them as --wired-limit-ratio, --memory-limit-ratio, and --cache-limit-ratio CLI flags in mlx_lm.lora. Defaults preserve existing behavior; users can opt in to lower limits to avoid OOM/hard crashes during training on memory-constrained systems.

Also fixes train_model() to forward args.clear_cache_threshold into TrainingArgs — it was previously a silent no-op from the CLI.

Fixes #828

### DIFF
--- a/mlx_lm/lora.py
+++ b/mlx_lm/lora.py
@@ -70,6 +70,9 @@ CONFIG_DEFAULTS = {
     "grad_checkpoint": False,
     "grad_accumulation_steps": 1,
     "clear_cache_threshold": 0,
+    "wired_limit_ratio": 1.0,
+    "memory_limit_ratio": None,
+    "cache_limit_ratio": None,
     "lr_schedule": None,
     "lora_parameters": {"rank": 8, "dropout": 0.0, "scale": 20.0},
     "mask_prompt": False,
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1313, opened 2026-05-26
Warn when speculative decoding may hurt throughput for MoE models

### LINKED ISSUE #1132: Feature request: warn when speculative decoding is unlikely to help (MoE models)
## Description

Speculative decoding with `--draft-model` can significantly hurt performance on Mixture-of-Experts models where the active parameter count is close to the draft model size. Users may enable it expecting a speedup and get the opposite.

## Benchmark

Target: Qwen3.5-397B-A17B (MoE, 17B active params per token)
Draft: Qwen3.5-9B
`--num-draft-tokens 3`, Mac Studio M4 Ultra 512GB

| Prompt type | Spec ON (tok/s) | Spec OFF (tok/s) | Delta |
|---|---|---|---|
| Technical writing | 19.87 | 31.59 | -37% |
| Python codegen | 20.11 | 36.65 | -45% |
| Creative writing | 20.22 | 36.57 | -45% |
| Analytical (longer ctx) | 26.85 | 35.84 | -25% |
| Rust codegen | 27.29 | 36.42 | -25% |
| **Average** | **22.94** | **35.30** | **-35%** |

Speculative decoding reduced throughput by 35% on average because the target model's active parameter count (17B) is close to the draft model (9B), making the overhead of draft + verify more expensive than just generating directly.

## Proposal

When a MoE model is loaded with `--draft-model`, check if the active parameter count is within some threshold of the draft model size (e.g., < 4x) and log a warning:

```
Warning: Target model active parameters (17B) are close to draft model size (9B). 
Speculative decoding may not improve throughput for MoE architectures. 
Consider benchmarking with and without --draft-model.
```

The model config already contains expert/active parameter metadata, so this check should be straightforward.

## Environment

- mlx-lm version: 0.31.2
- macOS

Shylin26:
Fixes #1132

## Problem
When using `--draft-model` with MoE target models (e.g. Qwen3.5-397B-A17B), 
speculative decoding can hurt throughput by 25-45% because the target model's 
active parameter count is close to the draft model size, making draft+verify 
more expensive than direct generation.

## Solution
Added `_warn_speculative_moe()` that runs after both models are loaded. It:
- Detects MoE architecture via `num_experts` and `num_experts_per_tok` in model config
- Calculates active parameters as `(num_experts_per_tok / num_experts) * total_params`
- Warns if draft model size is within 4x of active parameters

Non-MoE models are completely unaffected.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -40,7 +40,7 @@ from .models.cache import (
 )
 from .sample_utils import make_sampler
 from .tokenizer_utils import TokenizerWrapper
-from .utils import does_model_support_input_embeddings, load
+from .utils import does_model_support_input_embeddings, get_total_parameters, load
 
 DEFAULT_PROMPT = "hello"
 DEFAULT_MAX_TOKENS = 100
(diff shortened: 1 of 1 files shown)


zcbenz:

PR #1023, opened 2026-03-19
[transformers-to-mlx skill] Add OLMo Hybrid model support

pcuenca:
## TL;DR: for reviewers

This PR was created with an in-progress conversion Skill that we'd like to contribute to the community. We are trying to make it comprehensive and careful, so it only considers the job done if relevant tests pass. We have instructed it to pay attention to usual problem areas, like long-context generation (and rope), dtype/upcasting errors, and others.

Any feedback received in this PR will be applied to the Skill so it works better next time! Please, do not hesitate to point out any observations, including, for example:
- Missing tests
- Noisy / misleading comments in the PR description
- Deviations from mlx-lm best practices or idioms

Tests were run on M3 Ultra (512 GB).

---

## Summary

- Add MLX implementation of [OLMo Hybrid](https://huggingface.co/collections/allenai/olmo-hybrid) (GatedDeltaNet + Full Attention) architecture
- Supports all four official checkpoints: `allenai/Olmo-Hybrid-7B`, `allenai/Olmo-Hybrid-Instruct-SFT-7B`, `allenai/Olmo-Hybrid-Instruct-DPO-7B`, and `allenai/Olmo-Hybrid-Think-SFT-7B`
- Reuses the existing `gated_delta.py` Metal kernel infrastructure for the linear attention layers
- Supports conditional RoPE: base/SFT/Think-SFT models use RoPE (`rope_theta=10000`), DPO uses NoPE (no positional embeddings)

## Architecture

OLMo Hybrid uses the same transformer architecture as OLMo 3 7B, except that 75% of layers use GatedDeltaNet linear attention heads instead of standard attention. The layer pattern alternates: 3 linear attention layers followed by 1 full attention layer.

Key implementation details:
- **Linear attention layers** (GatedDeltaNet): separate q/k/v/a/b/g projections, per-projection depthwise conv1d, `allow_neg_eigval` scaling (beta × 2.0), `RMSNormGated` output normalization. Pre-norm residual style.
- **Full attention layers**: standard MHA with `q_norm`/`k_norm` (RMSNorm on full projections before head split), optional RoPE. Post-norm residual style.
- **NoPE mode**: the DPO variant sets `rope_theta: null` in `rope_parameters`, which disables RoPE entirely. This is handled by making RoPE conditional in the `Attention` class.
- **Float32 recurrent state**: GatedDeltaNet state is initialized in float32 for numerical stability, matching the upstream `gated_delta.py` convention.
[...]

PR #1315, opened 2026-05-27
Add olmo hybrid

cmurray1105:
Adds MLX support for [allenai/OLMo-Hybrid-7B](https://huggingface.co/allenai/OLMo-Hybrid-7B) — a 32-layer model combining GatedDeltaNet (GDN) recurrent layers with standard full attention in a 3:1 pattern (24 GDN + 8 Attention).

GDN is a fixed-memory alternative to the KV cache: instead of growing linearly with context length, it maintains a constant-size state matrix S regardless of sequence length. This makes 65k token contexts practical on consumer hardware. 

**Implementation notes:**
- `GatedDeltaNet`: sequential delta-rule recurrence with chunked `mx.eval()` (every 32 tokens) to prevent graph explosion at long contexts
- `ShortConv`: causal depthwise conv1d via `mx.conv1d` with `groups=C`; real prior-token context passed during decode to avoid zero-pad corruption after the first generated token
- Heterogeneous cache: `GDNCache` (fixed-size recurrent state + conv context) for GDN layers, `KVCache` for attention layers — `make_cache()` returns the right type per layer
- Two correctness bugs fixed vs naive port:
  - QK-norm applied to full `(H×D)` projection before reshape — not per-head `(D,)` — weight shape in the checkpoint is `(H*D,)`
  - `rope_theta = 500_000` (OLMo2 large-context default, not `10_000`)

**Tested on M4 Max 36GB:** 16.3 tok/s prefill, 4.9 tok/s decode (float16). Decode speed reflects GDN's sequential recurrence — a natural follow-up is a parallel prefix scan implementation for faster prefill.

### DIFF
--- a/mlx_lm/models/olmo_hybrid.py
+++ b/mlx_lm/models/olmo_hybrid.py
@@ -23,7 +23,7 @@
 import math
 import re
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple
 
 import mlx.core as mx
 import mlx.nn as nn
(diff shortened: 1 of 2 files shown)


nastya236:
Thanks for the contribution. PR #1023 seems to overlap with yours, which adds the same model. Could you take a look and see whether it covers your use case?

michalk8:
superseded by #1023

PR #1320, opened 2026-05-28
Fix Qwen3.5 MTP sanitize norm shift

xxxkkw:
## Summary
- Stop using stripped `mtp.*` weights as a signal to shift Qwen3.5 norm weights by `+1` during sanitize.
- Keep the raw-checkpoint norm shift tied to unsanitized Conv1d layout conversion.
- Cover both `qwen3_5` and `qwen3_5_moe` wrapper paths so already-converted checkpoints with MTP weights can be loaded without shifting norms twice.

## Environment
- OS: macOS Darwin 25.4.0 arm64
- Hardware: Apple M1 Max, 32 GB unified memory
- Python: local project virtual environment

## Testing
- `python -m unittest discover -s tests -p test_models.py -k qwen3_5_family_convert_then_load_norm_not_shift_twice`
  - Result: 1 passed

## Benchmark / profiling notes
- This is a sanitize/load correctness fix and does not change inference kernels.
- The regression test reproduces the load-sanitize path with already-converted norm weights plus stripped MTP weights; before this change the MTP key alone triggered an extra norm shift.

## Context
- Full native MTP decoding is already being discussed in #990; this PR only extracts the small sanitize correctness fix to keep it reviewable independently.

### DIFF
--- a/mlx_lm/models/qwen3_5.py
+++ b/mlx_lm/models/qwen3_5.py
@@ -305,11 +305,9 @@ class TextModel(nn.Module):
         return [ArraysCache(size=2) if l.is_linear else KVCache() for l in self.layers]
 
     def sanitize(self, weights):
-        has_mtp_weights = any("mtp." in k for k in weights)
         has_unsanitized_conv1d = any(
             "conv1d.weight" in k and v.shape[-1] != 1 for k, v in weights.items()
         )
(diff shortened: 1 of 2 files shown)


nastya236:
Thanks for flagging this! This one can be closed when we merge https://github.com/ml-explore/mlx-lm/pull/1198.
Closing this PR for now.

Issue #353, opened 2025-08-05
MLX to GGUF conversion


JoshShailes:
Hi, has anyone had issues with trying to convert mlx to gguf?

My workflow is as follows:

1. Fine tune with QLora (llama 3.2 1B instruct 4bit)
2. Fuse Lora adapters with the f16 version of Llama 3.2 1B instruct (convert to gguf)
3. Test the outputs from the fused model in mlx - all good.
4. Load into msty as gguf, get an error saying that it is missing tensor ‘token_embd.weight’. I think it might be missing all tensors though. Not sure what is causing it

I have also tried updating the version of mlx-lm to the latest version and I get the following error when running the fuse convert to gguf
```ValueError: [save_gguf] can only serialize row-major arrays``` 

Any help would be greatly appreciated, Thanks.

JoshShailes:
Something I have just thought is that I generated the fine-tune adapters with an old version of mlx-lm. I will also re-run the fine-tuning to check it isn't that.

JoshShailes:
> Something I have just thought is that I generated the fine-tune adapters with an old version of mlx-lm. I will also re-run the fine-tuning to check it isn't that.

I have tested this and still get the same ValueError as my initial post.

prabhu:
For GGUF conversion, I use llama.cpp convert_hf_to_gguf.py. Example [here](https://github.com/CycloneDX/cdxgen/blob/6e58ed3ca05c027218931c532e9b90fec7a2b907/contrib/fine-tuning/convert-gguf.sh#L52).

JoshShailes:
> For GGUF conversion, I use llama.cpp convert_hf_to_gguf.py. Example [here](https://github.com/CycloneDX/cdxgen/blob/6e58ed3ca05c027218931c532e9b90fec7a2b907/contrib/fine-tuning/convert-gguf.sh#L52).

Hi @prabhu , thanks for the reply, do you just feed your MLX fused model into that and the returned GGUF acts as you expected?

I have tested it again and still get sub-optimal responses from the gguf version compared to the MLX version. Have you tested your converted models in anything like Msty? I'm wondering if it could be something in that causing issues with the outputs if you haven't.

prabhu:
@JoshShailes Best to take this discussion elsewhere to avoid hijack. My email is prabhu at appthreat dot dev. Yes, after a lot of trial and errors, we get identical performance across both MLX and GGUF versions.

JoshShailes:
[...]

PR #1321, opened 2026-05-28
Keep model parameters for fuse GGUF export

abnormal749:
## Summary

- Forward `donate_model` from `utils.save()` to `save_model()` instead of always donating the model.
- Add regression coverage for both `utils.save(..., donate_model=...)` behavior and the `mlx_lm.fuse --export-gguf` path.

## Why

`utils.save()` accepted a `donate_model` argument but ignored it and always called `save_model(..., donate_model=True)`. That clears the in-memory model parameters even when callers explicitly need them after saving.

`mlx_lm.fuse --export-gguf` passes `donate_model=False` because it saves the fused model first, then reads `model.parameters()` to export GGUF weights. When the flag is ignored, GGUF export receives empty model tensors.

Related to #353, specifically the path where the generated GGUF is missing expected tensors. This does not claim to cover the separate row-major serialization symptom mentioned there.

## Validation

- `python -m pre_commit run --files mlx_lm/utils.py tests/test_utils.py tests/test_gguf.py`
- `python -m unittest -v tests.test_utils.TestUtils.test_save_respects_donate_model tests.test_gguf`
- Real-world smoke test with `mlx-community/Llama-3.2-1B-Instruct-bf16`: trained a one-step LoRA adapter, ran `mlx_lm fuse --export-gguf`, and confirmed the produced GGUF contains `token_embd.weight`.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -939,7 +939,7 @@ def save(
         hf_repo = None
 
     dst_path = Path(dst_path)
-    save_model(dst_path, model, donate_model=True)
+    save_model(dst_path, model, donate_model=donate_model)
     save_config(config, config_path=dst_path / "config.json")
     tokenizer.save_pretrained(dst_path)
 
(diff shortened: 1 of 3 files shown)


zcbenz:

PR #1322, opened 2026-05-28
Add BatchQuantizedKVCache

xxxkkw:
## Summary
- Add `BatchQuantizedKVCache` for non-rotating quantized KV histories.
- Allow quantized KV prompt caches to participate in continuous batching via merge, extract, filter, and extend operations.
- Fix quantized SDPA mask broadcasting for batched grouped-query attention with left-padded histories, which was exposed by a real Qwen3 batch decode smoke test.

## Reproduction
Before this change, `_merge_caches([[quantized_cache_1], [quantized_cache_2]])` fails because `QuantizedKVCache` has no batching support. After the initial cache support, a real Qwen3 batch decode with left-padded quantized histories exposed a second issue in quantized SDPA:

```text
ValueError: [broadcast_shapes] Shapes (2,1,1,66) and (2,8,2,1,66) cannot be broadcast.
```

The follow-up commit expands batched masks across the grouped-query repeat dimension before applying them to quantized attention scores.

## Environment
- macOS Darwin 25.4.0 arm64
- MLX / mlx-lm virtual environment on Apple Silicon
- Model: `mlx-community/Qwen3-0.6B-4bit`

## Real-weight benchmark

Protocol: two batched requests with existing prompt-cache histories, `prefill_step_size=2048`, 4-bit KV quantization with group size 64 for the quantized run. The 100K input / 10K output baseline run was attempted first and failed with Metal OOM, so the reported long-context comparison uses the requested fallback size: 50K input / 5K output per sequence.

| Mode | Input tokens | Output tokens | Prefill tokens | Prefill time | Decode time | Generated tokens | Decode TPS | KV cache bytes | Peak memory |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| BatchKVCache fp16 histories | 50,000 / 49,872 | 5,000 each | 99,870 | 131.87s | 389.12s | 10,000 | 25.70 tok/s | 11.48 GB | 25.00 GB |
| BatchQuantizedKVCache 4-bit histories | 50,000 / 49,872 | 5,000 each | 99,870 | 210.83s | 151.72s | 10,000 | 65.91 tok/s | 3.23 GB | 13.71 GB |

Observed deltas at 50K/5K:
- KV cache storage: 11.48 GB → 3.23 GB (71.9% lower)
- Peak memory: 25.00 GB → 13.71 GB (45.2% lower)
- Decode throughput: 25.70 → 65.91 tok/s (2.56x higher)
- Quantized prefill is slower because cache quantization is performed while building the histories.

## Test plan
- `python -m unittest discover -s tests -p test_models.py -k quantized`
- `python -m unittest discover -s tests -p test_prompt_cache.py -k quantized`
- Real Qwen3 smoke: batch decode two existing quantized prompt-cache histories with `mlx-community/Qwen3-0.6B-4bit`
- Real Qwen3 long-context benchmark above

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -30,6 +30,7 @@ from .models import cache
 from .models.cache import (
     ArraysCache,
     BatchKVCache,
+    BatchQuantizedKVCache,
     BatchRotatingKVCache,
     CacheList,
     KVCache,
@@ -842,6 +843,10 @@ def _make_cache(model, left_padding, max_kv_size):
     """
 
     def to_batch_cache(c):
+        if isinstance(c, QuantizedKVCache):
+            return BatchQuantizedKVCache(
+                left_padding, group_size=c.group_size, bits=c.bits
+            )
         if type(c) is KVCache:
             return BatchKVCache(left_padding)
         elif isinstance(c, ArraysCache):
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -90,6 +90,8 @@ def quantized_scaled_dot_product_attention(
             q_indices = mx.arange(kL - qL, kL)
             k_indices = mx.arange(kL)
             mask = q_indices[:, None] >= k_indices[None]
+        elif n_repeats > 1 and mask.ndim == scores.ndim - 1:
+            mask = mx.expand_dims(mask, axis=-3)
         if mask.dtype == mx.bool_:
             scores = mx.where(mask, scores, mx.finfo(scores.dtype).min)
         else:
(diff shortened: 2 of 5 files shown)


crackalamoo:
I independently implemented a similar thing for my own use with a subset of these functions and ended up with similar code (https://github.com/crackalamoo/qwen-crackalamoo/commit/2c3034c501eb1102f55c68884c2fe32fdeb314b5). I can vouch for this PR looking correct and useful

zcbenz:

PR #1323, opened 2026-05-28
Guard MLA caches from KV quantization

xxxkkw:
## Summary
- Add an explicit `MLACache` marker for DeepSeek-style latent MLA caches.
- Route DeepSeek V3 and Kimi K2.5 through MLA caches, and mark the latent side of DeepSeek V3.2's `CacheList` as MLA-specific.
- Skip `kv_bits` conversion for caches that do not support generic KV quantization, while keeping ordinary KV cache quantization unchanged.

## Reproduction
Before this change, a tiny DeepSeek V3 config with `kv_bits=4`, `kv_group_size=32`, prompt length 3, and `prefill_step_size=2` fails on the second step with:

```text
```

The failure happens because generic `KVCache.to_quantized()` turns the MLA rope side channel into a quantized tuple, but the DeepSeek V3 MLA attention path still needs an array for `k_pe.swapaxes(...)`.

After this change, the same repro generates successfully and the cache remains `MLACache`.

## Environment
- macOS Darwin 25.4.0 arm64
- Python virtual environment with MLX / mlx-lm test dependencies

## Test plan
- `python -m unittest discover -s tests -p test_models.py -k mla_models_make_mla_caches`
- `python -m unittest discover -s tests -p test_models.py -k deepseek_v3_kv_bits_skips_mla_cache`
- `python -m unittest discover -s tests -p test_models.py -k deepseek_v3`
- `python -m unittest discover -s tests -p test_models.py -k quantized`
- `python -m unittest discover -s tests -p test_prompt_cache.py -k cache_with_generate`
- Synthetic 1-layer DeepSeek V3.2 and Kimi K2.5 cache-forward check with `MLACache`

No throughput speedup is claimed here; this is a correctness guard for an invalid quantization path. Generic quantized KV cache coverage continues to pass.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -300,6 +300,8 @@ def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_
     if kv_bits is None:
         return
     for e, c in enumerate(prompt_cache):
+        if not getattr(c, "supports_quantized_kv_cache", True):
+            continue
         if hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
             prompt_cache[e] = c.to_quantized(group_size=kv_group_size, bits=kv_bits)
 
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -407,6 +407,13 @@ class KVCache(_BaseCache):
         return self.keys.nbytes + self.values.nbytes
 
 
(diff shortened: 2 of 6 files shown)


nastya236:
Do you mind sharing exact command to reproduce this issue? 
Like `mlx_lm.chat --model ...`

zcbenz:

PR #1324, opened 2026-05-29
Consolidate MLA kv_b_proj sanitize/shard into shared helpers

scyyh11:
## Summary

Several MLA models (DeepSeek-V3, DeepSeek-V3.2, GLM-4 MoE Lite, Kimi Linear, LongCat Flash) carry near-identical copies of two pieces of logic:

- **`sanitize()`** — splitting the fused `kv_b_proj` weight into the absorbed `embed_q` / `unembed_out` (`MultiLinear`) projections, including the quantized-checkpoint path.
- **`shard()`** — slicing those per-head projections across a distributed group.

This PR extracts both into shared helpers in `mlx_lm/models/mla.py` (`split_kv_b_proj_weights` and `shard_mla_projections`) and updates the five models to call them, removing ~230 lines of duplicated code.

## Behavior

Behavior-preserving refactor — no functional change for existing checkpoints:

- The splitter dequantizes → splits → **re-quantizes** the projections exactly as the per-model code did, so quantized checkpoints still load as `QuantizedMultiLinear` (no silent fall back to full precision).
- It additionally tolerates affine checkpoints saved without explicit `biases`.
- `shard_mla_projections()` raises a clear error when the head count is not divisible across ranks, instead of silently producing wrong shapes.

Forward passes, the KV cache, LoRA/DoRA, AWQ and DeepSeek-V2 are intentionally left untouched.

## Tests

- Added `test_mla_split_requantizes_affine_weight_without_biases` (covers the quantized split round-trip and the no-`biases` case).
- Added `test_mla_sharding_requires_divisible_heads`.
- Existing MLA model tests (`test_deepseek_v3`, `test_deepseek_v32`, …) pass.

```bash
python -m unittest discover tests/
```

### DIFF
--- a/mlx_lm/models/deepseek_v3.py
+++ b/mlx_lm/models/deepseek_v3.py
@@ -11,7 +11,7 @@ from mlx.nn.layers.distributed import shard_inplace, shard_linear, sum_gradients
 
 from .activations import swiglu
 from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
-from .mla import MultiLinear
+from .mla import MultiLinear, shard_mla_projections, split_kv_b_proj_weights
 from .pipeline import PipelineMixin
 from .rope_utils import initialize_rope
 from .switch_layers import SwitchGLU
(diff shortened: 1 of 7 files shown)


scyyh11:
@angeloskath would you mind taking a look when you get a chance?

This is a small, behavior-preserving refactor that pulls the duplicated MLA `kv_b_proj` sanitize/shard logic into shared helpers. Thanks!

zcbenz:

PR #1328, opened 2026-05-30
Fix Qwen3-Coder tool parser and harden server against mid-stream client disconnects

i1rr:
Two related server-resilience fixes triggered by an agent flow against `mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit` via `mlx_lm.server`.

### Problem

`_parameter_regex` used `<parameter=(.*?)</parameter>` with non-greedy `.*?`. When a value contained text like:

```
<parameter=file_text>
def example():
    # see <parameter=path>...</parameter> in the docs
</parameter>
```

the regex stopped at the first inner `</parameter>`, splitting the value mid-content. The truncated tail was then no longer valid JSON or a valid Python literal, so `_convert_param_value` raised `SyntaxError` from `ast.literal_eval`. That `SyntaxError` is not a `ValueError`/`JSONDecodeError`, so the `except` in `ToolCallFormatter.__call__` did not catch it. The exception escaped into `socketserver.process_request_thread`, producing a noisy traceback in the server log and a half-sent response on the client side.

### Fix

- Rewrite `_parameter_regex` with two capture groups and a trailing lookahead that anchors the closing `</parameter>` to one followed by the next `<parameter=`, the closing `</function>`, or end of input. Literal `<parameter=...>` / `</parameter>` substrings inside a value no longer terminate the match.
- The manual `match_text.index(">")` split is gone — `findall` now returns `(name, value)` directly.
- `except (ValueError, json.JSONDecodeError)` → `except Exception` in `ToolCallFormatter`. Future parser bugs degrade gracefully (warn + skip) instead of corrupting the response.

### Problem

Long generations (large prompt cache + big model) can take many minutes. If the client times out and closes the socket mid-generation, the next `self.wfile.write(...)` raises `BrokenPipeError` (or another `ConnectionError` subclass). The exception bypasses the access-log layer and prints a long traceback. Worse, every retry from the same client repeated the cycle, and each retry kept the model busy generating tokens that nobody was going to read — monopolizing the generation lock and starving subsequent requests.

### Fix

Add `except ConnectionError` (covers `BrokenPipeError`, `ConnectionResetError`, `ConnectionAbortedError`) right before the existing `finally: ctx.stop()`. The teardown then aborts the generation cleanly: `ctx.stop()` halts the token stream, the model becomes available, and a single concise INFO line replaces the traceback in the log.

## Test plan

- [x] `python -m unittest tests.test_tool_parsing` — 6/6, including the new `test_qwen3_coder_value_contains_parameter_tags`.
- [x] `python -m unittest tests.test_server` — 20/20.
- [x] `black --check` clean on all touched files.
- [x] Manually verified against the original failing prompts: parser returns the correct value, server logs no exception when a client disconnects mid-stream, and the generation actually stops (no orphaned compute after disconnect).

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1548,6 +1548,11 @@ class APIHandler(BaseHTTPRequestHandler):
                 self.end_headers()
                 self.wfile.write(response_json)
                 self.wfile.flush()
+        except ConnectionError as e:
+            logging.info(
+                f"Client disconnected before response completed "
+                f"({type(e).__name__}); aborting generation."
+            )
         finally:
             ctx.stop()
 
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -79,10 +79,10 @@ class ToolCallFormatter:
         for tool_text in tool_calls:
             try:
                 parsed = self._tool_parser(tool_text, self._tools)
-            except (ValueError, json.JSONDecodeError) as e:
+            except Exception as e:
                 logging.warning(
                     f"Failed to parse tool call ({type(e).__name__}: {e}) — "
-                    f"tool text was likely truncated mid-generation."
+                    f"tool text was likely malformed or truncated mid-generation."
                 )
                 continue
             if not isinstance(parsed, list):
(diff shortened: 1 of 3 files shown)


zcbenz:

PR #1331, opened 2026-05-30
Fix seeded stochastic completions in the server

### LINKED ISSUE #1245: mlx_lm.server: seed parameter silently ignored

## Repro

Start server:

```bash
```

```bash
done
```
```
bash repro.sh
```

## Expected

## My thoughts why

```python
if args.seed is not None:
    mx.random.seed(args.seed)
```

```python
```

### Update

## Also tried

## Workaround

lyonsno:
## Summary

Fixes #1245.

The server accepts a `seed` parameter, but stochastic completions could still ignore per-request reseeding in a long-lived server process. The sampled path could keep using compiled random sampling across requests, and generated prompt-cache entries could replay a previous sampled assistant continuation for a later seeded request.

This keeps the default compiled sampler path for normal generation, but has the server use uncompiled random sampling only when a request is both seeded and stochastic. It also bypasses prompt-cache reuse for seeded stochastic requests to avoid reusing sampled assistant state across seed changes, while preserving the existing cache behavior for deterministic requests and unseeded stochastic requests.

## Changes

- Add a `use_compiled_sampling` option to `make_sampler()`, defaulting to the current compiled behavior.
- Add uncompiled random-sampling paths for categorical sampling and XTC so per-request `mx.random.seed()` is honored in the server's generation thread.
- Seed the server generation stream for requests that pass `seed`.
- Bypass prompt-cache fetch/insert for seeded stochastic requests only.
- Add regression coverage for threaded sampler seeding and server cache replay behavior.

## Verification

- `pytest tests/test_sample_utils.py::TestSampleUtils::test_uncompiled_sampler_respects_seed_in_thread tests/test_server.py::TestServer::test_handle_completions tests/test_server.py::TestServer::test_handle_completions_different_seeds_do_not_replay_cache tests/test_server.py::TestServer::test_prompt_cache_policy_only_bypasses_seeded_stochastic_requests -q`
- `pytest tests/test_sample_utils.py tests/test_server.py -q`
- `git diff --check`

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -16,6 +16,7 @@ def make_sampler(
     xtc_probability: float = 0.0,
     xtc_threshold: float = 0.0,
     xtc_special_tokens: List[int] = [],
+    use_compiled_sampling: bool = True,
 ) -> Callable[[mx.array], mx.array]:
     """
     Make a sampler function for use with ``generate_step``.
@@ -37,6 +38,8 @@ def make_sampler(
             for being sampled.
         xtc_special_tokens (list(int), optional): List of special tokens IDs to
             be excluded from XTC sampling.
+        use_compiled_sampling (bool, optional): If ``False``, stochastic
+            sampling uses uncompiled random operations. Default: ``True``.
 
 
     Returns:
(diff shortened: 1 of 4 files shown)


IsmaelMartinez:
Confirmed this fixes it. Tested on Apple Silicon, patched build is this branch installed editable, baseline is stock mlx_lm 0.31.3.

Same model (mlx-community/Qwen3-0.6B-4bit), same prompt, temperature 1.0, seeds 1/2/3 on /v1/chat/completions:

```
# stock 0.31.3, all three identical (the bug)
seed=1: A lighthouse stands tall and gleaming, its beacon casting long shadows...
seed=2: A lighthouse stands tall and gleaming, its beacon casting long shadows...
seed=3: A lighthouse stands tall and gleaming, its beacon casting long shadows...

# this branch, three different outputs
seed=1: A lighthouse stood on the edge of a stormy sea, its beacon humbly casting...
seed=2: The lighthouse stands as an eternal beacon of faith and purpose, its towers...
seed=3: A lighthouse stands tall in the darkness, its beam illuminating the pathways...
```

Same seed twice still returns identical output, so seeded stays reproducible while different seeds diverge, which is what you want. Tests pass on the branch too, the four you added plus the full test_sample_utils.py and test_server.py (30 passed). Would be good to see this land.

Thanks again!

zcbenz:
This had been fixed by https://github.com/ml-explore/mlx/pull/3828 and I'm closing this in favor of https://github.com/ml-explore/mlx-lm/pull/1753.

PR #1260, opened 2026-05-08
Fix LFM2.5 tool parser inference

blairhudson:
LFM2.5 tokenizers expose `<|tool_call_start|>` and `<|tool_call_end|>` tokens, but do not provide chat-template metadata that lets mlx-lm infer a tool parser. As a result, server responses can emit raw tool-call markup in `message.content` instead of OpenAI-compatible `tool_calls`.

This adds tokenizer-vocab based parser inference for that marker pair and maps it to the existing pythonic tool parser.

Tests:
- `uv run --with-editable . --with unittest-xml-reporting python -m unittest tests.test_tool_parsing tests.test_tokenizers.TestTokenizers.test_tool_calling`

You can try it yourself with:

```sh
uv tool install "git+https://github.com/blairhudson/mlx-lm.git@fix/lfm25-tool-parser-inference"
mlx_lm.server --model LiquidAI/LFM2.5-1.2B-Instruct-MLX-8bit
```

PR #1333, opened 2026-05-31
Fix LFM2.5 pythonic tool parser auto-detection

ChristianWeyer:
## What

`_infer_tool_parser()` in `tokenizer_utils.py` detects the pythonic tool-call
format via the chat-template token `<|tool_list_start|>`. LFM2.5 renamed those
delimiters to `<|tool_call_start|>` / `<|tool_call_end|>` (the inner pythonic
call format is unchanged), so the detection misses for every LFM2.5
checkpoint — including the official `LiquidAI/LFM2.5-8B-A1B`,
`LiquidAI/LFM2.5-8B-A1B-MLX-8bit`, `LiquidAI/LFM2.5-1.2B-Thinking`, etc.

Symptom: loading any LFM2.5 model gives `tokenizer.has_tool_calling == False`,
so `mlx_lm.server`'s `/v1/chat/completions` returns tool calls as raw
`<|tool_call_start|>[func(arg='val')]<|tool_call_end|>` text inside
`message.content`, with `tool_calls = []` and `finish_reason = "stop"`.
OpenAI-compatible clients (Mastra, LangChain, ai-sdk, …) see no tool call and
the agent loop breaks.

## Fix

Match either token. The `pythonic` parser already handles both formats — only
the auto-detection needed updating.

```python
elif (
    "<|tool_list_start|>" in chat_template
    or "<|tool_call_start|>" in chat_template
):
    return "pythonic"
```

Same root cause and fix shape as the parallel `llama.cpp` issue
[ggml-org/llama.cpp#23838](https://github.com/ggml-org/llama.cpp/issues/23838),
and the same class of token-rename detection lag that prompted the Gemma 4
auto-detect addition.

## Test

Added `TestTokenizers.test_infer_tool_parser` in `tests/test_tokenizers.py`
covering:

- LFM2 original template (`<|tool_list_start|>`) — still returns `"pythonic"`
- LFM2.5 template (`<|tool_call_start|>`) — now returns `"pythonic"` (was `None`)
- Negative path (no tool tokens, and `None` input)

This is a pure unit test against `_infer_tool_parser`, so no HF download —
fast and deterministic. Verified locally that the new test fails on `main`
without the fix and passes with it.

End-to-end verified against the real `LiquidAI/LFM2.5-8B-A1B-MLX-8bit`
tokenizer:

```python
>>> from mlx_lm.utils import load_tokenizer
>>> tok = load_tokenizer("LiquidAI/LFM2.5-8B-A1B-MLX-8bit")
>>> tok.has_tool_calling
True
>>> tok.tool_call_start, tok.tool_call_end
('<|tool_call_start|>', '<|tool_call_end|>')
```

And via `mlx_lm.server` `/v1/chat/completions` with a `tools` payload — after
the fix, the response now correctly returns:

```json
{
  "finish_reason": "tool_calls",
  "message": {
    "tool_calls": [{
      "function": {"name": "...", "arguments": "{\"query\": \"...\"}"},
      "type": "function",
      "id": "..."
    }]
  }
}
```

All existing `tests/test_tool_parsing.py` and `tests/test_tokenizers.py` tests
still pass; ran `black` (25.1.0) and `isort --profile=black` per
`CONTRIBUTING.md`.

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -559,7 +559,9 @@ def _infer_tool_parser(chat_template):
         return "longcat"
     elif "<arg_key>" in chat_template:
         return "glm47"
-    elif "<|tool_list_start|>" in chat_template:
+    elif (
+        "<|tool_list_start|>" in chat_template or "<|tool_call_start|>" in chat_template
+    ):
         return "pythonic"
     elif (
         "<tool_call>\\n<function=" in chat_template
--- a/tests/test_tokenizers.py
+++ b/tests/test_tokenizers.py
@@ -9,6 +9,7 @@ from mlx_lm.tokenizer_utils import (
     BPEStreamingDetokenizer,
     NaiveStreamingDetokenizer,
     SPMStreamingDetokenizer,
+    _infer_tool_parser,
 )
 from mlx_lm.utils import load_tokenizer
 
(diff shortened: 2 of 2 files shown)


tdakhran:
Thanks for looking into this, @ChristianWeyer!

I ran internal tests against MLX, and it shows that one test case became green (top is pre PR commit, bottom is PR)
image

I'm attaching interactive HTML reports if you want to debug further for other test cases (click a cell, and it will show the request used for the test case).
[report-mlx-2b15e8b.html](https://github.com/user-attachments/files/28834600/report-mlx-2b15e8b.html)
[report-mlx-df1d3f3.html](https://github.com/user-attachments/files/28834601/report-mlx-df1d3f3.html)

For inspiration, here's how it looks with upstream llama.cpp😉 
image
and a report
[report-llama.cpp-9592.html](https://github.com/user-attachments/files/28835419/report-llama.cpp-9592.html)

michalk8:
closed via #1260

PR #1334, opened 2026-05-31
Support for Poolside LagunaXS open source coding model in nvfp4

tzachicohen:
The patch introduces support for LagunaXS model with Poolside's native nvfp4 quantization checkpoint.(https://huggingface.co/poolside/Laguna-XS.2-NVFP4)
The model definition is made from the standard primitives. 
Poolside's checkpoint defines two scaling factors, one global, per tensor, in fp32 and one local, per 16 elements tile. 
Since mlx supports only a single scale factor, the two are coalesced into one bf16 scale per tile.
This patch was also tested in integration with "vllm-metal", to enable seamless serving of LagunaXS with vLLM on Apple silicon Macos.

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -3251,6 +3251,57 @@ class TestModels(unittest.TestCase):
                 self.assertTrue(mx.allclose(y, y_gt, rtol=1e-4, atol=1e-4))
                 self.assertTrue(mx.allclose(st, st_gt, rtol=1e-4, atol=1e-3))
 
+    def test_laguna(self):
+        from mlx_lm.models import laguna
+
+        args = laguna.ModelArgs(
+            model_type="laguna",
+            vocab_size=10_000,
+            hidden_size=128,
+            intermediate_size=256,
+            num_hidden_layers=4,
+            num_attention_heads=4,
+            num_key_value_heads=2,
+            head_dim=32,
+            rms_norm_eps=1e-6,
+            sliding_window=4,
+            partial_rotary_factor=0.5,
+            num_experts=8,
+            num_experts_per_tok=2,
+            moe_intermediate_size=128,
+            shared_expert_intermediate_size=128,
+            moe_routed_scaling_factor=2.5,
+            # First layer full + dense, remaining sliding + sparse (MoE).
+            layer_types=[
+                "full_attention",
+                "sliding_attention",
+                "sliding_attention",
+                "sliding_attention",
+            ],
+            mlp_layer_types=["dense", "sparse", "sparse", "sparse"],
+            num_attention_heads_per_layer=[4, 4, 4, 4],
+            rope_parameters={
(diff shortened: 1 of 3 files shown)


tests/test_models.py
```diff
@@ -3220,6 +3220,57 @@ def test_gated_delta_masked(self):
                 self.assertTrue(mx.allclose(y, y_gt, rtol=1e-4, atol=1e-4))
                 self.assertTrue(mx.allclose(st, st_gt, rtol=1e-4, atol=1e-3))
 
+    def test_laguna(self):
```
nastya236: Is there a reason I'm missing for the dedicated `test_laguna` method? I think that your testing can be fully covered by adding laguna to the pool (`test_all_models`).  If you're planning to add laguna-specific assertions, then keeping the dedicated method makes sense, otherwise I'd move it into test_all_models.

nastya236 (changes requested):
Thank you for your contribution. 

If I understood everything correctly, you requantize the weights into nvfp4 format, which collapses the per‑tensor fp32 global scale because Metal qqmm doesn't accept it yet. Block‑scaled quantization in MLX is actively evolving and we are planning to add global‑scale support. For now I'd strongly suggest the following: 1) requantize once offline using `_dequantize_compressed_tensors` + `mx.quantize(..., mode="nvfp4")` and store the converted weights, 2) remove the format‑conversion code from sanitize leaving only the MoE expert stacking. 

Happy to iterate on this with you.

tests/test_models.py
tzachicohen: done.

tzachicohen:
Hi @nastya236  , 

We have generated an internal checkpoint that does not require sanitation at all.
Once this PR is approved we will publish the MLX compatible checkpoint in Huggingface. 
Other review comments are addressed as well.

nastya236:
Thank you! Will look at it as soon as possible.

tzachicohen:
Weights are available for testing. (together with this patch). 
[https://huggingface.co/poolside/Laguna-XS-2.1-NVFP4-mlx](https://huggingface.co/poolside/Laguna-XS-2.1-NVFP4-mlx)

nastya236 (approved):

tests/test_models.py
```diff
@@ -3252,5 +3294,6 @@ def test_gated_delta_masked(self):
                 self.assertTrue(mx.allclose(st, st_gt, rtol=1e-4, atol=1e-3))
 
 
+
```
nastya236:
```suggestion
```

nastya236:
Something failed on the checks, please fix and I will merge.

nastya236 (changes requested):

tzachicohen:
@nastya236  - Linting issue fixed.

nastya236 (approved):

PR #1336, opened 2026-05-31
Fix json_tools: match "<tool_call" prefix to survive ">" token merge

### LINKED ISSUE #1335: Tool calls dropped: json_tools start marker never matches when the tokenizer merges the `<tool_call>` closing `>`
### Summary

On the OpenAI-compatible server, tool calls from a model that emits the standard Hermes shape `<tool_call>\n{...}\n</tool_call>` are returned as assistant **content** with `tool_calls = null`, even though `tokenizer.has_tool_calling` is true and `tool_parser_type = json_tools`. The model's output is valid (correct JSON, right function/args) — it just never reaches the parser.

Root cause: tool-call detection matches markers as **exact token-id sequences**, and `json_tools`'s start marker `"<tool_call>"` encodes to a run ending in a **standalone `>` token** that the tokenizer **merges with the following byte** at generation time.

### Environment
- mlx-lm `main` (`df1d3f3`), mlx 0.31.2, Apple M4 Max / macOS 26.3
- `mlx-community/DeepSeek-V4-Flash-2bit-DQ` (any model where `<tool_call>`/`</tool_call>` are **not** single special tokens and the tokenizer merges `>`)

### What happens
```
POST /v1/chat/completions  {tools:[get_weather], messages:[{user:"weather in Tokyo?"}]}
->  finish_reason: stop
    tool_calls:  None
    content:    '<tool_call>\n{"name": "get_weather", "arguments": {"city": "Tokyo"}}\n</tool_call>'
```

### Root cause (with token IDs)

`server.py:_make_state_machine` wires `normal --tool_call_start_tokens--> tool`, and `SequenceStateMachine` (`generate.py`) matches those token-id sequences with an Aho-Corasick trie. `json_tools.tool_call_start = "<tool_call>"`, encoded **in isolation**:

```
encode("<tool_call>")   = (30, 72461, 112042, 32)      #  <  tool  _call  >    (trailing ">" = token 32)
```

But the `>` only survives as a standalone token at end-of-string. With any following byte it merges:

```
encode("<tool_call>\n") = (30, 72461, 112042, 1018)    # ">\n" is ONE token
encode("<tool_call>{")  = (30, 72461, 112042, 31923)   # ">{" is ONE token
```

The Hermes convention puts `\n` right after `<tool_call>`, so the model emits the merged `>\n` token and the precomputed marker `(…, 32)` is **never a subsequence** of the stream → the state machine stays in `normal` → the call is emitted as content.

### Reproducer (deterministic, tokenizer-only, no GPU)

```python
from pathlib import Path
from mlx_lm.tokenizer_utils import load as load_tokenizer
from mlx_lm.generate import SequenceStateMachine

tok = load_tokenizer(Path(MODEL))          # a model with non-special-token <tool_call> markers
enc = lambda s: tuple(tok.encode(s, add_special_tokens=False))
STOCK, PREFIX, END = enc("<tool_call>"), enc("<tool_call"), enc("</tool_call>")
emitted = enc('<tool_call>\n{"name":"get_weather","arguments":{"city":"Paris"}}\n</tool_call>')

def enters_tool(start):
    sm = SequenceStateMachine({"normal":[(start,"tool")], "tool":[(END,"normal")]}, initial="normal")
    st, seen = sm.make_state(), set()
    for x in emitted:
        st, _, s = SequenceStateMachine.match(st, x); seen.add(s)
    return "tool" in seen

print(enters_tool(STOCK))   # False  <- bug: stock "<tool_call>" marker never matches
print(enters_tool(PREFIX))  # True   <- "<tool_call" prefix matches
```

### Why it's gone unnoticed
Models that use `json_tools` successfully (Qwen2.5 / Hermes-style) either register `<tool_call>` so the runtime tokens match, or their tokenizer doesn't merge `>` with the newline. It bites tokenizers where `<tool_call>` is an ordinary multi-token string **and** `>` merges — increasingly common as new checkpoints get OpenAI-style tool templates added at conversion time.

### Proposed fix
Match the stable `<tool_call` prefix (drop the volatile `>`) and extract the first brace-balanced JSON object (tolerating the leftover `>`/newline and a trailing `</tool_call>`). Output is identical to `json.loads(text.strip())` for clean segments; streaming is unaffected (the server accumulates the whole tool segment before parsing). PR: **https://github.com/ml-explore/mlx-lm/pull/1336**.

Full analysis: [writeup](https://github.com/snagnever/macstudio-local-llm/blob/deepseek-v4-tool-template/docs/mlx-lm-tool-call-marker-merge-writeup.md). On `DeepSeek-V4-Flash-2bit-DQ` this took tool-call parsing from **0 → 82%** on a tool-calling bench.

snagnever:
Fixes https://github.com/ml-explore/mlx-lm/issues/1335.

### Cause
Tool-call detection matches markers as **exact token-id sequences** (`SequenceStateMachine`, Aho-Corasick). `json_tools.tool_call_start = "<tool_call>"` encodes to a run ending in a **standalone `>` token** — but many tokenizers merge that `>` with the following byte (`<tool_call>\n` → a single `>\n` token), so the precomputed marker is never a contiguous subsequence of the generated stream. The state machine never enters the `tool` state and a valid tool call is returned as assistant content with `tool_calls = null`. (Token IDs + a deterministic reproducer in the issue.)

### Change
`mlx_lm/tool_parsers/json_tools.py`:
- Start marker → `"<tool_call"` (drop the volatile `>`; the `<`,`tool`,`_call` tokens are stable).
- `parse_tool_call` extracts the **first brace-balanced JSON object**, tolerating the leftover `>`/newline that the wider capture now includes and a trailing `</tool_call>` if that marker likewise merges.

### Why it's safe
- **Backward-compatible:** `<tool_call` is a subsequence of what currently-matching models emit, and brace extraction returns output **identical** to `json.loads(text.strip())` for clean segments. The existing `json_tools` cases in `tests/test_tool_parsing.py` (clean JSON, nested braces) still pass.
- **Streaming unaffected:** `server.py` accumulates the entire `tool` segment (`tool_text`) while in the tool state and only calls `parse_tool_call` after the transition back to `normal`, so the leftover `>` never leaks into a streamed delta.

### Tests
Adds `test_json_tools_marker_merge` (clean JSON, leftover-`>`, trailing-`</tool_call>`, nested-brace/`}`-in-string). `python -m unittest tests.test_tool_parsing` → all pass.

### Verification (end-to-end)
On `mlx-community/DeepSeek-V4-Flash-2bit-DQ` (markers not special tokens; `>` merges): tool-call parse rate **0/4 → 4/4** on a probe and **8/40 → 33/40** on jdhodges tool-calling — same checkpoint, no other change.

### Note for reviewers
This changes the shared `json_tools` parser. I validated output-identity for clean input and the streaming path, but couldn't test against a live Qwen2.5/Hermes model — happy to gate this behind a separate parser (e.g. `json_tools_lenient`) or adjust the approach if you'd prefer. Full analysis: [writeup](https://github.com/snagnever/macstudio-local-llm/blob/deepseek-v4-tool-template/docs/mlx-lm-tool-call-marker-merge-writeup.md).

### DIFF
--- a/mlx_lm/tool_parsers/json_tools.py
+++ b/mlx_lm/tool_parsers/json_tools.py
@@ -2,10 +2,46 @@
 
 import json
 
-tool_call_start = "<tool_call>"
+# The start marker intentionally omits the closing ">". Tool-call markers are matched as
+# exact token-id sequences, and "<tool_call>" encodes to a run ending in a standalone ">"
+# token. Many tokenizers merge that ">" with the following byte (e.g. "<tool_call>\n" -> a
+# single ">\n" token), so the full-marker token run never appears in the generated stream
+# and the call is never captured. Matching the stable "<tool_call" prefix avoids this;
+# parse_tool_call() below tolerates the leftover ">" before the JSON.
+tool_call_start = "<tool_call"
 
 tool_call_end = "</tool_call>"
 
 
 def parse_tool_call(text, tools=None):
-    return json.loads(text.strip())
+    """Extract the first brace-balanced JSON object from the captured tool segment.
+
+    Robust to a leading ">"/newline left by the prefix start marker and to a trailing
+    "</tool_call>" if the end marker likewise merges. Output is identical to
+    json.loads(text.strip()) for a clean JSON-only segment.
+    """
+    start = text.find("{")
+    if start == -1:
+        raise ValueError("no JSON object in tool call segment")
+    depth = 0
+    in_str = False
+    esc = False
+    for i in range(start, len(text)):
+        c = text[i]
+        if in_str:
+            if esc:
+                esc = False
+            elif c == "\\":
+                esc = True
+            elif c == '"':
+                in_str = False
+            continue
+        if c == '"':
+            in_str = True
+        elif c == "{":
+            depth += 1
+        elif c == "}":
+            depth -= 1
+            if depth == 0:
+                return json.loads(text[start : i + 1])
+    raise ValueError("unbalanced JSON in tool call segment")
--- a/tests/test_tool_parsing.py
+++ b/tests/test_tool_parsing.py
@@ -207,6 +207,28 @@ class TestToolParsing(unittest.TestCase):
             {"settings": {"enabled": True, "name": "test"}},
         )
 
+    def test_json_tools_marker_merge(self):
+        # The start marker is "<tool_call" (without the closing ">", which many tokenizers
+        # merge with the next byte). The server therefore captures the tool segment starting
+        # at the leftover ">", and the segment may also include a trailing "</tool_call>" if
+        # that marker's tokens merge too. parse_tool_call must extract the JSON regardless.
+        expected = {"name": "get_weather", "arguments": {"city": "Paris"}}
+        variants = [
+            '{"name": "get_weather", "arguments": {"city": "Paris"}}',  # clean (backward-compat)
+            '>\n{"name": "get_weather", "arguments": {"city": "Paris"}}\n',  # leftover ">"
+            '>\n{"name": "get_weather", "arguments": {"city": "Paris"}}\n</tool_call>',  # + end marker
+        ]
+        for text in variants:
+            with self.subTest(text=text):
+                self.assertEqual(json_tools.parse_tool_call(text, None), expected)
+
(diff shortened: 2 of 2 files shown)


angeloskath:
Closed due to #1501.

PR #1337, opened 2026-05-31
Add deepseek_dsml tool parser for DeepSeek-V4's native DSML format

snagnever:
Adds a tool parser for **DeepSeek-V4 (Flash/Pro)**'s native **DSML** tool-call format, so tool calling works for these models through the OpenAI-compatible server.

### Why
DeepSeek-V4 emits tool calls in its own DSML format — not Hermes `<tool_call>` JSON or any of the existing parsers' formats — so there is currently no `tool_parser_type` that parses its output, and tool calls come back as plain content.

### Format
```
<｜DSML｜tool_calls>
<｜DSML｜invoke name="get_weather">
<｜DSML｜parameter name="city" string="true">Paris</｜DSML｜parameter>
<｜DSML｜parameter name="days" string="false">3</｜DSML｜parameter>
</｜DSML｜invoke>
<｜DSML｜invoke name="get_weather">...</｜DSML｜invoke>
</｜DSML｜tool_calls>
```
- Multiple `<｜DSML｜invoke>` per block → **native parallel calls**.
- `string="true"` → literal string value; `string="false"` → JSON value (number/bool/array/object).

### Change
- `mlx_lm/tool_parsers/deepseek_dsml.py` — new parser (modeled on `minimax_m2`).
- `_infer_tool_parser` — one entry so the official DSML chat template ([deepseek-ai/DeepSeek-V4-Flash#16](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash/discussions/16)) auto-selects it; no manual `tool_parser_type` needed.
- Tests — single call, **parallel calls**, mixed string/JSON params.

### Marker note
The start/end markers use the `<｜DSML｜tool_calls` **prefix** (dropping the trailing `>`). mlx-lm matches markers by exact token-id sequence, and on this tokenizer the `>` merges with the following byte (`...calls>\n` → one token), so the full marker never matches — the same class of issue as #1335. The `｜DSML｜` core is a special token, so the prefix is a stable anchor; the parser tolerates the leftover `>` before the first `<｜DSML｜invoke>`.

### Verification
On `mlx-community/DeepSeek-V4-Flash-2bit-DQ` (no tool template → tools dropped → **0** tool calls as a baseline), adding the official DSML template + this parser: **39/40 (98%)** on a jdhodges-style tool suite and **8/8 on the parallel multi-tool** cases — competitive with full-size locals. `python -m unittest tests.test_tool_parsing` passes (existing parsers unaffected; this is purely additive).

Pairs with the official template PR #16. Independent of #1335/#1336 (different module), though it uses the same prefix-marker insight.

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -549,6 +549,8 @@ def _infer_tool_parser(chat_template):
     """Attempt to auto-infer a tool parser from the chat template."""
     if not isinstance(chat_template, str):
         return None
+    elif "<｜DSML｜tool_calls>" in chat_template:
+        return "deepseek_dsml"
     elif "<minimax:tool_call>" in chat_template:
         return "minimax_m2"
     elif "<|tool_call>" in chat_template and "<tool_call|>" in chat_template:
--- a/mlx_lm/tool_parsers/deepseek_dsml.py
+++ b/mlx_lm/tool_parsers/deepseek_dsml.py
@@ -1,0 +1,79 @@
+# Tool parser for DeepSeek-V4's native DSML tool-call format. Deploy into
+# venv: mlx_lm/tool_parsers/deepseek_dsml.py, set tokenizer_config
+# "tool_parser_type": "deepseek_dsml", and use the official DSML chat template.
+#
+# Format (multiple invokes per tool_calls block => native parallel calls):
+#   <｜DSML｜tool_calls>
+#   <｜DSML｜invoke name="get_weather">
+#   <｜DSML｜parameter name="city" string="true">Paris</｜DSML｜parameter>
+#   <｜DSML｜parameter name="days" string="false">3</｜DSML｜parameter>
+#   </｜DSML｜invoke>
+#   ...
+#   </｜DSML｜tool_calls>
+#
+# string="true"  -> value is a literal string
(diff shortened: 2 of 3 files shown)


nh13:
Thanks for this, @snagnever — I hit the same DSML tool-calling gap on DeepSeek-V4-Flash and independently landed on the same parser shape (parallel invokes, `string="true|false"`, the no-trailing-`>` marker anchor). Two things from running it against the 4/5/6-bit MLX quants that may be worth folding in:

1. **Truncated invoke close tag** — those quants systematically emit `</｜DSML｜inv>` instead of `</｜DSML｜invoke>`, so the strict close in `_INVOKE` misses the call and it returns as content. A lenient close (`</｜DSML｜inv[^>]*>`) recovers it.
2. **Encode side** — DeepSeek ships no Jinja template (only `encoding_dsv4.py`), so there's currently nothing to pair the parser with under `chat_template_type`. I ported `encode_messages` to a small `chat_templates/deepseek_dsml.py` reusing the `deepseek_v32` helpers, byte-exact for single-turn prompts.

Both are on a branch stacked on this PR (#1760, draft) if useful — glad to fold them into yours instead if you'd rather. Also worth noting: #1501 (merged) makes the trailing-`>` marker workaround unnecessary on current `main`.

PR #1338, opened 2026-06-01
feat(gpt-oss): support unsloth BF16 layout + shared MoE LoRA adapters

electron-rare:
## Summary

Two independent fixes that together let mlx-lm consume the unsloth /
Hugging Face BF16 distribution of `openai/gpt-oss-20b` and fuse a PEFT
`target_parameters` MoE LoRA on top of it.

### 1. `gpt_oss.sanitize` — handle the BF16 expert weight layout

Two layouts exist for the fused `experts.gate_up_proj` tensor:

| layout            | shape         | key suffix |
|-------------------|---------------|------------|
| openai / MXFP4    | (E, 2H, D)    | `.weight` (post-dequant) |
| unsloth / HF BF16 | (E, D, 2H)    | (none, raw HF storage)   |

The old `sanitize()` assumed layout (a) and:
- split on axis -2 with `[::2, :]` / `[1::2, :]`, which on layout (b)
  halves the `hidden_size` axis and leaves the `2*intermediate` axis
  whole (`(32, 1440, 5760)` instead of `(32, 2880, 2880)`);
- emitted `experts.gate_proj` keys without the `.weight` suffix that
  `SwitchLinear` expects, so the weights silently fail to load.

The patch detects layout (b) by its trailing-axis signature
(`shape[-1] == 2*intermediate and shape[-2] == hidden`), normalises it
to (E, 2H, D) via `swapaxes(-1, -2)`, and then runs the existing
interleaved split — matching HF's forward (`gate, up = gate_up[..., ::2], gate_up[..., 1::2]`).
The `.weight` suffix is added where missing, including for the matching
`down_proj`. Layout (a) and the already-sanitized fast path are
unchanged.

### 2. `LoRASwitchLinear` — accept shared 2D adapters (PEFT `target_parameters`)

PEFT >= 0.13 supports fine-tuning the GPT-OSS MoE via
`target_parameters: ["<layer>.mlp.experts.gate_up_proj", ...]`. The
resulting adapter has a single shared low-rank update per expert
*parameter* rather than per expert:

```
lora_A.weight : (r, hidden_size)
lora_B.weight : (out_features, r)
```

`LoRASwitchLinear.from_base` builds 3-D tensors of shape
`(num_experts, r, in)` / `(num_experts, out, r)`, so assigning the
shared 2-D PEFT tensors and calling `fuse()` crashes at
`lora_b @ lora_a` (matmul shape mismatch).

`fuse()` and `__call__()` now detect 2-D adapter tensors:

- `fuse()` broadcasts `lora_a` and `lora_b` across the experts axis
  before the existing `(lora_b @ lora_a)` outer product, so every
  expert receives the same low-rank delta.
- `__call__()` uses a dense matmul instead of `gather_mm` when both
  tensors are 2-D — equivalent and avoids gathering identical rows
  `num_experts` times.

The pre-existing per-expert (3-D) layout continues to work through the
unchanged code path.

## Use case

We are fine-tuning gpt-oss-20b on a KiCad schematic-generation
dataset ("SchGen") via PEFT `target_parameters` on the MoE expert
fused projections, and serving the result through MLX-LM. Without
these two patches it is not possible to (a) convert the base
distribution at all, or (b) fuse the PEFT adapter into the converted
weights.

## Tests

```
$ python -m unittest tests.test_gpt_oss_sanitize tests.test_lora_switch_linear_shared -v
... 7 tests ok
$ python -m unittest tests.test_models tests.test_tuner_utils -v
... 78 tests ok (1 skipped)
```

New tests:
- `tests/test_gpt_oss_sanitize.py` — 4 cases covering both layouts,
  the already-sanitized fast path, and a value-level check that the
  interleaved split places even slots in `gate` and odd slots in `up`.
- `tests/test_lora_switch_linear_shared.py` — 3 cases covering the
  pre-existing per-expert layout, the new shared 2-D layout, and
  equivalence between the live forward pass and a fuse-then-call
  reference.

## Notes / scope

This PR is intentionally narrow:

- It does **not** change anything for openai's MXFP4 distribution.
- It does **not** add PEFT-adapter import (the caller is expected to
  produce mlx-lm-shaped tensors); it only ensures that a shared 2-D
  MoE LoRA tensor *does* fuse once present.
- The unsloth BF16 base model now loads, runs at native MLX speed
  (~80 tok/s on M3 Ultra) and accepts a SchGen MoE LoRA fuse; achieving
  bit-equivalent generation quality vs the HF transformers reference
  is tracked separately as it depends on attention-sinks / sliding-window
  details outside the `sanitize()` scope.

### DIFF
--- a/mlx_lm/models/gpt_oss.py
+++ b/mlx_lm/models/gpt_oss.py
@@ -244,6 +244,15 @@ class Model(nn.Module):
         if any("gate_proj.weight" in k for k in weights.keys()):
             return weights  # already sanitized
 
+        hidden_size = self.args.hidden_size
+        intermediate_size = self.args.intermediate_size
+
+        def _ensure_weight_suffix(key):
+            # HF/unsloth BF16 checkpoints omit .weight on experts.{gate,up,down}_proj.
+            if key.endswith(".weight") or key.endswith(".scales"):
+                return key
+            return key + ".weight"
+
         new_weights = {}
         for k, v in weights.items():
             if "gate_up_proj" in k and "bias" not in k:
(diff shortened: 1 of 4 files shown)


electron-rare:
Pushed two follow-up commits addressing the Copilot review and the runtime generation incoherence we documented when opening the PR.

**`6b7a319` fix(gpt-oss): bf16 down_proj swap when H==I**

This is also the fix for the runtime generation bug we flagged in the PR body. The previous `intermediate_size != hidden_size` guard on the `down_proj` transpose was skipped entirely on the default GPT-OSS config (`hidden_size == intermediate_size == 2880`), so unsloth/HF BF16 `down_proj` was silently misinterpreted as the openai `(E, hidden, intermediate)` layout — weights loaded fine, shapes matched, generation ran at ~78 tok/s but emitted semantically incoherent tokens. Copilot's first review comment diagnosed this precisely. Thanks!

Fix: detect the checkpoint layout up front from the bare `experts.gate_up_proj` key whose trailing dim is `2*intermediate_size` (an unambiguous unsloth signature), then drive both the `gate_up_proj` and `down_proj` swaps from that single signal. This addresses Copilot review comments #1 (`down_proj` H==I skip), #2 (per-tensor shape ambiguity), and #3 (missing values-based `down_proj` test).

Two new regression tests with distinct values:
- `test_unsloth_bf16_down_proj_values_unequal_dims` (H != I, the obvious case),
- `test_unsloth_bf16_down_proj_values_equal_dims` (H == I, the exact scenario that previously slipped through).

**`e498c94` fix(tuner): reject mixed-ndim lora on switch**

Addresses Copilot review comment #4: `LoRASwitchLinear.fuse()` and `__call__` now raise `ValueError` if `lora_a.ndim != lora_b.ndim`, instead of silently taking a partial code path that fails deep inside `gather_mm`/`broadcast_to`. New test `test_mixed_ndim_raises` covers both entry points.

**Status**

- 9/10 tests green on the PR's new test files. The one failing case (`test_call_shared_2d_matches_fuse_then_call`) is **pre-existing on the PR HEAD before these commits** (a numerical mismatch between the dense shared-2D forward path and the fused-then-call path), unrelated to the down_proj fix. Happy to address it in this PR or a follow-up — let me know your preference.
- Runtime end-to-end generation on M3 Ultra not re-verified in this round (the host that holds the fused BF16 checkpoint is currently down). The unit tests cover the exact regression scenario via values-based assertions on H==I. I will re-run the generate smoke test against an unsloth BF16 checkpoint once the host is back and report back here.

zcbenz:

PR #1341, opened 2026-06-03
tokenizer_utils: prefer template-inferred tool parser over generic json_tools label

eilidhmae:
## Problem

`mlx_lm.server` silently drops every tool call for some models. A
`<system>` example: `inferencerlabs/Qwen3-Coder-Next-MLX-9bit` (and other
Qwen3-Coder repos) ship

```json
"tool_parser_type": "json_tools"
```

in `tokenizer_config.json`, but their chat template emits Qwen3-Coder's
**XML** tool-call grammar:

```
<tool_call>
<function=NAME>
<parameter=KEY>value</parameter>
</function>
</tool_call>
```

In `tokenizer_utils.load()` the explicit `tool_parser_type` from the
config takes precedence over `_infer_tool_parser(chat_template)`. So the
generic `json_tools` parser is selected, `json.loads()` is called on the
XML, raises `JSONDecodeError`, and the server logs
`Failed to parse tool call` and drops the call. The model's own template
already tells us (via `_infer_tool_parser`) that this should be the
`qwen3_coder` parser.

## Fix

Prefer the template-inferred parser when the configured label is the
generic `json_tools` default **and** inference identifies a specific
parser. A deliberately-set, specific `tool_parser_type` is never
overridden, and a genuine Hermes/JSON model (`json_tools` config + no
specific marker in the template) is unchanged. Resolution is factored
into `_resolve_tool_parser_type` for testability.

## Tests

`tests/test_tool_parsing.py` gains `TestToolParserSelection` (precedence
matrix: generic-yields-to-specific, missing-label-infers,
hermes-unchanged, deliberate-specific-wins) and a `TestQwen3CoderServerHandoff`
case parsing `tool_text` exactly as the server's state machine hands it
off. All pass.

Verified end-to-end against `Qwen3-Coder-Next-MLX-9bit` on
`mlx_lm.server`: before the fix the first tool call dropped with the
JSONDecodeError above; after, the request returns `finish_reason:
"tool_calls"` with correctly-typed arguments.

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -575,6 +575,26 @@ def _infer_tool_parser(chat_template):
     return None
 
 
+def _resolve_tool_parser_type(configured, chat_template):
+    """Resolve which tool parser to use for a model.
+
+    Precedence: a deliberately-set, specific ``tool_parser_type`` from the
+    tokenizer config wins. But some repos ship the generic ``"json_tools"``
+    label even though their chat template emits a non-JSON tool-call grammar
+    (e.g. Qwen3-Coder's XML ``<tool_call>\\n<function=...>`` form). Parsing
+    that with ``json.loads`` raises ``JSONDecodeError`` and the tool call is
+    silently dropped. So when the explicit label is the generic default but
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1342, opened 2026-06-03
Fix load_model skipping to_quantized check for per-layer quantization config

### LINKED ISSUE #1266: MoEGate.to_quantized() missing in nemotron_h.py causes load failure when quantization config includes gate paths

robertlangdonn:
## Summary

`class_predicate` in `_quantize()` (inside `load_model`) checked whether a module path appeared in `config["quantization"]` **before** checking `hasattr(m, "to_quantized")`. For modules that lack `to_quantized()` — such as `MoEGate` in `nemotron_h.py`, which stores its routing weights as a raw `mx.zeros` tensor rather than `nn.Linear` — this caused the per-layer config dict to be returned, so `nn.quantize` would call `m.to_quantized(**kwargs)` and raise:

```
ValueError: Unable to quantize model of type <class 'mlx_lm.models.nemotron_h.MoEGate'>
```

## Fix

Move the `hasattr(m, "to_quantized")` guard before the per-layer config check so modules lacking `to_quantized()` are always skipped, regardless of quantization config entries. Two lines reordered in `utils.py`:

```python
def class_predicate(p, m):
    if not hasattr(m, "to_quantized"):   # check first
        return False
        return config["quantization"][p]
    return f"{p}.scales" in weights
```

## Test

Added `test_class_predicate_skips_non_quantizable_modules_in_per_layer_config` in `tests/test_utils.py`. Builds a small llama model, injects an `RMSNorm` path into the quantization config (RMSNorm has no `to_quantized`), saves and reloads — verifying no `ValueError` is raised and the model runs correctly.

Fixes #1266.

AI assistance was used for investigation and code review; all changes reviewed and tested by me on M3 / 16 GB / macOS.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -348,10 +348,10 @@ def load_model(
     def _quantize(quantization):
         def class_predicate(p, m):
             # Handle custom per layer quantizations
-            if p in config["quantization"]:
-                return config["quantization"][p]
             if not hasattr(m, "to_quantized"):
                 return False
+            if p in config["quantization"]:
+                return config["quantization"][p]
             return f"{p}.scales" in weights
 
         nn.quantize(
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -183,6 +183,54 @@ class TestUtils(unittest.TestCase):
             mx.eval(logits)
             self.assertEqual(logits.shape, (1, 3, args.vocab_size))
 
+    def test_class_predicate_skips_non_quantizable_modules_in_per_layer_config(self):
+        """load_model must not crash when per-layer quantization config lists a
+        module that lacks to_quantized() — regression for issue #1266 (MoEGate)."""
+        from mlx_lm.models import llama
+
+        args = llama.ModelArgs(
+            model_type="llama",
+            hidden_size=64,
+            num_hidden_layers=2,
+            intermediate_size=128,
+            num_attention_heads=2,
+            rms_norm_eps=1e-5,
+            vocab_size=256,
+        )
+        model_config = {
+            "model_type": "llama",
+            "hidden_size": 64,
+            "num_hidden_layers": 2,
+            "intermediate_size": 128,
+            "num_attention_heads": 2,
+            "rms_norm_eps": 1e-5,
+            "vocab_size": 256,
+        }
+        model = llama.Model(args)
+        model, config = utils.quantize_model(
+            model,
+            model_config,
+            group_size=32,
+            bits=4,
+        )
+
+        # Inject a path for a non-quantizable module (RMSNorm has no to_quantized).
+        # This simulates a pre-quantized config like nemotron_h whose quantization
+        # config includes MoEGate paths.
+        config["quantization"]["model.layers.0.input_layernorm"] = {
+            "group_size": 32,
+            "bits": 4,
+        }
+
+        with tempfile.TemporaryDirectory(dir=self.test_dir) as mlx_path:
+            utils.save_model(mlx_path, model)
+            utils.save_config(config, os.path.join(mlx_path, "config.json"))
+            # Should not raise: ValueError: Unable to quantize model of type RMSNorm
+            loaded, _ = utils.load_model(Path(mlx_path))
+            logits = loaded(mx.array([[1, 2, 3]], dtype=mx.int32))
+            mx.eval(logits)
+            self.assertEqual(logits.shape, (1, 3, args.vocab_size))
+
 
 if __name__ == "__main__":
     unittest.main()


nastya236:
Thanks for flagging this. 

Can you specify the HF model path that hits this error, and paste the quantization section from its config.json? Original Nemotron Ultra nvfp4 does not quantize MoE gate, so I'm curious how the config was generated for the checkpoint you're using.

robertlangdonn:
Model: [`JANGQ-AI/Nemotron-Cascade-2-30B-A3B-JANG_2L`](https://huggingface.co/JANGQ-AI/Nemotron-Cascade-2-30B-A3B-JANG_2L)

You're right that the original nvfp4 doesn't list gate paths — this is a community-quantized checkpoint. The tool used to generate it walked all modules and wrote 23 `backbone.layers.{N}.mixer.gate` entries explicitly into the config even though their values (`bits=8, group_size=64, mode=affine`) are identical to the global default. Redundant but not malformed.

Relevant excerpt from `config.json`:

```json
"quantization": {
  "group_size": 64,
  "bits": 8,
  "mode": "affine",
  "backbone.layers.1.mixer.gate":  { "bits": 8, "group_size": 64, "mode": "affine" },
  "backbone.layers.3.mixer.gate":  { "bits": 8, "group_size": 64, "mode": "affine" },
  "backbone.layers.6.mixer.gate":  { "bits": 8, "group_size": 64, "mode": "affine" },
  ...
  "backbone.layers.51.mixer.gate": { "bits": 8, "group_size": 64, "mode": "affine" }
}
```

263 entries total; 23 are `mixer.gate` paths at the odd/non-consecutive layer indices matching the `hybrid_override_pattern` MoE layout.

The crash is still a robustness gap in the loader: once a path appears in `config["quantization"]`, the old `class_predicate` returns the config dict before checking `hasattr(m, "to_quantized")`, so `nn.quantize` calls `to_quantized()` on `MoEGate` and raises. The fix applies to any pre-quantized config that lists a non-quantizable module path, not just this checkpoint.

nastya236:
Safetensors for the model that you listed has quantized gate weights: https://huggingface.co/JANGQ-AI/Nemotron-Cascade-2-30B-A3B-JANG_2L/blob/main/model-00001-of-00003.safetensors -- see backbone.layers.1.mixer.gate for example. If you apply your change, with quantized gate load_weights should error on shape mismatch. Have tried to use the model with proposed change?

robertlangdonn:
You're right — I pulled the safetensors header and confirmed that the checkpoint contains quantized gate tensors (`gate.weight` stored as packed U32 with companion `.scales` and `.biases`).

That means my proposed change is incorrect for this checkpoint. Skipping `MoEGate` in `class_predicate` leaves the module with its fp16 `weight`, so `load_weights(strict=True)` would then fail when loading the quantized gate tensors. My test only exercised a synthetic re-quantized model and never covered the load path for an already-quantized gate, so it missed this case.

Thanks for catching that.

One thing I noticed while investigating is that the same issue seems to apply to the no-op `to_quantized()` approach discussed in #1266: if the checkpoint contains packed gate weights, leaving the module in its fp16 form still doesn't allow those tensors to load correctly.

Given that the reference Nemotron recipe doesn't quantize the gate, I'm not sure whether the preferred direction here is:

- Support checkpoints with quantized gate weights by making `MoEGate` quantization-aware, or
- Treat gate quantization as unsupported and surface a clearer error.

Happy to work on either approach if there's a preferred direction. Otherwise I'm also fine closing this PR since the original fix doesn't address the actual loading behavior of this checkpoint.

nastya236:
> Treat gate quantization as unsupported and surface a clearer error.

I think it is current behaviour -- we through an error if gate is in the config. 

> Support checkpoints with quantized gate weights by making MoEGate quantization-aware

You definitely can, but it depends on your goal. If you simply want to use the model that you specified above, I would convert the checkpoint so that gate weight is dequantized.

nastya236:
I am closing this PR, if you have a different proposal, feel free to open a new one.

robertlangdonn:
Thanks — that settles it.

Agreed on both points: the current `Unable to quantize` error when the gate appears in the quantization config is reasonable behavior, and for this checkpoint the practical remedy is to dequantize the gate weight given that the reference Nemotron recipe keeps it in full precision.

My proposed change doesn't address the actual loading behavior for this checkpoint, so I'll close this PR rather than pursue it further.

Thanks for the quick review and for pointing out the quantized gate tensors — that was the piece I had missed.

PR #1343, opened 2026-06-03
Apply max_kv_size to KVCache layers returned by make_cache()

robertlangdonn:
## Motivation

Today, `max_kv_size` is not applied to models that implement `make_cache()` — which includes many modern architectures (Llama, Gemma, Qwen, Mistral, and others). The current behavior is documented and preserves each architecture's intended cache strategy. It also means that a user who explicitly requests `max_kv_size` does **not** receive a bounded KV cache for those models.

## What this changes

After calling `model.make_cache()`, plain `KVCache` instances are replaced with `RotatingKVCache(max_size=max_kv_size, keep=4)` — the same pattern already used for models without `make_cache()`. Entries that are already a `RotatingKVCache` (sliding-window layers) or non-KV caches (e.g. SSM/Mamba state) are left unchanged. The guard is `type(c) is KVCache`, so `KVCache` subclasses are intentionally preserved.

## Tradeoff (measured)

In interleaved-attention models, the plain `KVCache` entries are precisely the **global / full-attention layers**. Capping them replaces global attention with a sliding window.

Measured on `mlx-community/gemma-3-1b-it-4bit` (26 layers; global-attention layers at indices 5, 11, 17, 23), with a fact planted at the start of an 837-token prompt:

| Cache | KV memory | Recall of early-context fact |
|---|---|---|
| `make_cache()` default (current) | higher | **preserved** |
| `max_kv_size=128` (this PR) | bounded | **lost** — early tokens evicted |

This is expected when old KV entries are evicted, and it is the same tradeoff `max_kv_size` already makes for models without `make_cache()`. It is not a correctness bug; it is a behavioral consequence of bounding the cache.

## Question for maintainers

The real question is one of intent: when a user explicitly passes `max_kv_size`, should the requested **memory bound** win, or should **architecture-specific cache behavior** win for models that construct their own cache?

If bounding memory consistently (including global-attention layers) is the desired interpretation, this PR implements it. If the current scoping to non-`make_cache()` models is deliberate, I'm happy to gate it behind an opt-in, narrow it, or close it.

The included tests cover the type-replacement logic; the recall measurement above is from a manual run on M3.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -24,12 +24,25 @@ def make_prompt_cache(
 
     Args:
         model (nn.Module): The language model.
-        max_kv_size (Optional[int]): If provided and the model does not have a
-            ``make_cache`` method, a ``RotatingKVCache`` is used with a maximum
-            size of ``max_kv_size``
+        max_kv_size (Optional[int]): If provided, plain :class:`KVCache` entries
+            are replaced with :class:`RotatingKVCache` capped at ``max_kv_size``
+            tokens. This applies whether or not the model defines ``make_cache``.
+            Cache entries that are already a :class:`RotatingKVCache` (e.g.
+            sliding-window layers) or non-KV caches (e.g. SSM state) are left
+            unchanged.
     """
     if hasattr(model, "make_cache"):
-        return model.make_cache()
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1344, opened 2026-06-03
New UI for chat and lora

nastya236:
As the title.

Work for both light / dark themes.
Chat on a single machine:
<p align="center">
  Dark theme
  Light theme
</p>

Lora on 4 m3-ultra:
https://github.com/user-attachments/assets/c4bf85c8-8f13-400e-8a72-3c501a606056

**Code changes**: diff might look large but it is mostly because chat and training pipeline were wrapped with `TrainUI(…)` / with `ChatUI(…)` blocks , main training loop and generation are unchanged. 
Also in distibuted setting I log process (loading model, dataset) only on rank 0, because I felt that it is sufficient. 

**Important note on terminal width**:
For distributed launch, because processes a piped, we can't easily get terminal size. After experimenting with different approaches, I decided that the easiest way to set a correct terminal size will be by adding environmental variables in mlx.launch (see this [PR](https://github.com/ml-explore/mlx/pull/3618)).

### DIFF
--- a/mlx_lm/chat.py
+++ b/mlx_lm/chat.py
@@ -1,9 +1,16 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import sys
 
 import mlx.core as mx
 
+from .cli_ui import (
+    make_console,
+    make_corridor_prompt,
+    print_chat_help,
+    print_header_panel,
+)
 from .generate import stream_generate
 from .models.cache import make_prompt_cache
 from .sample_utils import make_sampler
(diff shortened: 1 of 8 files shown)


angeloskath (approved):
Love it.

Out of curiosity, why not use the `rich` input classes? No idea if they are a better choice. I think using `input` directly makes it easier to have it behave correctly in distributed situations but I was wondering.

ivanfioravanti:
Love it!!!

PR #1345, opened 2026-06-03
Add pipelining for Qwen 3.5 

angeloskath:
As title.


nastya236 (approved):
Looks great, thank you! Tested it with `mlx-community/Qwen3.6-35B-A3B-4bit` on 4 machines.
Second check should be fixed with https://github.com/ml-explore/mlx-lm/pull/1347

PR #1351, opened 2026-06-04
chore: improve mlx-lm maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/test_chat.py, tests/test_gguf.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.


zcbenz:

Issue #1308, opened 2026-05-24
Feature Request: Expose ⁠--kv-bits⁠ and thinking parameters in ⁠mlx_lm.server⁠ for large model inference


Wolfbane1:
Currently, there is a feature disparity between ⁠mlx_lm.generate⁠ and ⁠mlx_lm.server⁠. While the CLI generation tool supports ⁠--kv-bits⁠, ⁠--enable-thinking⁠, and ⁠--thinking-budget⁠, the OpenAI-compatible server (⁠mlx_lm.server⁠) does not expose these arguments in its CLI parser.

This becomes a critical blocker when using large models (like ⁠Qwen3.6-35B⁠ or ⁠32B⁠) on Apple Silicon devices (e.g., 48GB Unified Memory) connected to coding assistants like VS Code (GitHub Copilot, Cline, Continue). Since the server defaults to a 16-bit KV cache, passing large codebases as context quickly saturates the available GPU memory limit set by macOS, causing severe swap paging and hanging the machine. Furthermore, it prevents the use of the reasoning blocks for models that support it when accessed via the server API.

Please expose the following arguments in ⁠server.py⁠ so they can be passed down to the underlying generation pipeline during server startup:

Allowing KV cache quantization at the server level would make ⁠mlx_lm.server⁠ infinitely more viable for local coding agents and long-context RAG applications on Macs with 32GB/48GB of RAM, where 16-bit cache for a 30B+ model is simply too large.

Thanks a lot!

mabaeyens:
Splitting your two asks, since they are at very different stages.

**`--kv-bits`** is implemented and waiting for review in #1353, which adds `--kv-bits`, `--kv-group-size`, `--quantized-kv-start` and `--max-kv-size` to the server. That is the thread to watch.

One caveat that matters for your case specifically. You are on 48GB running Qwen3.6-35B, so you want the memory back. As #1353 currently stands, layers whose cache has no quantized variant are skipped silently, which for a sliding-window or hybrid model is most of them (Gemma 4 26B is 25 of 30 layers). Qwen3.6-35B-A3B is full-attention throughout, so this should not bite you, but if you switch to a hybrid model you would get much less memory back than the flag suggests, with no warning. #1584 implements the missing rotating-cache quantization so the skip stops being necessary, and #1618 adds a guard that fails loudly instead of skipping. I have left a note on #1353.

[...]

PR #1353, opened 2026-06-05
server: add KV cache quantization flags (--kv-bits, --kv-group-size, --quantized-kv-start)

### LINKED ISSUE #615: Feature Request: Add max-kv-size Support to MLX HTTP Server

### LINKED ISSUE #1043: Add KV cache quantization support to server
The server is missing --kv-bits, --kv-group-size, and --quantized-kv-start CLI arguments, so KV cache quantization (already supported by the generate function) can't be used when running the server.

soobrosa:
## Add KV cache quantization flags to `mlx_lm.server`

Closes #1043, closes #615. Covers the `--kv-bits` half of #1308.

Related: #1573 (same `RotatingKVCache` limitation reached from the hybrid-model
direction), #1584 (implements `RotatingKVCache.to_quantized()`), #1618
(library-level capability guard).

`mlx_lm.generate` has supported quantized KV cache for a while, but
`mlx_lm.server` did not expose it, so long-context serving was stuck at fp16 KV,
which OOMs well before the model's real context limit on memory-bound Apple
Silicon.

### What this does

Adds four CLI flags to `mlx_lm.server`:

- `--kv-bits` (default `None`, no quantization)
- `--kv-group-size` (default `64`)
- `--quantized-kv-start` (default `DEFAULT_QUANTIZED_KV_START`, 5000, same
  constant `mlx_lm.generate` and `mlx_lm.cache_prompt` use)
- `--max-kv-size` (default `None`)

These are threaded into the existing `stream_generate` path, reusing the
quantized-KV implementation `mlx_lm.generate` already uses. No change to
`mlx_lm/generate.py` or `mlx_lm/models/cache.py`.

### Reports what quantization will actually apply to

Not every cache class implements `to_quantized()`. `RotatingKVCache` raises
`NotImplementedError`, so on a hybrid model like Gemma 4 (25 sliding-window
layers against 5 full-attention ones) only some layers can be converted today.

Counting layers makes that look fatal, but the bytes say otherwise.
`gemma4_text.py:37` sets `sliding_window: int = 512`, so each rotating layer is
capped at 512 tokens forever while the 5 full-attention layers grow with the
context. Share of KV bytes held by the quantizable full-attention layers: 76% at
8K, 90% at 22K, 97.5% at 100K. @Git-on-my-level measured exactly that on a
Laguna S 2.1 at a 100K prefill (#1043): a 1.46 GB KV4 prefix against ~72 MiB of
bounded BF16 rotating state.

So partial quantization is worth serving, and the thing to avoid is silence
(#1573), not partiality. With `--kv-bits` set, `ModelProvider._load` probes the
cache and:

- **warns**, naming the layer counts and the reason each cache gave, when some
  layers cannot convert
- **raises** when *no* layer can convert, since the request would then have no
  effect at all

This sits next to the `hasattr(c, "merge")` batchability probe that was already
there:

```python
if self.cli_args.kv_bits is not None:
    probe_cache = make_prompt_cache(model, max_kv_size=self.cli_args.max_kv_size)
    if draft_model is not None:
        probe_cache += make_prompt_cache(
            draft_model, max_kv_size=self.cli_args.max_kv_size
        )
    _check_kv_cache_quantizable(
        probe_cache,
        self.cli_args.kv_group_size,
        self.cli_args.kv_bits,
        max_kv_size=self.cli_args.max_kv_size,
    )
```

It attempts the real conversion rather than inferring from the class, which
costs nothing because converting an empty cache short-circuits on
`if self.keys is not None`, and it reports the reason the cache itself raised
rather than guessing at a flag. Entries with no `to_quantized` at all, the
`ArraysCache` used for Mamba and gated-delta-net state, are skipped rather than
counted either way: those are not KV caches and `--kv-bits` does not apply to
them.

Once #1584 lands, rotating caches with `keep=0` convert too, so the warning
stops appearing for hybrid models with no change here.

### `--max-kv-size` and `--kv-bits` together

`make_prompt_cache(max_kv_size=N)` builds `RotatingKVCache(max_size=N, keep=4)`
(`cache.py:37`) for *every* layer, and sink tokens have no quantized variant
before or after #1584. Nothing is quantizable, so this is the case that raises
rather than warns. Filed separately as #1631, since `keep=4` is the only
`keep > 0` in the repo and every model that builds rotating caches itself passes
`keep=0`.

### Design note (batched path)

`BatchGenerator` accepts no `kv_bits`, so the batched path cannot honour a
quantization request at all. When `--kv-bits` is set, `_is_batchable()` returns
`False` and requests are served through the single-sequence generator, which
does support it. This trades batched throughput for a request that is actually
honoured, and `--help` says so.

Once `BatchGenerator` accepts `kv_bits` (#1584), this can narrow to only the
configurations whose caches still cannot be quantized, instead of serializing
every request. Narrowing it before that would silently drop quantization for the
batched requests, which is the failure mode this PR is trying to remove.

### Known interaction: `--prompt-cache-bytes`

`--prompt-cache-bytes` is only enforced on the batched path (`server.py:794-797`
calls `prompt_cache.trim_to`); the sequential path inserts into the LRU cache
with only the `--prompt-cache-size` count applied. Since `--kv-bits` forces
sequential serving here, quantized serving does not enforce
`--prompt-cache-bytes`. #1118 fixes that independently and is worth taking
alongside this. Raised by @Git-on-my-level on #1043.

### Validation

Tested against **Qwen3.6-35B-A3B** (hybrid GDN + MoE, MLX 4-bit) served with
`--kv-bits 4 --kv-group-size 64`:

- short-prompt correctness: OK
- long-context fact recall: a **22.3K-token** prompt (facts up front, question
  after ~16K filler, past `quantized_kv_start`) returned the correct answer, with
  no crash on the hybrid cache layout

That run predates the startup probe, but the generation path it exercised is
unchanged since, and this model's cache is `KVCache` plus `ArraysCache`
(`qwen3_next.py`/`qwen3_5.py`: `ArraysCache(size=2) if l.is_linear else
KVCache()`), so the probe passes for it and every attention layer is quantized.

Suite: `tests/test_server.py` 32 passed, `tests/test_generate.py` and
`tests/test_prompt_cache.py` 48 passed. New coverage: the probe warns and
serves on a hybrid cache, raises when no layer is quantizable, accepts plain
`KVCache`, stays silent for `ArraysCache`, raises on the `--max-kv-size` cache,
`_is_batchable()` is false with `--kv-bits`, and an end-to-end `/v1/completions`
request served with `kv_bits=8`. Formatted
with the pinned `black` 25.1.0 and `isort` 6.0.0 hooks.

This also fixes `tests/test_server.py`, which this PR had broken: `_is_batchable`
reads `cli_args.kv_bits` on every request and `DummyModelProvider` never defined
the new attributes.

### vs #1309

This PR additionally exposes `--kv-group-size`, `--quantized-kv-start` and
`--max-kv-size`, routes through the proven `stream_generate` path rather than
building caches manually, and reports up front which layers a
request will actually apply to.


sunpazed:
I noticed an issue with this PR, where `RotatingKVCache` quantization causes crash in hybrid cache models (ie; Gemma 4). When using models with hybrid caching strategies (like Gemma 4) that include RotatingKVCache instances, the server crashes with NotImplementedError: RotatingKVCache Quantization NYI if KV cache quantization is enabled. 

`mlx_lm.server --model mlx-community/gemma-4-26B-A4B-it-qat-4bit --kv-bits 8`

I believe that the Gemma 4 model architecture uses a hybrid cache strategy, ie; some layers use regular KVCache() (full attention), and some layers use RotatingKVCache() with a sliding window.

soobrosa:
@sunpazed this looks better to me now

mabaeyens:
Came here from #1573 and #1043. Three notes, one of which I think matters before this lands, and none of them are objections to the approach you took in June, which was right for the code as it stood then.

## This closes more issues than it claims

It also adds `--max-kv-size`, which is #615 (open since 2025-11), and covers the `--kv-bits` half of #1308. Worth listing those in the description so they get closed together and nobody files a fourth.

## The silent skip is the part I would change

```python
except NotImplementedError:
    # e.g. RotatingKVCache used by hybrid/sliding-window models has
    # no quantized variant yet; leave that layer unquantized.
    pass
```

For a hybrid model this is most of the network. Gemma 4 26B-A4B builds 25 `RotatingKVCache` layers against 5 `KVCache` (counted on the real model, not from the config), so `--kv-bits 8` would leave 25 of 30 layers at full precision and report nothing. The user asked for a smaller cache, got most of one, and has no way to tell.

That is also the exact complaint in #1573: someone set `--kv-bits`, believed it applied, and did not find out otherwise until much later. @PhilipJohnBasile is building the opposite behaviour in #1618, a capability guard that fails up front rather than skipping layers quietly, for the same reason.

If the skip has to stay as an interim, a one-time warning naming the layer count would at least make it visible.

## #1584 removes the reason for the skip

#1584 implements `RotatingKVCache.to_quantized()` plus `RotatingQuantizedKVCache` and `BatchRotatingQuantizedKVCache`, so the `NotImplementedError` you are catching stops being raised for rotating caches (the one remaining case is `keep > 0`, which Gemma 4 avoids since it passes `keep=0`). @PhilipJohnBasile validated that branch end to end on Gemma 4 26B: 25/5 layers converting to their quantized counterparts, `max_size` preserved, output identical to fp16 across a rotation boundary.

That also narrows your batching decision:

```python
if self.cli_args.kv_bits is not None:
    return False
```

Correct today. After #1584 it is over-broad in one direction and still right in the other. With `--max-kv-size` set the batching path has quantized rotating caches and does not need disabling. Without it, the caches become plain `QuantizedKVCache`, which implements no `merge()`, so batching-with-history cannot work at all and returning `False` remains correct. So the condition could become `kv_bits is not None and max_kv_size is None` rather than a blanket disable, which matters because the current form serialises every request on a server whose whole point is concurrency.

No ownership claim on any of this, it is your PR and you got there first. Happy to answer questions about the #1584 side, or to rebase #1584 around this if maintainers would rather take the server flags first.

soobrosa:
Thanks, this is a fair read and you're right that the silent skip has to go. I've dropped the `generate.py` change from this PR entirely rather than adding a warning to it. Two reasons: the skip is only reachable because `RotatingKVCache.to_quantized()` raises, which #1584 fixes properly, and #1618 is already replacing the `hasattr` check it lives in. A warning here would be a third implementation of the same idea that both of yours would then have to unpick.

What replaces it is a load-time probe in `ModelProvider._load`, right next to the `hasattr(c, "merge")` batchability probe that's already there:

```python
if self.cli_args.kv_bits is not None:
    probe_cache = make_prompt_cache(model, max_kv_size=self.cli_args.max_kv_size)
    ...
    _check_kv_cache_quantizable(
        probe_cache, self.cli_args.kv_group_size, self.cli_args.kv_bits
    )
```

It tries the real `to_quantized()` on each entry rather than guessing from the class, which costs nothing because converting an empty cache short-circuits on `if self.keys is not None`. If anything can't convert, the server refuses to start and names the class and count. Entries with no `to_quantized` at all, i.e. `ArraysCache` for Mamba and gated-delta-net state, are skipped, since those aren't KV caches and `--kv-bits` doesn't apply to them.

@sunpazed, this turns your crash into a startup error today: `--model mlx-community/gemma-4-26B-A4B-it-qat-4bit --kv-bits 8` now reports that 25 of 30 cache layers cannot be quantized instead of serving 5 of them quietly. Once #1584 lands the same probe passes and Gemma 4 gets real quantization, with no further change needed here.

The probe also caught something none of us had flagged. `make_prompt_cache(max_kv_size=N)` builds `RotatingKVCache(max_size=N, keep=4)` at `cache.py:37`, and #1584 refuses `keep > 0`. So `--max-kv-size` together with `--kv-bits` is unquantizable both before and after #1584 for any model without its own `make_cache`; Gemma 4 only escapes because it passes `keep=0`. There's a test for that case specifically. @mabaeyens, worth knowing on your side too, since `--max-kv-size` is the documented route to `RotatingKVCache` and it's hardcoded to `keep=4`.

On the batching condition I'm keeping the blanket disable, and I don't think `kv_bits is not None and max_kv_size is None` is right yet. Today's `BatchGenerator.__init__` takes no `kv_bits` at all, so sending a quantized request down that path would silently ignore quantization, which is the failure mode you're objecting to in the first place. The narrowing becomes correct in the commit that adds those parameters, so it belongs in #1584 or in a follow-up after it, and I'll take whichever you prefer. The comment now says so. Agreed that serialising requests is a real cost; I'd rather name it in `--help`, which it now does, than trade it for a silent one.

Separately, the server tests were broken on this branch and I hadn't noticed: `DummyModelProvider.cli_args` never got the new attributes and `_is_batchable` reads `cli_args.kv_bits` on every request, so `tests/test_server.py` was failing with `AttributeError`. Fixed, with coverage for the probe (unquantizable, quantizable, non-KV state, `--max-kv-size`) and an end-to-end `/v1/completions` request served with `kv_bits=8`.

`tests/test_server.py` 26 passed, `tests/test_generate.py` 26 passed, `tests/test_prompt_cache.py` 22 passed. `black` 25.1.0 and `isort` 6.0.0 per the pinned hooks, clean.

Description updated to close #1043 and #615 and to cover the `--kv-bits` half of #1308.

On ordering: this is now one source file plus tests, and it no longer touches `generate.py`, `models/cache.py`, or `tests/test_generate.py`, so it doesn't depend on #1584 and won't conflict with #1584, #1618, or #1595. Same argument you made to @PhilipJohnBasile about not being held hostage. @awni @angeloskath, this is the server-flag half of #1043 and is reviewable on its own.

PhilipJohnBasile:
Reviewed the current refactor. The load-time probe is the right server boundary and makes this PR cleanly independent. Proposed scope split: #1353 owns CLI/server startup validation and the sequential fallback; #1618 owns cache capability APIs, runtime validation, and `BatchGenerator` quantized-cache support. The `keep=4` case remains an intentional true negative: `make_prompt_cache(max_kv_size=...)` cannot be quantized until sink-token support exists. After #1584/#1618, server batching can be narrowed when the server wires the KV options into `BatchGenerator`. No duplicate code is needed here.

mabaeyens:
You are right and I was wrong on the batching condition. On main today `BatchGenerator.__init__` takes no `kv_bits`, so my narrowing would have sent a quantized request down a path that ignores the quantization, which is the exact failure I came here to complain about. Blanket disable is correct as this PR stands, and naming the cost in `--help` is the right trade.

One correction going the other way, in case you only read the `cache.py` half of #1584: the `generate.py` half already adds `kv_bits`, `kv_group_size` and `quantized_kv_start` to `BatchGenerator.__init__`, quantizes in `_make_new_cache()` while the cache is still empty, and quantizes externally supplied caches in `insert_segments()`. So the parameters you are waiting for arrive with #1584 itself, not after it.

On where the narrowing goes: **a follow-up, after both are in.** #1584 is already +1157 across four files with no review on it, and bolting `server.py` onto a cache PR would make the thing harder to review for no gain, since the wiring cannot be tested until your file lands anyway. I will open the follow-up once #1353 and #1584 are both merged, or earlier if a maintainer would rather see it stacked.

Two things the follow-up has to get right, both of which I only found while checking your probe.

The first is `--quantized-kv-start`. It defaults to `DEFAULT_QUANTIZED_KV_START`, 5000, in your argparse block, and #1584's `BatchGenerator` refuses `quantized_kv_start != 0` when `kv_bits` is set. Per-job caches are built once, empty, at insertion time, so there is no per-step re-check for the offset threshold to ever trigger; only immediate quantization works on that path. Which means a plain `mlx_lm.server --kv-bits 8`, no other flags, would hit that raise if the narrowing only looked at `max_kv_size`. The real condition is closer to `kv_bits is not None and quantized_kv_start != 0`, with `max_kv_size` not in it at all, and here is why.

The second is your `keep=4` catch, which is correct and which I confirmed (`cache.py:37` hardcodes `keep=4`, and #1584 raises on `keep > 0`, so `--max-kv-size` with `--kv-bits` is unquantizable in the sequential path before and after my PR). But it inverts on the batched path. `BatchGenerator._make_new_cache()` builds `RotatingKVCache(max_size=...)` with the default `keep=0`, so after #1584 that combination quantizes fine there. Your probe is right today, because `--kv-bits` forces sequential and sequential goes through `make_prompt_cache`. The moment the narrowing lands, the probe would refuse to start on a config the batched path can actually serve. Not asking you to change anything now (it would be dead code until the wiring exists), just flagging it so the `keep=4` assumption does not get baked into the error message as a permanent truth about the flag.

Sink tokens in `RotatingQuantizedKVCache` are the fix for the sequential half and I am not attempting them in #1584. Worth its own issue at some point.

Nothing here should hold you up. This PR is two files now, does not touch `cache.py` or `generate.py`, and closes #1043, #615 and half of #1308 on its own. It should go in whenever a maintainer gets to it, ahead of mine if that is what happens.

soobrosa:
Both fair, and the `BatchGenerator` correction is mine to take: I only read the `cache.py` half of #1584, so I had the parameters arriving after it rather than with it. Noted.

Pushed `3602862` for the one thing that was actionable now, the error message. It no longer asserts anything about a flag; it reports the `NotImplementedError` the cache itself raised, and mentions `--max-kv-size` only when that flag is actually set:

```
--kv-bits 8 was requested but 25 of 30 cache layers cannot be quantized:
25 x RotatingKVCache (RotatingKVCache Quantization NYI). Serving with only
the remaining layers quantized would ignore most of the request without
reporting it. Serve without --kv-bits.
```

So after #1584 the same probe reports whatever the new classes report, `keep>0` included, without this file having an opinion about it. There's a test asserting the flag hint is absent when the flag wasn't passed, so it can't creep back in.

On `--quantized-kv-start`: good catch, and it's the more dangerous of the two, since `mlx_lm.server --kv-bits 8` with no other flags is the common case and 5000 is the default that `mlx_lm.generate` and `mlx_lm.cache_prompt` already use. I'd rather not change the default here just to suit a path that can't be reached yet; matching the other entry points is worth more today than pre-adapting to the wiring. When the follow-up lands, the options are to pass `quantized_kv_start=0` on the batched path or to fall back to sequential when it's nonzero, and that's a decision for whoever writes it with both PRs in hand.

Agreed on follow-up-after-both for the narrowing, and it's yours. Two notes for whoever picks it up: `quantized_kv_start != 0` and `keep>0` pull in opposite directions across the two paths, so the probe likely has to know which generator will serve the request rather than validating one cache shape at load time. That's an argument for the capability API in #1618 over anything in `server.py`.

Sink tokens deserve their own issue, agreed. Happy to file it referencing #1584 and this thread if you'd rather not add another number to your stack.

@awni @angeloskath: this is now two files (`server.py`, `tests/test_server.py`), touches neither `generate.py` nor `models/cache.py`, and closes #1043 and #615 plus the `--kv-bits` half of #1308. Per CONTRIBUTING it needs one review to land. The one policy question worth a maintainer's opinion: with `--kv-bits` set and a model whose cache can't honour it, this refuses to start rather than warning and serving partially quantized. Happy to switch it to a warning if you'd rather, but the silent-partial behaviour is what #1573 was reported for.

@sunpazed if you still have the Gemma 4 checkout handy, `mlx_lm.server --model mlx-community/gemma-4-26B-A4B-it-qat-4bit --kv-bits 8` should now refuse at startup with the message above instead of crashing mid-request. A confirmation would be useful.

PhilipJohnBasile:
Confirming both open items from your comment:

**Sink tokens** — please file it, thank you for the offer. Referencing #1584 and
this thread is exactly right; no need to hold it for us.

**The narrowing follow-up is ours.** Tracking it against #1618 once #1584 and
this PR are both in — I'll rebase #1618 on top and use it for the wiring
decision, per your note that `quantized_kv_start != 0` and `keep > 0` pull in
opposite directions across the sequential and batched paths. Agreed that means
the probe needs to know which generator will actually serve the request rather
than validating one cache shape at load time, which is the argument for putting
that in the capability API rather than in `server.py`. I'll fold that
constraint into #1618's design before pinging for review.

Thanks for taking the `BatchGenerator` correction and for `3602862` — the
narrowed message (reporting the cache's own `NotImplementedError`, only
mentioning `--max-kv-size` when it's actually set) is the right shape, and a
regression test pinning the flag hint's absence is exactly the kind of guard
this file needs given how many entry points touch it.

soobrosa:
Note for reviewers, so pre-existing failures do not get attributed here.

On Apple GPU gen-17 (M5), `tests/test_generate.py` has 8 deterministic failures on vanilla `main`, caused by TF32 in fp32 GEMM rather than by any PR. `MLX_ENABLE_TF32=0 pytest tests/test_generate.py` takes it to 31 passed. Cause and analysis are in ml-explore/mlx#3897; @mabaeyens ran that down on #1584.

Relevant here only because six of the eight are batch-versus-single comparisons, and earlier revisions of this PR did touch `generate.py`. The current revision does not: it is `server.py` and `tests/test_server.py` only, so those failures cannot originate from it.

My runs were on an M3 Pro (`applegpu_g15s`, MLX 0.31.2), which does not exhibit the issue: `tests/test_server.py` 31 passed, `tests/test_generate.py` and `tests/test_prompt_cache.py` 48 passed, no env var needed. So I have not verified this PR on gen-17 silicon, and if you review there, the `test_server.py` numbers are the ones that cover it.

soobrosa:
Changed the policy in `d0d1479`: partial quantization now warns and serves, and only a request that can't be applied to **any** layer refuses.

@Git-on-my-level's report on #1043 is why. I had adopted refuse-on-any-unquantizable-layer from the argument that Gemma 4 leaves "25 of 30 layers at full precision", and I repeated that here without checking it. It counts layers, and layers are the wrong unit. `gemma4_text.py:37` sets `sliding_window: int = 512`, so each rotating layer is capped at 512 tokens forever while the 5 full-attention layers grow with the context. Share of KV bytes in the quantizable full-attention layers:

| context | quantizable share |
|---|---|
| 8K | 76% |
| 22K | 90% |
| 100K | 97.5% |

His measurement is that 97.5% on real hardware: Laguna S 2.1, 100,047-token prefill, 1.46 GB KV4 prefix against ~72 MiB of bounded BF16 rotating state. Laguna builds the same shape (`laguna.py:315`, rotating for sliding layers, `KVCache` for full attention), so my probe was refusing to start on a configuration he had already validated end to end through an agent client.

New behaviour:

```
WARNING: --kv-bits 8 applies to 5 of 30 cache layers. 25 will stay unquantized:
25 x RotatingKVCache (RotatingKVCache Quantization NYI). On sliding-window
architectures these hold window-bounded state, while the quantized layers grow
with the context.
```

and, when nothing is convertible at all, still an error:

```
ValueError: --kv-bits 8 cannot be applied to any of the 30 cache layers:
30 x RotatingKVCache (RotatingKVCache Quantization NYI). Serving would ignore
the request entirely. Serve without --kv-bits, or without --max-kv-size, which
builds a RotatingKVCache.
```

@mabaeyens, this doesn't reintroduce what you objected to in #1573. That report was about silence: someone set `--kv-bits`, it partly applied, and nothing said so. The warning names the counts and the cache's own reason. Refusing also had a cost I'd missed, which is that it denies the user the 76-97% they could have had.

Second thing from his comment, and it's specific to this PR: `--prompt-cache-bytes` is only enforced on the batched path (`server.py:794-797` calls `trim_to`); the sequential path inserts with only the `--prompt-cache-size` count applied. Since `--kv-bits` forces sequential here, quantized serving currently doesn't enforce it, which is exactly the control you want when you're holding a 1.46 GB prefix. #1118 fixes it in 8 lines and is worth taking alongside this. Noted in the description rather than fixed here, since it's an independent bug.

Also worth a maintainer's attention: #1309, #1476 and this PR all address #1043, and #1476 is what @Git-on-my-level actually ran. I'm not attached to mine winning. Happy to fold in whatever's best from the others or to close in favour of one, but four open PRs on one issue is the reason none of them have moved.

`tests/test_server.py` 32 passed on an M3 Pro. The new tests cover warn-and-serve on a hybrid cache, raise-when-nothing-convertible, and silence for recurrent state.

zcbenz:

PR #1354, opened 2026-06-05
Fix LFM2 MoE routing to match the HF reference (sigmoid gating)

ykhrustalev:
LFM2 MoE routing does not match the HF Transformers reference.

Lfm2Moe is a sigmoid-gated MoE, but the router uses softmax, folds expert_bias into the combination weights, and does not apply routed_scaling_factor.

This change makes it match the reference: the router scores are sigmoid(gate(x)), expert_bias is used only to select the top-k experts, the combination weights are gathered from the unbiased sigmoid scores, and they are scaled by routed_scaling_factor. routed_scaling_factor is added to ModelArgs with a default of 1.0, so existing configs are unaffected.

Validated against the HF reference on LiquidAI/LFM2.5-8B-A1B.

### DIFF
--- a/mlx_lm/models/lfm2_moe.py
+++ b/mlx_lm/models/lfm2_moe.py
@@ -36,6 +36,7 @@ class ModelArgs(BaseModelArgs):
     conv_bias: bool
     conv_L_cache: int
     rope_theta: float = 1000000.0
+    routed_scaling_factor: float = 1.0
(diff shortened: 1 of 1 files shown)


ykhrustalev:
cc @Blaizzy

nastya236 (approved):
Looks great, thank you!

PR #1355, opened 2026-06-05
tuner: fix NaN gradients in LoRA fine-tuning with mask_prompt=True

sunilquantumwire-web:
# Upstream PR Description

## Title

`Fix NaN gradients in LoRA fine-tuning with mask_prompt=True on chat-format datasets`

## Summary

Three small fixes in `mlx_lm/tuner/` to prevent `0/0` NaN losses when
fine-tuning chat models with `mask_prompt=True`:

1. `datasets.py::ChatDataset.process` — compute the loss-mask offset from
   the prompt-only tokenization (no `add_generation_prompt`) instead of
   the `add_generation_prompt=True` tokenization, which is not a prefix
   of the full tokenization for templates like Gemma 4.
2. `trainer.py::iterate_batches` — clamp the loss-mask offset to the
   truncated sequence length so a long prompt that gets truncated doesn't
   produce an empty loss mask.
3. `trainer.py::default_loss` — use `mx.maximum(ntoks, 1)` to avoid
   divide-by-zero (we can't call `.item()` inside the traced loss
   function).

## Problem

`ChatDataset.process` (and `CompletionsDataset.process`) compute the
loss-mask offset by re-applying the chat template to `messages[:-1]` with
`add_generation_prompt=True`. For chat templates where the
`add_generation_prompt=True` suffix is NOT a strict prefix of the full
template's prefix (notably Gemma 4, which appends
`<|channel>thought\n<channel|>` to nudge the model into a thinking
mode), the computed offset is **past the end of the full tokenization**.
The resulting loss mask is empty, and the trainer's `default_loss`
divides `0 / 0` → NaN.

Repro on Gemma 4 12B with a 28-token example:

```
>>> offset = len(apply_chat_template(messages[:-1], add_generation_prompt=True))
>>> len(apply_chat_template(messages))
28
>>> offset
29
```

A second, related issue: `iterate_batches` truncates sequences to
`max_seq_length` but never updates the per-example offset. So even with a
correct offset from `ChatDataset`, a long prompt that exceeds
`max_seq_length` produces an empty loss mask for that example, with the
same `0/0` NaN.

## Fix

`ChatDataset.process` now uses `add_generation_prompt=False`, which
returns a strict prefix of the full tokenization for any template that
follows the OpenAI / HuggingFace convention (Llama, Mistral, Gemma,
Qwen, Phi, etc.). A defensive `min(len(prompt_ids), len(tokens))` clamp
protects against custom templates that may render the prompt portion
longer than the full tokenization for any reason.

`iterate_batches` converts the `offsets` tuple to a list and clamps each
offset to the corresponding truncated length.

`default_loss` uses `mx.maximum(ntoks, 1)` to avoid the `0/0` division
in a way that stays inside MLX's tracing rules. The downstream
aggregation `all_losses += losses * toks` correctly drops empty batches
from the running average, so the max-with-1 trick is sound.

## Tests

Added `tests/test_chat_dataset.py` (in this PR) with 7 tests:

```
test_loss_mask_is_nonempty_and_loss_is_finite ... ok
test_long_example_also_correct ... ok
test_mask_prompt_false_is_unchanged ... ok
test_offset_is_in_bounds ... ok
test_offset_is_strictly_positive ... ok
test_offset_matches_prompt_only_tokenization ... ok
test_response_portion_decodes_to_assistant_turn ... ok
----------------------------------------------------------------------
Ran 7 tests in 9.608s
OK
```

4 of the 7 tests fail on the original code with messages like
`AssertionError: 29 not less than 28 : offset 29 is past end of tokens`.
The test file is included in the PR and uses only the
`transformers`/`mlx` packages that are already in `mlx_lm`'s test
dependencies (the Gemma 4 tokenizer is loaded from a local model
directory; the test can be made to use any chat-format model by changing
one path).

## End-to-end verification

Ran a 30-iteration LoRA training on `gemma-4-12B` with `mask_prompt=true`,
batch size 1, grad accum 8, lr 1e-4:

| Code version | Iter 1 | Iter 10 | Iter 20 | Iter 30 |
|---|---:|---:|---:|---:|
| Original | Val 1.327 | Val 2.062 | Val **NaN** | Val 2.534 |
| Patched  | Val 1.327 | Val 2.062 | Val **1.392** | Val 2.534 |

Train loss curves are identical (the patch affects only the loss mask,
not the model). Validation diverges at iter 20 on the original code
(NaN) and recovers a sensible value on the patched code. A full
500-iter run on the original code would NaN permanently and diverge the
model.

## Diff size

```
 datasets.py | 60 ++++++++++++++++++++++++++++++++--------------------
 trainer.py  | 27 +++++++++++++++++++++++++++
 2 files changed, 87 insertions(+), 21 deletions(-)
```

Most of the line count is comments explaining the fix. Functional code
changes are ~10 lines per file.

## Backward compatibility

- The `mask_prompt=True` path now produces correct loss masks for
  templates that follow the OpenAI / HuggingFace convention, which is
  the same set of templates the existing code intended to support.
- The defensive `min()` clamp ensures the fix is safe for any custom
  template, even ones where the prompt portion might be longer than the
  full tokenization for some reason.
- The `default_loss` `mx.maximum` change is mathematically equivalent
  for any batch with `ntoks > 0` (which is the overwhelming common case)
  and only differs for empty-mask batches (which previously produced
  NaN).

## Why split into two files?

`datasets.py` fix is the primary cause of the NaN. The `trainer.py` fix
addresses a related but distinct case (long prompts that get truncated).
Either fix alone is insufficient: the `datasets.py` fix is needed for
Gemma 4 specifically, and the `trainer.py` fix is needed for any chat
example with a long prompt (e.g. tool-calling or long-context
instruction-following). The `default_loss` change is a belt-and-suspenders
guard for any future case where an empty mask might still occur.

## Commit messages

Two suggested commits, in this order:

```
datasets.py: fix mask_prompt offset for chat templates with non-prefix gen prompts

ChatDataset.process (and CompletionsDataset.process) computed the
loss-mask offset by re-applying the chat template to messages[:-1]
with add_generation_prompt=True. For templates where the
add_generation_prompt=True suffix differs from the full template's
prefix (notably Gemma 4, which appends a <|channel>thought\n<channel|>
hint), the computed offset was past the end of the full tokenization.
The resulting loss mask was empty, and the trainer's default_loss
divided 0/0 -> NaN.

Use add_generation_prompt=False instead, which returns a strict prefix
of the full tokenization for templates that follow the OpenAI /
HuggingFace convention. Add a defensive min() clamp for custom
templates.

trainer.py: clamp loss-mask offset after truncation and avoid 0/0 in default_loss

iterate_batches truncates sequences to max_seq_length but never updates
the per-example offset, so a long prompt that exceeds max_seq_length
produces an empty loss mask for that example. Clamp the offset to the
truncated length.

default_loss divides by ntoks, which is zero when the entire batch has
an empty mask. Use mx.maximum(ntoks, 1) to avoid 0/0 -> NaN while
staying inside MLX's tracing rules. The downstream aggregation does
losses * toks, so empty batches are correctly dropped from the running
average.
```

### DIFF
--- a/mlx_lm/tuner/datasets.py
+++ b/mlx_lm/tuner/datasets.py
@@ -63,15 +63,37 @@ class ChatDataset:
             return_dict=False,
         )
         if self.mask_prompt:
-            add_generation_prompt = messages[-1].get("role") == "assistant"
-            offset = len(
-                self.tokenizer.apply_chat_template(
-                    messages[:-1],
-                    tools=tools,
-                    add_generation_prompt=add_generation_prompt,
-                    return_dict=False,
-                )
+            # Compute the loss-mask offset by tokenizing just the prompt
+            # portion (messages[:-1]) with add_generation_prompt=False.
+            # The result is a strict prefix of the full tokenization, and
+            # the offset points to the start of the assistant role header
+            # in the full sequence.
+            #
+            # We deliberately do NOT use add_generation_prompt=True here.
+            # Some chat templates (notably Gemma 4) inject extra tokens
+            # under add_generation_prompt=True (e.g. a
+            # "<|channel>thought\n<channel|>" hint) that are NOT present
+            # in the full tokenization. Computing the offset from such a
+            # tokenization gives a value PAST the end of the full
+            # sequence, which produces an empty loss mask and triggers
+            # NaN gradients in the trainer.
+            #
+            # Using add_generation_prompt=False works for any template
+            # where the prompt portion (without the role header) is a
+            # strict prefix of the full tokenization, which is the OpenAI
+            # / HuggingFace convention followed by Llama, Mistral, Gemma,
+            # Qwen, Phi, and most other modern templates.
+            prompt_ids = self.tokenizer.apply_chat_template(
+                messages[:-1],
+                tools=tools,
+                add_generation_prompt=False,
+                return_dict=False,
             )
+            # Defensive clamp: if the template renders the prompt portion
+            # longer than the full tokenization for any reason (it should
+            # not, but could for a custom template), fall back to no
+            # masking rather than producing an empty loss mask.
+            offset = min(len(prompt_ids), len(tokens))
             return (tokens, offset)
         else:
             return (tokens, 0)
@@ -114,14 +136,19 @@ class CompletionsDataset:
             messages, tools=tools, return_dict=False
         )
         if self.mask_prompt:
-            offset = len(
-                self.tokenizer.apply_chat_template(
-                    messages[:-1],
-                    tools=tools,
-                    add_generation_prompt=True,
-                    return_dict=False,
-                )
+            # See ChatDataset.process for the rationale. We compute the
+            # offset from the prompt-only tokenization
+            # (add_generation_prompt=False) and clamp to the length of
+            # the full tokenization. This is robust against templates
+            # whose add_generation_prompt=True suffix differs from the
+            # full template's prefix (e.g. Gemma 4).
+            prompt_ids = self.tokenizer.apply_chat_template(
+                messages[:-1],
+                tools=tools,
+                add_generation_prompt=False,
+                return_dict=False,
             )
+            offset = min(len(prompt_ids), len(tokens))
             return (tokens, offset)
 
         return (tokens, 0)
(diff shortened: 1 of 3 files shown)


nastya236:
Do you mind sharing dataset name you used to reproduce this issue?

zcbenz:

PR #1359, opened 2026-06-07
feat: add return_logprobs and return_token_ids to batch_generate

HaoXuAI:
Add two new options to batch_generate():

- return_logprobs: returns per-token log-probability of the sampled token for each generated sequence. Useful for reinforcement learning (RLOO, PPO) where behavior log-probabilities are needed for importance weighting.

- return_token_ids: returns the raw token IDs alongside the decoded text, avoiding an extra encode() round-trip when callers need both.

Both are off by default for backward compatibility. Results are stored in new optional fields on BatchResponse.


angeloskath (approved):
👍

PR #1361, opened 2026-06-07
fix(tokenizer): skip channel thinking detection for Gemma 4 (#1352)

### LINKED ISSUE #1352: Gemma 4 with thinking enabled produces only reasoning field, content is empty in API response
Description
    
    When enable_thinking is not explicitly set to false (or the model's default has_thinking is true), the /v1/chat/completions response for Gemma 4 models contains only the reasoning field. The content field is always missing, which breaks OpenAI API compatibility.
    
    Steps to Reproduce
    
    1. Start server with any Gemma 4 model (e.g. mlx-community/gemma-4-31b-it-4bit)
    2. Send a chat completion request WITHOUT chat_template_kwargs
    3. Observe response: content is absent, only reasoning is present
    
    Expected Behavior
    
    Gemma 4's thinking mode should produce both reasoning (thinking process) and content (final answer), similar to how OpenAI's o1 or DeepSeek R1 handle it.
    
    Workaround
    
    Pass "chat_template_kwargs": {"enable_thinking": false} in the request body to disable thinking. This works correctly and produces proper content.
    
    Related Issues
    
    - #1308 (Feature Request: Expose thinking parameters in mlx_lm.server CLI)
    - #1221 (enable_thinking injection breaks deepseek_v32 chat_template)
    - #914 (Adjusting thinking/reasoning on a per-message basis)

0xSoftBoi:
## Root cause

Gemma 4 has `<|channel>` and `<channel|>` in its vocabulary, so `_infer_thinking` sets `has_thinking=True` and the `TokenizerWrapper` auto-injects `enable_thinking=True` into every `apply_chat_template` call.

Unlike DeepSeek-R1 / Qwen3 — where only the reasoning portion is inside the thinking channel and the final answer follows after it — Gemma 4 with `enable_thinking=True` wraps its **entire response** inside `<|channel>thought...<channel|>`. The server's state machine then classifies all generated tokens as `reasoning_text` and returns an empty `content` field.

Confirmed by inspecting the Gemma 4 chat template:
- `enable_thinking=True` injects `<|think|>` into the system turn (mode switch)
- The model wraps its complete output in `<|channel>thought\n...\n<channel|>`
- Nothing is generated after `<channel|>`, so `text = ""`

The workaround (`chat_template_kwargs: {"enable_thinking": false}`) works because it bypasses the auto-inject and lets the template default (no channel wrapping) take effect.

## Fix

Guard the `<|channel>/<channel|>` thinking detection with a check for `<|think|>` in the vocabulary — the Gemma 4 model family's mode-switch token. This makes `has_thinking=False` for Gemma 4, preventing the auto-inject and restoring correct content output.

Users who explicitly want thinking output for Gemma 4 can still opt in via `--chat-template-args '{"enable_thinking": true}'` — they'll get the raw channel-wrapped output as content.

## Verification

```bash
# Before fix: content="" reasoning="2+2 equals 4"
# After fix:  content="2+2 equals 4" reasoning=null
mlx_lm.server --model mlx-community/gemma-4-e4b-it-4bit &
curl -s http://127.0.0.1:8080/v1/chat/completions \
  -d '{"messages":[{"role":"user","content":"What is 2+2?"}],"max_tokens":50}' \
  | python3 -c "import sys,json; r=json.load(sys.stdin); print(r['choices'][0]['message'])"
```

Fixes #1352

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -269,14 +269,22 @@ def _infer_thinking(tokenizer):
 
     # Multi token thinking modes
     if "<|channel>" in vocab and "<channel|>" in vocab:
-        think_start = "<|channel>thought"
-        think_end = "<channel|>"
-        return (
-            think_start,
-            think_end,
-            tuple(tokenizer.encode(think_start, add_special_tokens=False)),
-            tuple(tokenizer.encode(think_end, add_special_tokens=False)),
-        )
+        # Gemma 4 uses <|channel>/<channel|> for conversation structure but
+        # also has <|think|> as a mode-switch token. With enable_thinking=True,
+        # Gemma 4 wraps its entire response (not just the reasoning portion)
+        # inside <|channel>thought...<channel|>, which causes the server to
+        # classify all content as reasoning and return an empty content field.
+        # Skip thinking detection for these models; users can still opt in
+        # explicitly via chat_template_args.
+        if "<|think|>" not in vocab:
+            think_start = "<|channel>thought"
+            think_end = "<channel|>"
+            return (
+                think_start,
+                think_end,
+                tuple(tokenizer.encode(think_start, add_special_tokens=False)),
+                tuple(tokenizer.encode(think_end, add_special_tokens=False)),
+            )
 
     return (None, None, None, None)
 
--- a/tests/test_tokenizers.py
+++ b/tests/test_tokenizers.py
@@ -109,6 +109,30 @@ class TestTokenizers(unittest.TestCase):
         self.assertIsNone(tokenizer.think_start_id)
         self.assertIsNone(tokenizer.think_end_id)
 
+    def test_thinking_gemma4(self):
+        # Gemma 4 has <|channel>/<channel|> in vocab alongside <|think|> (a mode-switch
+        # token). With enable_thinking=True the model wraps its entire response inside
+        # <|channel>thought...<channel|>, leaving content empty. We skip the channel
+        # thinking detection for these models so has_thinking stays False by default.
+        tokenizer_repo = "mlx-community/gemma-4-e4b-it-4bit"
+        tokenizer = load_tokenizer(tokenizer_repo)
+        self.assertFalse(tokenizer.has_thinking)
+        self.assertIsNone(tokenizer.think_start)
+
+    def test_find_negative_start(self):
+        # Regression test for #1326: rfind_think_start should not raise IndexError
+        # when the prompt is shorter than the think-token lookahead window (11 tokens).
+        tokenizer_repo = "mlx-community/Qwen3-4B-4bit"
+        tokenizer = load_tokenizer(tokenizer_repo)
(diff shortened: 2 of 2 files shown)


0xSoftBoi:
Rebased onto current `main` — the branch had gone conflicted, so this was unmergeable as it stood. It is clean now.

While resolving, I dropped one of my own tests: `main` has since added `test_find_token`, a pure unit test covering the same `#1326` negative-start regression that my `test_find_negative_start` covered — but without downloading a model. `main`'s version is better, so I removed mine rather than keep a redundant networked test. What remains is just the subject of this PR:

```python
def test_thinking_gemma4(self):
    tokenizer = load_tokenizer("mlx-community/gemma-4-e4b-it-4bit")
    self.assertFalse(tokenizer.has_thinking)
    self.assertIsNone(tokenizer.think_start)
```

The fix is a single guard. Gemma 4 has `<|channel>`/`<channel|>` in vocab, so it currently trips the multi-token thinking path — but it also has `<|think|>` as a mode-switch token, and with `enable_thinking=True` it wraps its *entire* response in `<|channel>thought...<channel|>`. The server then classifies all of it as reasoning and returns empty content. Gating on `<|think|>` not being in vocab skips detection for those models only:

| vocab | behaviour |
|---|---|
| `<\|channel>` + `<\|think\|>` (Gemma 4) | detection skipped |
| `<\|channel>`, no `<\|think\|>` (gpt-oss) | **unchanged** |
| no `<\|channel>` | unchanged |

The middle row is the one that matters for regressions — gpt-oss style models keep the existing behaviour.

Happy to close this if channel-thinking is being reworked more broadly, or if it is simply not wanted — just let me know either way and I will get out of the queue.

0xSoftBoi:
Closing this myself — it has been open since June without review, and I would rather not leave it sitting in the queue while the project is prioritising elsewhere. No hard feelings; the triage call is a reasonable one.

Leaving the diagnosis here in case someone hits the same symptom and finds this thread. The symptom is that Gemma 4 returns an **empty `content` field** through the server, with everything landing in reasoning instead.

The cause is that Gemma 4 has `<|channel>` / `<channel|>` in its vocab, so `_infer_thinking` matches the multi-token thinking path and treats the channel markers as reasoning delimiters. But Gemma 4 also has `<|think|>` as a separate mode-switch token, and with `enable_thinking=True` it wraps its *entire* response — not just the reasoning portion — inside `<|channel>thought ... <channel|>`. So the whole response is classified as reasoning and the content comes back empty.

The one-line discriminator is `<|think|>`:

| vocab | correct behaviour |
|---|---|
| `<\|channel>` + `<\|think\|>` (Gemma 4) | skip channel-thinking detection |
| `<\|channel>`, no `<\|think\|>` (gpt-oss) | keep it — unchanged |
| no `<\|channel>` | unaffected |

Anyone wanting the workaround today can set it explicitly via `chat_template_args` rather than relying on inference.

The branch stays up if it is ever useful. Thanks for the project.

PR #1364, opened 2026-06-07
fix(server): keep tool_call arguments as JSON string (OpenAI spec)

### LINKED ISSUE #1065: Bug: tool_calls arguments converted to dict breaks chat templates that expect string (str + dict TypeError)
When a conversation history includes prior tool calls (assistant message with `tool_calls`), `process_message_content()` in `server.py` converts `arguments` from a JSON string
 to a Python dict via `json.loads()`. However, some model chat templates (e.g. DepSeek-R1) perform direct string concatenation with `arguments`, causing a `TypeError: can only
 concatenate str (not "dict") to str` → HTTP 404.

   ## Steps to Reproduce

   ```bash
   curl -X POST http://localhost:8081/v1/chat/completions \
     -d '{
       "model": "mlx-community/DeepSeek-R1-0528-Qwen3-8B-8bit",
       "messages": [
         {"role": "system", "content": "You are a helpful assistant."},
         {"role": "user", "content": "run ls"},
         {"role": "assistant", "content": null, "tool_calls": [{"id": "call_1", "type": "function", "function": {"name": "exec", "arguments": "{\"command\": \"ls\"}"}}]},
         {"role": "tool", "tool_call_id": "call_1", "content": "file1.txt"}
       ],
       "tools": [{"type": "function", "function": {"name": "exec", "description": "Run shell", "parameters": {"type": "object", "properties": {"command": {"type": "string"}},
 "required": ["command"]}}}],
       "max_tokens": 10
     }'
 ```

 Root Cause

 process_message_content() (server.py ~line 165):

 ```python
   func["arguments"] = json.loads(args)  # converts string → dict
 ```

 The DeepSeek-R1 Jinja2 template then does:

 ```
   '```json\n' + tool['function']['arguments'] + '\n```'
 ```

 Fails because arguments is now a dict, not a string.

 Fix

 Re-serialize back to string instead of leaving as dict:

 ```python
   if isinstance(args, str):
       func["arguments"] = json.dumps(json.loads(args), ensure_ascii=False)
   elif isinstance(args, dict):
       func["arguments"] = json.dumps(args, ensure_ascii=False)
 ```

 Environment

 - mlx-lm: 0.31.2
 - Model: mlx-community/DeepSeek-R1-0528-Qwen3-8B-8bit
 - macOS arm64

0xSoftBoi:
## Problem
`process_message_content()` called `json.loads()` on the `arguments` field of incoming assistant `tool_call` messages, converting the JSON string to a Python dict. Chat templates that call `json.loads()` themselves (deepseek_v32, DSML-format models) then received a dict and raised `TypeError`, silently breaking multi-turn tool use.

The OpenAI spec mandates `arguments` is always a JSON-encoded string. Chat templates that need a dict can use the `|fromjson` Jinja filter.

## Fix
Remove the pre-decode. The comment explains why for future readers.

Fixes #1065

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -143,11 +143,10 @@ def process_message_content(messages):
         elif content is None:
             message["content"] = ""
 
-        if tool_calls := message.get("tool_calls"):
-            for tool_call in tool_calls:
-                if func := tool_call.get("function"):
-                    if args := func.get("arguments"):
-                        func["arguments"] = json.loads(args)
+        # Leave tool_call arguments as a JSON string per the OpenAI spec.
+        # Pre-decoding to a dict breaks chat templates that call json.loads()
+        # themselves (e.g. deepseek_v32, DSML-format models) because
+        # json.loads() on an already-decoded dict raises TypeError.
 
 
 @dataclass


0xSoftBoi:
Closing to resubmit — branches got cross-contaminated during parallel commits, wrong diff on this PR.

PR #1366, opened 2026-06-07
fix(cache): bool('False') == True corrupts BatchRotatingKVCache on reload

### LINKED ISSUE #1250: BatchRotatingKVCache: rotated=False reloads as True from saved prompt cache

0xSoftBoi:
## Problem
`bool('False')` is `True` in Python. `BatchRotatingKVCache.meta_state.setter` stored the rotation flag as the string `'True'` or `'False'`, then loaded it back with `bool(v[3])`. Any cache saved with `rotated=False` reloaded as `rotated=True`, silently corrupting sliding-window KV reuse after every `save_prompt_cache` / `load_prompt_cache` round-trip.

## Fix
```python
- self.rotated = bool(v[3])
+ self.rotated = v[3] == 'True'
```

## Test
Added `test_batch_rotating_cache_meta_state_roundtrip` in `tests/test_prompt_cache.py` that round-trips both `rotated=False` and `rotated=True` and asserts the flag survives.

Fixes #1250

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1312,7 +1312,7 @@ class BatchRotatingKVCache(_BaseCache):
             int,
             v[:3],
         )
-        self.rotated = bool(v[3])
+        self.rotated = v[3] == "True"
 
     def is_trimmable(self):
         return self._offset < self.max_size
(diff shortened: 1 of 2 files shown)


0xSoftBoi:
Closing to resubmit — wrong diff, branches swapped.

nastya236:
Thanks, it was addressed in https://github.com/ml-explore/mlx-lm/pull/1251

PR #1369, opened 2026-06-07
fix(server): log confirmation when speculative decoding is active

### LINKED ISSUE #1131: Misleading 'Use --speculative-model to enable' log when --draft-model is active
## Description

When starting `mlx_lm.server` with `--draft-model`, the startup banner prints:

```
Speculative decoding: Use --speculative-model to enable
```

This is confusing because:
1. `--speculative-model` is not a valid flag (not in `--help`)
2. `--draft-model` IS the correct flag and IS active
3. The message implies speculative decoding is disabled when it's actually working

## Steps to reproduce

```bash
python -m mlx_lm server \
  --model <target_model> \
  --draft-model <draft_model> \
  --num-draft-tokens 3
```

Observe the startup log includes the misleading message.

## Expected behavior

When `--draft-model` is specified, the startup banner should confirm speculative decoding is active, e.g.:

```
Speculative decoding: enabled (draft model: <path>, num_draft_tokens: 3)
```

## Environment

- mlx-lm version: 0.31.2
- macOS

0xSoftBoi:
## Problem
When `--draft-model` is provided and tokenizers match, the server emits no startup message. Users have no way to confirm speculative decoding is actually running short of adding `--log-level DEBUG`.

## Fix
Add an `INFO` log confirming the draft model path when setup succeeds, symmetric with the existing warning when tokenizers don't match.

Fixes #1131

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -366,6 +366,10 @@ class ModelProvider:
                     "Draft model tokenizer does not match model tokenizer. "
                     "Speculative decoding may not work as expected."
                 )
+            else:
+                logging.info(
+                    f"Speculative decoding active: draft model '{draft_model_path}'"
+                )
 
         # Compute batchability
         is_batchable = draft_model is None


0xSoftBoi:
Closing after reading the full thread on #1131 — eyupcanakman confirmed the message described in that issue does not exist in mlx-lm itself (grepped all tags and the PyPI wheel). It appears to come from jjang-ai/vmlx, a separate wrapper. This PR adds logging for a different scenario and doesn't address what the reporter experienced.

PR #1377, opened 2026-06-08
[Documentation] top_k interval error

devYRPauli:
`apply_top_k` validates its argument with `0 < top_k < vocab_size`, but the error message it raises advertises the closed interval `(0, vocab_size]`:

```python
if not isinstance(top_k, int) or not (0 < top_k < vocab_size):
    raise ValueError(
        f"`top_k` has to be an integer in the (0, {vocab_size}] interval,"
        f" but is {top_k}."
    )
```

So `top_k == vocab_size` is rejected even though the message says it is valid. That value is a well-defined no-op: `mx.argpartition(-logprobs, kth=top_k-1)[..., top_k:]` selects an empty index set, so nothing is masked and the logprobs pass through unchanged. The server only lower-bounds `top_k` (`min_val=0`), and `make_sampler` wires in `apply_top_k` whenever `top_k > 0`, so a request with `temperature > 0` and `top_k == vocab_size` ("consider every token") aborts generation with a `ValueError` that contradicts its own message.

This changes the bound to inclusive (`<=`) so the guard matches the documented `(0, vocab_size]` interval. `top_k` greater than `vocab_size` is still rejected. Extended `test_apply_top_k` with the `top_k == vocab_size` no-op case (fails before, passes after).

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -141,7 +141,7 @@ def apply_top_k(
     vocab_size = logprobs.shape[-1]
     if not isinstance(top_k, int) or not (0 < top_k < vocab_size):
         raise ValueError(
-            f"`top_k` has to be an integer in the (0, {vocab_size}] interval,"
+            f"`top_k` has to be an integer in the (0, {vocab_size}) interval,"
             f" but is {top_k}."
         )
     mask_idx = mx.argpartition(-logprobs, kth=top_k - 1, axis=-1)[..., top_k:]


mlx_lm/sample_utils.py
```diff
@@ -139,7 +139,7 @@ def apply_top_k(
         top_k (int): Top k tokens to sample from.
     """
     vocab_size = logprobs.shape[-1]
-    if not isinstance(top_k, int) or not (0 < top_k < vocab_size):
+    if not isinstance(top_k, int) or not (0 < top_k <= vocab_size):
```
nastya236:
```suggestion
    if not isinstance(top_k, int) or not (0 < top_k < vocab_size):
```

mlx_lm/sample_utils.py
```diff
@@ -139,7 +139,7 @@ def apply_top_k(
...
         raise ValueError(
             f"`top_k` has to be an integer in the (0, {vocab_size}] interval,"
```
nastya236:
```suggestion
            f"`top_k` has to be an integer in the (0, {vocab_size}) interval,"
```

nastya236 (changes requested):
top-k with full vocabulary is not a no-op in mlx (see mx.argpartition implementation). Semantically it is, but it does a full sort under the hood. So it is better to fix the bracket in docs and raise an error.

devYRPauli:
Good point — you're right that `top_k == vocab_size` isn't actually free, since the `argpartition` still does a full sort. Reverted the guard to the exclusive `0 < top_k < vocab_size` and just corrected the error message to advertise the matching `(0, vocab_size)` interval, so the message no longer contradicts the bound. Updated the test to assert that `top_k == vocab_size` raises with the exclusive-interval message. Thanks!

nastya236:
Thanks. I would just remove the test.

devYRPauli:
Done, removed the test.

nastya236 (approved):

PR #1385, opened 2026-06-09
Fix CVE-2026-5843: gate model_file execution behind trust_remote_code

dogukanveziroglu:
If a model's config.json has a `model_file` key, `load_model` imports and runs
  that Python file straight from the model directory (added in #830). It happens
  on a plain `load()` with no way to turn it off, so loading a model can execute
  arbitrary code. This is CVE-2026-5843 / GHSA-9m9w-53g9-47c4: it was reported
  against Docker Model Runner, which embeds mlx-lm, but the root cause is here in
  the library and main is still unguarded. The tokenizer path already puts custom
  code behind `trust_remote_code`, so this does the same for `model_file`.

  `load_model` now takes `trust_remote_code` (default False) and raises with a
  clear message if a config wants a `model_file` without it. Setting
  `MLX_LM_TRUST_REMOTE_CODE=1` works too for the CLI tools. The flag is threaded
  through `load()` and `sharded_load()`, and models without a `model_file` are
  unchanged.

  Tests cover the four cases: blocked by default with the custom file never
  executed, enabled by the argument, enabled by the env var, and a normal model
  still loading.ed

### DIFF
--- a/mlx_lm/benchmark.py
+++ b/mlx_lm/benchmark.py
@@ -73,6 +73,11 @@ def setup_arg_parser():
         default=0,
         help="Delay between each test in seconds (default: 0)",
     )
+    parser.add_argument(
+        "--trust-remote-code",
+        action="store_true",
+        help="Enable trusting remote code for tokenizer/model loading.",
+    )
     return parser
 
 
@@ -94,7 +99,11 @@ def main():
 
     if group.size() > 1:
         model, tokenizer, config = sharded_load(
-            model_path, pipeline_group, tensor_group, return_config=True
+            model_path,
+            pipeline_group,
+            tensor_group,
+            return_config=True,
+            trust_remote_code=args.trust_remote_code,
         )
     else:
         model, tokenizer, config = load(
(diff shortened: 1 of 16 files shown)


kingroryg:
Worth flagging though that it's an mlx-lm only change and only covers the `model_file` import path. It does not address vulnerabilities in the mlx core loaders themselves (https://github.com/ml-explore/mlx/security/advisories/GHSA-mxqr-6xw9-3x8j) e.g., parser bugs reachable from `mx.load(...)` on an untrusted file, where `trust_remote_code` never enters the picture.

dogukanveziroglu:
> Worth flagging though that it's an mlx-lm only change and only covers the `model_file` import path. It does not address vulnerabilities in the mlx core loaders themselves (https://github.com/ml-explore/mlx/security/advisories/GHSA-mxqr-6xw9-3x8j) e.g., parser bugs reachable from `mx.load(...)` on an untrusted file, where `trust_remote_code` never enters the picture. 

Yeah, this one's just the model_file path (CVE-2026-5843). The heap overflow in the core loaders is a separate bug and trust_remote_code won't help there. Can't see the advisory tho, happy to help if you can add me.

angeloskath (approved):
Thanks! I removed the env var and added the argument everywhere. Also added it to all scripts and is now shared with the hf tokenizer when passed in the scripts.

PR #1392, opened 2026-06-11
fix(server): honor prompt cache byte limit

agisilaos:
## Summary

`mlx_lm.server` parsed `--prompt-cache-bytes`, but the server-created `LRUPromptCache` was only initialized with `prompt_cache_size`. That left the LRU cache itself with its default effectively-unbounded byte limit, so sequential serving could retain prompt cache entries beyond the configured byte budget.

This wires `prompt_cache_bytes` into `LRUPromptCache(max_bytes=...)` when the server constructs the prompt cache, keeping byte-limit enforcement centralized for all cache insertion paths.

## Changes

- Add `make_lru_prompt_cache()` to construct the server prompt cache with both sequence and byte limits.
- Use the helper from `run()` so `--prompt-cache-bytes` is honored by the LRU cache itself.
- Add a regression test showing a server-created prompt cache evicts entries according to the configured byte budget.

## Alternatives considered

- Adding another trim call only in the sequential serve path. This was rejected because passing the byte limit into `LRUPromptCache` keeps enforcement centralized and applies consistently to every insertion path.

## Notes

- Fixes the server prompt-cache byte-limit enforcement gap related to #883.
- Related to #854 and #1015, which track broader Metal OOM recovery behavior. This PR reduces one server-side OOM trigger but does not attempt to make arbitrary Metal OOMs recoverable.
- Overlaps with #1118.

## Testing

- `python3 -m unittest tests.test_server.TestLRUPromptCache`
- `black --check mlx_lm/server.py tests/test_server.py`
- `isort --profile=black --check-only mlx_lm/server.py tests/test_server.py`

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1740,12 +1740,19 @@ def run(
     handler_class=APIHandler,
 ):
     group = mx.distributed.init()
-    prompt_cache = LRUPromptCache(model_provider.cli_args.prompt_cache_size)
+    prompt_cache = make_lru_prompt_cache(model_provider.cli_args)
     response_generator = ResponseGenerator(model_provider, prompt_cache)
     if group.rank() == 0:
         _run_http_server(host, port, response_generator)
     else:
         response_generator.join()
+
+
+def make_lru_prompt_cache(args):
+    max_bytes = getattr(args, "prompt_cache_bytes", None)
+    if max_bytes is None:
+        max_bytes = 1 << 63
+    return LRUPromptCache(args.prompt_cache_size, max_bytes=max_bytes)
 
 
 def main():
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -17,6 +17,7 @@ from mlx_lm.server import (
     Response,
     ResponseGenerator,
     _process_control_tokens,
+    make_lru_prompt_cache,
 )
 from mlx_lm.utils import load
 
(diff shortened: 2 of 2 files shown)


Veritas-7:
I can independently confirm this gap, and that the fix direction in this PR is sufficient. Observations from running `mlx_lm.server` on the v0.31.x line and against current main:

**Confirmed unfixed upstream**

- On main (`f4f3b57`) and in v0.31.2/v0.31.3, `run()` still constructs `LRUPromptCache(model_provider.cli_args.prompt_cache_size)` without `max_bytes`.
- The in-insert byte eviction in `models/cache.py` (`while self._n_bytes > self.max_bytes`) therefore never runs when `--prompt-cache-bytes` is set; only the admission-time `trim_to(n_bytes=total - active)` applies, and segment/finish-time inserts push the stored cache past the flag.

**Reproduction**

- `mlx_lm.server --model <a 35B MoE hybrid-attention model, 4-bit> --prompt-cache-size 24 --prompt-cache-bytes 8589934592`
- Workload: agentic tool-loop traffic with ~80K-token prompts.
- Server log reached `Prompt Cache: 16 sequences, 13.64 GB`; the stats grew from 8.93 GB to 13.64 GB in about two minutes (assistant +2, user +1 entries after a single admission trim).

**Fix idea verified**

- We run a small wrapper that subclasses `LRUPromptCache` to pass `max_bytes=<--prompt-cache-bytes>` before calling `main()`. Under the same workload the stored cache peaked at 8.26 GB over 25 stats samples, at or below the 8.59 GB cap. Wiring the flag, as this PR does, is sufficient.
- `CacheOrder.pop()` evicts the assistant bucket first whenever it is at least as populated as the user bucket, which suits tool loops fine: the user checkpoint survives eviction.

**Notes**

- With the cap wired, the admission-time `trim_to` and the in-insert loop double-enforce harmlessly — the in-insert loop runs first, so the later trim mostly no-ops.

**Offer**

- We have a regression test ready: it constructs the cache the same way `run()` does, inserts fake non-trimmable leaves past the cap, and asserts `nbytes <= max_bytes` and that the user-type entry survives. Happy to push it to this branch if the author wants it.

Since v0.31.3 still ships the gap, it would be great if a maintainer could consider rebasing/merging #1392 (or #1118).

PR #1394, opened 2026-06-11
fix(mistral): parse JSON-list tool calls

### LINKED ISSUE #1374: Mistral tool parser rejects the JSON-list output format, dropping valid tool calls (e.g. Ministral-8B-Instruct-2410)
### Summary

`mlx_lm.server` drops valid tool calls from Mistral models whose chat template emits the call as a JSON list, for example `Ministral-8B-Instruct-2410`. The model produces a correct `[TOOL_CALLS] [{"name": ..., "arguments": ...}]` sequence, but `tool_parsers/mistral.py` only recognizes the newer `name[ARGS]{...}` shape. Parsing fails, the call is discarded, and the assistant message comes back with no `tool_calls`.

### Environment

- mlx-lm 0.31.3 and current `main`
- Python 3.12, macOS arm64
- Affected: Mistral checkpoints whose template renders tool calls as a JSON list. Confirmed on `mlx-community/Ministral-8B-Instruct-2410-4bit`; `Mistral-7B-Instruct-v0.3` and `Mixtral` use the same shape.
- Not affected: `Devstral-Small-2-...-2512`, which renders `[TOOL_CALLS]<name>[ARGS]<json>`.

### Reproduction (parser only, version-independent)

```python
from mlx_lm.tool_parsers.mistral import parse_tool_call

# What Ministral-8B-Instruct-2410's own template renders, i.e. what the model emits:
parse_tool_call('[{"name": "calculator", "arguments": {"expression": "2+3"}}]')
# -> ValueError: Could not parse tool call from: [{"name": "calculator", ...}]

# The newer shape parses fine:
parse_tool_call('calculator[ARGS]{"expression": "2+3"}')
# -> {'name': 'calculator', 'arguments': {'expression': '2+3'}}
```

You can confirm the model really emits the list shape by rendering its own template:

```python
from transformers import AutoTokenizer
tok = AutoTokenizer.from_pretrained("mlx-community/Ministral-8B-Instruct-2410-4bit")
tools = [{"type": "function", "function": {"name": "calculator",
          "properties": {"expression": {"type": "string"}}, "required": ["expression"]}}}]
msgs = [
    {"role": "user", "content": "What is 2+3?"},
    {"role": "assistant", "content": "", "tool_calls": [
        {"id": "abcdefghi", "type": "function",
         "function": {"name": "calculator", "arguments": "{\"expression\": \"2+3\"}"}}]},
]
print(tok.apply_chat_template(msgs, tools=tools, tokenize=False))
# -> ... [TOOL_CALLS] [{"name": "calculator", "arguments": "...", "id": "abcdefghi"}] ...
```

### Root cause

`mlx_lm/tool_parsers/mistral.py` matches a single regex:

```python
_tool_call_regex = re.compile(r"\s*(\w+)\[ARGS\]\s*(\{.*\})", re.DOTALL)
```

This only covers the `name[ARGS]{json}` form. The list form `[{"name": ..., "arguments": ...}]` never matches, so `parse_tool_call` raises `ValueError`. In `server.py`, `ToolCallFormatter.__call__` catches that and drops the call with a `Failed to parse tool call` warning, so the client gets an assistant message with no `tool_calls`.

### Note on visibility

On current `main`, Mistral tool calls are usually closed by EOS with no end marker, so they are dropped before the parser is ever reached (see #1307 and #1373). This parser gap is the next thing in line: once the call does reach the parser, the list-format output still fails. The reproduction above calls the parser directly, so it is independent of that path and of the mlx-lm version.

### Suggested direction

Accept both shapes in `parse_tool_call`: try the `[ARGS]` regex first, then fall back to parsing a JSON array of `{"name", "arguments"}` objects. I have a small patch and a `tests/test_tool_parsing.py` case ready and am happy to open a PR if that approach sounds right.

### Related

- #1307 and #1373 (tool calls dropped before the parser when the call is closed by EOS)
- #1375 (Mistral v3 tool-call id length constraint, the next step of the same round-trip)

agisilaos:
## Summary

Fixes #1374.

Adds support for Mistral-family tool call payloads emitted as JSON lists after `[TOOL_CALLS]`, while preserving the existing `name[ARGS]{...}` parser path.

## Changes

- Parse JSON-list Mistral tool call payloads.
- Decode `arguments` when templates render it as a JSON string.
- Preserve optional string `id` fields from parsed tool calls.
- Add regression tests for object and string argument forms.

## Alternatives considered

- Changing server-side tool call handling was broader than necessary because the failure is isolated to the Mistral parser.
- Replacing the existing regex path was rejected to preserve support for Devstral-style `name[ARGS]{...}` output.

## Notes

Verified the affected Mistral-family chat templates render assistant tool calls as JSON lists after `[TOOL_CALLS]`, including Ministral, Mistral-7B-Instruct-v0.3, Mistral-Nemo, and Mixtral-8x22B-Instruct-v0.1.

## Screenshots

Not applicable; no UI changes.

## Testing

- `python3 -m unittest tests.test_tool_parsing`
- `python3 -m unittest discover tests/`
- `pre-commit run --files mlx_lm/tool_parsers/mistral.py tests/test_tool_parsing.py`

### DIFF
--- a/mlx_lm/tool_parsers/mistral.py
+++ b/mlx_lm/tool_parsers/mistral.py
@@ -11,10 +11,48 @@ tool_call_start = "[TOOL_CALLS]"
 tool_call_end = ""
 
 
+def _parse_json_list(text: str):
+    try:
+        tool_calls = json.loads(text)
+    except json.JSONDecodeError as e:
+        raise ValueError(f"Could not parse tool call from: {text}") from e
+
+    if not isinstance(tool_calls, list):
+        raise ValueError(f"Could not parse tool call from: {text}")
+
+    parsed = []
+    for tool_call in tool_calls:
+        if not isinstance(tool_call, dict):
+            raise ValueError(f"Could not parse tool call from: {text}")
+
+        name = tool_call.get("name")
+        arguments = tool_call.get("arguments", {})
+        if not isinstance(name, str):
+            raise ValueError(f"Could not parse tool call from: {text}")
+
+        if isinstance(arguments, str):
+            try:
+                arguments = json.loads(arguments)
+            except json.JSONDecodeError as e:
+                raise ValueError(f"Could not parse tool call from: {text}") from e
+
+        if not isinstance(arguments, dict):
+            raise ValueError(f"Could not parse tool call from: {text}")
+
+        result = {"name": name, "arguments": arguments}
+        if "id" in tool_call:
+            if not isinstance(tool_call["id"], str):
+                raise ValueError(f"Could not parse tool call from: {text}")
+            result["id"] = tool_call["id"]
+        parsed.append(result)
+
+    return parsed
+
+
 def parse_tool_call(text: str, tools: Any | None = None):
     match = _tool_call_regex.search(text)
     if match is None:
-        raise ValueError(f"Could not parse tool call from: {text}")
+        return _parse_json_list(text)
     func_name = match.group(1)
     func_args = json.loads(match.group(2))
     return dict(name=func_name, arguments=func_args)
--- a/tests/test_tool_parsing.py
+++ b/tests/test_tool_parsing.py
@@ -313,6 +313,33 @@ class TestToolParsing(unittest.TestCase):
         ]
         self.assertEqual(tool_calls, expected)
 
+    def test_mistral_json_list_tool_call(self):
+        test_case = '[{"name": "calculator", "arguments": {"expression": "2+3"}}]'
+        tool_calls = mistral.parse_tool_call(test_case, None)
+        expected = [
+            {
+                "name": "calculator",
+                "arguments": {"expression": "2+3"},
+            }
+        ]
+        self.assertEqual(tool_calls, expected)
+
+    def test_mistral_json_list_tool_call_with_string_arguments(self):
+        test_case = (
+            '[{"name": "calculator", '
+            '"arguments": "{\\"expression\\": \\"2+3\\"}", '
+            '"id": "abcdefghi"}]'
+        )
+        tool_calls = mistral.parse_tool_call(test_case, None)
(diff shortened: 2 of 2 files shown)


u7k4rs6 (commented):
Reviewed the patch end-to-end — it's correct and ready to merge.

**The bug is real.** The existing parser only matches `name[ARGS]{json}` (Devstral style). Ministral-8B-Instruct-2410, Mistral-7B-v0.3, Mixtral, and Mistral-Nemo all emit `[{"name":..., "arguments":...}]`, which hits the `None` match branch, raises `ValueError`, and gets swallowed by `ToolCallFormatter.__call__` in `server.py` as a warning — silently dropping the tool call.

**The fix is correct.** Key checks:

| Concern | Status |
|---|---|
| Return type mismatch (dict vs list) | `ToolCallFormatter.__call__` already has `if not isinstance(parsed, list): parsed = [parsed]` — both paths work |
| `arguments` as JSON string | Handled — templates like Mistral-Nemo render arguments as an escaped JSON string |
| Optional `id` preservation | Handled — `_format` does `tc.pop("id", None) or uuid4()`, so preserving it is correct |
| Old `name[ARGS]{...}` path unchanged | Regex is still tried first; fallback only on `match is None` |
| Multiple calls in one list | `_parse_json_list` returns all of them; caller `extend`s correctly |
| Empty list `[]` | Returns `[]`, caller adds nothing — correct |
| Error messages consistent | All raise `ValueError` matching the original contract |
| Test coverage | Existing test (line 57) covers regex path; two new tests cover JSON-dict and JSON-string-arguments forms |

No issues found. The patch is minimal, surgical, and the tests are correct. LGTM.

michalk8 (approved):
I've added some extra tests, thank you, LGTM!

PR #1396, opened 2026-06-11
Fix Gemma 3n shared-KV layers with batched caches

### LINKED ISSUE #1384: Gemma 3n: mlx_lm.server returns garbage output (shared-KV layers incompatible with batched cache)
### Summary

### Reproduction

```bash

       "max_tokens":60}'
```

```bash
```

### Environment

- mlx-lm 0.31.3
- Apple Silicon (M3)

### Notes

discobot:
Fixes #1384.

The garbage output (symptom 2 there) turns out to be query RoPE positions rather than KV alignment, wrong in two independent ways:

- The shared-KV layers read `cache.offset` after the designated concrete layer has already advanced it in the same forward pass, so queries are roped `L` positions too far. This part isn't batch-specific: on main, one-shot prefill and stepwise decoding disagree even with the regular cache (max logit diff ~1.5 on a tiny random model; ~4e-6 with this fix).
- The non-shared layers rope queries after `update_and_fetch` using an `offset` captured beforehand — but the batch caches store `offset` as an `mx.array` and advance it in place, so the captured value aliases the post-update offset and queries again land at `position + L`. The int offsets of the non-batched caches are immune, which is why `mlx_lm.generate` stayed coherent.

Fix: `cache.state[:2]` and `cache.offset - L` in the shared branch; rope queries before updating the cache in the non-shared branch (the order other models use).

The regression test (tiny random gemma3n, no downloads) checks prefill == stepwise, batched B=1 == non-batched, and left-padded B=2 == per-row runs across sliding-window rotation. Re-ran the issue's repro with `mlx-community/gemma-3n-E2B-it-lm-4bit`: `batch_generate` now matches `mlx_lm.generate`.

### DIFF
--- a/mlx_lm/models/gemma3n.py
+++ b/mlx_lm/models/gemma3n.py
@@ -126,16 +126,20 @@ class Gemma3nAttention(nn.Module):
         queries = self.q_proj(x)
         queries = queries.reshape(B, L, -1, self.head_dim)
         queries = self.q_norm(queries)
+        queries = queries.transpose(0, 2, 1, 3)
 
-        offset = 0
         if self.is_kv_shared_layer and cache is not None:
-            # For shared layers, retrieve KV from the designated cache layer
-            keys, values = cache.state
-            offset = cache.offset
-
+            # For shared layers, retrieve KV from the designated cache layer.
+            # That cache was already updated earlier in this forward pass, so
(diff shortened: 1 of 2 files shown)


zcbenz:
Looks like automatic PR as author sent 70+ PRs in a few days.

PR #1399, opened 2026-06-12
fix: shorten generated tool-call ids to 9 alphanumeric chars (fixes #1375)

### LINKED ISSUE #1375: mlx_lm.server assigns 36-char UUID tool-call ids, which Mistral v3 templates reject (length-9 constraint)
### Summary

When a tool call comes back without an id, `mlx_lm.server` assigns `str(uuid.uuid4())`. Mistral v3 chat templates require tool-call ids to be exactly 9 alphanumeric characters and raise otherwise. A multi-turn tool exchange with these models therefore fails on the follow-up request: rendering the prior assistant tool call through the template raises, and the server returns HTTP 404.

### Environment

- mlx-lm 0.31.3 and current `main`
- Python 3.12, macOS arm64
- Affected: Mistral checkpoints whose template enforces the 9-char id rule. Confirmed on `mlx-community/Ministral-8B-Instruct-2410-4bit`; `Mistral-7B-Instruct-v0.3`, `Mixtral`, and `Mistral-Nemo` share the same template family.
- Not affected: `Devstral-Small-2-...-2512`, whose template has no id-length constraint.

### Reproduction (template only, version-independent)

```python
import uuid
from transformers import AutoTokenizer

tok = AutoTokenizer.from_pretrained("mlx-community/Ministral-8B-Instruct-2410-4bit")
tools = [{"type": "function", "function": {"name": "calculator",
          "properties": {"expression": {"type": "string"}}, "required": ["expression"]}}}]

bad = str(uuid.uuid4())          # 36 chars, what server.py assigns
msgs = [
    {"role": "user", "content": "What is 2+3?"},
    {"role": "assistant", "content": "", "tool_calls": [
        {"id": bad, "type": "function",
         "function": {"name": "calculator", "arguments": "{\"expression\": \"2+3\"}"}}]},
    {"role": "tool", "tool_call_id": bad, "name": "calculator", "content": "5"},
]
tok.apply_chat_template(msgs, tools=tools, tokenize=False)
# -> TemplateError: Tool call IDs should be alphanumeric strings with length 9!

# A 9-char alphanumeric id renders cleanly:
good = uuid.uuid4().hex[:9]      # e.g. "0819261ae"
# ... same messages with `good` instead of `bad` -> renders fine
```

End to end this shows up as HTTP 404 on the second `/v1/chat/completions` call (the one that carries the prior tool call and its result), because the server renders the history through `apply_chat_template` and the template raises.

### Root cause

`ToolCallFormatter._format` in `server.py`:

```python
tc_id = tc.pop("id", None) or str(uuid.uuid4())
```

`str(uuid.uuid4())` is 36 characters with hyphens, which violates the Mistral v3 constraint of 9 alphanumeric characters.

### Suggested direction

Two options, and I would like a maintainer's view before sending a patch:

1. Make the generated id 9 alphanumeric characters globally, for example `uuid.uuid4().hex[:9]`. This satisfies the Mistral templates and stays valid for templates that do not constrain the id.
2. Generate the constrained id only for templates that require it, leaving the current id elsewhere.

Option 1 is simpler and I have not found a template that rejects a short alphanumeric id, but option 2 is more conservative. Happy to open a PR with whichever you prefer, plus a `tests/test_server.py` case.

### Note on ordering

This is the last step of the Mistral tool-calling round-trip: a follow-up turn is only reached after the first turn returns a tool call, which today also depends on #1307 and #1373 (the EOS flush) and on the parser accepting the model's output (sibling issue). Reported separately because the root cause, the file, and the fix are distinct, and the reproduction above hits the template directly.

### Related

- #1307 and #1373
- #1374 (Mistral JSON-list tool-call parser format, the earlier step of the same round-trip)

u7k4rs6:
## Summary

- `ToolCallFormatter._format` in `server.py` falls back to `str(uuid.uuid4())` when a tool call has no `id` field
- `str(uuid.uuid4())` is 36 characters with hyphens; Mistral v3 chat templates enforce a 9-character alphanumeric constraint and raise `TemplateError` when this is violated
- This causes `apply_chat_template` to raise on the second turn of any multi-turn tool exchange with affected models, which the server surfaces as HTTP 404

## Change

```python
# before
tc_id = tc.pop("id", None) or str(uuid.uuid4())
# after
tc_id = tc.pop("id", None) or uuid.uuid4().hex[:9]
```

`uuid.uuid4().hex[:9]` yields 9 lowercase hex characters (e.g. `0819261ae`) — valid under the Mistral v3 constraint and accepted by all other templates tested.

## Affected models

`mlx-community/Ministral-8B-Instruct-2410-4bit`, `Mistral-7B-Instruct-v0.3`, `Mixtral`, `Mistral-Nemo` (any checkpoint sharing the v3 template family). `Devstral-Small-2-...-2512` is unaffected as its template has no id-length constraint.

Closes #1375

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -59,7 +59,7 @@ class ToolCallFormatter:
         self._streaming = streaming
 
     def _format(self, tc):
-        tc_id = tc.pop("id", None) or str(uuid.uuid4())
+        tc_id = tc.pop("id", None) or uuid.uuid4().hex[:9]
         tc["arguments"] = json.dumps(tc["arguments"], ensure_ascii=False)
         out = {
             "function": tc,


zcbenz:

PR #1400, opened 2026-06-12
feat(batch_generate): add return_logprobs for per-token log probabilities

### LINKED ISSUE #1358: feat: expose per-token logprobs from batch_generate()
## Motivation

`batch_generate()` currently returns decoded text but discards per-token log-probabilities. These are needed for reinforcement learning workflows (RLOO, PPO, DPO) where the behavior policy's log-probs are required for importance weighting.

The underlying `GenerationBatch.Response` already carries `logprobs` (full vocab log-prob vector) — this is just about surfacing them through the high-level API.

## Proposal

Add two optional flags to `batch_generate()`:

```python
response = batch_generate(
    model, tokenizer, prompts,
    max_tokens=256,
    return_logprobs=True,    # per-token log-prob of the sampled token
    return_token_ids=True,   # raw token IDs (avoids re-encoding)
)

# response.logprobs: List[List[float]] — one list per prompt
# response.token_ids: List[List[int]] — one list per prompt
```

Both default to `False` for backward compatibility.

## Use Case

I'm training RLOO (Leave-One-Out policy gradient) on Apple Silicon using MLX. Each training step samples K rollouts per prompt — `batch_generate` with continuous batching gives ~10× speedup over sequential generation. But I need the behavior log-probs for each sampled token to compute importance weights.

Currently I have to use `BatchGenerator` directly and manually collect `r.logprobs[r.token]` from each response. Exposing this through `batch_generate` would make RL training much more accessible.

u7k4rs6:
## Summary

- Adds `return_logprobs: bool = False` parameter to `batch_generate()` (non-breaking — defaults off)
- When enabled, `BatchResponse.logprobs` is `List[List[mx.array]]` — one list per prompt, one full-vocabulary log-prob array per generated token
- Adds `logprobs: Optional[List[List[mx.array]]] = None` field to `BatchResponse` (keyword arg with default, backward-compatible)

The logprobs are already computed inside `GenerationBatch` at every step and available on `GenerationBatch.Response.logprobs`; this PR simply surfaces them through `batch_generate()` without any extra computation.

Mirrors `GenerationResponse.logprobs` from `stream_generate()`.

## Motivation

Closes #1358. The primary use case is RL training (e.g. GRPO/PPO) where you need the log probability of each generated token under the current policy:

```python
response = batch_generate(model, tokenizer, prompts, return_logprobs=True)
# response.logprobs[i][t] → mx.array of shape [vocab_size]
# log p of sampled token: response.logprobs[i][t][token_id]
```

## Test plan

- [ ] Existing `test_batch_generate_function_with_logits_processors` still passes (no regression — `return_logprobs` defaults to `False`)
- [ ] `BatchResponse.logprobs` is `None` when `return_logprobs=False`
- [ ] `BatchResponse.logprobs[i]` has the same length as the number of generated tokens for prompt `i`
- [ ] Each `logprobs[i][t]` has shape `[vocab_size]` and sums (in prob space) to 1

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1875,13 +1875,20 @@ class BatchResponse:
     A data object to hold a batch generation response.
 
     Args:
-        texts: (List[str]): The generated text for each prompt.
+        texts (List[str]): The generated text for each prompt.
         stats (BatchStats): Statistics about the generation.
+        caches (Optional[List[List[Any]]]): The prompt caches, if requested.
+        logprobs (Optional[List[List[mx.array]]]): Per-token log probabilities
+            for each prompt. Each inner list contains one full-vocabulary
+            ``mx.array`` of log probabilities per generated token. Only
+            populated when ``return_logprobs=True`` is passed to
+            :func:`batch_generate`.
     """
 
     texts: List[str]
     stats: BatchStats
     caches: Optional[List[List[Any]]]
+    logprobs: Optional[List[List[mx.array]]] = None
 
 
 def batch_generate(
@@ -1892,6 +1899,7 @@ def batch_generate(
     max_tokens: Union[int, List[int]] = 128,
     verbose: bool = False,
     return_prompt_caches: bool = False,
+    return_logprobs: bool = False,
     **kwargs,
 ) -> BatchResponse:
     """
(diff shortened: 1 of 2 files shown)


zcbenz:

Issue #1391, opened 2026-06-11
Model type diffusion_gemma not supported.


WSINTRA:
Tried the following. 
$mlx_lm.server --model "mlx-community/diffusiongemma-26B-A4B-it-8bit"              

Got an error. 

> 2026-06-10 22:32:55,816 - INFO - HTTP Request: GET https://huggingface.co/api/models/mlx-community/diffusiongemma-26B-A4B-it-8bit/revision/main "HTTP/1.1 200 OK"
> Fetching 13 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 13/13 [00:00<00:00, 5221.29it/s]
> Download complete: : 0.00B [00:00, ?B/s]              Exception in thread Thread-1 (_generate):                                                                                                                                                          | 0/13 [00:00<?, ?it/s]
> Traceback (most recent call last):
>   File "**/.pyenv/versions/3.11.11/lib/python3.11/site-packages/mlx_lm/utils.py", line 188, in _get_classes
>     arch = importlib.import_module(f"mlx_lm.models.{model_type}")
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "**/.pyenv/versions/3.11.11/lib/python3.11/importlib/__init__.py", line 126, in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
>   File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
>   File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
> ModuleNotFoundError: No module named 'mlx_lm.models.diffusion_gemma'                      

$pip show mlx_lm                                                                                                                                                                                                                                    ~/mlx-lm
Name: mlx-lm
Version: 0.31.3
Summary: LLMs with MLX and the Hugging Face Hub
Home-page: https://github.com/ml-explore/mlx-lm
Author: MLX Contributors
Author-email: mlx@group.apple.com
License: MIT
Location: /Users/ws/.pyenv/versions/3.11.11/lib/python3.11/site-packages
[...]

PR #1402, opened 2026-06-14
Add DiffusionGemma (text): block-diffusion model + diffusion_generate

gilbert-barajas:
## DiffusionGemma (text) for mlx-lm

A text-only port of **DiffusionGemma** (`google/diffusiongemma-26B-A4B-it`) — Google's encoder-decoder **block-diffusion** MoE built on the Gemma 4 stack. Addresses the request in #1391 (text serving in mlx-lm; mlx-vlm already covers the multimodal path).

Instead of emitting tokens autoregressively, the decoder denoises a 256-token **canvas** over reverse-diffusion steps (entropy-bound acceptance + renoise, temperature schedule, adaptive stop), with the prompt prefilled into an encoder KV cache. A **block-autoregressive outer loop** chains canvases — commit a finalized canvas → extend the cache → denoise the next — for generation beyond `canvas_length`, stopping on EOS / `max_tokens` / `max_canvases`.

### Credit
Mirrors **@Blaizzy**'s working mlx-vlm `DiffusionGemma4Backbone` 1:1 (so the existing `mlx-community/diffusiongemma-*` conversions load directly, no reconvert), and uses the HF `transformers` reference for the sampler semantics. Grateful for both as parity targets.

### What works
- Full encoder-decoder model + a top-level `diffusion_generate` (a sibling to the AR loop, since the canvas-denoising contract doesn't fit `generate_step`), plus per-canvas `diffusion_stream_generate`.
- **Block-autoregressive outer loop** for `> canvas_length` generation (commit / extend / denoise; EOS / `max_tokens` / `max_canvases` stop). A single block (`max_tokens ≤ canvas_length`) reproduces single-canvas behavior.
- Loads the existing `mlx-community/diffusiongemma-26B-A4B-it` conversions (4-bit and bf16) with **zero key/shape mismatches** — `sanitize` drops the vision tower + handles the MoE weight names; `from_dict` flattens the nested `text_config`.
- CLI: `mlx_lm.generate` dispatches to `diffusion_generate` when the model exposes it (generic `hasattr` check; no model-specific import in `generate.py`).
- **Unit tests** (`tests/test_diffusion_gemma.py`) for the outer-loop control flow on a tiny random model: canvas chaining via cache-commit, `max_tokens` → block count, the `max_canvases` cap, EOS early-stop, and the streaming generator.

### Verification
- **MoE numerically parity-checked vs `transformers`**: router top-k indices exact; experts max abs diff ~1e-9 (machine precision). Verified offline against real weights; a fixture-backed **logits-parity test** is a planned follow-up (below).
- Adversarial code review against the HF + mlx-vlm reference caught and fixed two sampler bugs (committing the argmax canvas vs the renoised one; carrying temperature-scaled vs raw logits for self-conditioning).

### Receipts (Apple Silicon, M5 Max) — 256-token canvas, cold
| build | size | tok/s |
|---|---|---|
| 4-bit | 16.5 GB | ~250 |
| bf16 | 51.6 GB | ~121 |

```
mlx_lm.generate --model mlx-community/diffusiongemma-26B-A4B-it-4bit \
    --prompt "Why is the sky blue? Answer in one sentence."
# -> "The sky is blue because Earth's atmosphere scatters shorter wavelengths
#     of light (blue and violet) more effectively than other colors."
```
Coherent on prose and code (an `is_prime` with the 6k±1 rule).

### Open questions for maintainers
- Is a separate `diffusion_generate` the seam you'd prefer, or dispatch inside `generate`?
- Appetite for `mlx_lm.server` support here, or a follow-up?
- Any preferred encoder-decoder cache convention?

### Planned follow-ups (kept out of this PR to keep it focused)
- A fixture-backed **logits-parity test** (the offline ~1e-9 check, as a CI test).
- A **diffusion visualizer** (`mlx_lm.diffusion_viz` / `--diffusion-visual`) — watch the canvas denoise in place + offline trace replay. Built; holding for a separate PR so this one stays scoped to the model.

Marking ready for review — happy to iterate on the shape.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -2080,22 +2080,27 @@ def main():
     # instead of the token-by-token loop. Generic: any model exposing
     # `diffusion_generate` opts in; no model-specific import here.
     if hasattr(model, "diffusion_generate"):
+        import math
         import time
 
-        out = model.diffusion_generate(mx.array([prompt]))
+        # `--max-tokens` drives the block-autoregressive canvas count; EOS stops early.
+        eos_ids = set(getattr(tokenizer, "eos_token_ids", None) or [tokenizer.eos_token_id])
+        gen_kwargs = {"max_tokens": args.max_tokens, "eos_token_ids": eos_ids}
+        out = model.diffusion_generate(mx.array([prompt]), **gen_kwargs)
         mx.eval(out)
         tic = time.perf_counter()
-        out = model.diffusion_generate(mx.array([prompt]))  # timed (warm) pass
+        out = model.diffusion_generate(mx.array([prompt]), **gen_kwargs)  # timed (warm) pass
         mx.eval(out)
         dt = time.perf_counter() - tic
         toks = out[0].tolist()
-        eos_ids = set(getattr(tokenizer, "eos_token_ids", None) or [tokenizer.eos_token_id])
         cut = next((i for i, t in enumerate(toks) if t in eos_ids), len(toks))
         print(tokenizer.decode(toks[:cut]))
         if args.verbose:
             n = out.shape[1]
-            print(f"\n{'=' * 10}\nCanvas: {n} tokens in {dt:.3f}s ({n / dt:.1f} tok/s)",
-                  flush=True)
+            canvas_len = getattr(getattr(model, "config", None), "canvas_length", 0) or n
+            n_canvases = max(1, math.ceil(n / canvas_len))
+            print(f"\n{'=' * 10}\n{n} tokens across {n_canvases} canvas(es) in "
+                  f"{dt:.3f}s ({n / dt:.1f} tok/s)", flush=True)
         return
 
     if args.draft_model is not None:
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -2075,6 +2075,29 @@ def main():
     else:
         prompt = tokenizer.encode(prompt)
 
+    # Diffusion models (e.g. DiffusionGemma) denoise a fixed token canvas rather
+    # than emit tokens autoregressively — dispatch to the model's own generator
+    # instead of the token-by-token loop. Generic: any model exposing
+    # `diffusion_generate` opts in; no model-specific import here.
+    if hasattr(model, "diffusion_generate"):
+        import time
+
+        out = model.diffusion_generate(mx.array([prompt]))
+        mx.eval(out)
+        tic = time.perf_counter()
+        out = model.diffusion_generate(mx.array([prompt]))  # timed (warm) pass
+        mx.eval(out)
+        dt = time.perf_counter() - tic
+        toks = out[0].tolist()
+        eos_ids = set(getattr(tokenizer, "eos_token_ids", None) or [tokenizer.eos_token_id])
+        cut = next((i for i, t in enumerate(toks) if t in eos_ids), len(toks))
+        print(tokenizer.decode(toks[:cut]))
+        if args.verbose:
+            n = out.shape[1]
+            print(f"\n{'=' * 10}\nCanvas: {n} tokens in {dt:.3f}s ({n / dt:.1f} tok/s)",
+                  flush=True)
+        return
+
     if args.draft_model is not None:
         draft_model, draft_tokenizer = load(args.draft_model)
         if draft_tokenizer.vocab_size != tokenizer.vocab_size:
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -2079,28 +2079,30 @@ def main():
     # than emit tokens autoregressively — dispatch to the model's own generator
     # instead of the token-by-token loop. Generic: any model exposing
     # `diffusion_generate` opts in; no model-specific import here.
-    if hasattr(model, "diffusion_generate"):
-        import math
+    if hasattr(model, "diffusion_stream_generate"):
         import time
 
-        # `--max-tokens` drives the block-autoregressive canvas count; EOS stops early.
+        # Block-diffusion: STREAM each canvas as it commits (watch it generate) instead
+        # of freeze-then-dump. `--max-tokens` drives the canvas count; an EOS in a
+        # committed canvas ends generation.
         eos_ids = set(getattr(tokenizer, "eos_token_ids", None) or [tokenizer.eos_token_id])
         gen_kwargs = {"max_tokens": args.max_tokens, "eos_token_ids": eos_ids}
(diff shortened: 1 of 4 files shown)


dogukanveziroglu:
Don't make a single PR with 1100+ lines of changes and I think the overall qualty of the code is not great.

gilbert-barajas:
Thanks for taking a look. The diff is already trimmed to the core — the visualization/notebook tooling is held on a separate branch (`diffusion-gemma-viz`) so this PR is just the model, the >256-token diffusion loop, streaming, and tests. The bulk is one new model file mirroring mlx-vlm's `DiffusionGemma4Backbone` 1:1 so wire weights load directly (credit @Blaizzy).

Happy to split it further if the maintainers would prefer it staged (e.g. model + sampler first, tests second). And if there are specific spots where the quality concerns you, point me at them and I'll fix them directly.

dogukanveziroglu:
I think this PR still feels too much like vibe coded integration just enough to get this specific model running under specific conditions, but not yet enough to land as a maintainable implementation in mlx-lm.

Unfortunately, this is not the kind of issue where we can just ask Claude to “fix DiffusionGemma support” and merge the resulting code. DiffusionGemma changes the generation contract quite a bit compared to normal autoregressive models, so the API boundary needs to be designed deliberately.

My concern is not only that the model file is large. I understand that the core implementation may need to be large. The bigger issue is that the PR mixes the model implementation, diffusion sampling, streaming behavior, CLI dispatch, cache handling assumptions, and tests all in one shot, without clearly defining which parts are production API and which parts are model specific workaround code.

For example, the CLI integration currently looks like a special case hook rather than a clean generation interface. Some normal AR generation options may not be valid for DiffusionGemma, and that is fine, but they should not be silently accepted and bypassed. Unsupported options should produce a clear warning/error, and diffusion specific options should be exposed explicitly. And there are a lot more production grade qualitu issues on the code like local path referances on comments.

Ofc I might be wrong and maintainers might think different.

gilbert-barajas:
Thanks — this is the kind of specific feedback that's genuinely useful, and you're right about the shape problems. The generation contract is the real issue: DiffusionGemma's canvas-denoising `__call__` doesn't fit mlx-lm's autoregressive Model API, and bundling the model, sampler, streaming, CLI, and tests into one PR blurs which parts are stable API vs model-specific. I'd rather get that boundary right than land it fast.

Concrete next steps I'll take:
- CLI: replace the `hasattr`-based special-case hook with an explicit interface; expose diffusion-specific options (denoising steps, entropy bound, temperature schedule) explicitly, and raise a clear error on AR-only options that don't apply rather than silently ignoring them.
- The local-path comments are already removed in the latest push.

On structure, I'm glad to stage this into separate PRs — (1) model + sanitize conforming to the Model API, (2) the diffusion sampler, (3) the generation/CLI interface. Since the right shape depends on how mlx-lm wants a non-autoregressive model to expose its generation interface, I'd welcome @awni / @angeloskath's steer on that API boundary before I restructure, so I'm building toward what the repo actually wants.

pierre427:
**Proposing a concrete `diffusion_generate` boundary so the diffusion PRs can converge**

Four diffusion-LM PRs (#1402 DiffusionGemma, #1490/#1391 interface + LLaDA, and our #1560/#1561) each ship their own generation loop, and they've already diverged. Rather than re-litigate the design, here's a concrete, model-agnostic interface extracted from what two *shipped-and-validated* ports actually pass — every field is one a real port needs — so the PRs become a mechanical conform-and-rebase.

**Shape.** A frozen `DiffusionGenerationConfig` (policy) + a `diffusion_generate(model, prompt, config, *, tokenizer, return_stats)` / `stream_diffusion_generate(...)` pair (runtime handles). Models opt in by exposing a `diffusion_denoise` forward (one bidirectional pass over a windowed canvas → logits, optionally post-RoPE K/V) plus a `DiffusionCapabilities` descriptor; the generic loop lives once in `mlx_lm`, exactly as `stream_generate` relates to `Model.__call__`. An AR / tri-mode model stays a first-class `mlx_lm.generate` citizen — `diffusion_generate` is *added* surface, never a replacement.

**Config surface** (provenance in the RFC): canvas/block schedule (`gen_length`, `block_length`, `steps`, `mask_id`, `eos_id`, `eos_early_stop`); unmask policy (`UnmaskPolicy` enum: top-k schedule / confidence / adaptive, `threshold`, `dynamic_threshold_alpha`); remask/refine (`remasking`, `remask_refine`, `remask_conf`, `remask_rounds`); a capability-gated per-block KV-reuse contract (`kv_reuse`, `dual_cache`, `incremental_cache`, `allow_quantized_cache_drift`); sampler (`temperature`, `cfg_scale`, `sampler_row_chunk`); warm-start (`decoder_input_ids`); and an optional self-spec `verify_hook`.

**Ecosystem-answered sub-questions** (bank, don't re-derive): sampler-memory bound = vLLM's row-chunked O(K·V) (vllm-project/vllm#46155); block-slot lifecycle = SGLang FDFO in-place reuse (sgl-project/sglang#27877); canvas warm-start = mlx-vlm `decoder_input_ids` (Blaizzy/mlx-vlm#1589); streaming = per-finalized-block `(block_index, committed_ids)` (in-block commits are out of order, so token-by-token is ill-defined).

**The two decisions that unblock everything else** (the exact boundary #1402 asked for):
1. **Where does the loop live** — one top-level `diffusion_generate` (models expose only `diffusion_denoise` + `diffusion_capabilities`), or a per-model method + thin helper? We propose the former.
2. **Capability signaling / auto-dispatch** — should `mlx_lm.generate` auto-route when a model advertises `diffusion_capabilities`, or must the caller pick? Is a `DiffusionCapabilities` descriptor the right discovery mechanism, or a `model.is_diffusion` flag?

Secondary (answerable inline): KV-cache object (reuse `cache.py` vs a `DiffusionCache`; is the quantized-drift guard acceptable for v1?); unmask-policy surface (selectable enum vs one fixed policy — we recommend confidence default); `verify_hook` scope; streaming granularity; whether the row-chunked sampler bound is mandated in the shared loop.

If (1) loop location and (2) capability signaling get an answer, the remaining items are mechanical and #1402, #1490, #1560, #1561 can all land against one surface. Full RFC (interface, parameter provenance, per-path mapping, conforms-to checklist) available; happy to open it as its own thread if that's easier to review.

nastya236:
Thanks for your contribution! We recently introduced a new policy regarding the use of AI in pull request descriptions and discussions https://github.com/ml-explore/mlx-lm/pull/1770.

Issue #1178, opened 2026-04-22
Support `--prompt-cache-file` with mlx_lm.server


nathandaven:
I currently use mlx_lm locally to serve models only when I am using my local harness. The local harness has the same defined system prompt every time it starts up, I discovered the prompt cache file feature for generate and realized it is not supported when using the server. This would be helpful in allowing for the system prompt to be pre generated for faster initial prompt responses. Unless I am missing something, this seems like a useful feature!

kru2710shna:
I'd like to take this on. The `--prompt-cache-file` flag already exists in `mlx_lm.generate` (via `load_prompt_cache`), so this is largely a matter of wiring the same mechanism into the server's startup and request-handling path.

For your use case, I may suggest a fixed system prompt loaded once at startup and reused across requests. The natural approach is a `--prompt-cache-file` flag on `mlx_lm.server` that loads the cache at boot and uses it as a shared prefix for incoming requests.

One design question for the maintainers: when an incoming request's prefix doesn't match the cached prefix (e.g. a different system prompt than the one the cache was built with), should the server fall back to recomputing from scratch, or return an error? The generate path assumes a matching prefix, but the server is multi-request so it needs an  explicit policy. My instinct is graceful fallback to recompute, but I'll scope to whatever fits the server architecture best.

Will open a PR once the approach is confirmed.

PR #1405, opened 2026-06-15
server: persist LRU prompt cache to disk across restarts

javruben:
## Problem

`mlx_lm.server` already keeps an in-memory LRU prompt cache (`LRUPromptCache` in `mlx_lm/models/cache.py`), but the cache is lost on every server restart. For long shared prompts (system prompts, RAG contexts, multi-turn chats), this means re-prefilling thousands of tokens from scratch after every restart, even though the cached KV state is reusable.

## Change

Adds `--prompt-cache-dir` to `mlx_lm.server`. When set:

- On startup, the server restores any cached prompts whose model key matches the currently-loaded model.
- On `atexit` and `SIGTERM`, the in-memory LRU is re-saved to disk.
- Layout is one safetensors file per cached prompt (via the existing `save_prompt_cache` helper) plus a `manifest.json` carrying the LRU order, model keys, tokens, and cache type.
- Entries whose model is not currently loaded are skipped on load, so the cache survives model changes safely.
- Forward-compatible via `LRUPromptCache.MANIFEST_VERSION`.

Files: `mlx_lm/server.py` (+81), `mlx_lm/models/cache.py` (+107), `mlx_lm/SERVER.md` (+15), `tests/test_prompt_cache.py` (+73), `tests/test_server.py` (+59). Net +334 LOC.

Default behaviour is unchanged: omit `--prompt-cache-dir` and the server behaves exactly as before.

## Mechanism (asserted)

The on-disk layout is inspired by LuminaNAO's LSCKPT2 sidecar (`codeberg.org/LuminaNAO/llama-hdd.cpp`), which persists KV cache alongside the model file with a fixed magic header that lets the engine verify the sidecar before restoring. Here we apply the same idea at the Python server layer using `safetensors` files keyed by model identity, so a corrupted or model-mismatched sidecar is simply skipped rather than misapplied.

## Evidence

Measured on a 5-QA replay over a held prompt with the same model + flags, restart between cold and warm:

| Host             | Cold prefill | Warm mean | Drop  |
| ---------------- | -----------: | --------: | ----: |
| M2 Ultra (prime) |    1955.19ms |  209.42ms | 89.3% |
| M3 Max (MBP)     |    4834.93ms |  289.86ms | 94.0% |

(Numbers from a side-by-side reference implementation; for `mlx_lm.server` the per-prompt sidecar avoids the prefill on warm-start by the same mechanism.)

## Test

- `python -m pytest tests/test_prompt_cache.py tests/test_server.py`

New tests exercise round-trip save/load, model-mismatch skip, and the SIGTERM persist path.

## Notes

- Opt-in: no change unless `--prompt-cache-dir` is passed.
- Closes the resumable long-context gap requested in #1178.
- Happy to split into smaller PRs if preferred.

### DIFF
--- a/mlx_lm/SERVER.md
+++ b/mlx_lm/SERVER.md
@@ -140,6 +140,21 @@ curl localhost:8080/v1/chat/completions \
     - `completion_tokens`: The number of tokens generated.
     - `total_tokens`: The total number of tokens, i.e. the sum of the above two fields.
 
+### Persistent Prompt Cache
+
+The server maintains an in-memory LRU of recent prompt caches (see
+`--prompt-cache-size` and `--prompt-cache-bytes`). Pass `--prompt-cache-dir`
+to also persist that LRU to disk so it survives restarts:
+
+```shell
+mlx_lm.server --model <model> --prompt-cache-dir ~/.cache/mlx-lm/prompts
+```
+
+On startup the server restores any entries whose model key matches the
+currently-loaded model. The cache is re-saved on `atexit` and on `SIGTERM`.
+Entries are stored as one safetensors file per cached prompt plus a
+`manifest.json`; the format is forward-compatible via `MANIFEST_VERSION`.
+
 ### List Models
 
 Use the `v1/models` endpoint to list available models:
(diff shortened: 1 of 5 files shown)


zcbenz:

PR #1176, opened 2026-04-21
Lc/fix xtc special tokens server

micuentadecasa:
## Bug Description                                                                                                        
                                                                                                                             
   `_make_sampler` in `server.py` constructs `xtc_special_tokens` as a nested list `[int, [int, ...]]` instead of a flat     
   list of ints. When XTC sampling is triggered (`xtc_probability > 0`), `apply_xtc` passes this to `mask[...,               
   xtc_special_tokens] = False` which crashes:                                                                               
                                                                                                                             
   ```                                                                                                                       
   ValueError: Initialization encountered extra dimension.                                                                   
   ```                                                                                                                       
                                                                                                                             
   XTC sampling via the server is completely broken — any request with `xtc_probability > 0` returns an HTTP 500 on the      
   first token.                                                                                                              
                                                                                                                             
   ## Root Cause                                                                                                             
                                                                                                                             
   `server.py` lines 388-391:                                                                                                
   ```python                                                                                                                 
   xtc_special_tokens=[                                                                                                      
       tokenizer.eos_token_id,    # int                                                                                      
       tokenizer.encode("\n"),     # list -> creates [int, [int, ...]]                                                       
   ],                                                                                                                        
   ```                                                                                                                       
                                                                                                                             
   `tokenizer.encode()` returns a list. Wrapping it in `[...]` alongside an int creates a nested list. Additionally,         
   `eos_token_id` (singular) only includes one EOS token, missing others on multi-EOS models.                                
                                                                                                                             
   ## Fix                                                                                                                    
                                                                                                                             
   - Align with the correct pattern already used in `generate.py:2062` and `chat.py:155`                                     
   - `xtc_special_tokens=tokenizer.encode("\n") + list(tokenizer.eos_token_ids)`                                             
                                                                                                                             
   This produces a flat list and includes all EOS tokens.                                                                    
                                                                                                                             
   ## Validation                                                                                                             
                                                                                                                             
   **Before:**                                                                                                               
   ```                                                                                                                       
   xtc_special_tokens: [151645, [198]]                                                                                       
   apply_xtc -> ValueError: Initialization encountered extra dimension.                                                      
   ```                                                                                                                       
                                                                                                                             
   **After:**                                                                                                                
   ```                                                                                                                       
   xtc_special_tokens: [198, 151645]                                                                                         
   apply_xtc -> Success, shape (1, 151669)                                                                                   
   ```                                                                                                                       
                                                                                                                             
   6 unit tests in `tests/test_xtc_tokens.py` — all passing.

PR #1408, opened 2026-06-15
fix(server): build xtc_special_tokens as a flat list

devYRPauli:
### Bug

`_make_sampler` in `mlx_lm/server.py` builds `xtc_special_tokens` as a nested list:

```python
xtc_special_tokens=[
    tokenizer.eos_token_id,   # int
    tokenizer.encode("\n"),   # list, e.g. [198]
],
```

That produces `[int, [int]]` (e.g. `[50256, [198]]`). `apply_xtc` expects a flat `List[int]` — it does `mask[..., xtc_special_tokens] = False`. With the nested list it raises:

```
ValueError: Initialization encountered extra dimension.
```

So any server request with `temperature > 0` and `xtc_probability > 0` fails (at `temperature == 0` `make_sampler` short-circuits to argmax and never applies XTC, which is why it isn't always hit).

`generate.py` and `chat.py` already build this correctly:

```python
xtc_special_tokens=tokenizer.encode("\n") + list(tokenizer.eos_token_ids)
```

### Fix

Build the list the same flat way in `server.py`.

### Verification

Real tokenizer (gpt2) reproduces the exact construction and failure:

```
tokenizer.encode("\n")  -> [198]   (list)
tokenizer.eos_token_id  -> 50256   (int)
server builds           -> [50256, [198]]
apply_xtc(..., [50256, [198]])  -> ValueError: Initialization encountered extra dimension.
apply_xtc(..., [198, 50256])    -> OK
```

Added a network-free regression test (`tests/test_server.py::TestMakeSampler`) that calls `_make_sampler` with `temperature=0.6, xtc_probability=1.0` and runs the returned sampler. It fails on the current code with the `ValueError` above and passes with the fix. `tests/test_sample_utils.py` still passes (7); black + isort clean.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -386,10 +386,7 @@ def _make_sampler(args, tokenizer):
         min_p=args.sampling.min_p,
         xtc_probability=args.sampling.xtc_probability,
         xtc_threshold=args.sampling.xtc_threshold,
-        xtc_special_tokens=[
-            tokenizer.eos_token_id,
-            tokenizer.encode("\n"),
-        ],
+        xtc_special_tokens=tokenizer.encode("\n") + list(tokenizer.eos_token_ids),
     )
 
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -4,6 +4,7 @@ import http
 import io
 import json
 import threading
+import types
 import unittest
 
 import mlx.core as mx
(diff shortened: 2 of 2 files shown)


nastya236:
Thank you for the fix. I'm closing this as a duplicate of #1176.

PR #1413, opened 2026-06-18
Fix gradient accumulation to weight micro-batches by token count

Incheonkirin:
`default_loss` returns a per-micro-batch token-mean (`ce.sum() / ntoks`). With gradient accumulation, `step()` summed the per-micro-batch gradients and divided by a fixed `grad_accumulation_steps`. Since each micro-batch gradient is already normalized by its own token count, dividing the sum by K does not equal the gradient of the token-mean loss over the combined effective batch unless every micro-batch has the same number of trained tokens. With variable-length completions the accumulated update is biased toward shorter examples and depends on how examples are partitioned into micro-batches. `evaluate()` already token-weights its average, so the trainer was internally inconsistent.

This keeps a running token-weighted mean of the gradient over the accumulation window. For single-process training that running mean is already the gradient the optimizer needs, so no extra normalization is applied at the update; only with multiple ranks is it reduced across ranks weighted by each rank's token count. The accumulation denominators are guarded so an all-zero-token window cannot divide by zero. The final partial window (when `iters` is not a multiple of `grad_accumulation_steps`) is now applied instead of dropped. `default_loss` and `grad_accumulation_steps == 1` behavior are unchanged. All-zero-token windows still call `optimizer.update` with a zero gradient; the underlying zero-token loss NaN in `default_loss` is handled separately by #1355.

Tests: unequal-token micro-batches now match a single combined batch, the final partial window matches explicit windows, and an all-zero-token window keeps parameters finite. The first two fail on the current code and pass with this change.

### DIFF
--- a/mlx_lm/tuner/trainer.py
+++ b/mlx_lm/tuner/trainer.py
@@ -244,20 +244,37 @@ def train(
     state = [model.state, optimizer.state, mx.random.state]
 
     @partial(mx.compile, inputs=state, outputs=state)
-    def step(batch, prev_grad, do_update):
+    def step(batch, prev_grad, prev_ntoks, do_update):
         (lvalue, toks), grad = loss_value_and_grad(model, *batch)
 
-        if prev_grad is not None:
-            grad = tree_map(lambda x, y: x + y, grad, prev_grad)
+        # Token-weighted running mean of the gradient, so accumulation matches the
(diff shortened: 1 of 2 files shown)


Incheonkirin:
Replacement PR is now #1544 from a proper fork. It contains the rebased change and current validation; no reopen is needed.

PR #1417, opened 2026-06-20
Fix: qwen3_coder drops tool calls with float-formatted integer args

brianmeyers007:
`mlx_lm/tool_parsers/qwen3_coder.py` `_convert_param_value` coerces integer-typed tool arguments with bare `int(param_value)`. Models frequently emit integer arguments float-formatted (e.g. `offset="230.0"`), so `int("230.0")` raises `ValueError: invalid literal for int() with base 10: '230.0'`.

`server.py`'s `ToolCallFormatter.__call__` catches that ValueError and `continue`s, **silently dropping the otherwise-valid tool call** and logging a misleading "tool text was likely truncated mid-generation" warning. The result is a response with `finish_reason="tool_calls"` and an empty `tool_calls` array, which stalls tool-using agents. This is hard to diagnose because the warning points at the wrong cause and the model output is actually complete and correct.

### DIFF
--- a/mlx_lm/tool_parsers/qwen3_coder.py
+++ b/mlx_lm/tool_parsers/qwen3_coder.py
@@ -54,7 +54,7 @@ def _convert_param_value(param_value: str, param_name: str, param_config: dict) 
         or param_type.startswith("short")
         or param_type.startswith("unsigned")
     ):
-        return int(param_value)
+        return int(float(param_value))
     elif param_type.startswith("num") or param_type.startswith("float"):
         float_param_value = float(param_value)
         int_param_value = int(float_param_value)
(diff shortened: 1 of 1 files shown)


angeloskath (approved):

brianmeyers007:
Gentle ping on this one — it has been approved and MERGEABLE/CLEAN since 2026-06-25.

Since then the same defect has been reported independently several times, which suggests it is still costing people silently-dropped tool calls:

- #1627 — qwen3_coder rejects float-formatted integer args ("140.0"), dropping the call and causing client infinite loops
- #1604 — related unparseable-argument crash in the same parser
- #1635 and #1617 — independent PRs fixing the same code path

The approved diff here is @angeloskath's own revision (raise on a non-integral float rather than silently truncating), so it should already be in the preferred form. Happy to rebase or split it if that would help.

One note on why this is hard to diagnose in the wild: the ValueError is swallowed by ToolCallFormatter, so the symptom surfaces as finish_reason=tool_calls with an empty tool_calls array plus a "tool text was likely truncated mid-generation" warning — which points away from the actual cause.

PR #1422, opened 2026-06-21
feat: add batch_generate_same_prompt for SSM/hybrid models

### LINKED ISSUE #1403: feat: batched generation for Mamba/hybrid models (Nemotron-H)
## Problem

`batch_generate()` uses `BatchGenerator` which processes sequences via continuous batching — extract/filter/merge per finished sequence. For transformer models this works well since KV cache is read-only after prefill.

For **Mamba/SSM/hybrid models** (Nemotron-H, Jamba, Falcon Mamba), this is actually **slower** than sequential generation because:
1. Mamba's recurrent state must be updated each step — the "batch" overhead (extract/merge per token) exceeds any parallel benefit
2. The model's forward pass already supports batch dimension natively (`model(tokens_K×1)` works for K sequences simultaneously)

## What works instead

A simple batched decode loop that feeds all K sequences through one forward pass gives **3× speedup** on Nemotron-H (30B, 4-bit, M4 Max):

```python
# Prefill — all K sequences have same prompt
prompt_batch = mx.array([prompt_ids] * K)  # [K, prompt_len]
logits = model(prompt_batch, cache=cache)  # processes all K at once
logits = logits[:, -1, :]  # [K, V]

# Decode — generate all K tokens per step in one forward pass
for _ in range(max_tokens):
    tokens = sample(logits)           # [K]
    logits = model(tokens.reshape(K, 1), cache=cache)
    logits = logits[:, -1, :]         # [K, V]
```

Benchmark (M4 Max, 64GB, Nemotron-H 30B 4-bit, K=8, 512 tokens):
- Sequential: 72s
- Batched forward: 24s → **3× speedup**
- `BatchGenerator`: 2.4× **slower** than sequential (overhead)

## Suggestion

For models with `make_cache()` returning `ArraysCache`/`MambaCache`, a specialized code path in `batch_generate()` (or a new `batched_decode()` function) that:
1. Duplicates the prompt K times 
2. Prefills once with batch dim
3. Decodes with `model(tokens_K×1, cache=cache)` — one forward call per step for all K sequences

This would complement `BatchGenerator` rather than replace it — `BatchGenerator` is better for transformers with heterogeneous prompts and variable-length sequences.

## Context

Using this for GRPO training on Apple Silicon where each step generates K=8-16 rollouts from the same prompt. The generation is 91% of training time (per mlx-tracer profiling).

Related: #1358 (logprobs from batch_generate)

haoxu0:
## Summary

Add `batch_generate_same_prompt()` — generates K completions from one prompt in a single batched forward pass. Reads model weights once per decode step instead of K times.

## Motivation

For RL training (GRPO, RLOO, PPO) each step generates K=8-16 rollouts from the same prompt. Generation is the dominant cost (~91% of step time, profiled with mlx-tracer).

`batch_generate()` uses continuous batching which adds overhead for SSM/hybrid models (cache extract/merge per finished sequence). This function exploits the shared-prompt structure instead.

## Benchmark (M4 Max 64GB, Nemotron-H 30B 4-bit)

| Method | K=8, 512 tokens | Speedup |
|--------|-----------------|---------|
| Sequential (`generate()` × K) | 72s | 1× |
| `batch_generate()` (continuous batching) | slower | 0.4× ❌ |
| **`batch_generate_same_prompt()`** | **24s** | **3×** ✅ |

## API

```python
from mlx_lm.generate import batch_generate_same_prompt

response = batch_generate_same_prompt(
    model, tokenizer,
    prompt=token_ids,
    num_completions=8,
    max_tokens=1024,
    sampler=make_sampler(temp=1.0),
)
# response.texts, response.token_ids, response.logprobs
```

## Changes

- `mlx_lm/generate.py`: +147 lines (new function)
- `mlx_lm/__init__.py`: export
- `tests/test_generate.py`: 3 tests (basic, logprobs, deterministic)

Closes #1403

### DIFF
--- a/mlx_lm/__init__.py
+++ b/mlx_lm/__init__.py
@@ -7,13 +7,14 @@ from ._version import __version__
 os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 
 from .convert import convert
-from .generate import batch_generate, generate, stream_generate
+from .generate import batch_generate, batch_generate_same_prompt, generate, stream_generate
 from .utils import load
 
 __all__ = [
     "__version__",
     "convert",
     "batch_generate",
+    "batch_generate_same_prompt",
     "generate",
     "stream_generate",
     "load",
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1984,6 +1984,153 @@ def batch_generate(
     return BatchResponse(texts, stats, caches, token_ids, logprobs)
 
 
+def batch_generate_same_prompt(
+    model: nn.Module,
+    tokenizer: Union[PreTrainedTokenizer, TokenizerWrapper],
+    prompt: Union[str, List[int]],
+    num_completions: int = 4,
+    max_tokens: int = 256,
+    sampler: Optional[Callable[[mx.array], mx.array]] = None,
+    verbose: bool = False,
+) -> BatchResponse:
+    """
+    Generate multiple completions from the same prompt in a single batched
+    forward pass.
+
+    Unlike :func:`batch_generate` which uses continuous batching (optimal for
+    transformers with heterogeneous prompts), this function exploits the fact
+    that all sequences share the same prompt. It prefills once with batch
+    dimension K, then decodes K tokens per step in one ``model()`` call.
+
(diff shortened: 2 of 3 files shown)


zcbenz:

PR #1424, opened 2026-06-23
tests: add dedicated unit tests for Gemma3n model

Sreya8:
Gemma3n is covered by the generic `test_all_models` runner but its unique architectural features had no dedicated tests. This PR adds 13 tests to the `TestModels` class in `test_models.py` covering:

- **KV shared layers**: verifies `is_kv_shared_layer` flag is set correctly on each attention layer, and that `make_cache()` returns only non-shared layer entries with the correct cache types (`KVCache` for full attention, `RotatingKVCache` for sliding)
- **AltUp**: verifies `predict` and `correct` preserve output shape and dtype across float32 and float16
- **LauReL**: verifies residual shape contract and identity behavior with zeroed left projection
- **sanitize**: verifies multimodal towers (vision, audio) are stripped while all text weights are preserved
- **Forward pass**: verifies output shapes, incremental generation (prefill + decode step), and logit softcapping bounds

All 13 tests pass. No existing tests were broken (88 passed, 1 skipped).

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -3220,6 +3220,199 @@ class TestModels(unittest.TestCase):
                 self.assertTrue(mx.allclose(y, y_gt, rtol=1e-4, atol=1e-4))
                 self.assertTrue(mx.allclose(st, st_gt, rtol=1e-4, atol=1e-3))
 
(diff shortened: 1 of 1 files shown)


zcbenz:

PR #1427, opened 2026-06-23
Fix speculative decode with full rotating caches

### LINKED ISSUE #1423: Speculative decoding diverges from non-speculative output at temperature 0 (should be identical)
### Bug

At **temperature 0 (greedy)**, `mlx_lm.generate` with `--draft-model` (speculative decoding) produces **different output** than the same call **without** a draft model. Greedy speculative decoding should be lossless — the target verifies every drafted token and accepts only its own argmax, so the output must equal plain greedy decoding. Instead the two diverge, reproducibly.

### Versions

- `mlx` 0.31.2
- `mlx-lm` 0.31.3
- Apple M3 Ultra (512 GB), macOS 26.2 (arm64)

### Reproduce

Public models — 8-bit target, 4-bit draft, **byte-identical tokenizer** (both vocab 262144):

```bash
# (A) no speculative decoding
mlx_lm.generate --model mlx-community/gemma-4-31b-it-8bit \
  --temp 0 --max-tokens 500 \
  --prompt "Write a detailed essay (multiple paragraphs) explaining how a transformer neural network processes a sentence, from tokenization through attention to output."

# (B) with speculative decoding (same prompt)
mlx_lm.generate --model mlx-community/gemma-4-31b-it-8bit \
  --draft-model mlx-community/gemma-4-e2b-it-4bit --num-draft-tokens 2 \
  --temp 0 --max-tokens 500 \
  --prompt "Write a detailed essay (multiple paragraphs) explaining how a transformer neural network processes a sentence, from tokenization through attention to output."
```

### Expected

(A) and (B) identical (greedy spec-decode is exact).

### Actual

They diverge. Same prompt, temp 0:

> **(A) no draft:** …its **significance (replacing RNNs/LSTMs)**, and the general flow (Encoder-Decoder or Decoder-only).
>
> **(B) with draft:** …its **departure from RNNs/LSTMs (parallelism)**, and the general goal (mapping input sequences to output sequences).

Observations:
- **Reproducible & deterministic** across repeated isolated runs (not a flaky startup artifact).
- The divergence appears/accumulates with **longer generations**; short generations — or runs where an early `stop` sequence truncates output before the divergence point — can look identical, which makes it easy to miss.
- **Not a tokenizer mismatch** — draft and target `tokenizer.json` are byte-identical (vocab 262144).
- Also reproduces via `mlx_lm.server` `/v1/chat/completions` with the same model pair: spec-decode responses can drop the entire post-reasoning `content` on longer generations, consistent with the same underlying divergence.

cyq1017:
## Summary

Speculative decoding can diverge from deterministic non-speculative generation when a `RotatingKVCache` reaches the point where it can no longer be trimmed. If a drafted token is rejected after that point, the existing rewind path cannot remove the speculative cache entries.

This limits draft length around full rotating caches and falls back to target-only generation when rollback is not safe. Other trimmable cache types keep their existing speculative path.

Fixes #1423.

## Tests

- `uv run --isolated --python 3.12 --with-editable . --with pytest python -m pytest tests/test_generate.py::TestSpeculativeGenerateStep tests/test_generate.py::TestGenerate::test_stream_generate_speculative -q`
- `python3 -m py_compile mlx_lm/generate.py tests/test_generate.py`
- `git diff --check HEAD~2..HEAD`

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -595,11 +595,11 @@ def speculative_generate_step(
             if not c.is_trimmable():
                 return 0
 
-            max_size = getattr(c, "max_size", None)
-            offset = getattr(c, "offset", None)
-            if max_size is None or offset is None:
+            if not isinstance(c, (RotatingKVCache, BatchRotatingKVCache)):
                 continue
 
+            max_size = c.max_size
+            offset = c._offset if isinstance(c, BatchRotatingKVCache) else c.offset
             if hasattr(offset, "item"):
                 offset = offset.item()
             # Rotating caches stop being trimmable once the window fills, so
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -590,6 +590,25 @@ def speculative_generate_step(
         cache.trim_prompt_cache(model_cache, num_draft - num_accept)
         cache.trim_prompt_cache(draft_cache, max(num_draft - num_accept - 1, 0))
 
+    def _limit_draft_tokens(prompt_cache, num_draft, extra_target_tokens):
+        for c in prompt_cache:
+            if not c.is_trimmable():
+                return 0
+
+            max_size = getattr(c, "max_size", None)
+            offset = getattr(c, "offset", None)
+            if max_size is None or offset is None:
+                continue
+
+            if hasattr(offset, "item"):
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1429, opened 2026-06-24
server: add --lazy CLI flag for deferred model loading

### LINKED ISSUE #1428: server: add --lazy CLI flag for deferred model loading
## Description

The `mlx_lm.utils.load()` function accepts a `lazy: bool = False` parameter that defers model parameter evaluation to the first forward pass, reducing cold start time from ~7.67s to ~0.95s (measured on Apple M4, 10.99 GB Gemma 4 12B Q4). However, the `mlx_lm.server` CLI does not expose a `--lazy` flag, so server users always get the lazy=false (eager evaluation) code path.

## Evidence

1. `mlx_lm.utils.load()` signature (verified from installed v0.31.3):
   ```python
   def load(path_or_hf_repo, lazy: bool = False, ...)
   ```

2. Server `ModelProvider._load()` calls `load()` without `lazy`:
   ```python
   model, tokenizer = load(
       model_path,
       adapter_path=adapter_path,
       tokenizer_config=self._tokenizer_config,
       trust_remote_code=self.cli_args.trust_remote_code,
   )
   ```

3. Server argparse (verified from upstream HEAD via gh api) has no `--lazy` argument.

4. Measured impact: `load(lazy=False)` = 7,334 ms vs `load(lazy=True)` = 1,063 ms for Gemma 4 12B Q4 on Apple M4.

## Proposed fix

Add `--lazy` argparse argument in `server.py main()`, store it in CLI args, and pass it through `ModelProvider.load()` → `mlx_lm.utils.load(lazy=...)`.

## Source

- Upstream file: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/server.py
- `load()` declaration: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/utils.py#L467
- Server `_load()`: https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/server.py#L324

cyq1017:
## Summary
- add a `--lazy` flag to the OpenAI-compatible server CLI
- pass the lazy-load option through when loading the main model
- cover the parser flag and load wiring in server tests

Fixes #1428.

## Verification
- `.venv/bin/python -m unittest tests.test_server -v`
- `git diff --check origin/main..HEAD`
- changed-file credential-pattern scan

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -350,6 +350,7 @@ class ModelProvider:
                 adapter_path=adapter_path,
                 tokenizer_config=self._tokenizer_config,
                 trust_remote_code=self.cli_args.trust_remote_code,
+                lazy=getattr(self.cli_args, "lazy", False),
             )
 
         # Use the default chat template if needed
@@ -1748,7 +1749,7 @@ def run(
         response_generator.join()
 
 
-def main():
+def _build_arg_parser():
     parser = argparse.ArgumentParser(description="MLX Http Server.")
     parser.add_argument(
         "--model",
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1432, opened 2026-06-24
Add dedicated tests for AFM7 model

Sreya8:
## Summary

AFM7 (`mlx_lm/models/afm7.py`) is currently only covered by the generic `test_all_models` smoke test, which checks basic shape/dtype/cache correctness but doesn't exercise the model's two distinctive mechanisms: KV-cache reuse and fused/quantized linear projections. This PR adds dedicated tests for both.

## KV-cache reuse

AFM7 splits its layers into regular `TransformerBlock`s (which compute their own keys/values) and `KVReuseTransformerBlock`s (which borrow keys/values from the last regular layer instead of projecting their own). This PR adds four tests covering this mechanism.

`test_afm7_kv_reuse_cache_count` confirms `make_cache()` only allocates cache entries for the regular layers, not the reuse layers.

`test_afm7_kv_reuse_attention_has_no_kv_projections` confirms `KVReuseAttention` has its own query projection but no fused `qkv_proj`, unlike the standard `Attention` class.

`test_afm7_kv_reuse_attention` confirms the reuse block actually depends on the keys/values it's given. Passing zeroed keys/values produces a different output than the real ones, rather than the block silently ignoring them.

`test_afm7_incremental_generation_matches_full_forward` confirms step-by-step incremental generation, using a cache, produces output matching a single full forward pass. This is the key end-to-end correctness check for caching and KV-reuse working together.

## Fused/quantized linear projections

AFM7 packs multiple projections, such as q/k/v, into a single `FusedLinear` or `FusedQuantizedLinear` for efficiency, splitting the output afterward. This PR adds three tests covering this mechanism.

`test_afm7_fused_linear_split_shape` confirms `FusedLinear` splits its output into correctly shaped pieces matching `output_dims`.

`test_afm7_fused_linear_to_quantized_roundtrip` confirms `to_quantized()` preserves `output_dims` and produces outputs reasonably close to the unquantized version.

`test_afm7_fake_8bit_quant_clipping` confirms `fake_8bit_quant` correctly clips to the signed 8-bit range `[-128, 127]` at both bounds.

## Testing

All 7 new tests pass, along with the full existing test suite:

```
82 passed, 1 skipped
```

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -3219,6 +3219,188 @@ class TestModels(unittest.TestCase):
                 y = y[:, s:e]
                 self.assertTrue(mx.allclose(y, y_gt, rtol=1e-4, atol=1e-4))
                 self.assertTrue(mx.allclose(st, st_gt, rtol=1e-4, atol=1e-3))
+    
+
+    def test_afm7_kv_reuse_cache_count(self):
+        """
+        make_cache() should only allocate cache entries for non-reuse layers.
+        With 4 total layers and 1 reuse layer, expect 3 cache entries.
+        """
+
+        from mlx_lm.models import afm7
+
+        args = afm7.ModelArgs(
+            model_type="afm7",
+            vocab_size=32,
+            hidden_dim=32,
+            num_layers=4,
+            num_kv_reuse_layers=1,
+            num_heads=4,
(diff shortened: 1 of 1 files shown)


zcbenz:

PR #1437, opened 2026-06-26
Make RotatingKVCache trimmable so prefix cache reuse works for sliding-window models

amirarsalan90:
## Summary

`mlx-lm`'s automatic prompt-cache prefix reuse works for full-attention models but **silently fails for sliding-window-attention models** (Gemma 2/3/3n, GPT-OSS) once the prompt exceeds the attention window — every request reprocesses the whole prompt.

Root cause: `RotatingKVCache.is_trimmable()` returned `False` once its ring buffer wrapped. Since sliding-window models interleave layer types (e.g. Gemma 3 mixes `RotatingKVCache` with a full `KVCache` every 6th layer), the all-or-nothing `can_trim_prompt_cache()` gate then returned `False` for the entire model, so the server discarded the prompt cache and recomputed from scratch.

## Fix

- Split `RotatingKVCache`'s overloaded `offset` into two fields: `offset` (absolute chronological position, drives RoPE) and `_offset` (physical tokens resident in the ring buffer, capped at `max_size`). They diverge only after a trim; pre-wrap behavior is byte-identical. This mirrors the design `BatchRotatingKVCache` already uses.
- Implement a correct `trim()` for the **wrapped** case by linearizing the ring buffer into temporal order, bounded by `max_size` (older tokens are physically overwritten on wrap, so reuse is bounded by the window).
- Make the server reuse/eviction paths window-aware: `fetch_nearest_cache` only reuses a trimmed entry if the trim fully succeeded; `insert_cache` keeps prefixes older than the window as fallbacks; `trim_prompt_cache` returns the min trimmed across layers.

Scope: `RotatingKVCache` + `BatchRotatingKVCache`. SSM/Mamba (`ArraysCache`) is out of scope — recurrent state can't be split at a token boundary.

## Results

`gemma-3-1b-it-4bit`, 2802-token prompt (`[long fixed prefix] + [short changing tail]`), sliding window W=512:

| | Before | After |
|---|---:|---:|
| Warm `cached_tokens` | **0** | **2788** |
| Cold latency | 1.24 s | 1.24 s |
| Warm latency (median) | ~0.84 s | **0.15 s** |
| Speedup | 1.1× (none) | **~8×** |

Greedy (temperature=0) warm output is identical to cold.

## Tests

- `RotatingKVCache` / `BatchRotatingKVCache` now report trimmable after wrap and trim correctly back within the window.
- Added unit tests that a wrapped trim matches a fresh prefix build (state-level and through-model argmax identity), plus `_offset` save/load round-trip and backward-compat loading of the old 4-value `meta_state`.

`python -m unittest tests.test_prompt_cache tests.test_server tests.test_generate` passes.

Refs #980.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -108,7 +108,11 @@ def trim_prompt_cache(cache: List[Any], num_tokens: int) -> List[Any]:
     """
     if not can_trim_prompt_cache(cache) or len(cache) == 0:
         return 0
-    return [c.trim(num_tokens) for c in cache][0]
+    # Interleaved cache types (e.g. sliding-window models mixing RotatingKVCache
+    # and KVCache) may trim by different amounts. The number actually trimmed is
+    # the minimum across layers, since the cache is only consistent up to the
+    # least-trimmed layer.
+    return min(c.trim(num_tokens) for c in cache)
 
 
 def create_attention_mask(
@@ -414,7 +418,11 @@ class RotatingKVCache(_BaseCache):
         self.keep = keep
         self.keys = None
         self.values = None
+        # ``offset`` is the absolute chronological position (drives RoPE).
+        # ``_offset`` is the number of tokens physically resident in the ring
+        # buffer, capped at ``max_size``. The two diverge once the buffer wraps.
         self.offset = 0
+        self._offset = 0
         self.max_size = max_size
         self._idx = 0
 
@@ -434,7 +442,7 @@ class RotatingKVCache(_BaseCache):
         """
         if self._idx == v.shape[2]:
             return v
-        elif self._idx < self.offset:
+        elif self._idx < self._offset:
             return mx.concatenate(
                 [
                     v[..., : self.keep, :],
@@ -463,6 +471,7 @@ class RotatingKVCache(_BaseCache):
             self.keys = self._trim(trim_size, self.keys, keys)
             self.values = self._trim(trim_size, self.values, values)
         self.offset += keys.shape[2]
+        self._offset += keys.shape[2]
         self._idx = self.keys.shape[2]
         return self.keys, self.values
 
@@ -470,7 +479,7 @@ class RotatingKVCache(_BaseCache):
         # May not have hit the max size yet, so potentially
         # keep growing the cache
         B, n_kv_heads, S, k_head_dim = keys.shape
-        prev = self.offset
+        prev = self._offset
         if self.keys is None or (
             prev >= self.keys.shape[2] and self.keys.shape[2] < self.max_size
         ):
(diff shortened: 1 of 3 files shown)


nicedreamzapp:
Independent confirmation of this on **Gemma 4 31B (4-bit MLX)**, Apple Silicon, plus measurements in case they help make the case.

I hit this from the other direction: I maintain a small local coding agent and noticed long sessions were slow in a way short ones weren't. The cliff is exactly where you'd predict from this issue — `sliding_window` is 1024 on Gemma 4, and `make_cache()` hands 50 of 60 layers a `RotatingKVCache`, so `can_trim_prompt_cache()` goes `False` for the whole model once the transcript passes the window, and every turn re-prefills the entire conversation.

Measured on a 4.5k-token conversation, greedy, same model and machine:

| cache | time to first token |
|---|---|
| stock (`make_cache()`) | 6.5–7.2 s every turn |
| plain `KVCache` on all 60 layers | 0.36 s |

The part that might be useful for reviewing the fix: I also checked correctness rather than assuming. Running the same 4-turn conversation at temperature 0 with stock caches vs plain `KVCache` on every layer, the generated text was **identical on every turn** — which is what you'd expect given the sliding-window attention *mask* is what enforces the window, not the cache type. The cache type only decides what's retained. (Trade-off with the all-`KVCache` workaround is obvious: memory grows with the transcript instead of capping at `max_size`, which is exactly why a correct `trim()` like the one proposed here is the right fix rather than what I'm doing.)

Repro is straightforward: send turns until the transcript passes 1024 tokens, then check `can_trim_prompt_cache(cache)` — it flips to `False` and stays there. There's no error or warning, so the failure mode is just "it gets slow the longer you talk to it," which is easy to blame on the model.

Thanks @amirarsalan90 for writing this up — happy to run the patch against my setup and report numbers if that's useful.

amirarsalan90:
Thanks @nicedreamzapp, and yes please do. Numbers on a 31B with 50/60 rotating layers would be much better coverage than my 1B benchmark.

For anyone picking this up: #1502 independently landed on two of the server-side changes here (min trim count, and the full-trim guard in `fetch_nearest_cache`). I can rebase on it and drop those from this PR, leaving just the `offset`/`_offset` split and the wrapped-case `trim()`. Whichever order is easier.

One note on #980, which this references: it was closed as outdated, but exact-prefix caching doesn't cover this path. Once the ring wraps `is_trimmable()` returns False, and since these models interleave cache types the all-or-nothing gate then disables reuse for the whole model. If the scope isn't wanted I'm fine hearing that, just don't want it sitting.

@angeloskath CI has never run on this branch, I think workflows need approval for a first-time contributor.

zcbenz:

PR #1444, opened 2026-06-30
Fix frozen PRNG in categorical_sampling under repeated sampling

### LINKED ISSUE #1439: temperature > 0 is silently deterministic on the second call onward — categorical_sampling's @mx.compile(outputs=mx.random.state) doesn't advance the PRNG
Under inference with mlx-lm (and mlx-vlm, which reuses
`mlx_lm.sample_utils`), `temperature > 0` produces **byte-identical
output across repeat calls within a single process**, regardless of
`mx.random.seed()`, the `seed=` kwarg to `generate_step`, or any
sampling parameters.

Sampling appears to work for the first ~2 token positions in a process,
then silently collapses to a deterministic "stuck" sampler. Each token
position thereafter selects whichever token contains the same fixed
percentile of its distribution — confident decisions look greedy, less
confident decisions look weirdly off-distribution.

Server bounces are the only way to change output. Different processes
get different "stuck" sampler states, so behavior varies across
restarts but never within a session.

## Versions

- mlx-lm 0.31.3
- mlx-vlm 0.6.3 (also affected — reuses `mlx_lm.sample_utils`)
- mlx 0.31.2
- Python 3.12, macOS Darwin, Apple Silicon M-series

## Evidence

Two `/v1/chat/completions` calls, same prompt, `temperature=1.0`:

```
Request 1:  md5(content) = 7ac0dba2  "The user wants a haiku about coffee..."
Request 2:  md5(content) = 7ac0dba2  "The user wants a haiku about coffee..."
            ^^^^^^^^^^^^^ byte-identical
```

Tracing `mx.random.state` at each `categorical_sampling` call inside
the server shows the state advances once, then freezes:

```
call 1 state: [array([0, 2129127115])]                  ← initial
call 2 state: [array([1131398383, 4059925319])]         ← advanced
call 3 state: [array([1131398383, 4059925319])]         ← stuck
call 4 state: [array([1131398383, 4059925319])]         ← stuck
call 5 state: [array([1131398383, 4059925319])]         ← stuck
call 6 state: [array([1131398383, 4059925319])]         ← stuck
```

Same calls in a standalone Python script (no inference, just
`categorical_sampling(fake_logits, 1.0)` in a loop) DO advance state
correctly — so the bug only manifests in a real generation pipeline.
Something about `generate_step`'s context (compiled forward pass,
multiple `mx.compile`-decorated helpers sharing state, `mx.stream`
nesting, possibly thread affinity in a single-worker threadpool)
triggers the freeze.

## Workaround (confirmed working)

Replace `categorical_sampling` with a non-compiled version at process
startup:

\`\`\`python
import mlx.core as mx
import mlx_lm.sample_utils as su

def _cat_uncompiled(logits, temp):

su.categorical_sampling = _cat_uncompiled

# mlx-vlm imports the same symbol — patch it too if used:
try:
    import mlx_vlm.sample_utils as vsu
    vsu.categorical_sampling = _cat_uncompiled
except ImportError:
    pass
\`\`\`

After this patch:
- Same prompt + same `temperature > 0` → varied output across calls ✓
- `mx.random.seed()` actually changes output ✓
- Reasoning models with thinking blocks now produce varied actions, not
  just varied reasoning text

## Why this matters

CoT / thinking-mode models are particularly sensitive. They produce
varied `reasoning_content` (the thinking-phase sampler hits a
different code path) but byte-identical action/content output. So
the model "thinks" differently each turn while always emitting the
same final answer — sampling-driven exploration is silently lost.

For tool-calling agents this manifests as unbreakable loops: the model
re-thinks its decision each turn but always emits the same broken tool
call because the action distribution is peaked and the stuck percentile
always lands on the same top token.

For standard chat use, output across sessions feels "inconsistent"
(each process restart picks a different stuck state) while output
within a session is suspiciously deterministic. Easy to misattribute
to model variance or quantization noise.

## Suggested fix

The most direct option: remove the `@partial(mx.compile, ...)`
decorator from `categorical_sampling`. The body is a single
`mx.random.categorical` call — compile gains are negligible and the
correctness cost is total.

Longer-term: investigate whether `mx.compile` with
`outputs=mx.random.state` is supposed to bind state on first invocation
under heavy inference contexts. If yes, document the contract; if no,
that's an mlx-level issue worth filing separately.

## Related code

- `mlx_lm/sample_utils.py` — `categorical_sampling` decorator
- `mlx_lm/generate.py` (and `mlx_vlm/generate/`) — sampler call sites
  during decode loop
- `mlx_vlm/sample_utils.py` — re-exports the affected symbol

utkarshtiwari-24:
## Problem

With `temperature > 0`, repeated requests to `mlx_lm.server` return identical text. The sampling stops being random after the first request.

## Root cause

`categorical_sampling` is decorated with `@partial(mx.compile, inputs=mx.random.state, outputs=mx.random.state)`. In the server's generation loop the RNG state stops advancing after the first request, so every sample reuses the same state. Calling `categorical_sampling` or `generate()` directly does not hit this (the state advances normally), so it only appears in the long-running compiled server path.

## Fix

Remove the decorator. The function is a single `mx.random.categorical` call, so the compile gives little benefit, and without it the function uses the live global RNG state on each call.

## Repro

1. Start the server:
mlx_lm.server --model mlx-community/Qwen2.5-0.5B-Instruct-4bit

2. Run the below curl cmd twice:
curl -s localhost:8080/v1/chat/completions -H "Content-Type: application/json" \
   -d '{"messages":[{"role":"user","content":"Tell me a one-sentence story."}],"temperature":1.0,"max_tokens":40}'

Before, two identical requests return byte-identical completions:
1. Once upon a time, in a neighboring village, a young woman named Rose found herself lost ...
2. Once upon a time, in a neighboring village, a young woman named Rose found herself lost ...

After, they differ:
 1. In a land of empires ... a hero named Robin ... a mighty rebellion against the Elite Nobility ...
 2. Ah, a sample story for you: When Eliza first arrived in New York, she felt immense pressure ...

`cached_tokens` is still 36 on the second call after the fix, so prompt cache reuse is unaffected; only the RNG state changed.

## Testing

`tests/test_sample_utils.py` passes. No unit test added: the freeze only reproduces in the compiled server loop, not a direct call, so a unit test would pass on the unfixed code too. Happy to add an integration test if preferred.

Note: `apply_top_k`, `apply_top_p`, and `apply_min_p` carry the same `random.state` decorator but do not use randomness. Left unchanged here.

Fixes #1439

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -274,7 +274,6 @@ def apply_xtc(
     )
 
 
-@partial(mx.compile, inputs=mx.random.state, outputs=mx.random.state)
 def categorical_sampling(logits, temp):
     return mx.random.categorical(logits * (1 / temp))
 


nastya236:
Thanks for the fix. Can you please give tokens per second numbers with/without compile for categorical sampling?

utkarshtiwari-24:
> Thanks for the fix. Can you please give tokens per second numbers with/without compile for categorical sampling?

Thanks for your comment. Here are the numbers:

Benchmarks on an Apple M4 Pro (24GB), temp 1.0, 512 tokens, 5 runs each (generation tokens/sec):

Qwen2.5-0.5B-4bit
  with compile:      ~364 tokens/sec (354-376)
  without (this PR): ~353 tokens/sec  (350-358) -> ~3% drop

Qwen3.5-2B-4bit
  with compile:      ~175.8 tokens/sec (175.6-176.0)
  without (this PR): ~174.8 tokens/sec (174.6-175.4) -> ~0.6% drop

Qwen3.5-9B-4bit
  with compile:      ~50.1 tokens/sec (50.0-50.3)
  without (this PR): ~50.0 tokens/sec (49.9-50.0) -> ~0.3% drop

categorical_sampling only runs one op per token, so compiling it saves a tiny fixed amount per token. On a small model that's noticeable because each token is so fast (0.5B, ~3%), but on bigger models most of the time goes into the model itself, so it basically vanishes (~0.6% at 2B, ~0.3% at 9B). At normal model sizes it doesn't really affect speed.

utkarshtiwari-24:
@nastya236 please review and merge the PR if approved.

PR #1447, opened 2026-07-01
Fix dropped tool calls for models with empty tool_call_end (Mistral/Devstral)

DavidObando:
## Summary

Fixes #1307.

Models that use Mistral's native tool-call format (e.g. `Devstral-Small-2-24B`, `Ministral`, `Mistral-Small`) emit tool calls with **no explicit end marker** — their `tool_call_end` is `""`. As a result, the server's `SequenceStateMachine` has no `tool -> normal` transition, and the **final EOS** transitions the state out of `"tool"` to `None`.

At the end of `handle_completion`, the trailing flush is guarded by:

```python
if prev_state == "tool" and tool_text:
    tool_calls.append(tool_text)
    made_tool_call = True
```

Because the last token (EOS) already moved the state to `None`, `prev_state` is `None`, the guard never fires, and the buffered `tool_text` is silently discarded. The server returns an empty `message: {"role": "assistant"}` — neither `content` nor `tool_calls` — even though the model produced a valid `[TOOL_CALLS]name[ARGS]{json}` call.

## Fix

Key the end-of-generation flush off the buffered `tool_text` instead of `prev_state`. `tool_text` is only ever populated while in the `"tool"` state and is reset to `""` when a `tool -> normal` transition flushes it during the loop, so any leftover `tool_text` at the end is by definition an unflushed tool call. This is general and also covers models whose `tool_call_end` **is** defined (in that case `tool_text` is already empty here, so behavior is unchanged).

## Reproduction (before fix)

```bash
mlx_lm.server --model mlx-community/Devstral-Small-2-24B-Instruct-2512-8bit --port 8080
curl -s localhost:8080/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"mlx-community/Devstral-Small-2-24B-Instruct-2512-8bit",
  "messages":[{"role":"user","content":"List files using the tool."}],
  "tools":[{"type":"function","function":{"name":"run_shell","parameters":{"type":"object","properties":{"command":{"type":"string"}},"required":["command"]}}}]
}'
```

- **Before:** `message: {"role": "assistant"}` (call dropped), `finish_reason: "stop"`.
- **After:** `finish_reason: "tool_calls"`, `tool_calls: [{function: {name: "run_shell", arguments: "{\"command\": \"ls\"}"}}]`.

## Tests

[...]

Issue #1307, opened 2026-05-24
Devstral-Small-2-24B-Instruct-2512 (MXFP4): model emits valid [TOOL_CALLS] markup but mlx_lm.server returns empty message


aanjhan:
## Summary

When serving `mlx-community/mistralai_Devstral-Small-2-24B-Instruct-2512-MLX-MXFP4` via `mlx_lm.server` (v0.31.3), the model correctly generates Mistral's native `[TOOL_CALLS]name[ARGS]{json}` tool-call output (verified via `--log-level DEBUG`), but the server returns an empty `message: {"role": "assistant"}` — neither the structured `tool_calls` field nor the `content` field is populated. The model output is silently stripped.

This appears to be the same class of parser-mismatch bug as #1096 (Gemma 4) and a sibling of the issue fixed by #1151 / #1110, but for a different Mistral variant.

## Environment

- mlx-lm: **0.31.3**
- transformers: **5.9.0**
- Python: **3.14**
- OS: **macOS 26.5 (arm64)**
- Hardware: **M1 Max, 64GB unified memory**
- Model: **`mlx-community/mistralai_Devstral-Small-2-24B-Instruct-2512-MLX-MXFP4`** (14.1GB, converted from `mistralai/Devstral-Small-2-24B-Instruct-2512`)

## Reproduction

```bash
# 1. Serve the model with DEBUG logging
mlx_lm.server \
    --model mlx-community/mistralai_Devstral-Small-2-24B-Instruct-2512-MLX-MXFP4 \
    --port 8080 \
    --log-level DEBUG

# 2. In another terminal, send a basic tool-call request
curl -s http://localhost:8080/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model":"mlx-community/mistralai_Devstral-Small-2-24B-Instruct-2512-MLX-MXFP4",
    "messages":[{"role":"user","content":"What is on my calendar today?"}],
    "tools":[{"type":"function","function":{"name":"calendar_store","description":"Read calendar events.","parameters":{"type":"object","properties":{"mode":{"type":"string"},"range":{"type":"string"}}}}}],
    "tool_choice":"auto"
  }' | python3 -m json.tool
```

## Expected behavior

Response should contain a structured `tool_calls` array:

```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "tool_calls": [{
        "id": "...",
        "type": "function",
        "function": {
          "name": "calendar_store",
[...]

PR #1448, opened 2026-07-02
Fix Mistral tool parser dropping parallel/multiple tool calls

DavidObando:
## What

Fix the Mistral tool-call parser (`mlx_lm/tool_parsers/mistral.py`) dropping tool calls when a generation contains **multiple (parallel) tool calls** or a **single call followed by trailing text**.

## Why

Mistral/Devstral have an empty `tool_call_end`, so the server's state machine never leaves the `"tool"` state until EOS and accumulates the entire tail of the generation into a single string. That string can look like:

```
get_weather[ARGS]{"city": "Paris"}[TOOL_CALLS]get_weather[ARGS]{"city": "Tokyo"}
```

or a single call with trailing prose:

```
grep[ARGS]{"pattern": "TODO"}

Let me search for that.
```

The old parser used a greedy regex `(\{.*\})` and `json.loads` on the whole capture, which raises `json.JSONDecodeError: Extra data ...` in both cases. `ToolCallFormatter` catches that error and `continue`s, so **the call is silently dropped and the client receives an empty message** — an agent driving the server (e.g. via the OpenAI-compatible endpoint) then stalls with no tool call and no content.

## How

Locate each `name[ARGS]` header and decode **exactly one** JSON object with `json.JSONDecoder().raw_decode()`, which respects string/brace boundaries and ignores any trailing data. Scanning resumes after each decoded object, so all calls are collected. Returns a single dict for one call and a list for multiple — matching the existing `gemma4` and `kimi_k2` parsers, and `ToolCallFormatter` already normalizes both shapes.

## Tests

Adds `test_mistral` to `tests/test_tool_parsing.py` covering:
- single call followed by natural-language text,
- parallel calls with the `[TOOL_CALLS]` separator,
- multiple calls with no separator,
- nested JSON and literal braces / `[ARGS]` inside a string,
- the no-call `ValueError` path.

```
$ python -m unittest tests.test_tool_parsing -v
... test_mistral ... ok
Ran 6 tests OK
```

## Related

Complements #1447, which fixes the server-side flush so the trailing buffered call reaches the parser in the first place. This PR fixes the parser itself so multi/trailing calls are parsed rather than dropped. Together they address the empty-response behavior with Mistral/Devstral tool calling (see also #1307).

### DIFF
--- a/mlx_lm/tool_parsers/mistral.py
+++ b/mlx_lm/tool_parsers/mistral.py
@@ -5,16 +5,39 @@ from typing import Any
 
 import regex as re
 
-_tool_call_regex = re.compile(r"\s*(\w+)\[ARGS\]\s*(\{.*\})", re.DOTALL)
+# Matches a single tool-call header of the form "name[ARGS]" and any trailing
+# whitespace, positioned right before the JSON arguments object.
+_tool_call_header_regex = re.compile(r"(\w+)\s*\[ARGS\]\s*", re.DOTALL)
 
 tool_call_start = "[TOOL_CALLS]"
 tool_call_end = ""
 
 
 def parse_tool_call(text: str, tools: Any | None = None):
-    match = _tool_call_regex.search(text)
-    if match is None:
+    # Mistral has no explicit tool-call end token, so the server accumulates the
+    # whole tail of the generation into a single string. That string may contain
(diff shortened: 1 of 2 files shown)


michalk8 (approved):
Thanks, just made some minor changes, LGTM.

PR #1452, opened 2026-07-02
Fix DSA indexer LoRA-training crash: stop gradients through sparse-attention top-k indices

### LINKED ISSUE #1449: [BUG] LoRA fine-tuning of quantized GLM-5.2 (glm_moe_dsa) MoE crashes in backward: [scatter_axis] Cannot calculate VJP with respect to indices
## Describe the bug

LoRA fine-tuning of a **quantized GLM-5.2** MoE model (`glm_moe_dsa`) crashes **deterministically in the backward pass** with:

```
ValueError: [scatter_axis] Cannot calculate VJP with respect to indices.
```

The forward and loss compute fine; the crash is in `loss_value_and_grad` (the gradient), and it happens at the **same step every run** under a fixed seed (after iter 10 → step 11 in my case), i.e. it is triggered by the first batch that routes through the offending expert-scatter path, not by anything stochastic.

Root cause is upstream in MLX autograd: the MoE expert routing scatters/gathers by top-k indices that depend on the trainable router, and `ScatterAxis::vjp` throws for the index argument instead of returning a zero cotangent. See companion issue **ml-explore/mlx#3787** with a 5-line minimal repro and the exact `primitives.cpp` throw sites.

## Reproduction

- Model: GLM-5.2, 4-bit quantized, architecture `glm_moe_dsa`.
- `mlx_lm.lora` with a LoRA config over all layers (r=16), `mask_prompt: true`, `grad_checkpoint: true`, `max_seq_length: 4096`, `batch_size: 4`.
- Training descends normally, then crashes at the first backward that hits the MoE scatter.

Loss trajectory before the crash (healthy — this is **not** an LR/instability problem):

```
Iter 1: Val loss 1.789
Iter 2: Train 1.672   Iter 4: 1.490   Iter 6: 0.688 (LR reached 3e-5)
Iter 8: 0.554   Iter 10: 0.682
<crash in loss_value_and_grad>
```

Traceback:

```
File ".../mlx_lm/tuner/trainer.py", line 250, in step
File ".../mlx/nn/utils.py", line 35, in wrapped_value_grad_fn
ValueError: [scatter_axis] Cannot calculate VJP with respect to indices.
```

## Diagnostic — it is the MoE forward scatter, not the expert LoRA weights

I re-ran with the `switch_mlp` layers **excluded** from the LoRA target keys (trainable params dropped 7.65B → 102M). It crashed **identically at the same step**. So the failure is in the MoE forward scatter that gradients must flow *through* (router → top-k indices → `ScatterAxis`), and cannot be avoided by changing which LoRA keys are trained, nor by lowering LR / reducing `max_seq`.

## Environment

- mlx-lm **0.31.3**, mlx **0.31.2**
- Python 3.12.13, macOS 26.2, Apple Silicon (M3 Ultra, 512 GB)

## Ask

- Primary fix belongs upstream in MLX (zero VJP for scatter/scatter_axis index args) — linked companion issue.
- On the mlx-lm side: is there a supported way to make quantized MoE (GLM-5.2 / other `switch_mlp` models) trainable today (e.g. detaching/stop-grad on the routing indices in the model definition), or should GLM-5.2 LoRA be considered blocked until the MLX autograd fix lands? A stop-gradient on the computed expert indices in the MoE layer would sidestep the throw regardless of the core fix.

trevorgordon981:
## What

LoRA fine-tuning of DSA sparse-attention models (`deepseek_v32`, and `glm_moe_dsa` which subclasses it) crashes deterministically in the backward pass:

```
ValueError: [scatter_axis] Cannot calculate VJP with respect to indices.
```

Fixes #1449. Companion core issue: ml-explore/mlx#3787.

## Root cause

The DSA `Indexer` only activates when the key length exceeds `index_topk` (2048); below that it returns `None` and training works fine. Once a batch crosses that width, the indexer computes `topk_indices` via `mx.argpartition` over scores that derive from trainable projections (`wq_b`, `wk`, `weights_proj`, and the shared `q_a_proj` — all typical LoRA targets). Those integer indices then feed `mx.put_along_axis` (training/prefill path) and `mx.take_along_axis` (decode path) in the attention. Because the indices lie on the autodiff path, autograd requests a VJP with respect to the scatter's index argument, and `ScatterAxis::vjp` throws instead of returning a zero cotangent (ml-explore/mlx#3787).

This explains the reproduction in #1449 exactly: the crash lands on the first batch whose sequence width crosses 2048 — the same step every run under a fixed seed — and is unaffected by which LoRA keys are trained, the learning rate, or excluding `switch_mlp`. (My original report blamed the MoE routing scatter; that was wrong — the MoE gate's scatter is already protected by `mx.stop_gradient` on `group_idx`. The unprotected site is the indexer.)

## Fix

`mx.stop_gradient` on the top-k indices at two sites:

1. **Producer** — wrap the `Indexer.__call__` return. Integer top-k indices are a non-differentiable selection, so this is semantically exact: the forward pass is bit-identical, and inference never differentiates so serving is unaffected.
2. **Consumer** — re-assert the barrier where `topk_indices` is consumed in `DeepseekV32Attention.__call__`. This is defensive: if the indices ever reach the attention through a traced boundary (e.g. as a checkpointed function input under `grad_checkpoint` in a variant that passes indices between layers), a producer-side barrier is not visible to the re-traced graph and the crash returns. It is a no-op when the indices come straight from the producer, as here.

A zero-cotangent fix in core MLX (ml-explore/mlx#3787) would also resolve the throw, but the model-side barrier is exact regardless, avoids computing gradients that are discarded, and unblocks training today.

## Validation

- Applied the same two barriers to an mlx-lm 0.31.3 install and ran `mlx_lm.lora` on GLM-5.2 (4-bit, `glm_moe_dsa`) with `grad_checkpoint: true`, `max_seq_length: 4096` on real data: previously crashed at the first >2048-token batch, now trains cleanly (30-iteration smoke run, healthy loss descent).
- `python -m py_compile` and `black --check` pass on the changed file.

### DIFF
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -109,9 +109,16 @@ class Indexer(nn.Module):
         scores = scores.sum(axis=1, keepdims=True)
         if mask is not None:
             scores = mx.where(mask, scores, -float("inf"))
-        return mx.argpartition(scores, kth=-self.index_topk, axis=-1)[
-            ..., -self.index_topk :
-        ]
+        # The top-k indices are an integer selection and carry no gradient.
+        # Stop gradients here so that when training (e.g. LoRA) with
+        # sequences longer than index_topk, autograd does not request an
+        # indices VJP from the scatter/gather ops that consume them
+        # ("Cannot calculate VJP with respect to indices").
+        return mx.stop_gradient(
+            mx.argpartition(scores, kth=-self.index_topk, axis=-1)[
+                ..., -self.index_topk :
+            ]
+        )
 
 
 class DeepseekV32Attention(nn.Module):
@@ -206,6 +213,11 @@ class DeepseekV32Attention(nn.Module):
 
         topk_indices = self.indexer(x, qr, mask, cache=cache[1])
         if topk_indices is not None:
+            # Also stop gradients at the point of use: if the indices ever
+            # arrive through a traced boundary (e.g. as a checkpointed
+            # function input with grad_checkpoint), the producer-side
+            # stop_gradient is not visible to the re-traced graph.
+            topk_indices = mx.stop_gradient(topk_indices)
             if L == 1:
                 idx = topk_indices[:, :, 0, :, None]
                 kv_latent = mx.take_along_axis(
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -111,16 +111,8 @@ class Indexer(nn.Module):
         scores = scores.sum(axis=1, keepdims=True)
         if mask is not None:
             scores = mx.where(mask, scores, -float("inf"))
-        # The top-k indices are an integer selection and carry no gradient.
-        # Stop gradients here so that when training (e.g. LoRA) with
-        # sequences longer than index_topk, autograd does not request an
-        # indices VJP from the scatter/gather ops that consume them
-        # ("Cannot calculate VJP with respect to indices").
-        return mx.stop_gradient(
-            mx.argpartition(scores, kth=-self.index_topk, axis=-1)[
-                ..., -self.index_topk :
-            ]
-        )
+        indices = mx.argpartition(scores, kth=-self.index_topk, axis=-1)
+        return mx.stop_gradient(indices[..., -self.index_topk :])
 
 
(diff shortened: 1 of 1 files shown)


michalk8 (approved):
LGTM

PR #1454, opened 2026-07-02
DeepSeek-V3.2/GLM DSA: fix silent >128k top-k corruption + sparse-gather prefill

aidiffuser:
# PR: DeepSeek-V3.2/GLM DSA: fix silent >128k top-k corruption + sparse-gather prefill

Two independent improvements to the DSA (sparse-attention) models, split into
two commits so they can be discussed/landed separately.

## Commit 1 — fix: silent retrieval collapse past ~128k context

`Indexer.__call__` reduces the per-head score tensor with
`scores.sum(axis=1, keepdims=True)`. For long prompts under chunked prefill
this tensor reaches shapes like `[1, 32, 4096, 131072]`, where `mx.sum` over
the non-last axis returns **all-zeros** (MLX bug, filed with a minimal repro as
ml-explore/mlx#3784). Zeroed scores make `argpartition` select the trailing
`index_topk` positions — attention silently degrades to "last 2048 tokens
only" and retrieval quality collapses past ~128k. No crash, no warning.

**Repro (GLM-5.2, needle-in-haystack at 131,072-token contexts): 0/5 recall.
With this fix: 5/5.** DeepSeek-V3.2-Exp uses the same Indexer and tensor
shapes and is equally affected.

The fix reduces the head axis with elementwise adds during prefill (`s > 1`),
which never invokes the buggy reduction path; decode keeps the fused sum (its
tensor is orders of magnitude below the trigger size). Cost is negligible
relative to the `q @ k.T` that precedes it. Tagged with a TODO to revert once
the MLX fix lands.

## Commit 2 — perf: sparse-gather prefill attention

The indexer computes exact top-k indices per query, but the `L > 1` prefill
path then runs **dense** attention over all `S` keys with non-selected keys
`-inf`-masked — the computed sparsity is discarded and every chunk pays
O(L·S).

This commit attends over only the `index_topk` gathered keys (absorbed-latent
form, mirroring the existing `L == 1` decode path): O(L·topk) per chunk, and
the transient footprint stays flat in context length instead of growing
linearly (no `[.., L, S]` score/mask materialization).

- **Identical output:** it attends over the exact same selected key set;
  max-abs diff vs the dense-masked path on identical inputs is ~3e-7 (fp
  noise).
- **Gated, zero short-context regression:** engaged only above
  `SPARSE_PREFILL_MIN_CONTEXT = 16384`; the measured dense/sparse crossover is
  ~14k on M-series (the gather has a fixed overhead that loses to dense SDPA
  on short prompts).
- **Measured (GLM-5.2, bf16):** attention-op ~9× at 32k and flat in S where
  dense keeps growing; end-to-end prefill 1.39× at 41k and widening with
  depth. With both commits we've validated 100% needle recall at 300k+ token
  contexts.

Concept adapted from the sparse-MLA prefill work in jundot/omlx#1984
(Apache-2.0); this is a dependency-free pure-MLX formulation.

## Validation

- Parity harness: dense vs sparse on the real `DeepseekV32Attention` module,
  identical inputs — max-abs diff 2.7e-7.
- Long-context: needle recall 5/5 at exactly 131,072 tokens (the mx#3784
  trigger shape), 100% structured recall at 120k/200k/300k/350k on GLM-5.2.
- `python -m py_compile` + existing model loads unaffected (default path
  unchanged below 16k).

### DIFF
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -53,6 +53,13 @@ class ModelArgs(BaseModelArgs):
     indexer_rope_interleave: bool = False
 
 
+# Context length above which prefill switches from dense top-k-masked SDPA to
+# the sparse gather in _sparse_gather_attention. Below this the gather's fixed
+# overhead loses to dense SDPA (measured crossover ~14k on M-series); above it
+# the sparse path wins and keeps growing with context (~9x attention-op at 32k).
+SPARSE_PREFILL_MIN_CONTEXT = 16384
+
+
 class Indexer(nn.Module):
     def __init__(self, args: ModelArgs):
         super().__init__()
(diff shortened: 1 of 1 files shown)


aidiffuser:
Friendly ping — this has been open ~2 months without review, and it's a silent-corruption class bug, not a perf nit.

For anyone serving DSA models (DeepSeek-V3.2, GLM-5.2/5.3-Flash) on stock mlx-lm: past ~128k ctx the indexer's `scores.sum(axis=1, keepdims=True)` hits the non-last-axis `mx.sum` zeroing bug ([ml-explore/mlx#3784](https://github.com/ml-explore/mlx/issues/3784)) on shapes like `[1, 32, 4096, 131072]`. The zeroed scores make `argpartition` deterministically select the trailing `index_topk` positions — attention quietly becomes "last 2048 tokens only." No crash, no warning.

Measured on GLM-5.2 NIAH @131,072: **0/5 recall stock → 5/5 with commit 1's workaround.** Commit 2 (sparse-gather prefill, 2x @128k) is independent — glad to split it out if that eases review.

Status check: still unfixed as of `v0.32.2` — `mlx/backend/metal/reduce.cpp` hasn't been touched since this was filed (the only reduction commits since Jun 1 address `any`/`all` `-0.0` and 0-size `min`/`max`), and no duplicate fix has landed anywhere else.

Alternatively I can land the same workaround in `jundot/omlx`, where our glm-dsa kernels already live — but a fix here protects every stock user in the meantime.

michalk8:
Thanks for pointing this out and sorry for the delay. I've decided to keep your fix for the silent bug and drop the DSA performance (it's out of scope of this PR; feel free to submit a future PR).

michalk8 (approved):
LGTM

Issue #872, opened 2026-02-11
Internal MTP for Qwen3-based and Nemotron-based models


TomLucidor:
A lot of systems need 2 models, can we pick models that has their own MTP? Qwen3-Next have them already. Qwen3-32B and Qwen3-235B have support for it as well as Nemotron-3-Super (~100B) and Nemotron-3-Ultra (~500B)

Reference
- https://github.com/ml-explore/mlx-lm/issues/846
- https://github.com/waybarrios/vllm-mlx/issues/56#issuecomment-3881274619

WiseFarAI:
> A lot of systems need 2 models, can we pick models that has their own MTP? Qwen3-Next have them already. Qwen3-32B and Qwen3-235B have support for it as well as Nemotron-3-Super (~100B) and Nemotron-3-Ultra (~500B)
> 
> Reference
> 
> * [Speculative decoding produces incorrect output with Qwen3 models (skips tokens) #846](https://github.com/ml-explore/mlx-lm/issues/846)
> * [MTP Support? waybarrios/vllm-mlx#56 (comment)](https://github.com/waybarrios/vllm-mlx/issues/56#issuecomment-3881274619)

MTP support for Qwen 3.5 could be a meaningful speedup, especially if MLX gets proper end-to-end support for it. Right now, my understanding is that MTP-related pieces are not preserved in the current MLX conversion / quantization path, and the runtime stack for using them is not there yet. I do think some form of this will likely arrive over time, and not just here but in many inference apps like Llama.cpp and others.

AirRunner:
Following up on the discussion in #441 (@Goekdeniz-Guelmez @janhilgard @TomLucidor), I have a working implementation of native MTP speculative decoding for Qwen3.5 in mlx-lm.

Fork: https://github.com/AirRunner/mlx-lm/tree/feat/mtp-native

This covers what @janhilgard identified in #441:
- MTP weight preservation in sanitize() (also fixes missing +1 norm shift for mtp.pre_fc_norm_hidden/embedding and mtp.norm).
- model.mtp_forward(hidden, token_ids, mtp_cache) and model.make_mtp_cache() on the model class
- mtp_generate_step() in generate.py. draft/verify loop producing up to 2 tokens per backbone pass (no CLI flag yet, so auto-dispatched from stream_generate when the model has an MTP head).
- SSM state rollback on draft rejection (*).

[...]

PR #1456, opened 2026-07-03
qwen3_5: load in-checkpoint MTP head + speculative rollback for hybrid (GDN) caches

pierre427:
## What

Two related changes to `mlx_lm/models/qwen3_5.py`:

**1. Bugfix: mlx-community `-mtp` checkpoints load as garbage today.**
`sanitize()` treats `mtp.*` presence as a raw-HF-checkpoint signal and
applies the +1.0 zero-centered-norm shift. Converted checkpoints that
*keep* the MTP tensors (e.g. `mlx-community/Qwen3.5-122B-A10B-oQ4-mtp`,
converted with mlx-vlm) already have shifted norms — so every RMSNorm
gets double-shifted and the model emits fluent-speed gibberish. The
raw-checkpoint signal is now the unsanitized conv1d layout only, which
every real Qwen3.5 checkpoint has (kernel size 4). The existing
`test_qwen3_5_family_convert_then_load_norm_not_shift_twice` test is
updated to use a kernel-4 config so the conv signal is exercised, and
extended with the kept-mtp regression case.

**2. Feature: load the vendor-trained MTP head + support speculative
rollback on hybrid caches.**
- `mtp_num_hidden_layers` config field builds an `MTPModule`
  (`fc([norm(embed); norm(hidden)])` → full-attention decoder layer →
  norm → shared lm_head — the Qwen3-Next MTP structure). The module is
  dropped in `sanitize()` when a checkpoint has no `mtp.*` tensors, so
  non-mtp conversions keep loading strictly. Conversion now preserves
  and quantizes the MTP tensors instead of silently discarding them.
- `TextModel.mtp_step(hidden, tokens, mtp_cache)` runs one MTP forward
  (batched catch-up or single-token recursive drafting).
- `TextModel.rollback_speculative_cache(caches, gdn_states, keep,
  block_size)` rewinds after a verify forward: KV caches trim; the 36
  GatedDeltaNet recurrent states — which cannot trim — are rebuilt by
  replaying the captured layer inputs (`gdn_sink` kwarg on the forward
  path) on the accepted prefix, all linear layers batched into a single
  `gated_delta_update` call. Single-sequence scope; batched caches raise
  a clear error. (Replay pattern after mlx-vlm's qwen3_5 implementation
  and MTPLX, both Apache-2.0.)

## Why

Qwen3.5's hybrid linear-attention layers use `ArraysCache`, which is not
trimmable — stock speculative decoding cannot rewind rejected tokens, so
these models are currently locked out of speculation in mlx-lm entirely,
and the shipped MTP head (the natural, draft-model-free way to speculate)
is discarded at load. This PR provides the loading + rollback primitives;
generation-loop integration can follow separately if there's interest.

## Measured (M5 Max 128GB, Qwen3.5-122B-A10B-oQ4-mtp, greedy)

MTP self-speculative chain built on these primitives, vs interleaved
same-harness plain decode:

| workload | plain | MTP k=4 | speedup |
|---|---|---|---|
| structured list | 47.8 | 90.5 | 1.89× |
| doc-diff table | 49.4 | 104.3 | 2.11× |
| planning | 49.6 | 84.6 | 1.71× |
| open prose | 50.1 | 68.0 | 1.36× |

Depth-1 acceptance 83–100%. Losslessness: output ≡ plain greedy on
structured + prose probes, with divergences only at positions where
single-row recomputation shows a top-2 logit gap ≤ 1 bf16 ULP (0.000 /
0.125 observed) — the same batched-vs-sequential kernel tie behavior the
unmodified pipeline already exhibits. A forced-reject stress test
(every cycle rejects all drafts, exercising GDN replay on every cycle)
matches plain greedy under the same criterion.

## Tests

- `tests/test_qwen3_5_mtp.py` (new): rollback-replay ≡ sequential ground
  truth for keep = 0..block; mtp_step shapes + recursive chaining +
  trim; module dropped when checkpoint lacks mtp tensors; norm-shift
  fires for raw layout and not for converted layout (both directions).
- `tests/test_models.py`: norm-shift contract test updated as above.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -526,7 +526,9 @@ def speculative_generate_step(
         model_cache = prompt_cache[: len(model.layers)]
         draft_cache = prompt_cache[len(model.layers) :]
 
-    if not cache.can_trim_prompt_cache(model_cache):
+    if not cache.can_trim_prompt_cache(model_cache) and not getattr(
+        model, "supports_speculative_rollback", False
+    ):
         types = {type(c).__name__ for c in model_cache if not c.is_trimmable()}
         raise ValueError(
             f"Speculative decoding requires a trimmable prompt cache " f"(got {types})."
@@ -604,6 +606,12 @@ def speculative_generate_step(
         draft_y = _prefill(draft_model, draft_cache, y)
         y = _prefill(model, model_cache, y)
 
+    # After prefill (recording a rollback for the whole prompt would hold on to
+    # prompt-sized tensors), let rollback-capable caches start recording so
+    # verify steps can be trimmed exactly (no-op for regular KV caches).
+    for c in model_cache:
+        c.start_speculation()
+
     ntoks = 0
     # Set these so the finally block doesn't raise
     num_draft = 0
@@ -652,6 +660,8 @@ def speculative_generate_step(
             _rewind_cache(num_draft, n)
     finally:
         _rewind_cache(num_draft, n)
+        for c in model_cache:
+            c.stop_speculation()
 
 
 def stream_generate(
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -674,7 +674,10 @@ def speculative_generate_step(
             _rewind_cache(num_draft, n)
     finally:
         _rewind_cache(num_draft, n)
-        for c in model_cache:
+        # Symmetric with start_speculation above: draft caches must stop
+        # recording too, or a reused draft prompt cache keeps reporting
+        # is_trimmable() with stale rollback records.
+        for c in model_cache + draft_cache:
             c.stop_speculation()
 
 
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -526,10 +526,17 @@ def speculative_generate_step(
         model_cache = prompt_cache[: len(model.layers)]
         draft_cache = prompt_cache[len(model.layers) :]
 
-    if not cache.can_trim_prompt_cache(model_cache) and not getattr(
-        model, "supports_speculative_rollback", False
-    ):
-        types = {type(c).__name__ for c in model_cache if not c.is_trimmable()}
+    def _can_speculate(c):
+        # Trimmable directly, or able to record an exact rollback while
+        # speculating (see ArraysCache.record_rollback). The model must also
+        # declare support: recording is done by its recurrent layers.
+        return c.is_trimmable() or (
+            hasattr(c, "record_rollback")
+            and getattr(model, "supports_speculative_rollback", False)
+        )
+
+    if not all(_can_speculate(c) for c in model_cache):
+        types = {type(c).__name__ for c in model_cache if not _can_speculate(c)}
         raise ValueError(
             f"Speculative decoding requires a trimmable prompt cache " f"(got {types})."
         )
@@ -608,8 +615,10 @@ def speculative_generate_step(
 
     # After prefill (recording a rollback for the whole prompt would hold on to
     # prompt-sized tensors), let rollback-capable caches start recording so
-    # verify steps can be trimmed exactly (no-op for regular KV caches).
-    for c in model_cache:
+    # verify steps can be trimmed exactly (no-op for regular KV caches). The
+    # draft cache records too: hybrid draft models (e.g. a small Qwen3.5) need
+    # their recurrent state rewound just like the target's.
+    for c in model_cache + draft_cache:
         c.start_speculation()
 
     ntoks = 0
@@ -619,6 +628,11 @@ def speculative_generate_step(
     try:
         while True:
             num_draft = min(max_tokens - ntoks, num_draft_tokens)
+            # Until this round's verify step has advanced the caches there is
+            # nothing to rewind; keep n == num_draft so an exception here (or a
+            # generator close) makes the finally-block rewind a no-op instead
+            # of trimming with values from a previous round.
+            n = num_draft
             draft_tokens = _draft_generate(draft_y, num_draft)
             if prev_tokens is not None:
                 prev_tokens = prev_tokens[: prev_tokens.size - y.size - num_draft + 1]
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -146,6 +146,19 @@ class _BaseCache:
     def is_trimmable(self):
         return False
 
+    def start_speculation(self):
+        """Called by speculative decoding before draft/verify steps begin.
+
+        Caches that cannot trim directly but support an exact rollback (e.g.
+        recurrent-state caches, see ``ArraysCache``) use this to start recording
+        the information needed to roll a verify step back. No-op by default.
+        """
+        pass
+
+    def stop_speculation(self):
+        """Called by speculative decoding when it finishes. No-op by default."""
+        pass
+
     def size(self):
         """
         Return the size (i.e. sequence length) of the cache.
@@ -596,12 +609,57 @@ class ArraysCache(_BaseCache):
         instance = super().__new__(cls)
         instance.left_padding = None
         instance.lengths = None
+        instance.speculating = False
+        instance._rollback = None
         return instance
 
     def __init__(self, size, left_padding: Optional[List[int]] = None):
         self.cache = [None] * size
         if left_padding:
             self.left_padding = mx.array(left_padding)
+
+    def start_speculation(self):
+        self.speculating = True
+        self._rollback = None
+
+    def stop_speculation(self):
+        self.speculating = False
+        self._rollback = None
+
+    def record_rollback(self, num_tokens, fn, snapshot):
+        """Recorded by the owning layer during a forward while ``speculating``.
+
+        Args:
+            num_tokens (int): Sequence length of the forward being recorded.
+            fn (callable): ``fn(m) -> list`` returning the cache entries as they
+                would be had only the first ``m`` of ``num_tokens`` tokens been
+                processed. Must be exact (e.g. replay the recurrence from the
+                pre-forward state over the stashed per-token inputs).
+            snapshot (list): The cache entries from before the forward (returned
+                for a full rollback, ``m == 0``).
+        """
+        self._rollback = (num_tokens, fn, snapshot)
+
+    def is_trimmable(self):
+        # While speculating, every forward records an exact rollback, so the
+        # cache is trimmable within the last forward's window.
+        return self.speculating
+
+    def trim(self, n):
+        if n <= 0:
+            return 0
+        if self._rollback is None:
+            raise RuntimeError(
+                "ArraysCache.trim requires a rollback recorded by the model's "
+                "recurrent layers; this model does not support speculative "
+                "rollback."
+            )
+        num_tokens, fn, snapshot = self._rollback
+        n = min(n, num_tokens)
+        m = num_tokens - n
+        self.cache = list(snapshot) if m == 0 else fn(m)
+        self._rollback = None
+        return n
 
     @property
     def batch_size(self):
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -422,6 +422,17 @@ class KVCache(_BaseCache):
 
 class RotatingKVCache(_BaseCache):
     step = 256
+    # Tokens of exact-rollback history kept while speculating (mirrors
+    # ArraysCache; a speculative trim never exceeds the last verify window).
+    _ROLLBACK_WINDOW = 64
+
+    def __new__(cls, *args, **kwargs):
+        # Set on __new__ (not __init__) so caches reconstructed by
+        # load_prompt_cache (which bypasses __init__) still have them.
+        instance = super().__new__(cls)
+        instance.speculating = False
+        instance._rollbacks = deque()
+        return instance
 
     def __init__(self, max_size, keep=0):
         self.keep = keep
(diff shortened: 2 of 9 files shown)


lBroth:
Nice work — and a funny coincidence: we independently built the same capture-and-replay rollback this week, which I'd read as two votes for the approach being right. Your ULP-level near-tie observation matches our measurements exactly (top-2 gaps of 0.000–0.125 at divergence points, same class the unmodified batched-vs-sequential pipeline already shows).

Since your PR explicitly defers generation-loop integration ("can follow separately if there's interest") — that part exists, tested: https://github.com/lBroth/mlx-lm/tree/gdn-exact-replay

What it adds, complementary to this PR:
- an `ArraysCache`-level protocol (`start/stop_speculation`, `record_rollback`, `trim` applies the recorded replay) instead of model-file-local plumbing — the layer records a closure over the exact kernel inputs, cache.py stays model-agnostic;
- wired into the existing `speculative_generate_step`, so `--draft-model` speculation works today on hybrids (including **hybrid draft models** — rollback records stack across the draft's T=1 steps; the Qwen3.6-target + Qwen3.5-draft pair from #1446 runs end-to-end);
- covers `qwen3_next` as well as `qwen3_5*`;
- 8 tests, the key one being bit-exact tail-independence: two verify chunks sharing an m-token prefix with different tails must leave IDENTICAL caches and next-token logits after rollback-to-m — any off-by-one in the replay/slice leaks the tail. Validated on M2/M3/M5 Metal + CPU ops backend with real weights (greedy spec ≡ vanilla).

Your single-kernel batched replay (stacking layers along the head axis) is nicer than our per-layer replay loop — we'd adopt that with credit.

If you're open to it, I think your MTP head + this integration layer make one complete story (#872 + #1446 + unblocking #1297's strategies on hybrids): either fold the branch under this PR, or keep them stacked — whatever you prefer. Happy to coordinate.

pierre427:
Sure, go ahead and roll it all up into one.

pierre427:
@lBroth — thanks, that's a clean design and the independent convergence is reassuring. I took you up on "fold it into one" and pushed a consolidation to this PR's head (force-update):

**What's now on the branch**
- Your `gdn-exact-replay` (cache-level `start/stop_speculation` + `record_rollback`, wired into `speculative_generate_step`, `qwen3_5` + `qwen3_next`, your 8 tests incl. the bit-exact tail-independence one).
- Our MTP head loading + norm-shift fix (this PR's original half).
- Our model-file-local `gdn_sink`/`rollback_speculative_cache` **dropped** in favor of your cache-level protocol — it's cleaner (model files stay agnostic, the loop's existing `trim_prompt_cache` just works).
- Our single-kernel batched replay is **not** in yet — kept as a credited follow-up perf pass on your per-layer trim, since correctness-first is the right order.

All merged green: 8 exact-replay + 3 MTP loading + norm-shift contract, plus the item below.

**New: the same protocol for the *other* non-trimmable cache — `RotatingKVCache`.**
`ArraysCache` (recurrent GDN state) was one non-trimmable cache; sliding-window `RotatingKVCache` is the other — once it wraps (`offset >= max_size`) the tokens a verify step pushes out of the window are gone, so stock trim silently no-ops and corrupts them. I added the same protocol to it: `update_and_fetch` stashes the pre-forward window + new K/V while speculating; `trim` restores + re-appends the accepted prefix; `is_trimmable()` is True while speculating; no model-file hook (KV isn't recurrent). `gpt_oss` now declares `supports_speculative_rollback = True`, so `--draft-model` speculation covers gpt-oss-family sliding-window models too, not just GDN hybrids. 4 tests, same bit-exact tail-independence + stacked-draft. (`tests/test_rotating_rollback.py`.)

Honest caveat we measured: external-draft speculation *loses* on a fast MoE target (gpt-oss-puzzle-88B: 0.42x, ~45% accept with a 20B draft) — the rollback is correct and general, it just doesn't pay at that operating point.

**Other adjacent things we have, mention-only (not proposing to include here):**
- `gpt_oss_puzzle` model support (NVIDIA Puzzle NAS variant: per-layer heterogeneous experts/windows) — a separate new-model PR whenever there's interest.
- A reasoning-runaway guard (repetition + think-token-budget detector that force-closes a non-terminating think/analysis channel and harvests the answer; turned a 177s/14k-cap runaway into 26s on the Puzzle) — arguably a stopping-criteria/generation-wrapper concern rather than core, happy to shape into a reusable hook if wanted.
- Digit-strict relaxed thinking-channel acceptance for spec-decode.

Totally happy for you to drive the final shape — take any/all of this, or I'm glad to rebase onto your branch if you'd rather own it. Whatever lands it cleanest.

lBroth:
Consolidation validated on our side (M5 Pro, macOS, mlx 0.31.2):

- `tests.test_speculative_rollback` + `tests.test_rotating_rollback` + `tests.test_qwen3_5_mtp`: **15/15 OK** on this PR's head.
- Full `unittest discover` run: failure set is a strict subset of `main`'s on the same machine/env (only the two known network-flaky tests differ) — no regressions introduced by the consolidation.
- Attribution handled cleanly on your side — appreciated.

Two small notes for anyone testing with a fresh environment:
- pin `transformers<5.13` — 5.13.0 independently breaks `mlx_lm` import at this base (`AutoTokenizer.register` with a string config key raises `AttributeError`); unrelated to this PR.
- we'd previously validated the GDN-rollback core on M2 / M3 / M5 Metal + the CPU (ops) backend with real weights (greedy speculative ≡ vanilla, hybrid-draft pair included); we'll re-run that matrix against this consolidated head and report here.

The `RotatingKVCache` extension is exactly the kind of generalization the cache-level protocol was hoping to invite — nice. +1 on deferring the batched single-kernel replay as a follow-up perf pass.

lBroth:
As promised — cross-hardware validation of THIS consolidated head (`a33b773`), fresh environments each (python 3.13, mlx 0.31.2, `transformers<5.13` pinned):

| machine | unit tests (GDN + Rotating + MTP) | cache regression | e2e real weights¹ | e2e speed² |
|---|---|---|---|---|
| Apple M5 Pro 48GB | 15/15 | 22/22 (full-suite failure set ⊆ `main`'s) | identical ✓ (also 80B Coder-Next + 0.6B draft: identical, 1.39x) | — |
| Apple M2 Pro 16GB | 15/15 | 22/22 | identical ✓ | 1.10x / 0.78x |
| Apple M3 16GB | 15/15 | 22/22 | identical ✓ | 1.92x / 1.33x |
| CPU (ops backend, forced) | 15/15-equivalent³ | 22/22 | — | — |

¹ Qwen3.5-9B-MLX-4bit target + Qwen3.5-0.8B-4bit draft — both hybrid GDN (the exact pair class from #1446), greedy, `stream_generate --draft-model` path. "identical" = token-for-token vs non-speculative greedy.
² edit-style prompt / from-scratch prompt. Note the machine-dependence: on the slower M3 the draft pays on both workloads; on M2 Pro the from-scratch case is net-negative — consistent with the known external-draft economics (see also pierre427's 0.42x note above). Correctness is invariant; speed depends on the target/draft ratio and workload.
³ earlier run of the GDN-rollback tests on a forced-CPU (ops) backend, pre-consolidation commits which are unchanged in this head.

No regressions found anywhere. From our side this PR is validated and ready for review.

pierre427:
Following up on landing the speculative-rollback work — I think this PR is carrying too much for one review: it bundles the model-agnostic **rollback cache-API** (@lBroth's `start_speculation`/`record_rollback` protocol + my `RotatingKVCache` sibling) *and* the qwen3.5 in-checkpoint **MTP-head loading**.

**Proposal: split the rollback cache-API into its own focused PR and land it first**, with the MTP head as a follow-up on top. Rationale:

- The rollback protocol is a shared dependency — `--draft-model` speculative decoding on hybrid (GatedDeltaNet/Mamba) caches, prompt-lookup decoding on hybrids, and MTP self-speculation all build on it. Landing it standalone unblocks all of them.
- A small, focused PR is much easier to review and merge than the bundle.

Since the protocol is substantially @lBroth's work, I'd suggest **you drive the rollback PR**; I'll contribute the `RotatingKVCache` piece + review. To make it easy I've prepared a clean branch off current `main` with just the four rollback commits (`e871d09` / `b06bd13` / `6f2887b` + my `RotatingKVCache`), authorship preserved, **no MTP** — `cache.py` protocol + `generate.py` speculative wiring + representative GDN model support + our synthetic tests (12 tests, all green). Happy to push it somewhere you can take it over, or you can cherry-pick those four.

I'll then re-target this PR to just the MTP head on top of the landed API. Does that work for you?

pierre427:
Quick follow-up to keep the split moving. On my side the MTP-head changes (`qwen3_5.py` MTP module + `mtp_step`, the norm-double-shift bugfix, and the tests in `test_qwen3_5_mtp.py`/`test_models.py`) are cleanly separable from the rollback cache-API (`cache.py` + `generate.py` + `gpt_oss`/`qwen3_next` wiring + `RotatingKVCache` + the rollback test files) — the only file the two share is `qwen3_5.py`, and those hunks factor apart cleanly. So once your rollback PR is up I can re-target this one to sit on top with minimal churn.

Since you've already validated the consolidated head across M2/M3/M5, easiest is probably to cut the rollback PR straight from there. If a pre-separated rollback-only branch off `main` would save you the untangle, I'm glad to prepare and push one for you to take over — just say which you'd prefer. No rush; the MTP half is entirely downstream.

lBroth:
Opened the split as #1486 — the rollback cache-API alone (my three GDN/ArraysCache commits + your `RotatingKVCache` sibling), no MTP, 12 tests green. I led with the correctness / generality / no-regression points so a reviewer sees them up front.

One note on your piece: I brought it in via cherry-pick, so the **author** stays you (`Pierre Lamy`) but the committer/SHA changed as a side effect. If you'd rather your `RotatingKVCache` commit keep its exact identity, I'm happy to rebase #1486 onto the clean branch you prepared — just push it or point me at it and I'll re-base so your commit lands verbatim. Whatever keeps it cleanest on your end.

Then you re-target #1456 to just the MTP head on top of the landed API, as you suggested. Thanks for splitting it out — makes both much easier to review.

pierre427:
Heads up for reviewers on the current state of the split: this PR is now scoped to the **MTP-head loading** only (the `qwen3_5.py` MTP module + `mtp_step` and the norm-double-shift bugfix). The model-agnostic rollback cache-API has moved to #1486 (@lBroth). The rollback commits still show up on this branch for now because it predates the split — once #1486 lands on `main` I'll rebase this down so they drop out and only the MTP-head diff remains. Best reviewed after #1486.

pierre427:
Pushed a small follow-up (`d742b7e`): the wrapper `Model` (used by `qwen3_5_moe`, and the multimodal `qwen3_5` variant) didn't expose the MTP interface, so `self_mtp_generate_step` couldn't reach the head on wrapped checkpoints — `model.mtp` was `None` even though `language_model.mtp` had loaded. Delegating `mtp`/`logits`/`mtp_step`/`make_mtp_cache` to `language_model` fixes it. Verified on `Qwen3.5-122B-A10B-oQ4-mtp`: self-spec now runs against the top-level model (coding decode 44 -> 56 tok/s, ~1.26x at num_draft=1).

pierre427:
Requesting review of the rollback semantics and validation in this PR.

Speculative decoding normally discards rejected work by truncating the KV cache. Qwen3.5 also carries recurrent GDN state, which cannot be restored by slicing it at a token boundary. This implementation reconstructs the retained state through replay, while loading the model’s MTP head and correcting an extra normalization that otherwise changes the proposal path.

The current measurements show 83–100% acceptance and a 1.36–2.11× decoding speedup. Comparisons against ordinary target decoding remain lossless apart from equivalent low-precision ties.

The points that seem most important to check are:

- whether replay reconstructs the same GDN state across partial acceptance, complete rejection, and repeated speculation;
- whether the equivalence checks cover the relevant dtypes, quantized configurations, cache boundaries, and longer generations;
- whether the normalization change matches the intended checkpoint computation across Qwen3.5 variants;
- whether any model-specific assumptions have entered what should become a general recurrent-state rollback contract;
- whether the reported speedup still holds when replay frequency and lower acceptance regimes are considered separately.

The broader question is how speculative decoding should represent rollback for architectures whose decoding state is recurrent rather than token-indexed.

lBroth:
Went through the split again after your review request. Found one real
thing, in my half not yours: the finally block in
speculative_generate_step only stopped speculation on the model caches,
while start_speculation runs on model + draft. A hybrid or
sliding-window draft cache stayed speculating with stale rollback
records, so a reused draft prompt cache would keep reporting
is_trimmable() and trim through records from a finished generation.
Fixed on #1486 (8b98594) with the hybrid-draft test extended to catch
it (it needs a recording draft; the llama draft in the reuse test never
sets the flag, which is why it slipped). You'll want the same line when
you refresh this branch.

On your half: the conv1d-layout signal for the raw-checkpoint norm
shift reads right to me, and it just got independent confirmation from
production: youssofal/MTPLX#176 root-caused their 4B zero-acceptance
collapse yesterday to exactly this class (MTP RMSNorm tensors in the
raw zero-centered convention, never restored, every draft garbage).
Your double-shift guard covers the mirror image. The rope -1 shift
argument in mtp_step also checks out since the head only attends within
its own cache, relative distances are all that survive.

pierre427:
Applied in d548cab — thanks for flagging it. Same line as your 8b98594: the finally block now stops speculation on `model_cache + draft_cache`, with your rationale in the comment. test_generate shows the same 8 pre-existing batch/embedding failures before and after, nothing new.

PR #1459, opened 2026-07-03
Fix AutoTokenizer.register() for transformers 5.13.0+ compatibility

### LINKED ISSUE #1458: `transformers==5.13.0` breaks `mlx-lm` / `mlx-vlm` imports due to string auto-registration
After updating Transformers to `5.13.0`, importing `mlx_lm` and `mlx_vlm` fails.

The immediate failure is in `mlx_lm/tokenizer_utils.py`:

```python
```

Transformers 5.13.0 now expects the first argument to be a config class, and dereferences `config_class.__module__`. Passing a string now raises:

```text
```

Stack excerpt:

```text
File ".../mlx_lm/tokenizer_utils.py", line 503, in <module>

File ".../transformers/models/auto/tokenization_auto.py", line 984, in register
  TOKENIZER_MAPPING.register(config_class, tokenizer_class, exist_ok=exist_ok)

File ".../transformers/models/auto/auto_factory.py", line 680, in register
     ^^^^^^^^^^^^^^
```

This also prevents `mlx_vlm` from importing because it imports through `mlx_lm`.

I also noticed similar string registration patterns in `mlx-vlm`, for example:

```python
AutoProcessor.register("deepseekocr", DeepseekOCRProcessor)
```

So this may affect both `mlx-lm` and `mlx-vlm` under the newer Transformers auto-registration API.

**Environment**

```text
Python: 3.13
transformers: 5.13.0
mlx-lm: 0.31.3
mlx-vlm: 0.6.3
```

**Workaround**

Pinning Transformers below 5.13 restores imports:

```bash
pip install "transformers>=5.7.0,<5.13.0"
```

Verified working with:

```text
transformers: 5.12.1
import mlx_lm: ok
import mlx_vlm: ok
python -m mlx_vlm generate --help: ok
```

**Suggested fix**

Update the `AutoTokenizer.register` / `AutoProcessor.register` calls that pass strings so they match the Transformers 5.13+ API contract, or add an upstream dependency cap until that compatibility work lands.

jonpspri:
## Description

Fixes #1458 

This PR addresses the breaking change in transformers 5.13.0 where `AutoTokenizer.register()` now requires a config class instead of a string as the first parameter.

## Changes

- **Added `NewlineTokenizerConfig` class**: A proper `PreTrainedConfig` subclass with `model_type = "newline_tokenizer"`
- **Updated `AutoTokenizer.register()` call**: Changed from string `"NewlineTokenizer"` to `NewlineTokenizerConfig` class
- **Added test case**: `test_newline_tokenizer_registration()` to verify the fix and prevent regression

## Compatibility

✅ **Backward compatible**: Works with transformers 5.7.0+ (including 5.12.x and 5.13.0+)

The fix follows the proper API that was always documented - the string parameter was never officially supported but happened to work due to lax validation in older versions.

## Testing

- Verified import succeeds without `AttributeError: 'str' object has no attribute '__module__'`
- Confirmed `NewlineTokenizerConfig.__module__` = `"mlx_lm.tokenizer_utils"` (doesn't start with "transformers.")
- Added unit test to verify config class attributes and registration

## Related Issues

- Fixes #1458 
- Similar issue affects mlx-vlm with `AutoProcessor.register()`

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -5,7 +5,7 @@ from functools import partial
 from json import JSONDecodeError
 from typing import Any, Dict, List, Optional
 
-from transformers import AutoTokenizer, PreTrainedTokenizerFast
+from transformers import AutoTokenizer, PreTrainedConfig, PreTrainedTokenizerFast
 
 
 class StreamingDetokenizer:
@@ -474,6 +474,15 @@ class TokenizerWrapper:
             setattr(self._tokenizer, attr, value)
 
 
+class NewlineTokenizerConfig(PreTrainedConfig):
+    """Configuration for NewlineTokenizer."""
+
+    model_type = "newline_tokenizer"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+
 class NewlineTokenizer(PreTrainedTokenizerFast):
     """A tokenizer that replaces newlines with <n> and <n> with new line."""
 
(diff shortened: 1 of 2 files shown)


menisadi:
See #1465

angeloskath:
Fixed by #1465.

PR #1460, opened 2026-07-03
Fix import crash with transformers >= 5.13

Lazarus-931:
On the newest transformers (5.13.0), just running import mlx_lm crashes:

```
```

Since it fails on import, nothing in mlx-lm works — and any project that depends on mlx-lm breaks too.

mlx-lm registers its custom NewlineTokenizer like this:
```
```

That first argument is supposed to be a class, but we pass a plain text name ("NewlineTokenizer"). Older transformers didn't care. transformers 5.13 now assumes it's a class and tries to read .__module__ off it, leading to the error.

Fix: Wrap the call so that extra step failing can't take down the whole import:

```
try:
except (AttributeError, TypeError):
    pass
```

All tests pass

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -500,7 +500,13 @@ class NewlineTokenizer(PreTrainedTokenizerFast):
         return [self._postprocess_text(d) for d in decoded]
 
 
-AutoTokenizer.register("NewlineTokenizer", fast_tokenizer_class=NewlineTokenizer)
+try:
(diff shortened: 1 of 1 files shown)


Lazarus-931:
note #1459 already opened for same issue, will keep this open just incase

menisadi:
@Lazarus-931 also #1465

angeloskath:
Thanks! Fixed by #1465, which passes the tokenizer class to `AutoTokenizer.register` so the auto-mapping key has a valid `__module__`. Preferred over swallowing the exception. Closing as resolved by #1465.

angeloskath:
Fixed by #1465.

Issue #1453, opened 2026-07-02
glm_moe_dsa: output degenerates to gibberish above ~2-3K prompt tokens (DSA indexer suspected)


kkakkung2000:
## Summary
Models using the `glm_moe_dsa` architecture (GLM-5.2, and presumably GLM-5.1/GLM-5) produce coherent output for short prompts but degenerate into repetitive token/digit/punctuation soup once the prompt crosses roughly 3-3.5K tokens, and never reach a natural stop (always hits `max_tokens`). This reproduces across two independent quantizations and both single-node and multi-node (tensor-sharded) serving, which points at the DSA sparse-attention indexer path added in #879 / #1419 rather than a quantization or sharding artifact.

## Environment
- mlx-lm (via EXO, a distributed serving wrapper around mlx-lm — **not yet isolated against vanilla `mlx_lm.server`/`mlx_lm.generate`**, see Caveats)
- Hardware: Apple Silicon (M3 Ultra 512GB), Metal backend
- Models tested:
  - `mlx-community/GLM-5.2-mxfp4` (uniform 4-bit, OCP microscaling, group_size=32, no per-tensor precision overrides, 395GB) — single-node
  - GLM-5.2 DQ4plus-q8 quant (affine/integer scale-zero-point, mixed precision — 4-bit floor with attention projections/indexer weights/MLP/embed table overridden to 8-bit, 432GB) — 2-node tensor-sharded
- Architecture per `config.json`: `model_type=glm_moe_dsa` (`GlmMoeDsaForCausalLM`), 753B total params, `num_hidden_layers=78`, `hidden_size=6144`, `num_key_value_heads=64`, `index_topk=2048`

## Reproduction
Single-turn, tool-less chat completion, `temperature=1.0`, `top_p=0.95` (model's own sampling defaults), `max_tokens=200`. Prompt = repeated neutral filler sentence padded to a target token count, followed by "Summarize the above in exactly one sentence."

## Results (mxfp4, single-node)
| Prompt tokens | Output | finish_reason |
|---|---|---|
| 62 | Correct, coherent | `stop` |
| 2,227 | Correct, coherent | `stop` |
| 3,422 | Gibberish — digit/punctuation soup, no recognizable words | `length` |
| 7,381 | Gibberish, more severe | `length` |

[...]

PR #1431, opened 2026-06-24
Fix deepseek indexer

pcuenca:
_This description was written by a human_ 🙋‍♂️ 

Rope `traditional` was changed for the Indexer [here](https://github.com/ml-explore/mlx-lm/pull/867/changes/f921946dc630698a3f0cafd171f036c9661b9bde) as part of #867. This fixed a very obvious long-sequence generation error with GLM 5: the model started producing gibberish after a while.

I think that the indexer of Deepseek 3.2 requires `False`, but using `True` does not result in immediately observed garbage output. Outputs remain coherent, but quality silently degrades for long sequences.

I tested by generating 6 JavaScript implementations for different games using the prompts shown in the script below. We want generation to go beyond the 2048 sequence length where the indexer kicks in. In all 6 cases, the implementations from the current mlx-lm package are coherent, but the games are not playable. Using the patched version, all games are playable (with confusing visual imperfections in two of the games, but playable).

I found this issue while working on [GLM 5.2](https://github.com/ml-explore/mlx-lm/pull/1410). I think that silent degradation + superclass being used as the reference for other implementations is worth fixing as robustly as we can. Happy to run additional tests to confirm behavior!

<details><summary>Generation script</summary>

Covers 5 prompts, there's a sixth one I ran outside the script.

```bash
#!/bin/bash

declare -a prompts=(
  "Write a complete, polished, single-file HTML+JavaScript implementation of Tetris, with comments. Include the 7 tetromino shapes, rotation, line clearing, scoring, increasing speed by level, and a next-piece preview."
  "Write a complete, polished, single-file HTML+JavaScript 2048 game, with comments. Arrow-key controls, tile sliding and merging, score, win at 2048, and game-over detection."
  "Write a complete, polished, single-file HTML+JavaScript implementation of Snake, with comments. Grid movement, growth on eating food, self- and wall-collision death, score, and a restart."
  "Write a complete, polished, single-file HTML+JavaScript implementation of Breakout/Arkanoid, with comments. A paddle, a grid of destructible bricks, ball physics with wall/paddle/brick bounces, lives, score, and level-clear."
  "Write a complete, polished, single-file HTML+JavaScript implementation of Minesweeper, with comments. Random mine placement, left-click reveal with flood-fill on zero-adjacency cells, right-click flagging, number hints, and win/lose detection."
)

typeset -i id=1
for p in "${prompts[@]}"
do
  for version in "mlx-lm" "mlx-lm.patched"
  do
    uv pip install -e "./$version"
[...]

PR #1419, opened 2026-06-21
GLM-5.2 (glm_moe_dsa) inference: IndexShare, indexer RoPE/eps, int8 MLA-KV

avlp12:
Resolves #879 — adds inference support for **GLM-5.2** (`model_type: glm_moe_dsa`), Zhipu/Z.ai's 744B (~40B active) MoE. It is architecturally DeepSeek-V3.2 (MLA + DSA "lightning indexer" sparse attention), so this extends the existing `deepseek_v32.py` rather than adding a new model.

## What was missing

The stock `glm_moe_dsa`/`deepseek_v32` port runs the DSA indexer on **every** layer and threads no indices between layers. GLM-5.2 computes top-k on a "full" layer and **reuses** it on the following "shared" layers (IndexShare). At ≤2048 ctx this is invisible (top-k selects all tokens → dense attention), but beyond that the port diverges from the reference and forfeits the sparse-attention savings.

## Changes (inference only)

- **IndexShare** — `derive_indexer_types()` marks layers full/shared from `index_topk_freq`; "full" layers run the indexer, "shared" layers reuse the previous full layer's top-k (threaded through the decoder). `index_topk_freq=1` (the default) makes every layer "full", i.e. **stock DeepSeek-V3.2 behavior is unchanged**.
- **Indexer RoPE / eps** — GLM-5.2's indexer uses non-interleaved (half-split) RoPE + LayerNorm eps `1e-6`, distinct from its interleaved main attention (verified against the HF reference to ~1e-7 on post-RoPE `q`). `deepseek_v32` defaults are unchanged (`traditional=True`, eps `1e-5`); `glm_moe_dsa` overrides them.
- **int8 MLA latent KV cache** — `MLACacheList` (a `CacheList` subclass used only by these models) quantizes only the large compressed-latent cache and keeps the tiny DSA indexer cache full precision; the latent is dequantized on read. This lets a 1M-token context fit on a single machine. The generic `CacheList` is left untouched, so other hybrid-cache models are unaffected.
- `KVCache`/`ChunkedKVCache.state` tolerate an unwritten sub-cache (the indexer KV on shared layers) so `mx.eval(cache.state)` during generation doesn't dereference `keys=None`.
- `glm_moe_dsa` `ModelArgs` defaults the fields GLM-5.2's `config.json` omits.

## Tests

- `test_all_models` gains a `glm_moe_dsa` config with `index_topk_freq=4` (exercises full + shared layers, fp32/fp16 prefill, decode, batch, deepcopy).
[...]

Issue #1418, opened 2026-06-20
GLM-5.2 (glm_moe_dsa with IndexShare) fails to load — missing per-layer indexer params


anjansastry-hash:
### Summary

`GLM-5.2` (architecture `glm_moe_dsa` with **IndexShare**) fails to load in mlx-lm. The current loader expects a DeepSeek-V3.2-style indexer on **every** layer, but GLM-5.2 only places indexers on a subset of layers, so loading aborts with missing parameters.

### Reproduction

Loading `mlx-community/GLM-5.2-mxfp4`:

```
Missing 285 parameters:
  model.layers.8.self_attn.indexer.*
  model.layers.9.self_attn.indexer.*
  model.layers.11.self_attn.indexer.*
  ... (and similar for other layers)
```

### Cause

GLM-5.2 uses **IndexShare** (see arxiv:2603.12201). Its `config.json` carries an `indexer_types` field listing, per layer, whether the attention indexer is `"full"` or `"shared"` — in the released checkpoint this is **21 `full` layers + 57 `shared` layers**. Indexers are therefore materialized only on the ~1-in-4 `full` layers; `shared` layers reuse a neighbor's indexer.

The current `mlx_lm/models/glm_moe_dsa.py` is a thin shim that delegates to the `deepseek_v32` attention path, which instantiates a per-layer indexer unconditionally. As a result it tries to bind 285 indexer params that don't exist in the GLM-5.2 weights.

### Note: GLM-5.1 is unaffected

GLM-5.1 is the same `glm_moe_dsa` family but has **no** `indexer_types` field (no IndexShare), so it loads natively today. The gap is specific to the IndexShare variant introduced in GLM-5.2.

### Question

Would a PR adding IndexShare support (honoring `indexer_types` so `shared` layers reuse the upstream `full` layer's indexer) be welcome? Or is GLM-5.2 / IndexShare support already planned or in progress? Happy to work on it if it's wanted.

### Environment

- mlx-lm 0.31.3
- Apple Silicon (M3 Ultra, 512 GB unified memory)
- Model: `mlx-community/GLM-5.2-mxfp4`

bibproj2:
PR #1410 is working

anjansastry-hash:
[...]

PR #1463, opened 2026-07-04
GLM-5.2: full/shared indexer typing for glm_moe_dsa (DSA schedule + interleaved indexer rope)

machiabeli:
GLM-5.2 ships a per-layer DSA indexer schedule (`indexer_types`: 21 `full` layers own a lightning indexer; 57 `shared` layers reuse the previous full layer's top-k and ship **no** indexer weights). The current `glm_moe_dsa.py` subclasses DeepSeek-V3.2 directly, so every layer expects indexer weights — strict loading of the real checkpoint fails (#1418), and past `index_topk` tokens the schedule semantics are lost (#1453).

## What this adds

- **`GlmMoeDsaAttention` / decoder / model honoring `indexer_types`**: full layers construct the `Indexer`; shared layers `del` it (param tree matches the checkpoint — 0 missing / 0 unexpected on strict load of the real 743B mxfp4 checkpoint, 21 full + 57 shared) and consume `prev_topk_indices` threaded layer-to-layer (mirrors HF's `next_skip_topk`).
- **Asymmetric `make_cache`**: `CacheList(kv, indexer)` on full layers, `CacheList(kv)` on shared.
- **`indexer_rope_interleave: bool = True` for this arch** — carries forward the merged #1431 resolution (GLM-5.2 keeps interleaved; per #1431 discussion and huggingface/transformers#46842, the HF GLM-5.2 reference is the suspected outlier here). Independent validation below agrees with interleaved.
- **Tests**: `test_all_models` entry with a `["full","shared","full","shared"]` schedule (B=1/B=2), plus `test_glm_moe_dsa_shared_schedule` covering schedule construction, the cache asymmetry, the rope flag, and the **engaged** sparse path (prefill past `index_topk`, then a decode step threading `prev_topk`).

## Validation beyond the unit tests

- Strict `load_model` of the real GLM-5.2 743B mxfp4 checkpoint: 78 layers, zero random weights, schedule correct.
- An independent C/Metal engine (glmx) was gated against this implementation as its oracle: engaged prefill + decode logits match at ~1e-6 for **both** all-`full` and `full/shared` schedules at tiny config, and per-layer parity probes on the real checkpoint at 3015-token context reproduce the top-2048 selections exactly (interleaved rope), with coherent long-context generation.

## Relation to #1410 / #1412 / #1419

#1410 (@pcuenca) implements the same IndexShare core and has broad tester traction; #1410's remaining open item is numerical-parity evidence, which is exactly what the engaged-path tests and the independent-engine validation here provide. Happy to consolidate in whichever direction the maintainers prefer — these tests and the validation story are offered either way.

https://claude.ai/code/session_011YpPndSoRJZRbZeCNs8KYL

### DIFF
--- a/mlx_lm/models/glm_moe_dsa.py
+++ b/mlx_lm/models/glm_moe_dsa.py
@@ -57,6 +57,12 @@ class ModelArgs(BaseModelArgs):
     index_share_for_mtp_iteration: bool = False
     index_topk_freq: int = 4
     index_skip_topk_offset: int = 3
+    # GLM-5.2's indexer rope is interleaved (traditional), unlike the
+    # DeepSeek-V3.2 default introduced in #1431. Validated against the real
+    # 743B mxfp4 checkpoint: engaged-DSA generation at 3k context is coherent
+    # and an independent C reference reproduces the selection exactly with
+    # interleaved rope (glmx engine parity gates, 2026-07-03).
+    indexer_rope_interleave: bool = True
 
     def __post_init__(self):
         self.rope_scaling = self.rope_parameters
--- a/mlx_lm/models/glm_moe_dsa.py
+++ b/mlx_lm/models/glm_moe_dsa.py
@@ -1,9 +1,15 @@
 # Copyright © 2025 Apple Inc.
 
 from dataclasses import dataclass
-from typing import Any, Dict, Optional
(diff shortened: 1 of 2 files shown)


zcbenz:

Issue #1450, opened 2026-07-02
Decode performance optimizations for MLX-LM inference


dahai80:
## Decode Performance Optimizations for MLX-LM Inference

We have implemented and benchmarked a set of decode-time optimizations for MLX-LM on Apple Silicon. On an M2 Ultra 137GB, these combine to deliver up to **162% faster decode throughput** over the baseline mxfp8 quantization. We are proposing these for upstream integration into mlx-lm.

---

### 1. Mixed-bit Quantization Recipes

**Problem:** Uniform quantization wastes bits on layers that tolerate aggressive quantization and over-quantizes sensitive layers.

**Proposal:** Pre-tuned per-layer bit allocation plans (`mixed_3_4`, `mixed_2_6`, `quant2_all`, etc.) that apply different bit widths to different layer groups based on their sensitivity. Exposed via a `--quant-recipe` flag in `mlx_lm.convert`.

Example usage:
```bash
python -m mlx_lm.convert --model Qwen/Qwen2.5-27B --quant-recipe mixed_3_4
```

Recipes assign bit widths to layer groups (embeddings, attention projections, MLP projections, output head) based on empirically measured sensitivity. This avoids the need for per-model calibration while still capturing most of the mixed-precision benefit.

---

### 2. Skip-logsumexp in Decode

**Problem:** During autoregressive decode, only one new token position is computed per step. The full logsumexp normalization over the vocabulary is unnecessary for correctness when sampling from a single-token logit vector.

**Proposal:** Skip the logsumexp normalization in the decode path. Since we only sample one token, the relative ordering of logits is preserved without normalization, and the sampling step (argmax or random) produces identical results.

---

### 3. Fused Sampler

**Problem:** Sampling (argmax for greedy, or random sampling with temperature/top-p/top-k) is done in Python with multiple dispatches to Metal — one to apply temperature, one for top-p masking, one for the random draw, etc.

**Proposal:** Combine the entire sampling operation into a single Metal kernel. This eliminates multiple Python-to-Metal round trips per token and reduces kernel launch overhead.

---

### 4. FP8 KV Cache

**Problem:** The KV cache is stored in float16, consuming 2 bytes per element. For long sequences and large models, this dominates memory bandwidth during attention.

[...]

PR #1466, opened 2026-07-04
Add quant2 / quant2_128 mixed-bit quant recipes

dahai80:
## Add `quant2` / `quant2_128` mixed-bit recipes

Adds two opt-in mixed-bit quantization recipes to `--quant-predicate`:

- **`quant2`** — 2-bit decoder body, 3-bit first/last layer, 4-bit `embed_tokens`/`lm_head`, group_size=64.
- **`quant2_128`** — same layer allocation, group_size=128 for a better quality/speed balance.

### Rationale

Refs #1450. That issue explored aggressive 2-bit quantization and benchmarked a pure 2-bit body on a 27B model at **+162% decode speedup** vs mxfp8 with coherent output. The downside of pure 2-bit is degraded token resolution at the embedding/output and the boundary layers. This recipe keeps those at 3–4-bit while quantizing the bulk of the decoder body at 2-bit, preserving coherence where it matters most while retaining the decode speedup.

### Layer allocation

| Layer                         | bits | group_size        |
|-------------------------------|------|-------------------|
| `embed_tokens`, `lm_head`     | 4    | caller-supplied   |
| first / last decoder layer    | 3    | 64 (`quant2`) / 128 (`quant2_128`) |
| decoder body                  | 2    | 64 (`quant2`) / 128 (`quant2_128`) |

### Usage

```bash
mlx_lm.convert --model <hf-or-local> -q --quant-predicate quant2 --q-group-size 64 --mlx-path out-quant2
mlx_lm.convert --model <hf-or-local> -q --quant-predicate quant2_128 --q-group-size 128 --mlx-path out-quant2-128
```

### Tests

`tests/test_quant_recipes.py` — 5 tests covering bit allocation per layer, head/embed preservation, group_size selection, first/last layer index detection, and end-to-end inference. All pass.

### Benchmark (Qwen2.5-7B-Instruct, Apple M5 Max, 60-token decode)

| recipe      | decode tok/s | vs mxfp8 speed | size (GiB) | vs mxfp8 size | bpw   |
|-------------|--------------|----------------|------------|---------------|-------|
| mxfp8       | 29.1         | 1.00x          | 7.31       | 1.00x         | 8.250 |
| quant2      | 97.7         | 3.36x (+236%)  | 2.53       | 0.35x         | 2.848 |
| quant2_128  | 139.1        | 4.78x (+378%)  | 2.30       | 0.31x         | 2.598 |

The recipe converts cleanly, runs end-to-end, and delivers a large decode speedup with a ~3x memory reduction. `quant2_128` is both faster and smaller than `quant2` thanks to the wider group.

### Quality note (important)

2-bit quantization has a coherence threshold tied to model scale. At 7B the recipe runs correctly and the speedup holds, but the 2-bit body degrades output coherence (English-shaped but nonsensical text); at 0.5B output is fully degraded. Coherent 2-bit output requires the larger model regime — #1450 reports coherent output at 27B with +162% speedup. The recipe is intended for that ≥27B scale where 2-bit is viable; on smaller models users should prefer `mixed_3_4` / `mixed_4_6` or uniform 4-bit.

Refs #1450.

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -77,7 +77,43 @@ def mixed_quant_predicate_builder(
     return mixed_quant_predicate
 
 
-QUANT_RECIPES = ["mixed_2_6", "mixed_3_4", "mixed_3_6", "mixed_4_6"]
+def quant2_predicate_builder(
+    recipe: str, model: nn.Module, group_size: int = 64
+) -> Callable[[str, nn.Module, dict], Union[bool, dict]]:
+    mode = "affine"
+    down_keys = [k for k, _ in model.named_modules() if "down_proj" in k]
+    if len(down_keys) == 0:
+        raise ValueError("Model does not have expected keys for quant2.")
+    for layer_location, k in enumerate(down_keys[0].split(".")):
+        if k.isdigit():
+            break
+    num_layers = len(model.layers)
+    # quant2_128 uses a wider group size (128) for a better quality/speed
+    # balance; quant2 uses the caller-supplied group_size.
+    gs = 128 if recipe == "quant2_128" else group_size
+
+    def quant2_predicate(path: str, module: nn.Module) -> Union[bool, dict]:
+        index = (
+            int(path.split(".")[layer_location])
+            if len(path.split(".")) > layer_location
+            else 0
+        )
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1468, opened 2026-07-04
Attach Qwen3.6 MTP head for speculative decoding + fix cache-rewind for hybrid recurrent models

### LINKED ISSUE #1462: mlx_lm.server --draft-model fails to load any Qwen3.6 MTP checkpoint (model_type qwen3_5_mtp not registered)
### Summary

`mlx_lm.server --draft-model` fails immediately at load time for every Qwen3.6 MTP drafter checkpoint I've tried, regardless of which target model it's paired with. The drafter's `config.json` declares `model_type: "qwen3_5_mtp"`, which has no corresponding module in `mlx_lm.models` and no entry in `MODEL_REMAPPING`, so `_get_classes()` raises immediately, before any inference is attempted.

This is a different failure mode from #1292 (which is about truncated completions on an already-loaded MTP model); this one never gets the model loaded at all.

### Affected versions

- `mlx-lm 0.31.3` (pip, current latest at time of writing)
- `mlx 0.31.2`
- macOS 26.4.1, Apple Silicon (M5 Max, 128GB)
- Python 3.13

### Affected models

- Target: `unsloth/Qwen3.6-27B-UD-MLX-6bit` (loads fine standalone — `model_type: qwen3_5`)
- Draft: `mlx-community/Qwen3.6-27B-MTP-bf16` (`model_type: qwen3_5_mtp` — this is what fails)

### Minimal reproduction

```bash
pip install mlx-lm==0.31.3
mlx_lm.server \
    --model unsloth/Qwen3.6-27B-UD-MLX-6bit \
    --host 127.0.0.1 --port 8082 \
    --draft-model mlx-community/Qwen3.6-27B-MTP-bf16
```

### Observed output

```
  ...
  File ".../mlx_lm/server.py", line 695, in _generate
    self.model_provider.load_default()
  File ".../mlx_lm/server.py", line 385, in load_default
  File ".../mlx_lm/server.py", line 363, in _load
    draft_model, draft_tokenizer = load(draft_model_path)
  File ".../mlx_lm/utils.py", line 491, in load
    model, config = load_model(model_path, lazy, model_config=model_config)
  File ".../mlx_lm/utils.py", line 334, in load_model
    model_class, model_args_class = get_model_classes(config=config)
  File ".../mlx_lm/utils.py", line 191, in _get_classes
    raise ValueError(msg)
ValueError: Model type qwen3_5_mtp not supported.
```

### Root cause

`mlx_lm/utils.py`'s `_get_classes()` does:
```python
model_type = config["model_type"]
model_type = MODEL_REMAPPING.get(model_type, model_type)
arch = importlib.import_module(f"mlx_lm.models.{model_type}")
```
`MODEL_REMAPPING` has no `qwen3_5_mtp` entry, and there's no `mlx_lm/models/qwen3_5_mtp.py`. Checked the full registry directly (`pkgutil.iter_modules(mlx_lm.models.__path__)`, 118 entries): nothing containing `mtp` exists at all, for any model family. So this isn't specific to this one checkpoint, it affects every Qwen3.6 MTP drafter on the Hub today, since none of them have a registered loader.

### Expected output

The draft model should load and speculative decoding should proceed (or fail with something actionable at generation time, not at import time for a documented CLI flag).

### Workaround

None found. `mlx_vlm` (a separate package) does have `gemma4`-family multimodal classes registered, but nothing for `qwen3_5_mtp` either. The MTPLX project (github.com/youssofal/MTPLX) reportedly works around this by implementing its own inference path for MTP heads instead of going through `mlx_lm`'s standard model registry, rather than fixing the registry gap itself.

h9q2cyxvgm-ui:
Fixes #1462.

Two things bundled together because the second was needed to actually exercise/verify the first end to end:

## 1. Attach the MTP head instead of loading it as a peer draft model

The community-published `qwen3_5_mtp` drafter checkpoints (e.g. `mlx-community/Qwen3.6-27B-MTP-bf16`) are DeepSeek-V3-style MTP heads: no embedding table or output head of their own (they borrow the target model's), and their forward pass needs the *target's hidden state* as input, not just token ids. That's architecturally incompatible with mlx_lm's generic `--draft-model` mechanism, which only knows how to call an independent peer model with token ids — hence the `Model type qwen3_5_mtp not supported` error in #1462.

- `models/qwen3_5.py`: `MTPHead` module, lazily attached in `sanitize()` only when a checkpoint's weights actually bundle `mtp.*` keys — not eagerly from config alone, since the config can declare `mtp_num_hidden_layers > 0` as an architecture capability while the specific file doesn't carry the tensors (every currently-published Qwen3.6 checkpoint does this).
- `utils.py`: `load_mtp_head(model, mtp_path)` attaches a *separately published* MTP checkpoint onto an already-loaded target model; `is_mtp_head_checkpoint(config)` detects one.
- `generate.py`: `mtp_speculative_generate_step()`, parallel to `speculative_generate_step` but chaining the attached MTP head against the target's own hidden states. `stream_generate` dispatches here when `model.has_mtp` and no separate `draft_model` was given.
- `server.py`: `ModelProvider._load()` detects an MTP-head checkpoint and attaches it instead of peer-loading it.

## 2. Give `ArraysCache` checkpoint/restore so speculative decoding can actually run

Once loading worked, a second and *independent* bug showed up: **any** speculative decoding on Qwen3.6 — including the original, unmodified peer-model mechanism, no MTP code involved — fails with `"Speculative decoding requires a trimmable prompt cache (got {'ArraysCache'})"`. `ArraysCache` holds this model family's hybrid linear-attention layers' recurrent state as a fixed-size array, not an append-only buffer like `KVCache`, so `is_trimmable()` correctly returns `False` — there's nothing to slice back to an earlier position after a rejected draft token.

- `models/cache.py`: `is_checkpointable()` / `checkpoint()` / `restore()` added to `_BaseCache` (default: unsupported) and implemented on `ArraysCache` (a shallow list-copy of `self.cache` is a safe, cheap snapshot — verified across every hybrid-model file that uses `ArraysCache` that cache slots are always replaced wholesale via `__setitem__`, never mutated in place). New module functions `can_rewind_prompt_cache()`, `checkpoint_prompt_cache()`, `rewind_prompt_cache()`.
- `generate.py`: both `speculative_generate_step()` and `mtp_speculative_generate_step()` gate on `can_rewind_prompt_cache()` instead of `can_trim_prompt_cache()`. When the cache mixes trimmable KV layers with checkpoint-only recurrent layers, a rejected round can't just trim (recurrent state has no notion of "state as of token k" other than recomputing it) — instead it checkpoints every non-trimmable layer before each round, and on partial rejection undoes the *whole* round (trims KV back to the pre-round offset for free, restores checkpointed layers) and replays exactly the tokens that turned out correct through the full model in one extra forward call. This is unavoidable — a monolithic forward touches every layer together, so cheaply trimming KV while separately patching recurrent layers isn't possible; they have to move in lockstep. This is generic infra shared by every hybrid linear-attention model in mlx_lm, not just Qwen3.6.
- Also fixes a tensor-rank bug in `mtp_speculative_generate_step` (only reachable once the cache fix let execution get this far for the first time) and adds a guard so an exception during draft generation doesn't crash the `finally`-block cleanup.

## Verified

Temp=0 speculative output is token-for-token identical to plain generation, for both the peer-draft-model path (small `Qwen3-0.6B` draft) and the attached-MTP-head path, on `unsloth/Qwen3.6-27B-UD-MLX-6bit`. Full existing test suite has the same pre-existing failures before and after this change (confirmed via `git stash` on a clean checkout).

## Honest caveat on speedup

Modest and inconsistent right now — roughly 1.2x on a 40-token generation, ~0.9x (slightly *slower*) on 200 tokens, at ~66% draft acceptance either way (with the MTP head; see the follow-up PR for how that acceptance rate was fixed). The reason: `full_attention_interval=4` means 3 of every 4 layers are the non-trimmable recurrent kind, so any rejected round costs a full extra replay forward through most of the network. Correctness is solid; whether this specific rewind strategy is the right performance tradeoff for hybrid-recurrent models in general is a fair thing to discuss — happy to iterate (e.g. tuning `num_draft_tokens` down, or a cheaper reject-recovery path) based on feedback.

## Also open

A follow-up PR fixes a real accuracy bug in `MTPHead` itself (wrong concat order + wrong hidden-state norm stage) that took MTP draft acceptance from 0% to ~66% — separated out since it's a standalone correctness fix independent of the cache work above, and depends on the `MTPHead` class this PR introduces.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -517,11 +517,23 @@ def speculative_generate_step(
         model_cache = prompt_cache[: len(model.layers)]
         draft_cache = prompt_cache[len(model.layers) :]
 
-    if not cache.can_trim_prompt_cache(model_cache):
-        types = {type(c).__name__ for c in model_cache if not c.is_trimmable()}
+    if not cache.can_rewind_prompt_cache(model_cache):
+        types = {
+            type(c).__name__
+            for c in model_cache
+            if not (c.is_trimmable() or c.is_checkpointable())
+        }
         raise ValueError(
-            f"Speculative decoding requires a trimmable prompt cache " f"(got {types})."
+            "Speculative decoding requires a trimmable or checkpointable "
+            f"prompt cache (got {types})."
         )
+    # Some hybrid linear-attention models mix trimmable KV layers with
+    # fixed-size recurrent state that can only be checkpointed/restored, not
+    # trimmed to an arbitrary position -- rejecting draft tokens on those
+    # models means rewinding the whole round and replaying the accepted
+    # prefix (see _rewind_cache below), not just an O(1) trim.
+    needs_checkpoint = not cache.can_trim_prompt_cache(model_cache)
+    model_checkpoint = None
 
     sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
 
@@ -577,8 +589,29 @@ def speculative_generate_step(
             mx.clear_cache()
         return y
 
-    def _rewind_cache(num_draft, num_accept):
-        cache.trim_prompt_cache(model_cache, num_draft - num_accept)
+    def _rewind_cache(num_draft, num_accept, batch_y):
+        if needs_checkpoint:
+            # model_checkpoint is None if this round never got far enough
+            # to take one (e.g. an exception during draft generation, which
+            # only touches the draft cache) -- nothing to rewind in that
+            # case, model_cache is still exactly where the last round left it.
+            if num_accept < num_draft and model_checkpoint is not None:
+                # A fixed-size recurrent state can't be trimmed back to the
+                # accepted position the way a KV cache can -- undo the
+                # whole round (trimmable layers roll back for free, the
+                # checkpointed ones are restored) and replay exactly the
+                # tokens that turned out to be correct so every layer ends
+                # up in sync again.
+                cache.rewind_prompt_cache(model_cache, model_checkpoint, num_draft + 1)
+                accepted = batch_y[: num_accept + 1]
+                with mx.stream(generation_stream):
+                    model(accepted[None], cache=model_cache)
+                    quantize_cache_fn(model_cache)
+                    mx.eval([c.state for c in model_cache])
+            # else: every draft token was accepted, so model_cache already
+            # reflects the true trajectory -- nothing to rewind.
+        else:
+            cache.trim_prompt_cache(model_cache, num_draft - num_accept)
         cache.trim_prompt_cache(draft_cache, max(num_draft - num_accept - 1, 0))
 
     def _draft_generate(y, num_draft):
@@ -599,6 +632,7 @@ def speculative_generate_step(
     # Set these so the finally block doesn't raise
     num_draft = 0
     n = 0
+    batch_y = y
     try:
         while True:
             num_draft = min(max_tokens - ntoks, num_draft_tokens)
@@ -606,6 +640,9 @@ def speculative_generate_step(
             if prev_tokens is not None:
                 prev_tokens = prev_tokens[: prev_tokens.size - y.size - num_draft + 1]
             y = mx.concatenate([y, draft_tokens])
+            batch_y = y
+            if needs_checkpoint:
+                model_checkpoint = cache.checkpoint_prompt_cache(model_cache)
             tokens, logprobs = _step(model, model_cache, y, num_draft + 1)
             mx.eval(tokens, draft_tokens)
             draft_tokens = draft_tokens.tolist()
@@ -640,9 +677,9 @@ def speculative_generate_step(
 
             if prev_tokens is not None:
                 prev_tokens = prev_tokens[: -max(num_draft - n, 1)]
-            _rewind_cache(num_draft, n)
+            _rewind_cache(num_draft, n, batch_y)
     finally:
-        _rewind_cache(num_draft, n)
+        _rewind_cache(num_draft, n, batch_y)
 
 
 def mtp_speculative_generate_step(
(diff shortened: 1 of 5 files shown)


h9q2cyxvgm-ui:
Follow-up data point on the speedup caveat above, since it's worth being precise about what's actually causing it.

I ran a controlled comparison: same prompt, same `num_draft_tokens` sweep, temp=0, 150 tokens — but against a **dense** Qwen3 model (`mlx-community/Qwen3-32B-6bit`, plain `KVCache` only, no hybrid recurrent layers) with a small peer draft model (`Qwen3-0.6B`), using the *existing, unmodified* speculative decoding path (this model was never affected by the `ArraysCache` issue in the first place).

| `num_draft_tokens` | Dense Qwen3-32B (cheap reject via `.trim()`) | Hybrid Qwen3.6-27B + MTP head (expensive reject via checkpoint+replay) |
|---|---|---|
| 1 | **1.21x** (43% accepted) | 1.04x (48% accepted) |
| 2 | **1.13x** (54% accepted) | 1.00x (60% accepted) |
| 3 | **1.00x** (61% accepted) | 0.93x (67% accepted) |

At every setting, the dense model comes out ~0.15–0.2x ahead of the hybrid model — despite the hybrid model's MTP head actually guessing right *more often* at each setting. That's consistent with the theory in the description above: it isn't that this rewind approach is poorly implemented, it's that any rejected round on a hybrid model pays a real, structural cost (a full extra forward pass to rebuild the recurrent state) that a pure-KV model simply doesn't pay. A better guesser doesn't fully compensate for a more expensive miss.

Practical takeaway for anyone picking this up: tuning `num_draft_tokens` down helps a bit (1 is best of the three tested) but doesn't get hybrid-recurrent models to a clear win on its own — closing that gap further would need a cheaper way to recover from a rejected round than a full-model replay, which is a separate, harder problem than what's fixed in this PR.

Also flagging, found while running this comparison and unrelated to anything in this PR: `speculative_generate_step` with `num_draft_tokens=4` on the dense Qwen3-32B + Qwen3-0.6B pairing above produced output that didn't match plain generation at temp=0 (should be lossless). Reproduces on `main` too (checked before any of this branch's changes), so it's a separate, pre-existing bug — will file it as its own issue rather than mixing it in here.

zcbenz:
Closing per https://github.com/ml-explore/mlx-lm/pull/990#pullrequestreview-4989083778

PR #1469, opened 2026-07-04
Fix MTPHead: correct hidden-state/embedding concat order and use pre-norm hidden

h9q2cyxvgm-ui:
**Stacked on #1468** — this branch is that PR's branch plus one more commit (`06d99ad`). Please review just the last commit / the `models/qwen3_5.py` diff; everything else duplicates #1468 and will resolve itself once that one merges (happy to rebase this down to just the new commit at that point, whatever's easiest for you).

## The bug

While verifying #1468 end to end, the attached MTP head's drafted tokens had ~0% agreement with the target model's actual output — not occasionally wrong, essentially unrelated (e.g. drafting `'ikum' 'i' 'lech'` where the target was actually generating `'s '\n' '\n'`). That pointed at a real bug rather than just a low-quality drafter, and turned out to be two of them in `MTPHead`:

1. **Wrong concat order.** `MTPHead.__call__` concatenated `[hidden, embedding]` before the `fc` projection; the correct (DeepSeek-V3-style) order is `[embedding, hidden]`. Confirmed empirically: ran both orderings against the same hidden state/token and compared the draft output to the target model's actual next tokens at that position — only the swapped order matched.

2. **Wrong hidden-state stage.** `Qwen3_5TextModel.__call__` only ever returned the **post-final-norm** hidden state via `output_hidden_states=True`, but the MTP head needs the **pre-final-norm** residual stream (the final norm is calibrated for the LM head, not meant to be fed into an auxiliary module). Now returns the pre-norm hidden state instead, threaded up through `TextModel.__call__`. This is safe to change: `output_hidden_states=True` has no other callers anywhere in the codebase (only `mtp_speculative_generate_step` uses it).

Both bugs silently produce a *structurally valid but semantically scrambled* vocabulary projection rather than an error, which is why they were easy to miss — garbage in, plausible-looking garbage out.

## Verified

Together these took MTP draft acceptance from 0% to ~66% on `unsloth/Qwen3.6-27B-UD-MLX-6bit` + `mlx-community/Qwen3.6-27B-MTP-bf16`, with temp=0 output still exactly matching plain generation (see #1468 for the correctness-test details). No regressions in the existing test suite.

This is a correctness bug in the MTP head itself, independent of the cache-rewind work in #1468 — worth landing even if that PR needs more discussion.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -517,11 +517,23 @@ def speculative_generate_step(
         model_cache = prompt_cache[: len(model.layers)]
         draft_cache = prompt_cache[len(model.layers) :]
 
-    if not cache.can_trim_prompt_cache(model_cache):
-        types = {type(c).__name__ for c in model_cache if not c.is_trimmable()}
+    if not cache.can_rewind_prompt_cache(model_cache):
+        types = {
+            type(c).__name__
+            for c in model_cache
+            if not (c.is_trimmable() or c.is_checkpointable())
+        }
         raise ValueError(
-            f"Speculative decoding requires a trimmable prompt cache " f"(got {types})."
+            "Speculative decoding requires a trimmable or checkpointable "
+            f"prompt cache (got {types})."
         )
+    # Some hybrid linear-attention models mix trimmable KV layers with
+    # fixed-size recurrent state that can only be checkpointed/restored, not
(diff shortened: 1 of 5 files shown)


zcbenz:
Closing per https://github.com/ml-explore/mlx-lm/pull/990#pullrequestreview-4989083778

PR #1473, opened 2026-07-04
fix(vl): strip model.visual.* in qwen2_vl/qwen3_vl/qwen3_vl_moe sanitize

Jonathangadeaharder:
## Problem

HuggingFace `ForConditionalGeneration` VL checkpoints (Qwen3_5MoeForConditionalGeneration, Qwen3VLForConditionalGeneration, Qwen2VLForConditionalGeneration) nest the language model and vision tower under a top-level `model` dict:

```
model.language_model.*   (language model weights)
model.visual.*           (vision tower weights)
```

The existing `sanitize()` helpers only popped bare `visual` / `vision_tower` top-level keys — which matches the mlx-vlm conversion layout but **not** the HF layout. On HF checkpoints:

- **`qwen3_vl_moe.sanitize` crashes** with `KeyError: 'model'` because `weights["language_model"]["model"]` doesn't exist (the language model is nested under `model.language_model.model.*`, so after `tree_unflatten` it lives at `weights["model"]["language_model"]["model"]`, not `weights["language_model"]["model"]`).
- **`qwen3_vl` and `qwen2_vl` leak** `model.visual.*` into the sanitized weights, re-prefixed as `language_model.model.visual.*` — causing `"parameters not in model"` errors downstream.

## Reproducer

```python
from mlx.utils import tree_unflatten
flat = [
    ("model.visual.pos_embed.weight", None),
    ("model.language_model.model.layers.0.self_attn.q_proj.weight", None),
    ("model.language_model.lm_head.weight", None),
]
tree = tree_unflatten(flat)
# top keys: ["model"] — no bare "visual" or "language_model"
tree.pop("visual", None)  # no-op
tree["language_model"]["model"]  # KeyError: "model"
```

## Fix

After `tree_unflatten`, pop the top-level `model` dict, re-root its `language_model` subtree to the top level, and drop its `visual` / `vision_tower` subtrees. Three model classes changed identically:

- `mlx_lm/models/qwen3_vl_moe.py`
- `mlx_lm/models/qwen3_vl.py`
- `mlx_lm/models/qwen2_vl.py`

## Tests

`TestVLSanitize` in `tests/test_models.py` covers all three model classes against the HF `ForConditionalGeneration` layout (9-key fixture mirroring Ornith-1.0's real weight_map). Each test asserts:
1. `sanitize()` does not crash,
2. zero `visual` / `vision_tower` keys remain,
3. the language model tensors survive, re-rooted under `language_model.*`.

All three fail on `main` (Red), pass after the fix (Green). Full `tests/test_models.py` suite: 75 passed, 1 skipped (pre-existing `test_ssm` failure unchanged, unrelated).

## Verification

Verified end-to-end on `deepreinforce-ai/Ornith-1.0-35B` (333 `model.visual.*` tensors, full `vision_config`): with this fix, the model loads cleanly where it previously crashed with `KeyError: 'model'` during `sanitize()`.

### DIFF
--- a/mlx_lm/models/qwen2_vl.py
+++ b/mlx_lm/models/qwen2_vl.py
@@ -44,6 +44,12 @@ class Model(nn.Module):
         weights = tree_unflatten(list(weights.items()))
         weights.pop("visual", None)
         weights.pop("vision_tower", None)
+        # HF ForConditionalGeneration nests the language model under
+        # ``model.language_model.*`` and vision under ``model.visual.*``.
+        nested = weights.pop("model", None)
+        if isinstance(nested, dict):
+            if isinstance(nested.get("language_model"), dict):
+                weights["language_model"] = nested["language_model"]
         weights = dict(tree_flatten(weights))
 
         sanitized = {}
--- a/mlx_lm/models/qwen2_vl.py
+++ b/mlx_lm/models/qwen2_vl.py
@@ -44,12 +44,9 @@ class Model(nn.Module):
         weights = tree_unflatten(list(weights.items()))
         weights.pop("visual", None)
         weights.pop("vision_tower", None)
-        # HF ForConditionalGeneration nests the language model under
-        # ``model.language_model.*`` and vision under ``model.visual.*``.
-        nested = weights.pop("model", None)
-        if isinstance(nested, dict):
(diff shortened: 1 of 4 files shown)


michalk8 (approved):
I've updated the code based on yours, thanks; LGTM!

PR #1474, opened 2026-07-05
Fix NewlineTokenizer registration with Transformers v5

kime541200:
## Summary
- register NewlineTokenizer with Transformers v5 class-name lookup when available
- keep the existing AutoTokenizer.register fallback for older Transformers versions
- add a regression test for the NewlineTokenizer class-name registration

## Testing
- python -m unittest tests.test_tokenizers.TestTokenizers.test_newline_tokenizer_is_registered
- mlx_lm.server --help

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -500,7 +500,23 @@ class NewlineTokenizer(PreTrainedTokenizerFast):
         return [self._postprocess_text(d) for d in decoded]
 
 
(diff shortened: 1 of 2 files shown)


menisadi:
See also #1465 which might be related

angeloskath:
Thanks! The transformers 5.13 import crash was fixed by #1465, which passes the tokenizer class (rather than a string) to `AutoTokenizer.register`. That preserves the exact name-based lookup behavior with a one-line change. Closing as resolved by #1465.

PR #1476, opened 2026-07-05
Add KV cache quantization to mlx_lm.server (#1043)

### LINKED ISSUE #1043: Add KV cache quantization support to server

katlun-lgtm:
# Add KV cache quantization to `mlx_lm.server` (closes #1043)

## Status: partly superseded by #1584 — rebasing on top of it

**Correction to an earlier version of this section, which said #1584 superseded the design outright.** It doesn't, and the difference decides what this PR still has to do.

#1584 makes quantized caches batchable **when `--max-kv-size` is set**: it implements `RotatingKVCache.to_quantized()` / `RotatingQuantizedKVCache` (on main today both `RotatingKVCache.to_quantized` and `BatchRotatingKVCache.to_quantized` raise `NotImplementedError("… Quantization NYI")`) and threads them into `BatchGenerator`. For that case the blanket `_is_batchable → False` opt-out described below is the wrong call, and this PR should defer to #1584's path.

It does **not** cover plain `--kv-bits` with no `--max-kv-size`, which is the default server shape. `QuantizedKVCache` on main implements neither `merge()` nor `to_quantized()`, and #1584 adds `merge()` only via `RotatingQuantizedKVCache` — as its author [confirmed while isolating it from his own diff](https://github.com/ml-explore/mlx-lm/pull/1584#issuecomment-5082799557), `kv_bits` without `max_kv_size` cannot do batching-with-history at all. So the sequential route stays necessary, and the rebase **narrows** the opt-out to that case rather than dropping it.

Split [agreed with @mabaeyens](https://github.com/ml-explore/mlx-lm/pull/1584#issuecomment-5035969478): #1584 lands the primitive and deliberately leaves `server.py` untouched; this PR is the server surface — the three CLI flags, plus routing `--quantized-kv-start > 0` to the sequential path, since the batch path rejects a nonzero start by design (per-job caches begin at offset 0, with no per-step re-check, so only immediate quantization is well-defined there). #1618 is the complementary fail-early validation and also deliberately adds no server flag.

I'm holding the rebase until #1584 merges, since it may still change in review (its author has asked whether to split it); it's a sub-hour change once it lands. Everything below this section describes the **current** commit, which still assumes the pre-#1584 world.

## What
Exposes the KV-cache quantization already supported by `generate`/`stream_generate`
as server CLI flags:

- `--kv-bits` (default `None` → off)
- `--kv-group-size` (default `64`)
- `--quantized-kv-start` (default `0`)

When set, they are threaded into the server's sequential `stream_generate` path.

## Why the batched path is handled by fallback, not plumbing
The server uses continuous batching (`BatchGenerator`) whenever a request
`_is_batchable` — which requires **mergeable** caches
(`all(hasattr(c, "merge") ...)`). `QuantizedKVCache` intentionally does **not**
implement `merge()`, so a quantized cache cannot participate in continuous
batching without new batch-membership support for quantized caches (a larger,
separate feature).

Rather than silently ignore `--kv-bits` in the default (batched) path, this PR
makes `_is_batchable` return `False` when `--kv-bits` is set, so those requests
use sequential generation, which applies the quantization. This keeps the
existing mergeable-cache invariant intact and makes the flag behave correctly.

Trade-off worth a maintainer call: with `--kv-bits`, requests no longer batch.
If you'd prefer, I can instead scope a follow-up that teaches the batched path
to carry quantized caches. Happy to go either way.

## Changes
- `mlx_lm/server.py`: 3 argparse flags; pass to `stream_generate`; `_is_batchable`
  returns `False` when `kv_bits` is set.

## Test plan
Verified on Apple Silicon (M3 Max, mlx 0.31.2, mlx-lm 0.31.3) with
`mlx-community/Qwen2.5-0.5B-Instruct-4bit`:

- [x] `--kv-bits` flags parse with correct defaults (`None` / `64` / `0`).
- [x] `stream_generate(..., kv_bits=4)` produces a `QuantizedKVCache` and coherent
      output; `kv_bits=None` stays a plain `KVCache` (regression check).

```
kv_bits=None: cache=KVCache          quantized=False coherent=True
kv_bits=4:    cache=QuantizedKVCache quantized=True  coherent=True
```

- [ ] End-to-end HTTP smoke against a running `mlx_lm.server --kv-bits 4` (recommended
      before merge; core generation path is exercised above).

## Notes
Mirrors the existing `generate.py` flag names/help text verbatim for consistency.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -683,6 +683,11 @@ class ResponseGenerator:
         return sm, sequences
 
     def _is_batchable(self, args):
+        # Quantized KV caches are not mergeable, so they cannot participate in
+        # continuous batching. When --kv-bits is set, fall back to sequential
+        # generation (which applies the quantization) instead.
+        if self.cli_args.kv_bits is not None:
+            return False
         return self.model_provider.is_batchable and args.seed is None
 
     def _generate(self):
@@ -985,6 +990,9 @@ class ResponseGenerator:
                 num_draft_tokens=args.num_draft_tokens,
                 prompt_progress_callback=progress,
                 prefill_step_size=self.cli_args.prefill_step_size,
+                kv_bits=self.cli_args.kv_bits,
+                kv_group_size=self.cli_args.kv_group_size,
+                quantized_kv_start=self.cli_args.quantized_kv_start,
             ):
                 finish_reason = gen.finish_reason
                 sm_state, match_sequence, current_state = sm.match(sm_state, gen.token)
@@ -1878,6 +1886,25 @@ def main():
         "--prompt-cache-bytes",
         type=_parse_size,
         help="Maximum size in bytes of the KV caches",
+    )
+    parser.add_argument(
+        "--kv-bits",
+        type=int,
+        default=None,
+        help="Number of bits for KV cache quantization. Defaults to no quantization.",
+    )
+    parser.add_argument(
+        "--kv-group-size",
+        type=int,
+        default=64,
+        help="Group size for KV cache quantization.",
+    )
+    parser.add_argument(
+        "--quantized-kv-start",
+        type=int,
+        default=0,
+        help="When --kv-bits is set, start quantizing the KV cache "
+        "from this step onwards.",
     )
     parser.add_argument(
         "--pipeline",


katlun-lgtm:
Verified end-to-end on M3 Max: `mlx_lm.server --kv-bits 4` boots, `/v1/chat/completions` returns HTTP 200 and generates through the sequential (quantized-cache) path. The core `stream_generate` path is also confirmed to produce a `QuantizedKVCache` with `kv_bits=4` (plain `KVCache` without). Output quality in any short snippet here reflects a 0.5B/4-bit test model, not the change itself.

katlun-lgtm:
Heads-up on a dependency: #1584 implements the batched-path piece I flagged in the "Why the batched path is handled by fallback" section above — it gives quantized caches a real `merge()` (via `BatchRotatingQuantizedKVCache`), so the `_is_batchable → False` opt-out here isn't needed once it lands. #1584 doesn't touch `server.py` though, so #1043 stays open. If #1584 merges first I'll rebase this to just pass `kv_bits`/`kv_group_size` into `BatchGenerator` and keep quantized requests batched (routing `--quantized-kv-start > 0` to the sequential path, since #1584 rejects it on the batch path). Flagging so whoever reviews either PR sees the interaction.

Git-on-my-level:
Tested an equivalent server patch end-to-end on a 128 GB M5 Max with `mlx-lm 0.31.3`, using `pipenetwork/Laguna-S-2.1-MLX-4bit` and OMP as the OpenAI-compatible client.

One mixed-cache compatibility issue is worth covering before merge: Laguna has 12 full-attention `KVCache` layers and 36 fixed 512-token `RotatingKVCache` layers. `maybe_quantize_kv_cache()` sees `RotatingKVCache.to_quantized`, calls it, and gets `NotImplementedError("RotatingKVCache Quantization NYI")`. The PR as written therefore fails on the first request for this model.

Our local compatibility wrapper quantizes supported ordinary caches and leaves rotating caches unquantized. That is a good trade for Laguna because the rotating caches remain bounded at roughly 72 MiB total. Results:

- Cache-level test: 1,048,576 bytes BF16 -> 294,912 bytes KV4 (3.56x); rotating cache preserved.
- HTTP text and GLM tool-call parsing passed; the tool call included the required `command` argument.
- Native OMP passed a roughly 17.9K-token tool session and reused the cached prefix on following tool turns.
- A 100,047-token prefill completed; the retained KV4 prefix was 1.46 GB and produced no new swap growth.

I suggest either having `maybe_quantize_kv_cache()` skip caches whose conversion raises `NotImplementedError`, or adding an explicit quantization-capability contract rather than relying on `hasattr`. A mixed `KVCache`/`RotatingKVCache` test would protect hybrid architectures.

Also, because this PR deliberately routes KV-quantized requests through `_serve_single`, #1118 is a practical dependency: otherwise `--prompt-cache-bytes` remains unenforced in the sequential path.

katlun-lgtm:
Thanks for running it that hard — the 100k-token prefill and the cached-prefix reuse across tool turns are well past what I'd tested here.

You're right about the crash, and the mechanism is worth spelling out because the guard is the wrong shape rather than just missing a case. `maybe_quantize_kv_cache` gates on `hasattr(c, "to_quantized")` (`generate.py:294`), and `RotatingKVCache` *does* define `to_quantized` — it just raises `NotImplementedError("RotatingKVCache Quantization NYI")` (`cache.py:552`). So the `hasattr` check passes and the exception propagates on the first request. Any hybrid model hits it; your 12 full-attention + 36 rotating layout is exactly the shape that does.

The good news is that #1584 already fixes this at the source. On its current head (`443dd99b`) both NYI raises are gone — `RotatingKVCache.to_quantized` returns a real `RotatingQuantizedKVCache`, and `BatchRotatingKVCache.to_quantized` returns `BatchRotatingQuantizedKVCache`. This PR is already parked waiting to rebase onto #1584 (agreed split upthread: #1584 owns the batched-quantized primitive, this one owns the server flag surface), so the Laguna case should resolve itself when that lands rather than needing a separate compatibility wrapper here.

On preferring an explicit capability contract over `hasattr` — agreed, and that's the same conclusion @pierre427 and I landed on over in #1558, framed around #1323's `quantizable=False` flag. Better to have one answer for both than an `except NotImplementedError` in one place and a capability flag in the other.

I'll add the mixed `KVCache`/`RotatingKVCache` test with the rebase regardless of whether #1584 makes the crash itself unreachable — that's the case that would silently regress.

The #1118 point I haven't checked yet, so I'll verify rather than take it on trust before this goes up for review again. If `--prompt-cache-bytes` really is unenforced on the sequential path that `--kv-bits` routes to, that's worth noting in the flag docs here even if the fix belongs in #1118.

zcbenz:

PR #1477, opened 2026-07-06
Fix crash when batching requests with and without logits processors

### LINKED ISSUE #1472: mlx_lm.server: generation thread dies with TypeError ('NoneType' object is not iterable) when a batch mixes requests with and without logits processors — server then hangs forever
## Summary

In `mlx_lm.server` (continuous batching), if a plain request (no `repetition_penalty` / `logit_bias` / penalties) is in the generation batch and a request **with** a logits processor joins it (or vice versa), `BatchGenerator._step` raises `TypeError: 'NoneType' object is not iterable` in the `_generate` worker thread. The thread dies, but the HTTP server stays up — so the server keeps accepting connections while no request ever completes again, until it is externally restarted. We hit this roughly daily in production (a health-check probe colliding with periodic large summarization calls that set `repetition_penalty: 1.1`) and captured the traceback below.

## Traceback (production, v0.31.3)

```
  File ".../python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File ".../python3.12/threading.py", line 1010, in run
  File ".../site-packages/mlx_lm/server.py", line 853, in _generate
  File ".../site-packages/mlx_lm/generate.py", line 1855, in next
    return self._next()
  File ".../site-packages/mlx_lm/generate.py", line 1775, in _next
    generation_responses = self._generation_batch.next()
  File ".../site-packages/mlx_lm/generate.py", line 1415, in next
    tokens, logprobs = self._step()
  File ".../site-packages/mlx_lm/generate.py", line 1346, in _step
TypeError: 'NoneType' object is not iterable
```

## Root cause (code walk, v0.31.3 — same code present on `main` as of 2026-07-04)

Three interacting pieces in `mlx_lm/generate.py`:

1. **The None factory** — when a batch's entries are all plain, the per-entry list is canonicalized to **`None` placeholders** before extending with new entries:

```python
# generate.py:1065-1066 (batch extend path)
if not any(self.logits_processors):
    self.logits_processors = [None] * len(self.uids)
...
# generate.py:1078
self.logits_processors.extend(logits_processors)
```

2. **The mix** — a request that *does* carry processors (e.g. `repetition_penalty` via `make_logits_processors`, which returns a non-empty list) extends in, producing e.g. `[None, [rep_penalty_fn]]`. Now `any(self.logits_processors)` is truthy.

3. **The unguarded iteration** — `_step` gates on batch-level `any()` but iterates per-entry with no per-entry guard:

```python
# generate.py:1337, 1344-1346
if any(self.logits_processors):
    ...
    for e in range(len(self.uids)):
        sample_logits = logits[e : e + 1]
        for processor in self.logits_processors[e]:   # <- entry can be None
```

The `None` entry raises, killing the generation thread. Note the samplers loop just below already handles exactly this case per-entry (`self.samplers[e] or self.fallback_sampler`), and the sibling filter path normalizes the all-plain case to `[[]] * len(keep)` (line 1117-1120) rather than `[None] * n` — the two normalizations disagree, and `_step` only survives one of them.

Homogeneous traffic never crashes: all-plain batches keep `any()` falsy (block skipped), all-processor batches have no None entries. The crash requires mixing — which is routine for an OpenAI-compatible endpoint serving multiple clients.

## Reproduction

1. `mlx_lm.server --model <any model> --host 0.0.0.0 --port 8080`
2. Send a plain streaming chat completion (no penalties) with enough `max_tokens` that it is still decoding in the generation batch.
3. While it decodes, send a second request with `"repetition_penalty": 1.1`.
4. When the second request joins the generation batch, `_step` hits the first request's `None` entry → traceback above; from then on the server accepts connections but never completes any generation request.

## Impact

- Single unlucky collision permanently wedges the server (alive process, dead inference) — needs external watchdog/restart.
- Because the trigger is per-request sampling-parameter *diversity*, any deployment with heterogeneous clients will eventually hit it.

## Suggested fix

Either (or both, for defense in depth):

- Normalize with empty lists instead of `None` at generate.py:1066 (`[[]] * len(self.uids)`), matching the line-1120 path; or
- Per-entry guard at generate.py:1346: `for processor in (self.logits_processors[e] or []):` — the same pattern the samplers loop already uses.

We are running the one-line 1346 guard as a local patch in production — verified by deliberately recreating the collision (a plain long completion with concurrent `repetition_penalty` requests joining mid-decode, and the reverse): no crashes, all requests complete, server stderr clean. Happy to submit it as a PR if useful.

## Environment

- mlx-lm 0.31.3 (pip), Python 3.12.3 (pyenv)
- macOS, Apple M4 Max 128 GB
- Server: `mlx_lm.server --model mlx-community/Qwen3.6-35B-A3B-8bit --host 0.0.0.0 --port 8080 --prefill-step-size 8192`

dasashreeya:
Fixes #1472

The server dies when a request using `repetition_penalty` or `logit_bias` ends up in the same batch as one that doesn't use either: the generation thread crashes with `TypeError: 'NoneType' object is not iterable` but the HTTP server stays up, so it accepts connections forever without answering.

The cause is a mismatch between how the batch classes represent "no logits processors". `PromptProcessingBatch.extend()` normalizes an all-plain batch to `[None] * n`, but `GenerationBatch._step()` iterates each entry with no per-entry guard:

`    for processor in self.logits_processors[e]:   # entry can be None`

Timing matters, which is probably why this survived so long: the plain request has to still be mid-prefill when the processor carrying request is inserted. If it finishes prefill first, `filter()` rewrites the all-None list to `[[]]` and nothing goes wrong. When they overlap, the mixed list `[None, [proc]]` passes the batch-level `any()` check and the None reaches the loop.

The fix is a one-line guard at the consumption point, same as the samplers loop a few lines down already does (`self.samplers[e] or self.fallback_sampler`). This also covers `None` passed per-request through `insert()` directly.

The regression test replays the interleaving deterministically (17-token plain prompt with `prefill_step_size=8`, insert the processor request after one step). It fails on main with the traceback from #1472 and passes with this change. It also asserts the logit bias still applies to the request that carries it and not to the plain one, so the guard isn't just skipping processing.

One thing I deliberately left alone: `extend()` pads with `None` while `filter()` pads with `[]` for the same state. Normalizing that would also fix this, but it's a bigger change — happy to do it here or as a follow-up if you'd prefer.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1343,7 +1343,7 @@ class GenerationBatch:
             processed_logits = []
             for e in range(len(self.uids)):
                 sample_logits = logits[e : e + 1]
-                for processor in self.logits_processors[e]:
+                for processor in self.logits_processors[e] or []:
                     sample_logits = processor(token_context[e], sample_logits)
                 processed_logits.append(sample_logits)
             logits = mx.concatenate(processed_logits, axis=0)
--- a/tests/test_generate.py
+++ b/tests/test_generate.py
@@ -438,6 +438,35 @@ class TestGenerate(unittest.TestCase):
         generated_token = self.tokenizer.encode(response.texts[0])[0]
         self.assertEqual(generated_token, 0)
 
+    def test_batch_generate_mixed_logits_processors(self):
+        """A request with logits processors joining a batch of requests
+        without them must not break generation (issue #1472)."""
+        tok = self.tokenizer.encode("hello")[-1]
+        batch_gen = BatchGenerator(self.model, max_tokens=5, prefill_step_size=8)
+
+        # Plain request long enough to need two prefill steps
+        (uid0,) = batch_gen.insert([[tok] * 17])
+        batch_gen.next()
+
+        # Request with a logits processor joins while the plain one is
+        # still prefilling
+        processors = make_logits_processors(logit_bias={0: 2000.0})
+        (uid1,) = batch_gen.insert([[tok] * 9], logits_processors=[processors])
+
+        first_logprobs = {}
+        finished = set()
+        while responses := batch_gen.next_generated():
+            for r in responses:
+                first_logprobs.setdefault(r.uid, r.logprobs)
+                if r.finish_reason is not None:
+                    finished.add(r.uid)
+
+        self.assertEqual(finished, {uid0, uid1})
+        # The processor was applied to the request that carries it
+        self.assertEqual(first_logprobs[uid1][0].item(), 0.0)
+        # ... and not to the one that doesn't
+        self.assertNotEqual(first_logprobs[uid0][0].item(), 0.0)
+
     def test_batch_generate_with_samplers(self):
         """Test that batch_generate with logits_processors produces correct results."""
         batch_gen = BatchGenerator(


zcbenz:

PR #1482, opened 2026-07-06
Fix mixed-bit quantized load for sanitize()-derived MLA projections

### LINKED ISSUE #1451: Mixed-bit MLA checkpoints fail to load — sanitize() re-quantizes derived embed_q/unembed_out at kv_b_proj's bits, but the modules are built at the global default
Loading a GLM-4.7-Flash checkpoint with a non-uniform per-tensor `quantization` map fails with a shape mismatch on the first derived MLA projection:

### Reproduction

```python
from mlx_lm import load
load("kwonhan/GLM-4.7-Flash-mixed-4-6-g32-mlx")
```

```
ValueError: Expected shape (20, 512, 24) but received shape (20, 512, 36)
for parameter model.layers.0.self_attn.embed_q.weight
```

`mlx-lm 0.31.3` (current PyPI); the relevant code is unchanged on current `main`.

The checkpoint is a 4-bit / g32 quant with 6-bit overrides on some tensors, including `kv_b_proj` on 23 of 47 layers:

```jsonc
"quantization": {
  "group_size": 32, "bits": 4, "mode": "affine",
  "model.layers.0.self_attn.kv_b_proj": {"group_size": 32, "bits": 6},
  ...
}
```

### Root cause

`glm4_moe_lite.py`'s `sanitize()` derives the absorbed-MLA projections `embed_q` / `unembed_out` from the fused `kv_b_proj` at load time. For quantized checkpoints it infers `kv_b_proj`'s bits/group_size from the packed shapes, dequantizes, splits, and **re-quantizes the two derived tensors at those same inferred bits** (`mlx_lm/models/glm4_moe_lite.py` L424 and L436-441).

But the *modules* are quantized separately by `class_predicate` in `mlx_lm/utils.py` (L363-369): the derived paths (`...self_attn.embed_q`, `...self_attn.unembed_out`) never exist in `config["quantization"]` — the converter never saw them, they only come into existence inside `sanitize()` — so the lookup misses, the fallback `f"{p}.scales" in weights` returns `True` (sanitize just created those scales), and the module is allocated at the **global default bits**.

Result: on every layer where `kv_b_proj`'s bits differ from the global default, the weight produced by `sanitize()` (packed at 6 bits → `(20, 512, 36)`) doesn't match the module allocated at 4 bits (`(20, 512, 24)`).

Uniform-bit quants can't expose this by construction — inferred bits always equal the default.

### Scope

- The identical `sanitize()` pattern exists in `deepseek_v3.py`, `deepseek_v32.py`, `kimi_linear.py`, and `longcat_flash.py`, so any mixed-bit quant of those models with a bit override on `kv_b_proj` should hit the same mismatch.
- #1324 consolidates these five copies into a shared helper but is explicitly behavior-preserving (it re-quantizes at the inferred source bits), so it neither fixes nor worsens this.
- Related but distinct from #1214: same failure class (per-tensor `quantization` map disagreeing with the post-`sanitize` namespace), different mechanism — there the keys are *renamed* by `sanitize`, here they are *created* by it.

### Workaround

Add explicit per-tensor overrides for the derived modules to the checkpoint's `config.json`, inheriting each layer's `kv_b_proj` bits:

```jsonc
"model.layers.0.self_attn.embed_q":     {"group_size": 32, "bits": 6},
"model.layers.0.self_attn.unembed_out": {"group_size": 32, "bits": 6},
// ... one pair per layer whose kv_b_proj has a non-default override
```

With those 46 entries added, the checkpoint loads and runs cleanly.

### Suggested fix

Have the load path honor the bits `sanitize()` actually used for derived tensors — e.g. `sanitize()` (or the shared helper from #1324) could also register the inherited `{bits, group_size}` for the derived paths into the quantization map before `class_predicate` is built, or `class_predicate` could infer bits from the packed weight/scales shapes when a path is missing from the map. Happy to send a PR if a direction is preferred.

rajanshxrma:
## Summary

Loading a mixed-bit quantized checkpoint fails with a shape mismatch when the source tensor a model's `sanitize()` derives another tensor from carries a per-tensor quantization override.

Concretely: `deepseek_v3.py`, `deepseek_v32.py`, `glm4_moe_lite.py`, `kimi_linear.py`, and `longcat_flash.py` all derive the absorbed MLA projections `embed_q`/`unembed_out` from `kv_b_proj` inside `sanitize()`, dequantizing and re-quantizing them at whatever bits/group_size `kv_b_proj` itself used. But `embed_q`/`unembed_out` never appear in `config["quantization"]` — the converter never saw those paths, since they only come into existence inside `sanitize()`. So `load_model`'s `class_predicate` (`mlx_lm/utils.py`) falls through to `f"{p}.scales" in weights`, which is `True` (sanitize just created those scales), and `nn.quantize` allocates the module at the **global default** bits/group_size instead of the bits it was actually packed at. On any layer where the source tensor's override differs from the default, `load_weights` then raises a shape mismatch.

Repro (from #1451):
```python
from mlx_lm import load
load("kwonhan/GLM-4.7-Flash-mixed-4-6-g32-mlx")
# ValueError: Expected shape (20, 512, 24) but received shape (20, 512, 36)
# for parameter model.layers.0.self_attn.embed_q.weight
```

Fixes #1451.

## Fix

When a path is missing from `config["quantization"]` but its weight already carries `.scales`, infer the actual `group_size`/`bits` from the packed weight and scales shapes (`in_dims = m.weight.shape[-1]`, same formula every affected `sanitize()` already uses on the source tensor) instead of assuming the top-level default. This is generic — it isn't specific to any one model file — and a no-op for the common uniform-bit case, since the inferred bits equal the default that was used to produce the checkpoint in the first place.

## Testing

- Added `test_load_model_with_mixed_bit_derived_mla_projection` in `tests/test_utils.py`: builds a tiny `deepseek_v3` model, quantizes it uniformly, then re-packs `embed_q`/`unembed_out` at a different bit width (without adding their paths to the quantization map) to simulate a `kv_b_proj` override — confirms the old code raises the shape-mismatch error from #1451, and the fix loads cleanly with the modules correctly reporting the overridden bits.
- `python -m unittest tests.test_models` — 76 passed.
- `python -m unittest tests.test_utils tests.test_prompt_cache` — all pass except two pre-existing failures in `test_convert`/`test_load_model_with_custom_get_classes` caused by an incomplete local HF cache with network disabled, unrelated to this change.
- `pre-commit run --files mlx_lm/utils.py tests/test_utils.py` (black, isort) — clean.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -366,7 +366,25 @@ def load_model(
                 return config["quantization"][p]
             if not hasattr(m, "to_quantized"):
                 return False
-            return f"{p}.scales" in weights
+            if f"{p}.scales" not in weights:
+                return False
+            # This path has no entry in the per-tensor quantization map but
+            # the checkpoint already carries a quantized weight for it (e.g.
+            # a model's sanitize() can derive and re-quantize tensors, such
+            # as the absorbed MLA projections in deepseek_v3/deepseek_v32/
+            # glm4_moe_lite/kimi_linear/longcat_flash, from a source tensor
+            # that has its own per-tensor bits/group_size override). Infer
+            # the actual packing from the weight/scales shapes instead of
+            # assuming the top-level default, so the module nn.quantize
+            # allocates matches the weight that will be loaded into it.
+            in_dims = m.weight.shape[-1]
+            group_size = in_dims // weights[f"{p}.scales"].shape[-1]
+            bits = (weights[f"{p}.weight"].shape[-1] * 32) // in_dims
+            return {
+                "group_size": group_size,
+                "bits": bits,
+                "mode": quantization.get("mode", "affine"),
+            }
 
         nn.quantize(
             model,
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,5 +1,6 @@
 # Copyright © 2024 Apple Inc.
 
+import dataclasses
 import json
 import os
 import tempfile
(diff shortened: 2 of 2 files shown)


michalk8 (approved):
Thanks for spotting this, I've modified your code a bit, happy to merge now.

PR #1483, opened 2026-07-06
Add opt-in checkout_cache/return_cache to avoid deep-copy on fetch (#1395)

### LINKED ISSUE #1395: `fetch_nearest_cache` deep-copies the cached KV, doubling peak memory exactly when a cached conversation is reused

rajanshxrma:
Fixes #1395.

## Problem

`LRUPromptCache.fetch_nearest_cache` deep-copies the matched entry's full KV cache on every fetch. For the duration of the fetch, the process holds two full copies of that conversation's KV -- on memory-constrained machines this can OOM-crash the server exactly when reusing the longest-running conversation (see #1395 for the full repro: ~6.3GB cached sequence dies instantly on fetch on a 16GB Mac).

## Why not just remove the deep copy

`KVCache.update_and_fetch` writes new tokens into a slice of a pre-allocated buffer **in place** (`self.keys[..., prev:offset, :] = keys`), not via functional/immutable concatenation. So two references sharing that buffer would corrupt each other the moment either one continues generation -- the deep copy is load-bearing, not incidental.

It's also not a safe drop-in change: the existing test suite (`test_lru`, `test_caching`) explicitly exercises repeated fetches of the same/overlapping-prefix entry with no re-insertion in between, and asserts the entry is still found each time. A naive pop-on-fetch would break that documented, tested contract, and the batched server path (`server.py`, `batch_generator`) can have multiple concurrent requests sharing a prefix, which a blind pop would also mishandle.

## This PR

Adds `checkout_cache` / `return_cache` as a new, **opt-in** API alongside the existing `fetch_nearest_cache` / `insert_cache`, which are completely unchanged:

- `checkout_cache(model, tokens)` returns the same object reference (no copy) when the matched entry isn't currently checked out by anyone else, or falls back to the existing safe deep-copy behavior if it's contended.
- `return_cache(checkout_token)` releases the checkout, making the no-copy path available again. Safe to call with `None`, and safe to call if the entry was evicted while checked out (both are no-ops).
- The trim-and-continue path (`can_trim_prompt_cache`) always uses the deep-copy behavior regardless of checkout state, since trimming's continuation writes into a region the untrimmed entry still needs.

Nothing about the default `fetch_nearest_cache`/`insert_cache` behavior changes, so this is zero-risk to existing callers. Wiring `checkout_cache`/`return_cache` into `server.py`'s actual request-handling call sites (with a `try`/`finally` around generation) is a natural follow-up once this primitive lands -- happy to send that as a second PR, or fold it into this one if preferred.

## Testing

- All existing `tests/test_server.py` cache tests pass unmodified (verified before and after).
- Added tests covering: no-copy on the uncontested path (verified via object identity, not just equality), fallback-to-copy when contended, reuse of the no-copy path after `return_cache`, and safe no-op behavior when returning a checkout for an entry that was evicted in the meantime.

- Added regression tests for two release-safety cases: a doubled `return_cache` of a stale token must not release a later caller's checkout, and a token from before an evict-and-reinsert of the same key must not release the new entry's checkout (both verified failing before the hardening commit that scoped tokens per checkout and added the lock).

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1,6 +1,7 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import copy
+import threading
 from collections import deque
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional
@@ -1626,7 +1627,10 @@ class LRUPromptCache:
         prompt_cache: List[Any]
         nbytes: int
         cache_type: str
-        checked_out: bool = False
+        # Nonce identifying the currently active checkout, or None when the
+        # entry isn't checked out. Comparing identities in return_cache means
+        # only the checkout that set it can clear it.
+        checked_out: Optional[object] = None
 
     class CacheOrder:
         def __init__(self, ordering: List[str] = ["assistant", "user", "system"]):
@@ -1664,6 +1668,9 @@ class LRUPromptCache:
         self._lru = LRUPromptCache.CacheOrder()
         self._n_bytes = 0
         self._n_bytes_by_type = {k: 0 for k in self._lru._ordering}
+        # Serializes the checked_out check-then-set/clear so two concurrent
+        # checkout_cache calls can never both take the no-copy path.
+        self._checkout_lock = threading.Lock()
 
     def __len__(self):
         return len(self._lru)
@@ -1701,10 +1708,15 @@ class LRUPromptCache:
         # sharing that reference would corrupt the first's history. Only
         # hand out the shared reference (skipping the copy) when nothing
         # else currently holds it; otherwise fall back to the safe copy.
-        if cache_entry.checked_out:
-            return copy.deepcopy(cache_entry.prompt_cache), None
-        cache_entry.checked_out = True
-        return cache_entry.prompt_cache, (model, entry_tokens)
+        # The lock makes the check-then-set atomic, and the per-checkout
+        # nonce means a stale or doubled return_cache can't release a
+        # checkout it doesn't own. The deep copy happens outside the lock.
+        with self._checkout_lock:
+            if cache_entry.checked_out is None:
+                nonce = object()
+                cache_entry.checked_out = nonce
+                return cache_entry.prompt_cache, (model, entry_tokens, nonce)
+        return copy.deepcopy(cache_entry.prompt_cache), None
 
     def checkout_cache(self, model: Any, tokens: List[int]):
         """Like fetch_nearest_cache, but avoids deep-copying the returned KV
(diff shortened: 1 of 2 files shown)


rajanshxrma:
Flagging that #1409 is also open against this same issue, from a different approach: it modifies \`fetch_nearest_cache\` directly with a consume-mode parameter and also wires in \`--prompt-cache-bytes\` enforcement on the server, whereas this PR adds a separate, additive \`checkout_cache\`/\`return_cache\` API and leaves the existing functions untouched.

Didn't see #1409 before opening this one -- apologies for the overlap. Happy to close this in favor of #1409 if that's the preferred direction, or to fold the opt-in/no-copy angle into it as a follow-up if there's appetite for both. Deferring to maintainer judgment on which (if either) to take.

rajanshxrma:
Pushed a hardening commit after another review pass: checkout tokens are now scoped to the specific checkout that produced them (per-checkout nonce) rather than just `(model, tokens)`, so a doubled or stale `return_cache` — including one from before an evict-and-reinsert of the same tokens — is a no-op instead of releasing a checkout it doesn't own. The `checked_out` check-then-set is now also guarded by a lock so two threads can't both take the no-copy path. Added regression tests for both (verified failing before the fix). Also rebased on main.

zcbenz:

PR #1486, opened 2026-07-06
Exact speculative rollback for hybrid caches (model-agnostic cache API)

lBroth:
Up front: I'm not a professional ML engineer and this was done with heavy AI
assistance — I've validated everything I could on-device (the matrix below is
all real measured runs), and I'd welcome expert scrutiny on anything I've missed.

Split of #1456 as @pierre427 proposed: this is the **rollback cache-API alone**
(the shared dependency), landed standalone. The qwen3.5 MTP-head loading follows
as a separate PR on top. Co-developed with @pierre427, who contributed the
`RotatingKVCache` support and reviews.

## What & why

Speculative decoding needs to rewind the KV/state cache when a draft is
rejected. mlx-lm's existing `--draft-model` path does this by `trim()`ing the
cache — but that **silently corrupts** any cache that isn't trimmable:
- **hybrid GDN/SSM models** (qwen3-next, qwen3.5): the GatedDeltaNet recurrent
  state (`ArraysCache`) has no valid `trim()` (#980, #1446) — a multi-token
  draft advances the recurrence and there is no way to undo it by truncation;
- **`RotatingKVCache`**: `trim()` only adjusts `offset`/`_idx`, it does **not**
  shrink the key buffer, so a multi-token block feed that crosses the sliding
  window leaves stale keys.

This PR adds an **exact** rollback protocol so speculative decoding is
lossless on these caches.

## The API (model-agnostic — proven on two cache families)

A small protocol on the cache objects: `start/stop_speculation()`,
`record_rollback(num_tokens, fn, snapshot)`, `is_trimmable()`, strict `trim()`.
Two independent implementations demonstrate the abstraction isn't cache-specific:

1. **GDN / ArraysCache** (lBroth): rollback by **exact state replay** — snapshot
   the pre-block `conv_state` + recurrent `S`, capture (via hooks) the exact
   tensors the GatedDeltaNet kernel consumed, and reconstruct `S_m` by re-running
   `gated_delta_update` on the committed prefix. Exact by construction (no
   approximation), because the recurrence can't be truncated.
2. **RotatingKVCache** (pierre427): rollback by **copying** the (≤ window) buffers
   at snapshot time, since `trim()` can't shrink them.

Same protocol, opposite mechanisms — that's the evidence the API is right.

## Correctness (the first thing you'll ask)

- Rollback is **exact**, not approximate. Contract tests assert token-for-token
  identity between speculative and non-speculative greedy decode.
- Losslessness bar: every emitted token equals the target's **batched-greedy
  argmax** at that position (the accept rule always commits the target's own
  token) — not bit-identity with sequential `generate_step`, which differs by
  inherent FP noise regardless of speculation.
- Tests: `tests/test_speculative_rollback.py` (8) + `tests/test_rotating_rollback.py`
  (4) — 12 total, all green. Cover cache-reuse across turns, early-close, and
  offset-exactness after each step.

## No overhead when unused, no regressions

- The path is inert unless speculation is active: `start_speculation()` is only
  called on the speculative branch; normal generation is byte-identical.
- Cross-hardware validation (fresh envs, python 3.13, mlx 0.31.2): M5 Pro 48GB /
  M2 Pro 16GB / M3 16GB / CPU(ops) — unit tests 15/15, cache regression 22/22
  (failure set ⊆ main's), real-weights e2e identical (80B Coder-Next + 0.8B
  draft, token-for-token vs non-speculative). No regressions found anywhere.

## Scope

Rollback cache-API + speculative wiring + GDN & RotatingKVCache support + tests.
**No** MTP head (separate follow-up). Diff is +639/-6 across 7 files.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -517,7 +517,9 @@ def speculative_generate_step(
         model_cache = prompt_cache[: len(model.layers)]
         draft_cache = prompt_cache[len(model.layers) :]
 
-    if not cache.can_trim_prompt_cache(model_cache):
+    if not cache.can_trim_prompt_cache(model_cache) and not getattr(
+        model, "supports_speculative_rollback", False
+    ):
         types = {type(c).__name__ for c in model_cache if not c.is_trimmable()}
         raise ValueError(
             f"Speculative decoding requires a trimmable prompt cache " f"(got {types})."
@@ -595,6 +597,12 @@ def speculative_generate_step(
         draft_y = _prefill(draft_model, draft_cache, y)
         y = _prefill(model, model_cache, y)
 
+    # After prefill (recording a rollback for the whole prompt would hold on to
+    # prompt-sized tensors), let rollback-capable caches start recording so
+    # verify steps can be trimmed exactly (no-op for regular KV caches).
+    for c in model_cache:
+        c.start_speculation()
+
     ntoks = 0
     # Set these so the finally block doesn't raise
     num_draft = 0
@@ -643,6 +651,8 @@ def speculative_generate_step(
             _rewind_cache(num_draft, n)
     finally:
         _rewind_cache(num_draft, n)
+        for c in model_cache:
+            c.stop_speculation()
 
 
 def stream_generate(
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -665,7 +665,10 @@ def speculative_generate_step(
             _rewind_cache(num_draft, n)
     finally:
         _rewind_cache(num_draft, n)
-        for c in model_cache:
+        # Symmetric with start_speculation above: draft caches must stop
+        # recording too, or a reused draft prompt cache keeps reporting
+        # is_trimmable() with stale rollback records.
+        for c in model_cache + draft_cache:
             c.stop_speculation()
 
 
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -587,8 +587,18 @@ def speculative_generate_step(
         return y
 
     def _rewind_cache(num_draft, num_accept):
-        cache.trim_prompt_cache(model_cache, num_draft - num_accept)
-        cache.trim_prompt_cache(draft_cache, max(num_draft - num_accept - 1, 0))
+        # trim_prompt_cache is all-or-none: if any cache reports untrimmable
+        # (e.g. a layer that declared rollback support but never armed it),
+        # it trims nothing and returns 0. Continuing would leave ghost tokens
+        # in every layer, silently. Fail loudly instead.
+        for c, n in ((model_cache, num_draft - num_accept),
+                     (draft_cache, max(num_draft - num_accept - 1, 0))):
+            if n > 0 and cache.trim_prompt_cache(c, n) != n:
+                raise RuntimeError(
+                    "Speculative rewind failed: cache refused to trim "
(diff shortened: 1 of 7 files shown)


pierre427:
Thanks for cutting this out into its own PR, @lBroth — clean split. Confirming from my side: the `RotatingKVCache` half is validated on an M5 Max (gpt-oss-family sliding-window layers, bit-exact tail-independence check against a fresh sequential recompute, mod the bf16 ULP ties we both saw). Reads good to me and matches what I had on the staged branch. 👍

lBroth:
A real-world datapoint on whether this API is needed downstream.

MTPLX (youssofal/MTPLX, the MTP speculative decoding server for Apple
Silicon) loads its trunks through mlx_lm.utils.load and runs speculative
verify over the resulting mlx-lm cache objects. Since mlx-lm offers no
rollback for untrimmable caches, it hand-rolls exactly the protocol this
PR proposes, from the outside: snapshot_untrimmable_cache copies the GDN
recurrent state before every verify forward, and rollback_after_verify
restores it and trims the attention entries when drafts are rejected
(mtplx/cache_state.py lines 3424, 3442 and 3474 on current main, wired
into the decode loop at generation.py 4739 and 4857). It is duck-typed
over .state/.meta_state and a trimmability check, so it has to track
mlx-lm cache internals release by release.

I came across this while landing a context-copy PR there
(youssofal/MTPLX#151); I don't speak for that project. But it seems like
a concrete answer to "who needs exact rollback on hybrid caches": a
downstream with real users already ships a private reimplementation of
it, and with this API in mlx-lm (MTPLX pins mlx-lm>=0.31,<0.32) that
external surgery could become calls into the cache objects themselves.

lBroth:
Rebased on current main (cf10f96); the two commits since my base are the XTC sampler fixes and touch nothing this PR touches, so the rebase is clean and the diff is unchanged.

On the rebased tree: both rollback suites pass, 12 tests, test_prompt_cache passes all 22, and test_models is 75 passed with test_ssm failing, which fails identically on stock main on this machine, so it is pre-existing and unrelated to this change.

Two datapoints on what this unblocks, both measured on the 80B hybrid (Qwen3-Coder-Next 3-bit, M5 Pro).

Through the upstream path directly: on current main these models cannot speculate at all, `ValueError: Speculative decoding requires a trimmable prompt cache (got {'ArraysCache'})`, reproduced today. On this branch the same call runs: with a Qwen3-0.6B draft on a verbatim-copy prompt, 81.0 to 106.7 tok/s decode, 1.32x, and the two greedy outputs agree on 537 of 538 tokens, the one difference being a newline placement of the same near-tie class the batched-verify drift produces.

Through a standalone prompt-lookup loop over the same hooks, re-validated today: 3.6x on verbatim re-emission, 3.1x on edit-heavy passes, 2.1x on small edits, 1.04x on from-scratch generation, token-identical to plain decode everywhere except a documented near-tie case on scratch. Harness, gates and result JSONs: https://github.com/lBroth/mlx-lm/tree/pld-validation/pld

It also bears on #1297: on hybrid caches that PR has to snapshot by deepcopy and re-forward accepted tokens, and this API replaces both with an exact in-place rollback, so it is not just my loop's dependency.

The RotatingKVCache half has @pierre427's M5 Max validation upthread; the hybrid GDN half carries the same falsification tests in tests/test_speculative_rollback.py. If anything is blocking review that I can change, tell me and I will.

zcbenz:

PR #1490, opened 2026-07-07
Add LLaDA-8B masked diffusion language model + diffusion decoding

pierre427:
# Add LLaDA-8B (masked diffusion LM) + training-free diffusion decoding

## Summary

This adds **LLaDA-8B** — a masked **diffusion** language model — to mlx-lm, the
first non-autoregressive model in the tree, plus a self-contained diffusion
sampler with a suite of **opt-in, training-free acceleration levers** ported from
the diffusion-LLM literature (Fast-dLLM, BD3-LM).

- New module: `mlx_lm/models/llada.py` (`model_type: "llada"`, auto-discovered).
- Loads `GSAI-ML/LLaDA-8B-Instruct` / `-Base` via `mlx_lm.load` with **no
  changes to the loader** (`sanitize` remaps the HF `transformer.*` keys).
- A pure-MLX `generate()` implementing the LLaDA denoising loop. **The default
  path (fixed schedule, greedy) is exact and correct**; every accelerator below
  is opt-in and off by default.

> **Heads-up for reviewers — interface gap.** mlx-lm's `generate`/`server`/`chat`
> are autoregressive (causal KV cache, token-by-token) and **cannot drive a
> diffusion model** — cf. #1391 (diffusion_gemma fails in `mlx_lm.server`). LLaDA
> therefore ships its own `generate()` rather than plugging into the standard
> loop. **This PR is also a concrete proposal to discuss how diffusion decoding
> should integrate** (a sanctioned `diffusion_generate`, or a model-declared
> custom-generate hook). Happy to reshape around whatever interface you prefer.

## Provenance (this is a reproduction, not new research)

The acceleration techniques are established; the contribution is the **MLX
implementation + measurements on Apple Silicon**, not the methods:

- **Fast-dLLM** ([2505.22618](https://arxiv.org/abs/2505.22618), NVIDIA) —
  training-free block-wise KV cache + DualCache + confidence-aware parallel
  decoding on LLaDA/Dream. Official (PyTorch) impl: NVlabs/Fast-dLLM.
- **Fast-dLLM v2** ([2509.26328](https://arxiv.org/abs/2509.26328)) — block-diffusion.
- **Block Diffusion / BD3-LMs** ([2503.09573](https://arxiv.org/abs/2503.09573),
  ICLR 2025) — AR-across-blocks / diffusion-within; finalized-block KV reuse.
- **dKV-Cache** ([2505.15781](https://arxiv.org/abs/2505.15781)), **dLLM-Cache**
  (feature caching) — related.

The one small original bit is `remask_refine` (below), an order-aware re-masking
fix for a parallel-decoding failure mode.

---

## The model (`models/llada.py`)

LLaDA-8B is OLMo-derived but with two defining differences from a causal LM:

1. **Bidirectional attention** — `mask=None` everywhere, no causal mask.
2. **No KV cache in the base forward** — masked diffusion re-runs a full
   bidirectional forward over the whole sequence each denoising step.

Otherwise standard: pre-norm RMSNorm (ε=1e-5), RoPE θ=5e5, SwiGLU 4096→12288
(three matrices), 32 layers / 32 heads / no-GQA, untied output head (`ff_out`),
vocab 126464, `mask_token_id=126336`. `sanitize()` remaps
`model.transformer.{wte,blocks.N.*,ln_f,ff_out}` → the standard mlx-lm layout
(the one trap: `blocks.N.ff_out` = MLP down-proj vs top-level `transformer.ff_out`
= LM head).

### Sampler interface

```python
generate(model, prompt,
    steps=128, gen_length=128, block_length=128,          # schedule
    temperature=0.0, cfg_scale=0.0, remasking="low_confidence",
    parallel_threshold=None,                              # lever 1
    remask_refine=False, remask_conf=0.9, remask_rounds=2,# lever 2
    kv_cache=False, dual_cache=False,                     # levers 3, 4
    incremental_cache=False,                              # lever 5
    mask_id=126336, tokenizer=None, return_stats=False)
```

All levers default off → identical to the plain fixed-schedule LLaDA sampler.

---

## Performance levers (all opt-in) + measurements

All numbers: M-series (M5), bf16, LLaDA-8B-Instruct, `block_length=32`, greedy
unless noted.

### First: the operating point (why these levers, in this order)

Profiling a single denoising step (prompt_len=25, total_len=153):

| component | ms | share |
|---|---|---|
| **transformer forward** | 78.66 | **97.9%** |
| sampler (softmax/argmax/argsort/assemble over 126k vocab) | 1.69 | 2.1% |

The 8B forward **is** the step; sampler-op optimization is pointless. And a quant
sweep (q8/q6/q4) moved throughput <5% (q4 slightly *slower*) → **decode is
compute-bound, not weight-bandwidth-bound**: each step is a prefill-shaped
full-sequence forward. Forward scaling: `forward(L) ≈ 29ms + 0.236ms/token`. So
the levers that pay are (a) **fewer forwards** (parallel decoding) and (b)
**cheaper forwards** (caching) — quant is memory-only (q6: 6.8 GB, no quality/speed
cost).

### Lever 1 — Confidence-aware parallel decoding (`parallel_threshold`)

Reveal every masked token whose clean-softmax confidence clears τ, instead of a
fixed per-step count → cuts the number of forwards.

| prompt | baseline #fwd | τ=0.9 #fwd | speedup |
|---|---|---|---|
| capital-of-France | 128 | 5 | ~26× |
| n-th Fibonacci | 128 | 62 | ~2.1× |

Content-dependent (short/low-entropy collapses; dense factual only ~1.1–1.3×).
**Not bitwise-lossless** (Fast-dLLM Thm 1 approximation bound).

<details><summary><b>Tuning: τ and a real failure mode</b></summary>

A 12-prompt sweep (gen 128 + 256) found parallel decoding corrupts one
multilingual case at **gen=256**: `Buenos días` → `Buenos mañ` (a garbled
word-fragment collision between *buenos días* / *buenas mañanas*), reproduced on
bf16 **and** q6. It is **not** fixed by raising τ (0.9 ≡ 0.95 both break) nor by
`block_length` (non-monotonic). Narrow (1 of 4 multilingual prompts; all math /
code / English clean) but real. → default is the fixed schedule; parallel is
opt-in; recommended τ=0.95; multilingual/high-fidelity runs should enable
`remask_refine` (lever 2).
</details>

### Lever 2 — Order-aware re-masking (`remask_refine`) — the corruption fix

*(the one original piece)* Tracks each token's **reveal-time** confidence; after a
block fills, **re-masks** (not flips) the low-confidence span (±1 neighbour) and
**re-decodes it one-token-per-step, most-confident-first**, conditioned on the
fixed high-confidence tokens. Re-masking removes the locked-in fragment so the
model re-predicts correctly.

- **Fixes the repro** (`Buenos mañ` → `Buenos días`, matches greedy), **zero
  regressions** across a 6-prompt suite. Minimal `remask_conf=0.99, remask_rounds=1`.
- Cost is real: on decode-heavy prompts it approaches the fixed-schedule forward
  count. → opt-in high-fidelity lever.

<details><summary><b>Why a naive corrector fails (documented negative)</b></summary>

An argmax-disagreement corrector (re-forward, flip tokens where the full-context
argmax disagrees) was tried first and **does not work**: once `mañ` is placed the
full-context forward is 99.8% confident in *keeping* it — it's a self-consistent
local optimum, so nothing is ever flipped. The bug is reveal-**order**, which is
why re-masking (removing the token) succeeds where flipping cannot.
</details>

### Levers 3–5 — Block KV caching (`kv_cache`, `dual_cache`, `incremental_cache`)

Diffusion K/V are near-stable across adjacent steps (cosine≈1), so cache them.
Because decode is compute-bound, these FLOPs reductions translate to wall-clock.

- **`kv_cache`** (Fast-dLLM prefix): per block, prime the fixed prefix `[0,
  block_start)` K/V once; steps forward only the active window. A single primed
  forward is **bitwise-exact** (RoPE offsets verified). **1.40×@128 / 1.68×@256.**
- **`dual_cache`** (Fast-dLLM DualCache): also cache the masked suffix →
  active = just the block. **1.51×@128 / 2.27×@256** (~the full per-forward
  ceiling; captures the early-block win prefix-only can't).
- **`incremental_cache`** (BD3-LM inter-block reuse): at gen≥512 the cost is the
  per-block *primes* (16 at gen512/block32). Instead, **append** each finalized
  block's exact K/V to the prefix cache (captured from its last forward) and slice
  the suffix — 16 primes → 1. **1.43× fixed / 2.29× parallel over DualCache @512**;
  and it's *more faithful* than DualCache (exact append vs approximate re-prime).

<details><summary><b>Long-context (gen 512 / 1024) table + caveats</b></summary>

gen=512, incr vs dual: fixed mean 1.43×, parallel mean 2.29× (up to 2.9× where
primes dominate; a 584-tok-prompt case = 2.91× parallel). gen=1024 confirmation
(coherent, 0 mask leakage, correct answers) — `parallel+incr` vs plain parallel:

| prompt | parallel | parallel+incr | speedup |
|---|---|---|---|
| capital | 93 t/s | 410 t/s | 4.4× |
| fibonacci | 30 t/s | 137 t/s | 4.6× |
| diffusion | 28 t/s | 160 t/s | 5.8× |

Caches are **approximate** (bidirectional drift) → benign paraphrase, never
garbling; all outputs stayed correct/coherent to total_len ~1049. `incr` freezes
finalized K/V (no refresh); held clean at 1024, unverified beyond. `dual_cache`
requires `kv_cache`; `incremental_cache` requires both; all require `cfg_scale=0`.
</details>

### Composition

The levers compose. Best measured: **`parallel + incremental_cache` = 5.8× over
plain parallel at gen=1024**; `parallel + kv_cache` = 5.67× vs fixed (fib@256).

---

## Testing

- `test_llada_smoke.py` — **offline, no weights**: tiny-config forward shapes,
  mask-free generation on every lever + combo, and a strict `sanitize` weight-load
  on synthetic HF keys. Fast; suitable for CI.
- Real-weight benches (parallel sweep, quant sweep, profilers, cache/long-context)
  are included as scripts for reproducibility.

## Reviewer checklist / risk surface

- [x] Default path (all levers off) = exact fixed-schedule diffusion; no behavior
      change to any existing model.
- [x] Auto-discovered by `model_type`; no loader/registry edits.
- [x] `sanitize` verified (strict load).
- [ ] **Open design question**: how should diffusion decoding integrate with
      `generate`/`server`? (see #1391) — this PR is a starting point, not a
      finished interface.
- Non-lossless levers (parallel, caches) are opt-in and documented; a known
  multilingual failure mode has a fix (`remask_refine`).
- Techniques are reproductions (Fast-dLLM / BD3-LM); happy to trim to a minimal
  core (model + fixed-schedule generate) if you'd prefer the accelerators as
  follow-ups.

### DIFF
--- a/mlx_lm/models/llada.py
+++ b/mlx_lm/models/llada.py
@@ -1,0 +1,966 @@
+# LLaDA-8B: masked-diffusion language model (MDM).
+#
+# Structurally OLMo-derived (pre-norm RMSNorm + SwiGLU MLP), but with two
+# defining differences from a standard causal LM:
+#   1. Attention is BIDIRECTIONAL / non-causal (mask=None everywhere).
+#   2. Generation is diffusion-based, re-running a full forward over the whole
+#      sequence at every denoising step, so there is NO KV cache.
+#
+# HF checkpoints prefix everything with ``model.transformer.``; ``sanitize``
+# remaps those keys to the standard mlx-lm layout so the weights are
+# quantization-friendly.
+
+from dataclasses import dataclass
+from typing import Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .activations import swiglu
+from .base import BaseModelArgs
+
+# LLaDA-8B special token ids (from the official config.json).
+DEFAULT_MASK_TOKEN_ID = 126336
+DEFAULT_EOS_TOKEN_ID = 126081
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    # Field names mirror the official config.json keys so ``from_dict`` maps
+    # them with no manual translation.
+    model_type: str = "llada"
+    d_model: int = 4096
+    n_layers: int = 32
+    n_heads: int = 32
+    n_kv_heads: int = 32
+    mlp_hidden_size: int = 12288
+    vocab_size: int = 126464
+    embedding_size: Optional[int] = None
+    rms_norm_eps: float = 1e-5
+    rope_theta: float = 500000.0
+    mask_token_id: int = DEFAULT_MASK_TOKEN_ID
+    eos_token_id: int = DEFAULT_EOS_TOKEN_ID
+    pad_token_id: int = DEFAULT_EOS_TOKEN_ID
+    weight_tying: bool = False
+    include_bias: bool = False
+    include_qkv_bias: bool = False
+
+    def __post_init__(self):
+        # ``embedding_size`` is the true vocab / head size in LLaDA configs;
+        # keep ``vocab_size`` as an alias when only one is provided.
+        if self.embedding_size is None:
+            self.embedding_size = self.vocab_size
+        else:
+            self.vocab_size = self.embedding_size
+
+    @property
+    def head_dim(self) -> int:
+        return self.d_model // self.n_heads
+
+
+class Attention(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+
+        dim = args.d_model
+        self.n_heads = args.n_heads
+        self.n_kv_heads = args.n_kv_heads
+        self.head_dim = head_dim = args.d_model // args.n_heads
+        self.scale = head_dim**-0.5
+
+        bias = args.include_qkv_bias
+        self.q_proj = nn.Linear(dim, self.n_heads * head_dim, bias=bias)
+        self.k_proj = nn.Linear(dim, self.n_kv_heads * head_dim, bias=bias)
+        self.v_proj = nn.Linear(dim, self.n_kv_heads * head_dim, bias=bias)
+        self.o_proj = nn.Linear(self.n_heads * head_dim, dim, bias=args.include_bias)
+
+        self.rope = nn.RoPE(head_dim, traditional=False, base=args.rope_theta)
+
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        prefix_kv: Optional[tuple] = None,
+        pos_offset: int = 0,
+        return_kv: bool = False,
+        suffix_kv: Optional[tuple] = None,
+    ):
+        """Bidirectional attention with an optional Fast-dLLM prefix/suffix KV cache.
+
+        Modes:
+
+        * **Plain** (``prefix_kv=None``, ``return_kv=False``): the historical
+          path — project Q/K/V over all of ``x``, RoPE at absolute position 0,
+          attend with no cache. Byte-identical to the original implementation.
+
+        * **Prime** (``return_kv=True``): same full forward, but ALSO return the
+          post-RoPE ``(keys, values)`` for the whole ``x`` so the caller can
+          slice out a prefix window ``[0, block_start)`` and (for DualCache) a
+          suffix window ``[block_end, total_len)`` and cache them.
+
+        * **Cached** (``prefix_kv`` given): ``x`` holds ONLY the active window's
+          hidden states. Q/K/V are projected for the active positions, RoPE is
+          applied at ``offset=pos_offset`` (their absolute start), the cached
+          prefix K/V are prepended, and active queries attend over the full
+          ``[prefix ++ active]`` K/V bidirectionally. Only active outputs are
+          returned.
+
+        * **DualCache** (``prefix_kv`` AND ``suffix_kv`` given): as cached, but
+          the cached suffix K/V (post-RoPE at absolute positions
+          ``[block_end, total_len)``) are ALSO appended, so active queries attend
+          over ``[prefix ++ active ++ suffix]``. Attention is permutation-
+          invariant over keys and RoPE is baked in at prime time, so the concat
+          order is fine as long as each cached K carries its correct absolute
+          position. This shrinks the active window to just the current block.
+        """
+        B, L, _ = x.shape
+
+        queries = self.q_proj(x)
+        keys = self.k_proj(x)
+        values = self.v_proj(x)
+
+        queries = queries.reshape(B, L, self.n_heads, -1).transpose(0, 2, 1, 3)
+        keys = keys.reshape(B, L, self.n_kv_heads, -1).transpose(0, 2, 1, 3)
+        values = values.reshape(B, L, self.n_kv_heads, -1).transpose(0, 2, 1, 3)
+
+        # RoPE at the active window's absolute offset (0 in the plain/prime
+        # paths; block_start in the cached path).
+        queries = self.rope(queries, offset=pos_offset)
+        keys = self.rope(keys, offset=pos_offset)
+
+        if prefix_kv is not None or suffix_kv is not None:
+            # Prepend the cached prefix K/V and append the cached suffix K/V
+            # (each already post-RoPE at its absolute position) so active
+            # queries see the whole sequence bidirectionally.
+            key_parts, val_parts = [], []
+            if prefix_kv is not None:
+                key_parts.append(prefix_kv[0])
+                val_parts.append(prefix_kv[1])
+            key_parts.append(keys)
+            val_parts.append(values)
+            if suffix_kv is not None:
+                key_parts.append(suffix_kv[0])
+                val_parts.append(suffix_kv[1])
+            attn_keys = mx.concatenate(key_parts, axis=2)
+            attn_values = mx.concatenate(val_parts, axis=2)
+        else:
+            attn_keys, attn_values = keys, values
+
+        # Bidirectional attention: no causal mask.
+        output = mx.fast.scaled_dot_product_attention(
+            queries, attn_keys, attn_values, scale=self.scale, mask=mask
+        )
+        output = output.transpose(0, 2, 1, 3).reshape(B, L, -1)
+        out = self.o_proj(output)
+        if return_kv:
+            # Post-RoPE K/V for all of x (caller slices the prefix window).
+            return out, (keys, values)
+        return out
+
+
+class MLP(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        dim = args.d_model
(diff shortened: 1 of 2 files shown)


pierre427:
Cross-linking #1402 and #1391 — I opened this before catching up on the DiffusionGemma PR, and it overlaps directly.

#1402 is the right place for the generation-interface question, and I don't want to fork that discussion. LLaDA is a *different* diffusion paradigm — full masked diffusion (whole-sequence denoising) rather than DiffusionGemma's block-diffusion — so it's a useful second data point for whatever `diffusion_generate` / API boundary the maintainers land on there, but it shouldn't be a parallel design.

Marking this draft and reshaping it accordingly:

- Keep the core: the `llada` model + `sanitize`, conforming to the `Model` API.
- Replace the bespoke in-model `generate()` with a `diffusion_generate` matching whatever interface #1402 converges on.
- Pull the acceleration levers out of the initial PR (they're all opt-in / default-off): confidence-aware parallel decoding, block-wise prefix/Dual/incremental KV caching, and an order-aware re-masking quality fix. I'm flagging them mainly as concrete evidence of what a diffusion `generate` needs to support — parallel unmasking and cross-step / inter-block KV reuse — i.e. input to the interface design, not something to land in one shot.

Happy to follow the same model → sampler → interface staging being discussed on #1402.

pierre427:
**A second masked-diffusion model as input to the `diffusion_generate` interface question (#1391, #1402).**

We've ported **NVIDIA Nemotron-Labs-TwoTower-30B-A3B** to MLX — a *block-wise* diffusion LM on a **frozen autoregressive** Nemotron-H (Mamba-2 / attention / MoE hybrid) backbone. A frozen "context" tower produces per-layer KV + Mamba boundary states; a trained "denoiser" tower fills one block at a time via mask diffusion with **adaLN time-conditioning**, bidirectional in-block attention over the context KV, and Mamba layers **seeded from the context SSM/conv state**.

I'm not proposing a PR for it — flagging it as a third concrete data point for what a shared diffusion-generate interface has to support:

- **LLaDA** (this PR): full masked-diffusion, no KV cache
- **DiffusionGemma** (#1402): block-diffusion
- **TwoTower**: block-diffusion **+** per-step time conditioning **+** reuse of *both* attention KV *and* recurrent (Mamba/SSM) state across committed blocks

So beyond block size / steps / confidence threshold, the interface needs to let a model carry **arbitrary per-block state (KV *and* recurrent)** forward between committed blocks. Two fragments make the requirement concrete (condensed from our `nemotron_twotower.py`):

**1. The block loop — the frozen context cache holds KV *and* Mamba state, is seeded into each denoiser step, and is *extended in place* by the committed block:**

```python
ctx_cache = self._new_context_cache()          # per-layer KVCache (attn) + ArraysCache (Mamba conv/ssm)
self.context_tower(input_ids, cache=ctx_cache)  # prefill once

for _ in range(num_blocks):
    seeds = self._extract_seeds(ctx_cache)       # per-layer: KV[:offset]  OR  (conv, ssm) state
    xt = mx.full((B, block_size), mask_token_id, dtype=mx.int32)
    for step in range(steps_per_block):
        t = (xt == mask_token_id).astype(mx.float32).mean()   # diffusion time = mask fraction
        logits = self._denoiser_step(xt, seeds, t)            # adaLN(t)-conditioned denoiser forward
        xt = self._commit(xt, logits, ...)                    # confidence-based unmask
    out_ids = mx.concatenate([out_ids, xt], axis=1)
    self.context_tower(xt, cache=ctx_cache)      # extend: advances Mamba state AND appends KV, in place
```

**2. Seeding a *recurrent* layer from carried state — the part a KV-only interface can't express** (a fresh cache is seeded with the context tower's conv/SSM state, so one block chunk-scan resumes exactly from the committed boundary):

```python
def _denoiser_mamba(self, mixer, x, conv_seed, ssm_seed):
    c = ArraysCache(2)
    c[0], c[1] = conv_seed, ssm_seed     # seed conv + SSM state from the context tower
    return mixer(x, mask=None, cache=c)  # one-shot block chunk-scan resuming from that state
```

**Dependency / what's next:** this stays behind the interface design. Once #1391 gets a maintainer steer and an interface lands (whatever #1402 converges to), we'd slim this LLaDA PR to conform and follow with TwoTower as a conforming model port — its backbone already loads on the existing `nemotron_h` path. Happy to share the full TwoTower generate loop as a reference.

*(Scoping note: on Apple Silicon this is a correctness/parity port, not a throughput win — flagging so it isn't read as a perf proposal.)*

pierre427:
Requesting review of the generation interface and sequence-state semantics in this PR.

mlx-lm’s generation path assumes that each step appends one token and advances a monotonic cache. LLaDA instead begins with a partially masked sequence and repeatedly changes multiple positions. This PR therefore adds both the model and a separate diffusion-generation path rather than expressing the sampler as an autoregressive special case.

The implementation reproduces the established masked-diffusion method and also includes an optional order-aware refinement procedure. The main question is whether the resulting interface captures the common structure of diffusion generation or still embeds assumptions specific to LLaDA.

The points that seem most important to examine are:

- which parts of the masking, confidence, remasking, and update schedule belong to the model and which belong in a shared generator;
- whether fixed-seed comparisons against the reference implementation show equivalent transition behaviour despite differences in random-number generation;
- whether attention masks, stopping conditions, and partially resolved sequences remain well-defined at every step;
- whether `remask_refine` is sufficiently separated from the reference sampler and evaluated as an optional modification;
- whether incremental or cached execution preserves the intended denoising process rather than introducing stale sequence state;
- whether the interface can support other block-diffusion and masked-diffusion models without adding model-family branches to the main generation loop;
- whether existing autoregressive generation remains unchanged by the new abstractions.

The model itself follows direct prior work. The part needing the closest review is the boundary between model-specific denoising rules and a reusable non-autoregressive generation contract for mlx-lm.

PR #1491, opened 2026-07-07
Fix float32 promotion in BatchKVCache/BatchRotatingKVCache extend()

ethanelasky:
When a new sequence is inserted into a running batch, `extend()` creates the placeholder for the empty side with `mx.array([]).reshape(...)`, which is float32. The `mx.concatenate` right after then promotes the whole merged K/V buffer to float32.

That doubles KV memory for the batch, and it also makes generation depend on scheduling: whether a given request runs attention against bf16 or fp32 K/V depends on whether another request happened to join its batch mid-flight. For most models the numeric difference is negligible, but I hit this with a model with large activation outliers where it flips entire generations — same prompt, temp 0, different output depending on batch insertion order. Took a while to track down.

Repro:

```python
import mlx.core as mx
from mlx_lm.models.cache import BatchKVCache, KVCache

c = KVCache()
kv = mx.ones((1, 8, 5, 64), mx.bfloat16)
c.update_and_fetch(kv, kv)
batch = BatchKVCache.merge([c])
print(batch.keys.dtype)  # bfloat16

batch.extend(BatchKVCache.merge([KVCache()]))  # a new sequence joins the batch
print(batch.keys.dtype)  # float32 before this fix
```

The fix creates the placeholder with the dtype of the non-empty side, which is how `merge()` and `ArraysCache.extend()` already handle it (the both-None case returns early, so one side always has keys). Same change in `BatchRotatingKVCache.extend`.

Added a regression test in `tests/test_prompt_cache.py` covering both cache classes and both extend directions; it fails on current main. `test_prompt_cache.py` and `test_generate.py` both pass with the fix.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1059,6 +1059,7 @@ class BatchKVCache(_BaseCache):
             B, H, L2, D = other.keys.shape
             M = other.values.shape[3]
         max_size = max(L1, L2)
+        dtype = (self.keys if self.keys is not None else other.keys).dtype
 
         # Pad the keys and values so they are right-justified
         # with the index and the same size
@@ -1066,8 +1067,8 @@ class BatchKVCache(_BaseCache):
             k, v = c.keys, c.values
             if k is None:
                 Bc = c.offset.shape[0]
-                k = mx.array([]).reshape(Bc, H, 0, D)
-                v = mx.array([]).reshape(Bc, H, 0, M)
+                k = mx.zeros((Bc, H, 0, D), dtype=dtype)
+                v = mx.zeros((Bc, H, 0, M), dtype=dtype)
             left = max_idx - c._idx
             right = max_size - k.shape[2] - left
             if right < 0:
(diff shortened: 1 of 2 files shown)


mloiterman:
Independent confirmation, with measurements of how much this costs in practice. Hardware: M1 Ultra 128 GB; mlx 0.32.0; mlx-lm at `ed1fca4` (v0.31.3) and current `main` (the two `pad()` placeholders are unchanged). Model: `mlx-community/Qwen3.6-35B-A3B-4bit` (head_dim 256), `prefill_step_size=4096`, `prefill_batch_size=8`.

**It is not only 2× KV bytes — the whole prefill attention transient doubles.** With head_dim 256 the prefill goes through the unfused SDPA fallback (`use_fallback` returns true for 256 at query length > 8), so fp32 K/V makes the `[B, H, N, K]` scores and softmax fp32 as well. Bare `BatchGenerator`, 4 × 24,884-token prompts, per-step `mx.get_peak_memory()` over the resident weights:

| insertion pattern | peak transient | wall |
|---|---|---|
| all 4 inserted before step 0 (no `extend`) | 24.3 GB | 99 s |
| one inserted per `next()` (the normal case when requests arrive during a prefill) — current main | **57.1 GB** | 192 s |
| same, with this PR's diff applied | **29.3 GB** | 152 s |

Same shape on `Qwen3.5-0.8B-4bit`, 4 × 16,384: 17.5 GB → 9.4 GB (lockstep 9.3 GB). Bisected by building the identical ragged batch through `merge()` instead of `extend()`: `merge()`-built costs the lockstep number; the only differing attribute of the `extend()`-built batch is `keys.dtype == float32`.

**It propagates into the prompt cache.** `extract()` keeps the dtype, so `LRUPromptCache` entries stored from a batch that ever had a join are fp32 — measured 1.07 GB per 24,918-token entry vs 0.575 GB after the fix — and `merge()` then infers fp32 from those entries on a cache hit, so batches that never had a join also run fp32.

**In a serving process built on `BatchGenerator`** (4 concurrent 24,884-token requests): process peak +52 GB → +27 GB, per-request wall 164 s → 130 s, decode throughput on batches that had a join 24–26 tok/s → 32–37 tok/s.

**Output determinism:** confirms the scheduling dependence described above — 4 concurrent 12k-token prompts at temperature 0, all 8 completions (cold pass and prompt-cache-hit pass) differ between current `main` and this diff.

We applied this diff verbatim locally and it holds; a regression test of the same shape as the one here fails on `main` for both classes.

zcbenz (approved):

PR #1496, opened 2026-07-07
Fix LRUPromptCache: return one-token prefix matches and refresh LRU recency on fetch

### LINKED ISSUE #1495: LRUPromptCache: one-token prefixes never match in PromptTrie.search; eviction ignores fetch recency (FIFO, not LRU)
*All code links and repro results below reference `main` @ [`2ed2231`](https://github.com/ml-explore/mlx-lm/commit/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5).*

## Summary

Two defects in the server's `LRUPromptCache` cause avoidable cache misses and incorrect eviction order. Neither appears to produce wrong text by itself, but both make cached prefixes unavailable when they should be reused.

## Defect 1 — `PromptTrie.search` never returns a one-token prefix match

[`PromptTrie.search`](https://github.com/ml-explore/mlx-lm/blob/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5/mlx_lm/models/cache.py#L1578-L1603):

```python
shorter = None
if last_index > 0:
    shorter = tokens[: last_index + 1]
```

`last_index` is the *index* of the deepest node holding a value (`-1` = none), so a stored one-token key matches at `last_index == 0` and the condition should be `>= 0`. Repro:

```python
from mlx_lm.models.cache import PromptTrie
trie = PromptTrie()
trie.add("m", [7], "entry")
print(trie.search("m", [7, 8, 9]))
# PromptTrieResult(model='m', exact=None, shorter=None, longer=[7], common_prefix=1)
```

`shorter` should be `[7]`. Instead, the entry is returned as `longer=[7]`, even though `PromptTrieResult.longer` is documented as a value that extends beyond the request tokens.

For trimmable caches this is mostly hidden: the longer branch deep-copies the entry and trims zero tokens. For non-trimmable caches the entry becomes unreachable — the longer branch is skipped, `shorter` is `None`, and `fetch_nearest_cache` returns `(None, tokens)`, a full recompute of a cached prefix. Non-trimmable (hybrid/recurrent) models are exactly the ones that depend on the shorter path (cf. #980):

```python
class NotTrimmable(KVCache):
    def is_trimmable(self):
        return False

lru = LRUPromptCache(max_size=10)
lru.insert_cache("m", [7], [NotTrimmable()])
print(lru.fetch_nearest_cache("m", [7, 8, 9]))  # (None, [7, 8, 9]) — miss
```

## Defect 2 — fetches never refresh recency, so eviction is FIFO, not LRU

[`fetch_nearest_cache`](https://github.com/ml-explore/mlx-lm/blob/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5/mlx_lm/models/cache.py#L1674-L1694) returns exact/shorter/longer hits without ever touching `self._lru`, and [`CacheOrder`](https://github.com/ml-explore/mlx-lm/blob/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5/mlx_lm/models/cache.py#L1630-L1657) has no touch operation at all (only `push`/`remove`/`pop`). Position in the deque is fixed at insert time, so eviction is insertion-order within the type priority. Repro:

```python
lru = LRUPromptCache(max_size=2)
lru.insert_cache("m", [1], [KVCache()])         # A
lru.insert_cache("m", [2], [KVCache()])         # B
for _ in range(5):
    lru.fetch_nearest_cache("m", [1])           # A is hot
lru.insert_cache("m", [3], [KVCache()])         # C -> evicts...
print(lru.fetch_nearest_cache("m", [1])[0])     # None  — hot A evicted
print(lru.fetch_nearest_cache("m", [2])[0])     # cache — cold B survives
```

On a busy server with a small `--prompt-cache-size`, frequently used shared prefixes (e.g. a common system prompt) can be evicted before colder entries. A fix would refresh recency (`remove` + `push`) on every hit, including shorter/longer partial hits.

## Why the existing tests catch neither

[`MockCache`](https://github.com/ml-explore/mlx-lm/blob/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5/tests/test_server.py#L75-L92) in `tests/test_server.py` has a no-op `trim()` that always claims success and an `__eq__` comparing only `.value`. As a result, the LRU tests ([`test_lru`](https://github.com/ml-explore/mlx-lm/blob/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5/tests/test_server.py#L571-L628)) exercise trie navigation but cannot detect whether a returned cache's state matches its key (this also hides the mismatched-entry class in #1494). Suggested: give the mock real content semantics — store a token list, have `trim(n)` slice it and return the true count, compare content in `__eq__` — then add the two repros above as regression tests.

## Prior art (checked)

- #1042 (closed, unmerged PR) — related; covered the index-0 match and recency refresh, but was not merged.
- #1078 (merged) — fixed a *different* off-by-one, in `PromptTrie.pop_prefixes()`, not `search()`.
- #1314 / #1409 / #1483 (open PRs) — adjacent cache work; none covers these.

T0mSIlver:
Fixes #1495.

Two small fixes in the server's `LRUPromptCache`, plus a `MockCache` upgrade so the tests can catch this class of bug:

1. **`PromptTrie.search` never returned a one-token prefix match.** `last_index` is the index of the deepest node holding a value (`-1` = none), so a stored one-token key matches at `last_index == 0` and the shorter-prefix check needs `>= 0`, not `> 0`. Before this change the entry leaked out as `longer=[t]` instead, which the longer branch happens to serve for trimmable caches (deepcopy + trim 0) — but for non-trimmable caches the entry was unreachable: the longer branch is skipped and the fetch returned `(None, tokens)`, a full recompute of a cached prefix.

2. **Fetches never refreshed recency, so eviction was FIFO, not LRU.** `CacheOrder` only recorded insertion order and `fetch_nearest_cache` never touched it, so a hot entry (e.g. a shared system prompt) could be evicted while cold entries survived. `fetch_nearest_cache` now refreshes the hit entry's position on exact, shorter, and longer hits via a new `CacheOrder.touch`. The per-type eviction priority is unchanged.

3. **`MockCache` now has real content semantics**: its state is the token list it was computed from, `trim()` actually slices it and returns the true trimmed count, and `__eq__` compares content. The existing `test_lru` assertions now verify that a fetched cache's contents match the tokens it is served under (previously `trim()` was a no-op and `__eq__` compared an opaque value, so trie navigation was tested but key↔state correspondence was not). Existing tests updated accordingly; two regression tests added (`test_lru_one_token_prefix`, `test_lru_fetch_refreshes_recency`).

Both defects and the repros are described in detail in #1495. `python -m unittest tests.test_server.TestLRUPromptCache` passes (8 tests); `black` clean.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1599,7 +1599,7 @@ class PromptTrie:
 
         # Check if we found a prefix at any point
         shorter = None
-        if last_index > 0:
+        if last_index >= 0:
             shorter = tokens[: last_index + 1]
 
         # Check for sequences that are longer
@@ -1646,6 +1646,16 @@ class LRUPromptCache:
                 except ValueError:
                     pass
 
+        def touch(self, model: Any, tokens: List[Any]):
+            for cache_type in self._ordering:
+                lru = self._lrus[cache_type]
+                try:
+                    lru.remove((model, tokens))
+                except ValueError:
+                    continue
+                lru.append((model, tokens))
+                break
+
         def pop(self):
             i = 0
             while i + 1 < len(self._ordering):
@@ -1675,6 +1685,7 @@ class LRUPromptCache:
         result = self._trie.search(model, tokens)
         if result.exact is not None:
             cache_entry = self._trie.get(result.model, result.exact)
+            self._lru.touch(result.model, result.exact)
             return copy.deepcopy(cache_entry.prompt_cache), []
 
         short_length = len(result.shorter) if result.shorter is not None else 0
@@ -1685,10 +1696,12 @@ class LRUPromptCache:
                 prefix = min(len(tokens) - 1, result.common_prefix)
                 num_to_trim = len(result.longer) - prefix
                 trim_prompt_cache(cache, num_to_trim)
+                self._lru.touch(result.model, result.longer)
                 return cache, tokens[prefix:]
 
         if short_length > 0:
             cache_entry = self._trie.get(result.model, result.shorter)
+            self._lru.touch(result.model, result.shorter)
             return copy.deepcopy(cache_entry.prompt_cache), tokens[short_length:]
 
         return None, tokens
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -73,22 +73,29 @@ class DummyModelProvider:
 
 
 class MockCache:
-    def __init__(self, value, is_trimmable: bool = True):
-        self.value = value
+    """A mock KV cache whose state is the list of tokens it was computed from,
+    so tests can check that a fetched cache's contents match the tokens it is
+    keyed and served under."""
+
+    def __init__(self, tokens, is_trimmable: bool = True):
+        self.tokens = list(tokens)
         self._is_trimmable = is_trimmable
 
     @property
     def nbytes(self):
-        return len(self.value)
+        return len(self.tokens)
 
     def __eq__(self, other):
-        return other.value == self.value
+        return other.tokens == self.tokens
 
     def is_trimmable(self):
         return self._is_trimmable
 
     def trim(self, n):
         assert self._is_trimmable
+        n = min(n, len(self.tokens))
+        if n > 0:
+            self.tokens = self.tokens[:-n]
         return n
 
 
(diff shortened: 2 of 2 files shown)


rajanshxrma (approved):
Following up from #1495 — reviewed the diff and verified this branch independently.

**Verification:**
- Full `TestLRUPromptCache` passes locally (8/8).
- The new tests are genuine regression coverage: with the `cache.py` changes reverted and the new tests kept, `test_lru_one_token_prefix` and `test_lru_fetch_refreshes_recency` both fail at their documented assertions, and pass again with the fix restored.
- `black --check` is clean on both files.

**Notes on the changes:**
- `last_index >= 0`: exact matches are still caught earlier by `last_index == len(tokens) - 1 >= 0`, and an empty-token query returns before the walk, so widening the shorter branch adds only the previously dropped one-token case. It also has a nice side effect: when a one-token entry and a longer trimmable entry offer the same reusable length (`common_prefix == short_length`), the fetch now serves the smaller entry directly instead of deep-copying and trimming the longer one.
- `touch()` mirrors `remove()`'s tier scan, so recency refreshes within an entry's own tier and `pop()`'s per-type eviction priority is preserved. All three hit paths touch the key of the entry actually served — including `result.longer` on the trim path, which stays the correct stored key since trimming happens on a deepcopy.
- The `MockCache` upgrade makes the `test_lru` assertions verify real key/content correspondence — e.g. the post-trim `MockCache([])` and `MockCache([1])` expectations weren't expressible with the old opaque-value mock.

LGTM — approving as a non-maintainer; happy to defer to the maintainers on anything further.

zcbenz:

Issue #1497, opened 2026-07-08
N-gram self-speculative decoding for hybrid GDN/SA models


dahai80:
## Motivation

Single-token decode on Apple Silicon is **bandwidth-bound**: each step streams the full model weights for one token. Speculative decoding can beat this ceiling by verifying multiple draft tokens per weight load — but standard speculative decoding needs a separate draft model (extra weights, extra memory).

**N-gram self-speculative decoding** generates draft tokens on the **CPU** from an n-gram table built from the prompt + generated tokens, with **zero GPU overhead**. The model's own forward pass verifies the draft. On a hit, one weight load yields 2 tokens; on a miss, we fall back to 1 token with cache rollback. No draft model required.

This is especially effective for **hybrid GDN/SA architectures** (e.g. Qwen3.5/3.6), where the GatedDeltaNet (GDN) layers have O(1) recurrent state that must be checkpointed and rolled back on draft rejection — KV-cache-only speculative decode would corrupt GDN state.

## Proposal

Add an opt-in n-gram self-speculative decode path:

1. **`NgramDraftTable`** (`generate.py`) — CPU-side n-gram table (trigram → bigram → unigram fallback). O(1) hash probe, zero GPU cost. Built from the prompt, rolling-updated on each accepted token.

2. **`ngram_speculative_generate_step()`** (`generate.py`) — self-speculative decode generator. S=2 verification: forward pass produces 1–2 tokens per step. On acceptance: yield 2; on rejection: yield 1 + rollback cache. CLI flags `--ngram-spec` and `--ngram-n` (default n=3).

3. **`ArraysCache.checkpoint()` / `rollback()` / `trim()`** (`cache.py`) — generic cache state save/restore for GDN recurrent state during speculative verification. `trim` is a no-op for `ArraysCache` (state-based, not offset-based like `KVCache`); rollback is used instead. ~18 lines, no behavior change for existing caches.

4. **`GatedDeltaNet._conv1d_decode_multi()`** (`qwen3_5.py`) — vectorized S>1 decode path for the conv1d in GDN layers, using `mx.stack` + batched multiply instead of a sequential Python loop. Dispatched only when S>1 and a cache exists (the speculative verification step).

## Correctness

- S=2 verification: the model's own logits verify the draft token, so output is **identical to greedy decoding** (verified: ngram-spec output matches baseline token-for-token).
[...]

PR #1498, opened 2026-07-08
feat: n-gram self-speculative decoding for hybrid GDN/SA models

dahai80:
Refs #1497.

## Summary

Add **n-gram self-speculative decoding**: generate draft tokens on CPU from an n-gram table (zero GPU overhead), then verify them with the model's own forward pass. On a hit, one weight load yields 2 tokens; on a miss, fall back to 1 token with cache rollback. No separate draft model required.

This beats the single-token bandwidth ceiling on Apple Silicon and is especially effective for **hybrid GDN/SA architectures** (e.g. Qwen3.5/3.6), where GatedDeltaNet layers have O(1) recurrent state that must be checkpointed and rolled back on draft rejection.

## Changes

**`mlx_lm/generate.py`** (+301)
- `NgramDraftTable` — CPU-side n-gram table (trigram → bigram → unigram fallback). O(1) hash probe, zero GPU cost. Built from the prompt, rolling-updated on each accepted token.
- `ngram_speculative_generate_step()` — self-speculative decode generator. S=2 verification: forward pass produces 1–2 tokens per step. On acceptance: yield 2; on rejection: yield 1 + rollback cache. Draft hit-rate tracking.
- CLI flags `--ngram-spec` and `--ngram-n` (default n=3).

**`mlx_lm/models/cache.py`** (+18)
- `ArraysCache.checkpoint()` / `rollback()` / `trim()` / `is_trimmable()` — generic save/restore of cache state for speculative verification. `trim` is a no-op for `ArraysCache` (state-based, not offset-based like `KVCache`); rollback is used instead. No behavior change for existing caches when speculative decode is not used.

**`mlx_lm/models/qwen3_5.py`** (+279)
- `GatedDeltaNet._conv1d_decode_multi()` — vectorized S>1 decode path for the conv1d in GDN layers, using `mx.stack` + batched multiply instead of a sequential Python loop. Dispatched only when S>1 and a cache exists (the speculative verification step). `_conv1d_decode` (S=1) factored out symmetrically.

## Correctness

- **S=2 verification**: the model's own logits verify the draft token, so output is identical to greedy decoding (verified ngram-spec output matches baseline token-for-token on Qwen3.6-27B).
- **Cache rollback on rejection**: `ArraysCache.rollback()` restores GDN recurrent state + KV cache to the pre-draft checkpoint, so the next step proceeds correctly from the rejected position.
- **Opt-in**: the fast path is gated behind `--ngram-spec`; without it, `generate`/`generate_step`/`batch_generate`/`stream_generate` behavior is unchanged.

## Performance

Measured on Qwen3.6-27B (hybrid: 48 GDN + 16 SA layers) on M2 Ultra 137GB:

| Mode | Speed | Notes |
|------|-------|-------|
| Baseline (S=1) | 44.6 tok/s | bandwidth-bound |
| N-gram spec (1 draft) | 52.1 tok/s | **+17%**, 44% draft hit rate |
| N-gram spec (2 drafts) | 34.4 tok/s | 0.77× — S=3 overhead exceeds gain |

The 1-draft (S=2) configuration gives a reliable speedup; higher draft counts are counterproductive due to quadratic multi-token forward overhead on bandwidth-bound models.

**Note on quantization:** the benchmark used `quant2-all` (an experimental 2-bit mixed recipe, see #1466). The +17% is **architecture-level** — it reduces GPU forward passes per token by verifying CPU-generated drafts, independent of weight precision. On standard 4-bit / bf16 the speedup should reproduce (lower-precision models are *more* bandwidth-bound, so the relative gain should hold or grow). I have not yet re-measured on standard upstream quant formats; happy to rerun on request.

## Testing

- Static verification: `import mlx_lm` and `from mlx_lm.generate import NgramDraftTable, ngram_speculative_generate_step` succeed on a clean checkout of `upstream/main` + this commit (cherry-picked cleanly, no conflicts).
- GDN `__call__` three-branch dispatch (S=1 / S>1 / prefill) verified self-consistent with upstream's existing prefill path and `cache.lengths` usage.
- I was unable to run `tests/test_generate.py` locally (Hugging Face Hub unreachable in my environment); relying on upstream CI to confirm no regression in `generate`/`batch_generate`/`stream_generate`. The opt-in gate means existing tests exercise only the unchanged path.

## Relationship to prior work

- #851 (n-gram hashing) — related but distinct: this PR implements full self-speculative **verification with GDN recurrent-state rollback** for hybrid architectures, where the hard part is safely checkpointing/rolling back GDN O(1) state (not just KV cache).
- #1450 (decode performance omnibus) — this is one focused item split out for a clean trail.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -36,7 +36,9 @@ from .models.cache import (
     QuantizedKVCache,
     RotatingKVCache,
     TokenBuffer,
+    can_trim_prompt_cache,
     load_prompt_cache,
+    trim_prompt_cache,
 )
 from .sample_utils import make_sampler
 from .tokenizer_utils import TokenizerWrapper
@@ -217,6 +219,17 @@ def setup_arg_parser():
         "--num-draft-tokens",
         type=int,
         help="Number of tokens to draft when using speculative decoding.",
+        default=3,
+    )
+    parser.add_argument(
+        "--ngram-spec",
+        action="store_true",
+        help="Use n-gram self-speculative decoding (no draft model needed).",
+    )
+    parser.add_argument(
+        "--ngram-n",
+        type=int,
+        help="N-gram order for self-speculative decoding (default: 3).",
         default=3,
     )
     return parser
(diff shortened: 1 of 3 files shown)


dahai80:
Friendly nudge for review when you have a moment.

This adds n-gram self-speculative decoding for hybrid GDN/SA models (e.g. Qwen3.6), measured at +17% (44.6 → 52.1 tok/s on Qwen3.6-27B). The gain is architecture-level: n-gram drafts run on CPU, so fewer GPU forwards are needed, which helps beat the bandwidth-bound single-token decode ceiling. It should reproduce on standard 4bit/bf16 (even more bandwidth-bound) — happy to rerun benchmarks on a standard quant if that would help the review.

The branch is a clean cherry-pick onto current `main` (zero conflicts), touching only `generate.py` / `cache.py` / `qwen3_5.py`. CI needs maintainer approval to run (external contributor) — let me know if there's anything I can adjust.

zcbenz:

PR #1499, opened 2026-07-08
Add layerwise DWQ (--layers-per-round) for very large students

avlp12:
### Problem

DWQ trains the quantization `scales`/`biases` of **every** quantized layer in a single backward pass. For very large students the first training step runs out of memory regardless of sequence length, because the quantized-matmul backward materializes large per-layer intermediates for all trainable layers at once.

Concrete case (M3 Ultra 512 GB): a 242 GB / 745B-parameter MoE student (256 experts, 75 quantized MoE layers) fails in the first step with `[METAL] Command buffer execution failed: Insufficient Memory` at `--max-seq-length` 1024, 512, and 256, with `--grad-checkpoint` and `--batch-size 1`. Pipeline-sharding the student across two machines does not help either: `Send` has no VJP (`[Primitive::vjp] Not implemented for Send`), and even forward-only distributed passes previously hit the GPU watchdog on cross-node recv-waits.

### Approach

`dwq_quantize_layerwise` trains rounds of at most `--layers-per-round` layers at a time:

- collect the layer indices that contain quantized (affine, `bits < 8`) modules; embeddings/head train with the first round,
- rounds run **deepest layers first** (their gradients are the best-conditioned; shallow-first ordering diverged in our runs),
- per round: freeze everything, unfreeze that round's `scales`/`biases`, train one pass over the data with a fresh optimizer,
- **validate after each round and revert the round if validation got worse** (monotone by construction).

Since `value_and_grad` only differentiates the unfrozen parameters, backward memory is bounded by the round size. The default (`--layers-per-round 0`) keeps the current single-pass behavior unchanged.

### Results

Full-scale run on the failing case above (2.56-bpw GLM-5.2 student, 78 quantized layers, `--layers-per-round 8`, seq 512, 145 samples, lr 1e-6, single M3 Ultra 512 GB):

- **Peak memory ~334 GB** (the same configuration OOMs in the first step without layerwise rounds).
- **10/10 rounds accepted** (the rollback never had to fire); training-set KL fell monotonically.
- Held-out validation KL **0.458 → 0.322 (−30%)**.
- Downstream 3-corpus KL vs a 4.5-bpw reference of the same base model: overall **0.655 → 0.379 (−42%)**, top-1 flip 24.4% → 15.9%; the worst language slice (ZH) improved **0.987 → 0.562 (−43%)**.

~12 s/iteration; the full 10-round run took ≈5 h.

### Testing

- `tests/test_dwq.py`: end-to-end layerwise rounds on a tiny quantized Llama (runs in <1 s).
- Full-scale run on the 745B MoE student described above.

### DIFF
--- a/mlx_lm/quant/dwq.py
+++ b/mlx_lm/quant/dwq.py
@@ -2,6 +2,7 @@
 
 import argparse
 import copy
+import re
 import time
 import types
 from pathlib import Path
@@ -17,13 +18,8 @@ from mlx_lm.tuner.datasets import load_dataset
 from mlx_lm.tuner.losses import kl_div_loss
 from mlx_lm.tuner.trainer import grad_checkpoint, iterate_batches
 from mlx_lm.tuner.utils import print_trainable_parameters
-from mlx_lm.utils import (
-    load,
-    load_tokenizer,
-    quantize_model,
-    save,
-    sharded_load,
-)
+from mlx_lm.utils import (load, load_tokenizer, quantize_model, save,
+                          sharded_load)
 
 
 def compute_dwq_targets(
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1501, opened 2026-07-08
Text-based state machine

angeloskath:
The token based SequenceStateMachine has a design flaw that makes it impossible to identify the state change because substrings can be encoded in different ways.

This replaces it with TextStateMachine so we switch state on the actual string and not the tokens. It also introduces a StopSequenceMatcher that can be used to make sure that generation stops on specific token sequences.

This fixes #1373, #1447, #1406, #1336, #1160 .

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -10,16 +10,7 @@ import time
 from collections import deque
 from dataclasses import dataclass
 from functools import partial
-from typing import (
-    Any,
-    Callable,
-    Generator,
-    List,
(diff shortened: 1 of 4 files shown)


nastya236:
Beautiful fix, thank you! Splitting it into two machines makes a lot of sense: one on decoded text for stripping thinking / tool-calling, one on token ids to stop generation. Nice buffering approach to ensure that we find a full match.

nastya236 (approved):

Issue #1494, opened 2026-07-07
Prompt-cache reuse can return KV state that doesn't match the keyed prefix (windowed-cache trim contract)


T0mSIlver:
*All code links and repro results below reference `main` @ [`2ed2231`](https://github.com/ml-explore/mlx-lm/commit/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5).*

## Summary

`mlx_lm.server`'s prompt-cache reuse (`LRUPromptCache.fetch_nearest_cache`) can return KV state that no longer represents the token prefix it is keyed under.

The reuse path assumes that:

1. a stored cache's KV state corresponds exactly to its token key; and
2. `is_trimmable() == True` means `trim(n)` can exactly remove the requested suffix.

`KVCache` satisfies both. `ChunkedKVCache` (llama4-family chunked attention) and `ConcatenateKVCache` do not, and nothing verifies either assumption at reuse time. On a warm server, a longer cached prompt can then be trimmed and reused for a shorter request even though the returned KV no longer covers that request's prefix — silently wrong output instead of a fallback to recompute (Defect A). The server's segment checkpointing then stores the mismatched state under the new request's key, so subsequent exact hits reuse the bad entry instead of recomputing (Defect B). Defect C is a related trim-contract problem in `trim_prompt_cache` itself.

## Reproduce

```bash
pip install mlx-lm  # tested at main @ 2ed2231
python repro_prompt_cache_reuse.py  # exits 1 while the bugs are present, 0 once fixed
```

The script uses only cache/reuse-layer code and does not load model weights. It prefills 1024 position-tagged tokens through a `[KVCache, ChunkedKVCache(chunk_size=512)]` layer pair (llama4's `attention_chunk_size=8192`, scaled down; window sliding driven exactly as [`llama4.py` does](https://github.com/ml-explore/mlx-lm/blob/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5/mlx_lm/models/llama4.py#L227-L232)), stores that cache in an `LRUPromptCache`, then issues a request sharing only the first 32 tokens. Checks are labeled by defect: A1, A2, B1, C1–C3.

<details>
<summary><code>repro_prompt_cache_reuse.py</code></summary>

```python
# Repro: prompt-cache reuse serves KV that doesn't match the keyed prefix
# when a ChunkedKVCache window has slid (llama4-style chunked attention).
#
[...]

PR #1502, opened 2026-07-08
Prompt cache: never reuse KV that fails to trim to the keyed prefix

T0mSIlver:
Part 1 of 2 addressing #1494. This PR is the general trim/reuse correctness guards; the follow-up (#1503) adds the slid-window prefix guard on top.

`mlx_lm.server`'s prompt-cache prefix reuse (`LRUPromptCache.fetch_nearest_cache`) can serve KV state that no longer matches the token prefix it is keyed under, producing silently wrong output instead of falling back to recompute. This PR adds three defensive guards so a partial or lossy trim can never be reused:

1. **`trim_prompt_cache` returns the minimum tokens trimmed across all layers**, not layer 0's count. With mixed cache types (e.g. a full `KVCache` beside a windowed `ChunkedKVCache`), layer 0 can trim the full requested amount while another layer cannot. Returning layer 0 alone hides the shortfall.
2. **The `fetch_nearest_cache` longer branch only reuses when the trim removed the full requested amount** (`trim_prompt_cache(cache, n) == n`); otherwise it falls back to a shorter hit or recompute. This is co-dependent with (1): the guard is only meaningful once the return value reflects every layer.
3. **`ConcatenateKVCache.trim` now slices the stored `keys`/`values`** instead of only decrementing `offset`. Previously `update_and_fetch` recomputed `offset` from `keys.shape[-2]`, so a later update resurrected the "trimmed" tokens and their KV — a standalone bug independent of the reuse path.

### Tests
`TestPromptCacheReuseContract` in `tests/test_prompt_cache.py`:
- a slid `ChunkedKVCache` sharing a short prefix falls back to recompute (the min + trim-count guard catch the under-trim);
- a synthetic partial-trim cache is refused;
- `ConcatenateKVCache.trim` drops the KV so a subsequent update cannot resurrect it.

### Note
This PR closes the specific repro in #1494. There is a further case — a prefix that lands inside a slid window but behind the chunk boundary, where the trim count *does* fit — that the follow-up PR closes with a fetch-scoped guard. Splitting so the general guards can land independently.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -104,11 +104,13 @@ def trim_prompt_cache(cache: List[Any], num_tokens: int) -> List[Any]:
         num_tokens (int): The number of tokens to trim.
 
     Returns:
-        (int): The number of tokens that were trimmed.
+        (int): The number of tokens that were trimmed from every layer. If
+        this is less than ``num_tokens`` the trim was incomplete and the cache
+        should not be reused as an exact prefix.
     """
     if not can_trim_prompt_cache(cache) or len(cache) == 0:
         return 0
-    return [c.trim(num_tokens) for c in cache][0]
+    return min(c.trim(num_tokens) for c in cache)
 
 
 def create_attention_mask(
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1503, opened 2026-07-08
Prompt cache: don't reuse a slid windowed cache as a keyed prefix

T0mSIlver:
Part 2 of 2 addressing #1494. **Stacks on #1502** — until that merges this PR shows both commits; review the second commit ("Don't reuse a slid windowed cache as a keyed prefix"), or merge #1502 first and GitHub will narrow this to the delta.

#1502's trim-count guard is necessary but not sufficient. Even when every layer trims the *full* requested amount, a `ChunkedKVCache` whose window has slid (`start_position > 0`) can still be wrong to reuse: if the shared prefix `P` lands inside the retained window but behind the current attention-chunk boundary, the trimmed window `[start_position, P)` is missing the leading in-chunk tokens `[0, start_position)` that a fresh prefill of `P` tokens would attend to (llama4 builds its chunk mask from `start = cache[0].start_position`). The trim count matches, so #1502 alone would reuse it and emit silently wrong output.

This PR refuses the trim-based longer-branch reuse when any layer reports `start_position > 0` (new `_cache_window_has_slid`), falling back to a shorter hit or recompute.

### Why a fetch-scoped guard and not `ChunkedKVCache.is_trimmable()`
Making `is_trimmable()` return `start_position == 0` looks simpler but is unsafe: `is_trimmable`/`trim` are also on the speculative-decoding rewind path (`generate.py` `_rewind_cache`), which checks trimmability once at startup and then calls `trim_prompt_cache` on every draft rejection **ignoring the return value**. Once a llama4 chunked window slides mid-generation, a global `is_trimmable() -> False` would make those rewinds silently trim nothing and leave rejected KV resident — corrupting output in exactly the model family this is meant to protect. A slid chunked cache *can* still do the small suffix rewinds spec decoding needs; only arbitrary-prefix reuse is unsafe. Keeping the guard inside `fetch_nearest_cache` leaves the global trim/rewind contract untouched. (This also relates to #1437, which moves `RotatingKVCache` in the trimmable-after-wrap direction; a rotating ring keeps its logical prefix recoverable, a slid chunked window does not.)

### Tests
Added to `TestPromptCacheReuseContract`:
- an end-to-end slid `ChunkedKVCache` where the prefix is behind the chunk boundary and the trim fully succeeds — must recompute;
- a structural case (a fake slid cache) that passes the trim-count guard but is refused.

### Known follow-up (not in this PR)
`insert_cache` still evicts shorter prefix entries when a slid cache is inserted (it passes `can_trim_prompt_cache`), so fetch then recomputes where a shorter entry could have helped. Correctness is unaffected; this is a reuse-efficiency improvement for a later change.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -104,11 +104,13 @@ def trim_prompt_cache(cache: List[Any], num_tokens: int) -> List[Any]:
         num_tokens (int): The number of tokens to trim.
 
     Returns:
-        (int): The number of tokens that were trimmed.
+        (int): The number of tokens that were trimmed from every layer. If
+        this is less than ``num_tokens`` the trim was incomplete and the cache
+        should not be reused as an exact prefix.
     """
     if not can_trim_prompt_cache(cache) or len(cache) == 0:
         return 0
-    return [c.trim(num_tokens) for c in cache][0]
+    return min(c.trim(num_tokens) for c in cache)
 
 
 def create_attention_mask(
@@ -213,6 +215,9 @@ class ConcatenateKVCache(_BaseCache):
 
     def trim(self, n):
         n = min(self.offset, n)
+        if n > 0 and self.keys is not None:
+            self.keys = self.keys[..., : self.offset - n, :]
+            self.values = self.values[..., : self.offset - n, :]
         self.offset -= n
         return n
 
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1508, opened 2026-07-09
Prompt-lookup decoding (draft-free speculative decoding)

pierre427:
## What

Adds an opt-in, **draft-free** speculative decoder: instead of a second draft
model, a pluggable proposer suggests a continuation of the running sequence from
the prompt/generation history (tail n-gram or suffix-automaton retrieval), and
the target verifies the whole proposal in one batched forward. No draft model, no
extra weights, no training.

- New module `mlx_lm/prompt_lookup.py` — the proposers (`SuffixAutomaton`,
  `NgramProposer`, `SuffixAutomatonProposer`, `make_proposer`) and a small stats
  object. Pure Python, no mlx dependency.
- New `prompt_lookup_generate_step(...)` in `mlx_lm/generate.py` — the
  verify/rewind loop. It follows the shape of `speculative_generate_step`
  (prompt, model, sampler, prompt_cache, yields per-token) but is a separate
  library entry point; it does not yet support `logits_processors` or the
  kv-quantization kwargs (unknown kwargs raise), and it is not wired into
  `stream_generate`/CLI in this PR — that can follow if the interface is
  agreeable.

## Why it matters / when you benefit

On copy-heavy generation — code edits and refactors, structured/JSON output, long
shared context, "repeat/transform this" tasks — the next tokens frequently already
exist in the context, so retrieval proposals get accepted in long runs, collapsing
many bandwidth-bound target forwards into one. It is architecture-agnostic (it
rides the model's own KV path). On free-form prose there is nothing to retrieve;
an optional one-way `adaptive` latch detects that (accepted-fraction below a gate
after warmup) and falls back to plain `generate_step` for the tail, so worst case
is ≈ plain decoding (never-lose).

This fills a real gap: mlx-lm today only has model-draft speculation
(`--draft-model`); there is no draft-free path.

## Provenance

Prompt-lookup decoding (Apoorv Saxena, 2023; the `transformers`
`prompt_lookup_num_tokens` feature). Established technique — the contribution is
the MLX implementation, the suffix-automaton proposer, and the never-lose latch.

## How it works

At each step the proposer emits up to `num_draft` tokens; the target runs once
over `[pending] + proposal`; each position is accepted iff it equals the target's
own sample at that position (speculative-sampling acceptance for a deterministic
drafter), stopping at the first mismatch. On a partial reject the cache is rewound
to the pre-proposal state. **Losslessness:** the emitted sequence equals the
target's own (batched) greedy/sampled decode — verified by
`test_matches_target_batched_greedy`. The `prompt_cache` is left representing
exactly `prompt + emitted` at every stop boundary, so callers that persist/reuse
it (e.g. an LRU prompt cache) stay correct.

**Cache support / scope.** Snapshots/rewinds are handled for `KVCache`,
`RotatingKVCache` (buffer-copy, safe past the window), any cache exposing
`offset`/`trim` (e.g. `QuantizedKVCache` — untested beyond the offset/trim
contract), and `CacheList` of those. Recurrent (`ArraysCache` / SSM) caches are
**not** supported — their state cannot be rewound to an earlier token after a
rejected proposal — and raise a clear error.

## Measured

Standalone, greedy, output **identical** to plain decode:
- Qwen1.5-0.5B-Chat-4bit, copy-heavy prompt: **1.87×** (589 → 1103 tok/s) —
  reproducible with the included test model.
- Larger targets amortize the fixed proposer cost better (a tiny fast target is
  the worst case for it); in our internal runs, retrieval-heavy code editing on
  30B–105B-class models reached ~2.0–2.7×, and novel prose stayed at parity via
  the latch.

## Tests

`tests/test_prompt_lookup.py` — 13 tests: 8 fully offline (no weights), 5 in
one class that loads Qwen1.5-0.5B-Chat-4bit (one download, shared `setUpClass`):
- proposer logic; cache-offset helpers; `ArraysCache` rejection;
- a tiny synthetic `CacheList(KVCache, KVCache)` model runs end-to-end and leaves
  the cache exact (no weights);
- on Qwen1.5-0.5B: every backend runs and leaves the cache exact, **output equals
  the target's own batched greedy** (losslessness), `max_tokens` boundary is
  exact, a reused cache's base is preserved, and the adaptive latch runs.

`black`/`isort` clean (pinned 25.1.0 / 6.0.0). Default behavior of every existing
API is unchanged (this is a new, separate entry point).

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -867,11 +867,8 @@ def prompt_lookup_generate_step(
             if prop:
                 stats.retrieval_cycles += 1
                 stats.retrieval_proposed += len(prop)
-                stats.retrieval_accepted += n_acc
-                stats.bonus_tokens += 1
             else:
                 stats.plain_cycles += 1
-                stats.plain_tokens += 1
 
             if prop and n_acc < len(prop):
                 _pld_rewind(prompt_cache, snaps)
@@ -880,6 +877,16 @@ def prompt_lookup_generate_step(
                 pending = [emit[-1][0]]
 
             for tok, row, from_draft in emit:
+                # Count delivered tokens at the yield boundary. The caller may
+                # stop on EOS in the middle of an accepted verify batch; eager
+                # accounting would then overstate retrieval/output telemetry.
+                if prop:
+                    if from_draft:
+                        stats.retrieval_accepted += 1
+                    else:
+                        stats.bonus_tokens += 1
+                else:
+                    stats.plain_tokens += 1
                 seq.append(tok)
                 history_seq.append(tok)
                 proposer.observe(tok)
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -38,6 +38,7 @@ from .models.cache import (
     TokenBuffer,
     load_prompt_cache,
 )
+from .prompt_lookup import NgramProposer, PromptLookupStats, make_proposer
 from .sample_utils import make_sampler
 from .tokenizer_utils import TokenizerWrapper
 from .utils import does_model_support_input_embeddings, load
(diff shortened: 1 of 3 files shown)


pierre427:
Pushed one addition: `NgramProposer` now bounds its backward scan (`max_lookback=4096`, `0` = unbounded). On novel text the miss case walked the full sequence per verify cycle for every n-gram size — quadratic over a long generation at long context. Proposals are verified against the target anyway, so the bound is lossless (worst case: fewer proposals). Test included; 15/15 on the branch.

PR #1509, opened 2026-07-09
deepseek_v2: absorbed-MLA decode (cache the latent, like deepseek_v3)

pierre427:
## What

Make `deepseek_v2`'s attention cache the **MLA latent** (`kv_lora_rank` +
`qk_rope_head_dim`) instead of the decompressed per-head K/V, folding `kv_b_proj`
into `embed_q`/`unembed_out` so decode runs in latent space. This is the **same
absorbed-MLA pattern `deepseek_v3` and `glm4_moe_lite` already use upstream**;
`deepseek_v2` is the flagship MLA family still materializing K/V (`minicpm3` and
`youtu_llm` also materialize — the same recipe applies there if this lands).

Uses the existing `mla.MultiLinear` helper (already imported by `deepseek_v3`).
Prefill stays materialized (a fully-fused absorbed prefill is slower); only the
`L == 1` decode path runs absorbed — the same approach as
`deepseek_v3.Attention`, with one deliberate difference: the forward
transparently handles a `QuantizedKVCache` latent, **preserving `deepseek_v2`'s
existing `--kv-bits` support** (which works today on the materialized path, and
which the absorbed `deepseek_v3` currently lacks — it errors with a quantized
cache). That is why this file keeps two attention call sites instead of v3's
one; the quantized branches are inert on a normal cache.

## Why it matters / when you benefit

Decode is memory-bandwidth-bound; the cached latent is ~8.9× smaller per token
than the materialized per-head K/V on V2-Lite (16 heads: 5120 vs 576 elements
per token per layer) — and ~71× on full DeepSeek-V2 (128 heads) — so the dominant
KV read shrinks. The win is a **long-context** one and grows with context length.

## Measured (DeepSeek-V2-Lite-Chat-4bit, isolated A/B — only `deepseek_v2.py` changed)

| context | materialized | absorbed | decode speedup | peak mem saved |
|---|---|---|---|---|
| 4k  | 112.1 tok/s / 12.4 GB | 111.0 / 11.0 | 0.99× | 1.4 GB |
| 8k  | 79.7 / 15.4 | 100.4 / 14.1 | 1.26× | 1.3 GB |
| 16k | 53.9 / 25.9 | 87.8 / 20.6 | **1.63×** | 5.3 GB |
| 32k | 31.5 / 59.6 | 75.2 / 49.8 | **2.39×** | 9.8 GB |

So: a tie at short context, **~2.4× decode and ~10 GB less peak memory at 32k**,
growing with length. Short-context users are unaffected (the split keeps prefill
materialized).

## Correctness — read this carefully

Absorbed MLA is *mathematically* equivalent to materialized, but **not
bitwise-identical on quantized weights**, and this PR does not claim otherwise.
On the 4-bit model, greedy output diverges from the materialized path after a few
tokens (both remain correct and coherent) because the fused `embed_q`/`unembed_out`
projections are **requantized** and the matmuls are reordered. Measured at the
first position: **logit correlation 0.9998**, top-5 token ids identical, only a
top-2 near-tie swaps. This is the **same numerical tradeoff `deepseek_v3` already
ships** (it is absorbed too); on bf16 weights the two are much closer. If exact
bit-parity with the current materialized path matters to you, this is the
expected, documented difference.

## Provenance

The MLA absorption identity is from the DeepSeek-V2 paper (arXiv:2405.04434 §2.1)
and the official DeepSeek inference code; the implementation mirrors the absorbed
`deepseek_v3` already in this repo. Faithful MLX port, not new research.

## Tests

- The existing offline `tests/test_models.py::test_deepseek_v2` (tiny synthetic
  config, no weights) now exercises the absorbed `embed_q`/`unembed_out` path at
  both prefill (`L>1`) and decode (`L==1`) and passes — no new test needed.
- `bench_absorbed_mla.py` reproduces the real-weight A/B (decode tok/s + peak GB
  at {4,8,16,32}k, and the logit correlation) by running it on this branch vs a
  pristine upstream checkout.

One migration note: prompt-cache snapshots saved with the old materialized layout
are shape-incompatible with the latent cache (the same one-time cutover
`deepseek_v3` went through).

### DIFF
--- a/PORT_RECIPE.md
+++ b/PORT_RECIPE.md
@@ -1,0 +1,55 @@
+# PR A — deepseek_v2 absorbed-MLA decode (port recipe)
+
+**Status: staged, GPU-verification pending** (the 48/48 bitwise gate needs
+DeepSeek-V2-Lite loaded; a GPU session is currently in use).
+
+## The finding that defines this PR
+
+- Upstream **`deepseek_v3`** (deepseek_v3.py:145) and **`glm4_moe_lite`**
+  (glm4_moe_lite.py:151) **already cache the latent** — `cache.update_and_fetch(
+  kv_latent, k_pe)` — i.e. they are already absorbed-MLA.
+- Upstream **`deepseek_v2`** (deepseek_v2.py:219–237) is the **only** MLA model
+  left that decompresses via `kv_b_proj` and caches full per-head K/V.
+
+So this PR is a **consistency fix**: bring `deepseek_v2` in line with the
+absorbed pattern upstream already ships in `deepseek_v3`. The technique is already
+accepted upstream (it's in v3), so review friction is low; the payoff is the
+lossless long-context win v2 currently leaves on the table.
+
+## The port (mirror deepseek_v3's `Attention.__call__`)
+
+Template = `deepseek_v3.py:118–170` (captured). Apply the same structure to
+`deepseek_v2.py::Attention`:
+
+1. **Projections:** add the absorbed projections v3 uses — `embed_q` (folds
+   `W_UK` into the query path) and `unembed_out` (folds `W_UV` into output).
+   v2 keeps `kv_b_proj` weights on disk; absorb them at load into these (see how
+   v3 builds `embed_q`/`unembed_out`; v2 has no `q_lora_rank`, so the q path is
+   just `q_proj`).
+2. **Cache the latent:** replace the decompress-then-cache block with
+   `kv_latent = self.kv_a_layernorm(compressed_kv)` → `cache.update_and_fetch(
+   kv_latent, k_pe)` (cache 512-dim latent + rope key, not per-head K/V).
+3. **Decode fast path (`L == 1`):** `q_nope = self.embed_q(q_nope)`,
+   `k = v = kv_latent`, run SDPA in latent space, then `output =
+   self.unembed_out(output)`. This is the absorbed decode path — the win.
+4. **Prefill (`L > 1`):** materialize `k = embed_q(kv_latent, transpose=False)`,
+   `v = unembed_out(kv_latent)` exactly as v3 — keeps prefill cheap (do NOT run
+   the fully-fused absorbed prefill; it is 3.3× slower, per our measurements).
+5. **`sanitize`:** keep the `kv_b_proj` → `embed_q`/`unembed_out` absorption in
+   the model's weight loading, mirroring v3.
+
+Net: v2's `Attention` becomes v3's `Attention` modulo the q-lora difference.
+Do NOT add quantized-latent KV here (memory-only, measured slower — separate).
+
+## Verification (GPU, pending)
+
+`bench_absorbed_mla_v2.py` (to write): load
+`mlx-community/DeepSeek-V2-Lite-Chat-4bit-mlx`, greedy(48) on the absorbed port
+vs a materialized reference, assert **48/48 bitwise-identical**, and record the
+decode-tok/s + KV-GB at {8k,32k,64k} for the PR table (expected ~3–7× decode,
+~2.5× KV vs cold, per experiments/deepseek-v2-lite-results.md).
+
+## Test (offline, to add)
+
+`tests/test_models.py`: a tiny synthetic deepseek_v2 config forward-shape check
+(absorbed path produces correct shapes at L==1 and L>1) — CPU, no weights.
--- a/PR_BODY.md
+++ b/PR_BODY.md
@@ -1,0 +1,72 @@
+# deepseek_v2: absorbed-MLA decode (cache the latent, like deepseek_v3)
+
+## What
+
+Make `deepseek_v2`'s attention cache the **MLA latent** (`kv_lora_rank` +
+`qk_rope_head_dim`) instead of the decompressed per-head K/V, folding `kv_b_proj`
+into `embed_q`/`unembed_out` so decode runs in latent space. This is the **same
+absorbed-MLA pattern `deepseek_v3` and `glm4_moe_lite` already use upstream**;
+`deepseek_v2` is the flagship MLA family still materializing K/V (`minicpm3` and
+`youtu_llm` also materialize — the same recipe applies there if this lands).
+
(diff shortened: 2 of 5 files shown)


pierre427:
Pushed a small defensive fix (`0cd3860`): the absorbed-MLA path now rejects **asymmetric key/value KV-cache bits** with a clear `NotImplementedError` instead of crashing deep inside `mx.quantized_matmul`.

**Why it's needed.** Absorbed MLA caches the compressed latent once and reuses it as *both* the attention key (`q_nope · latent`) and the value (`softmax @ latent`), and the rope key rides in the cache's value slot. So there is no independent value tensor to carry a distinct `value_bits`. If a cache does expose `key_bits ≠ value_bits`, the generic quantized SDPA dequantizes the *key-quantized* latent using `value_bits`, and you get a packed-weight-vs-scales shape mismatch (e.g. `w.shape == (1,1,4096,16)` against `scales.shape == (1,1,4096,1)` at `group_size=64, bits=4`) — cryptic and far from the cause.

**Scope.** The guard is `getattr(cache, "key_bits", None) != getattr(cache, "value_bits", None)`, so it's completely inert for this PR on its own (a plain `QuantizedKVCache` has no such attributes). It only fires when this path is composed with the asymmetric-K/V-bits cache from #1550 — which is exactly where I hit it while testing the two together.

**Validated** on `DeepSeek-V2-Lite-Chat-4bit` (2048-token teacher-forced ppl): symmetric `--kv-bits` and Hadamard-rotated configs are unchanged (fp16 14.72, K8/V8 14.71, K8/V8-rotated 14.68, K4/V4-rotated 16.84); the asymmetric config now raises the clear message rather than the shape crash.

Cross-ref #1550: asymmetric bits remain fully supported on the standard (non-absorbed) attention path, which has genuinely separate K and V tensors — this only scopes out the absorbed path where the distinction is undefined.

rohash123:
Thanks for putting this together. I opened a smaller overlapping PR (#1715) before noticing this one, so I’m going to close mine as superseded.

One small observation from that branch may still be useful.

For quantized `kv_b_proj` checkpoints, this PR needs to split `kv_b_proj` into two pieces:

- the K-side piece becomes `embed_q`
- the V-side piece becomes `unembed_out`

The K-side piece has to be transposed before it can become `embed_q`. For 4-bit weights, that means:

`dequantize -> transpose -> quantize again`

That final requantization can slightly change the attention scores.

I tested a variant that avoids requantizing K:

- K-side: `dequantize -> transpose -> keep fp16` for `embed_q`
- V-side: slice the original packed 4-bit weights and keep them quantized for `unembed_out`

This is still compatible with the existing code because `embed_q` can be a normal dense `MultiLinear`, while `unembed_out` remains a `QuantizedMultiLinear`.

In a small `DeepSeek-V2-Lite-Chat-4bit-mlx` greedy smoke test:

Prompt:
`The capital of France is`

Generated tokens, `max_tokens=64`:

- materialized upstream path: starts `[8913, 13, 185, 549, ...]` -> `"Paris.\nThe official..."`
- fully requantized absorbed path: starts `[8913, 11, 588, 317, ...]` -> `"Paris, which is also..."`
- dense-K absorbed variant: starts `[8913, 13, 185, 549, ...]` -> `"Paris.\nThe official..."`

So in this smoke:

- fully requantized absorbed path diverged from upstream at generated token **2** (`13` / `"."` vs `11` / `","`)
- dense-K absorbed variant matched upstream for all **64 / 64** generated tokens

This is only one smoke test, not a full correctness proof. The tradeoff is that dense `embed_q` costs more memory/compute, so I’m not sure it should be the default here. But it may be worth considering as a correctness-oriented option, or at least documenting when discussing the expected quantized-output drift.

pierre427:
Thanks — this reproduced, and I've adopted your variant. I ported the dense-K fold onto this branch and ran a three-way comparison on `DeepSeek-V2-Lite-Chat-4bit-mlx` (M5, mlx 0.32.0): upstream materialized `main`, this PR's requantized fold, and your dense-K fold (fp16 `embed_q`, packed-row slice for `unembed_out`).

Beyond greedy smokes, I teacher-forced upstream's greedy continuations through the absorbed decode path (8 prompts × 64 steps = 512 single-token steps, identical cache state), so the variants are scored on identical inputs rather than on diverged trajectories:

| vs upstream, 512 forced decode steps | requantized | dense-K |
|---|---|---|
| top-1 disagreements | 17 (3.3%) | 4 (0.8%) |
| mean abs Δlogprob of the forced token | 0.037 | 0.014 |
| largest upstream top-2 gap flipped | 0.42 nats | 0.094 nats |

Two observations worth putting on the record:

1. The requantized fold's flips are not just tie-breaking noise — it flips tokens where upstream's top-2 gap was up to 0.42 nats. Your diagnosis holds: the transpose regroups the quantization blocks, so a requantized `embed_q` no longer round-trips to the checkpoint weights. (The packed V-side row slice is bit-exact by construction; the new test pins both properties.)
2. The 64/64 token-exact match doesn't generalize, though. On your exact smoke prompt in my environment, the dense-K path flips the same near-tie at generated token 2 (`.` vs `,` behind a 0.016-nat gap) that the requantized path does. All four residual dense-K disagreements sit on gaps ≤ 0.094 nats — that's the float-reassociation of absorbed attention itself (`(q·Wk)·c` vs `q·(Wk·c)`, rope scores applied as an additive mask), which no weight handling removes, and the same tradeoff `deepseek_v3` already ships. So dense-K removes the requantization error, and token-exactness on any single prompt stays environment-dependent.

On cost, dense-K measured *faster* at decode than the requantized fold here (148 vs 122 tok/s, short context, single-pass numbers; an M=1 per-head fp16 GEMM beats the quantized matmul at these shapes), at +41 MB on V2-Lite. The memory cost scales as `heads × kv_lora_rank × qk_nope_head_dim × layers`, so order ~0.5–1 GB on 236B-class V2 — still small next to the weights.

Given strictly better fidelity and no speed penalty, I made it the default rather than an option: 53bb7ee switches the quantized fold to your scheme and adds the sanitize contract test (adapted from your #1715 — thanks for that as well).

PR #1514, opened 2026-07-10
Fail fast when the generation thread stops

dogukanveziroglu:
If the generation thread crashes, the HTTP server currently stays up but can no longer answer completion requests. This change makes requests fail quickly, returns 503 from /health, and closes active streams with an error so a supervisor can safely restart the process.

Addresses #1505 and complements #1513


dogukanveziroglu:
I guess this is not a priority so I am closing this PR for pushing another one.

PR #1515, opened 2026-07-10
Anchor-stride prefix reuse for non-trimmable prompt caches

pierre427:
### What

`LRUPromptCache` (the request-level prefix cache) can reuse a stored longer prompt across a divergent tail only through its **trim** path, which requires `can_trim_prompt_cache`. For **non-trimmable** caches that is `False`, so `fetch_nearest_cache` finds no reusable shorter entry and the request re-prefills the entire shared prefix from scratch.

Non-trimmable covers a large, widely-used class of models:
- recurrent / hybrid architectures whose caches are `ArraysCache` + Mamba/GDN state (e.g. the Qwen3-Next / Qwen3.5-MoE family, Nemotron-H), and
- a sliding-window `RotatingKVCache` once it is outside its window.

These models are locked out of request-level prefix reuse entirely today: a multi-turn or branched conversation re-prefills the whole shared prefix every request.

### The fix

Store exact **shorter-prefix snapshots** so the existing shorter-match lookup can serve them. Retrieval is unchanged — the trie already returns the longest stored prefix of a query — the only gap is that nothing stores the intermediate prefixes for non-trimmable caches.

- **`PrefixAnchorCapture`** — a `prompt_progress_callback(processed, total)`-compatible callable. During the prefill the caller is already doing, it deep-copies the cache once its processed length has advanced at least `stride` past the previous anchor, recording a `PrefixAnchor(length, prompt_cache)`. No extra forward passes — the capture is amortized into the one prefill. It is a **no-op for trimmable caches and for `stride <= 0`** (`enabled == False`), so wiring it is safe on any model.
- **`LRUPromptCache.insert_anchors`** — stores captured anchors as ordinary shorter entries, skipping any trimmable snapshot (the trim path already covers those). `fetch_nearest_cache` then finds the nearest anchor below a divergent request's shared prefix and returns it plus the remaining tail.

### Correctness

Anchors are **exact** prefixes of the original prompt, so continuing from an anchor is next-token-identical to a cold full prefill — no trimming, no approximation. The trimmable no-op gate means trimmable models keep their existing behavior exactly (they already reuse the full prefix via `trim_prompt_cache`, and anchors would only take space).

### Scope of this PR (please read)

This PR adds the **library primitive + tests + a benchmark**. It intentionally does **not** wire the server flag yet. Reason: the target non-trimmable models are *batchable*, so the server runs them through `BatchGenerator`, whose per-request prefill does not currently expose the snapshot hook `PrefixAnchorCapture` needs. Wiring a `--prompt-cache-anchor-stride` flag only onto the non-batched path would advertise a feature that doesn't apply to the models it targets. The server wiring lands in a follow-up PR:
- non-batched path: pass `PrefixAnchorCapture` as (part of) `prompt_progress_callback`, then call `insert_anchors` after generation;
- batched path: add an optional prefill-boundary snapshot hook to `BatchGenerator`.

This PR is split out so the primitive can be reviewed on its own.

### Benchmark

`benchmarks/bench_prompt_cache_anchor_stride.py` compares cold full prefill vs upstream `LRUPromptCache` vs anchor-stride on a shared-prefix-then-divergent request pair, gating every arm on next-token equivalence.

The default synthetic mode runs on CPU with no download: a tiny non-trimmable recurrent model (`ArraysCache` per layer), prefilled through `generate_step` with `PrefixAnchorCapture` as the progress callback — the intended production wiring. With a 600-token shared prefix, stride 128, and divergent tails:

```
reused tokens   : upstream=0   anchor=512   (nearest anchor <= 600)
next-token equiv: upstream=True anchor=True
```

Upstream reuses **0** tokens across the divergent tail (the non-trimmable longer-match path is skipped); anchor mode reuses **512** (the nearest stride boundary ≤ the shared prefix) and the continuation is next-token-identical to a cold full prefill. The tiny-model TTFT deltas are illustrative of the mechanism only — real prefill-time wins should be measured with `--model` against an actual recurrent/hybrid checkpoint (the script supports it).

Prior measurement against the real `LRUPromptCache` (separate harness, non-trimmable models): upstream reused 0 tokens across the divergent tail (≈ cold), while anchor-stride cut TTFT 4.6×–24.9× @4k on `nemotron_h` and 3.3×–10.6× @4k on `qwen3_5_moe`. Numbers are prefill/TTFT only — this is not a decode-throughput change.

### Memory tradeoff

Each anchor is a cache snapshot, bounded by `max_anchors` (default 32) per prompt and by the existing `LRUPromptCache` byte budget. For recurrent caches the state is constant-size regardless of prefix length, so anchors are cheap; for hybrid caches the attention component grows, so the byte budget does the bounding. Anchors are only stored for non-trimmable caches.

### Tests

`tests/test_prompt_cache.py`: capture boundaries, the trimmable no-op gate, end-to-end reuse on a non-trimmable cache (upstream reuses 0; anchor mode reuses the nearest anchor and the continuation is an exact prefix), and the trimmable-skip path. `black` and `isort --profile black` clean.

### DIFF
--- a/benchmarks/bench_prompt_cache_anchor_stride.py
+++ b/benchmarks/bench_prompt_cache_anchor_stride.py
@@ -1,0 +1,489 @@
+"""Benchmark the prompt-cache prefix-anchor-stride API (upstream PR).
+
+Exercises the NEW public API in ``mlx_lm.models.cache``:
+
+  * ``PrefixAnchorCapture(cache, stride)`` -- a
+    ``prompt_progress_callback(processed, total)``-compatible callable wired into
+    ``generate_step``. It deep-copies the live cache at stride boundaries during
+    the prefill the server is doing anyway, recording ``PrefixAnchor`` snapshots.
+  * ``LRUPromptCache.insert_anchors(model, tokens, anchors)`` -- stores each
+    snapshot as an ordinary *shorter* trie entry. Retrieval is unchanged:
+    ``fetch_nearest_cache`` finds the longest stored shorter prefix + tail.
+
+The scenario that distinguishes the anchor API from plain upstream
+``LRUPromptCache``: a long SHARED prefix followed by DIVERGENT tails --
+``req_a = P + tail_a``, ``req_b = P + tail_b``. Upstream's ``fetch_nearest_cache``
+can reuse ``P`` for ``req_b`` only by *trimming* the stored ``req_a`` cache back
+to ``|P|``; for a NON-TRIMMABLE cache (recurrent/hybrid models whose caches
+report ``can_trim_prompt_cache == False``, e.g. ``ArraysCache``) that path is
+skipped, so ``req_b`` re-prefills from scratch (reused == 0). Anchor-stride
+pre-saved an exact shorter snapshot at each stride boundary of ``req_a``, so
+``req_b`` reuses up to the nearest anchor ``<= |P|``.
+
+Two modes:
+
+  synthetic (default, CPU, no download)
+      Builds a genuinely tiny NON-TRIMMABLE recurrent nn model whose
+      ``make_prompt_cache`` yields ``ArraysCache`` per layer, and drives a real
+      prefill through ``mlx_lm.generate.generate_step`` with a
+      ``PrefixAnchorCapture`` wired as ``prompt_progress_callback`` (the intended
+      usage). Proves: upstream reuses 0 across the divergent tail; anchor mode
+      reuses the nearest-anchor prefix; and the anchor-reused continuation is
+      next-token-IDENTICAL to a cold full prefill.
+
+  --model PATH (optional, real recurrent/hybrid model)
+      Mirrors ``bench_prefix_cache_vs_upstream.py``: cold full prefill of req_b
+      vs upstream ``LRUPromptCache`` reuse vs anchor-stride reuse, reporting
+      reused-tokens and prefill/TTFT ms per arm, each gated on next-token
+      equivalence vs cold. The anchor arm captures during req_a's prefill (not a
+      separate manual re-prefill). Prefill/TTFT is reported separately from
+      decode throughput; no decode-token/s win is claimed.
+
+  # synthetic (default)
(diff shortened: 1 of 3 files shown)


pierre427:
Requesting review of the checkpoint exactness and prefix-matching rules in this PR.

Recurrent and hybrid caches cannot generally be shortened to an arbitrary token boundary. This implementation records exact intermediate states at selected prefill anchors, then allows the prefix trie to resume from the nearest anchor preceding the point where a new prompt diverges.

The anchors are captured during the original prefill rather than by running additional forward passes. The reported 4K-prefix measurements show TTFT reductions of 4.6–24.9× on Nemotron-H and 3.3–10.6× on Qwen3.5 MoE.

- whether an anchor’s identity includes every input that affects recurrent state, including tokenization, positions, model configuration, dtype, and quantization;
- whether the trie can ever select an anchor beyond the last exactly matching token;
- whether recurrent state and ordinary KV layers remain aligned at the same logical position after resumption;
- whether captured MLX arrays are fully materialized and isolated from later mutation by the request that created them;
- whether repeated reuse, trimming, and re-anchoring remain equivalent to uninterrupted full prefill;
- whether exactness checks cover long continuations, chunked prefill, quantized models, and multiple hybrid architectures;
- how anchor stride should be chosen when checkpoint memory, lookup probability, and recomputation distance are considered together;
- whether the TTFT measurements include checkpoint lookup, state restoration, and any copy or materialization costs.

Intermediate recurrent-state checkpoints have precedent elsewhere. The question for mlx-lm is whether restartable state should become an explicit cache capability, with exact-prefix requirements defined by the cache implementation rather than inferred by the server.

pierre427:
Superseded by #1596, which reaches the same landings at strictly lower cost: instead of deep-copying the whole cache at stride positions, it checkpoints only what cannot be re-derived — recurrent state (context-independent, ~77 MB on an 80B) and the bounded sliding window — inside the existing LRU entry, with the batch path handled natively (per-lane histories survive lane churn) rather than via the separate capture in #1517. Closing this series in its favor.

PR #1534, opened 2026-07-10
Add olmo_hils: HiLS-Attention hierarchical landmark sparse attention (OLMo3 backbone)

pierre427:
Add support for HiLS-Attention models (`olmo_hils`), starting with
[tencent/HiLS-Attention-7B](https://huggingface.co/tencent/HiLS-Attention-7B)
(arXiv [2607.02980](https://arxiv.org/abs/2607.02980)) — an OLMo3-7B
continued-trained with chunk-wise sparse attention learned end-to-end. The
architecture cannot be expressed with the existing attention/mask helpers, so
this is a self-contained model file plus a consistency test.

### How the architecture works

- A **landmark token** (id = vocab_size, embedded via a standalone
  `lmk_embed` parameter) follows every `chunk_size-1` (63) real tokens,
  forming 64-token chunks. Landmarks are masked as attention targets
  everywhere; their hidden states act as chunk-retrieval queries.
- 3 of every 4 layers are **sliding-window attention** (512, landmarks
  masked). Every 4th layer is **HiLS attention**: per-chunk landmark keys are
  attention-pooled from the chunk's keys, each query retrieves its top-32
  chunks per head, and the output fuses per-chunk attention with the window
  branch through a softmax over `[chunk scores + prior bias, window LSE]`.
  HiLS layers use HoPE partial RoPE (long-period frequency pairs zeroed).

### Implementation notes

- **Landmark insertion is internal to `Model.__call__`**: it takes real
  token ids and returns logits aligned to them, so `generate`, chunked
  prefill at any split points, and the server work unchanged.
- **Caches expose real-token `offset`/`trim()`** over landmark-inserted
  storage (the invariant: position `p` is a landmark iff `(p+1) % 64 == 0`),
  so `trim_prompt_cache` and speculative-decoding rollback compose without
  knowing about landmarks. Trimming also rolls back the pooled landmark-key
  cache.
- **Sliding-window layers keep only the window band** plus a rollback
  margin (fail-closed: a trim that would truncate the re-decode window
  raises rather than silently degrading). KV for those 24 layers is
  O(window); only the 8 HiLS layers retain full KV, which retrieval needs.
- Decode gathers just the selected chunks, so per-token attention cost is
  bounded (`topk*chunk_size + window` keys) at any context length; prefill
  uses an equivalent dense weight-scatter formulation. The two paths are
  chosen by KV volume moved.
- `save_prompt_cache`/`max_kv_size`/`input_embeddings`/batch>1 are not
  supported and fail with explicit errors (landmark bookkeeping is
  per-sequence; HiLS layers must retain full KV).

### mlx 0.31.x note

`quantized_matmul` on mlx 0.31.x (the current minimum pin) returns wrong
results for **unaligned N** (this model's vocab is 100278, not a multiple of
32) when `rows >= 10 and rows % 32 != 0` — all bit widths; fixed on mlx
main. Single-token decode (`rows == 1`) is unaffected, but batched-verify
logits (speculative decoding) and mid-batch prefill logits are corrupted.
The model pads lm_head rows to a multiple of 32, gated on `mlx < 0.32`.
Minimal repro:

```python
import mlx.core as mx
W = mx.random.normal((100278, 4096)).astype(mx.bfloat16) * 0.02
wq, s, b = mx.quantize(W, group_size=64, bits=6)
x = mx.random.normal((1, 4096)).astype(mx.bfloat16)
r1 = mx.quantized_matmul(x, wq, s, b, transpose=True, group_size=64, bits=6)
X = mx.broadcast_to(x, (10, 4096))
r10 = mx.quantized_matmul(X, wq, s, b, transpose=True, group_size=64, bits=6)
print(mx.abs(r10 - r1).max())  # ~7-15 on 0.31.2; ~0 on main
```

### Validation (M5, q6 unless noted)

- Coherent generation; needle retrieval at 6.6k context (needle far outside
  the window, i.e. through learned chunk retrieval) passes in bf16 and q6.
- Decode is near-flat with context: 48.6 t/s @8k → 50.6 t/s @32k (bf16:
  28.6 → 25.5 over 1k→32k); peak memory 18.3 GB @32k.
- Consistency test (in `tests/test_models.py`): a single full forward vs
  chunked prefill across boundary-crossing splits vs token-by-token decode
  vs trim-and-redecode agree on a small fp32 model.
- Works with prompt-lookup / speculative-style verify batches: batched
  verify logits match single-token decode (0 argmax flips at fp32; bf16
  matches to kernel noise).

Base model, no chat template (no instruct variant released yet).

### DIFF
--- a/mlx_lm/models/olmo_hils.py
+++ b/mlx_lm/models/olmo_hils.py
@@ -1,0 +1,791 @@
+# Copyright © 2026 Apple Inc.
+#
+# HiLS-Attention (Hierarchical Landmark Sparse attention) on an OLMo3
+# backbone — port of tencent/HiLS-Attention-7B (arXiv:2607.02980,
+# github.com/Tencent-Hunyuan/HiLS-Attention).
+#
+# Structure (values for the 7B checkpoint in parentheses):
+#   * A landmark token (id = vocab_size, embedded via the standalone
+#     `lmk_embed` parameter) is inserted after every chunk_size-1 (63)
+#     real tokens. Landmarks share the position id of the following real
+#     token, are masked out as attention targets everywhere, and their
+#     hidden states act as chunk-retrieval queries.
+#   * Layers with idx % full_attn_interleave == interleave-1 (every 4th)
+#     use HiLS attention; the rest are sliding-window (512) attention
+#     with landmark keys masked.
+#   * HiLS layers factorize attention into a sliding-window part and a
+#     top-k (32) retrieved-chunk part, fused by a softmax over
+#     [per-chunk retrieval score + Taylor prior bias, window LSE].
+#     Per-chunk landmark keys are attention-pooled from the chunk's keys
+#     against the landmark query at the chunk's last position.
+#   * HiLS layers use HoPE in-range RoPE: standard rope frequencies with
+#     every pair whose period exceeds rope_context_length /
+#     rope_period_multiplier zeroed (NoPE tail).
+#
+# Landmark insertion is handled inside `Model.__call__`: it takes real
+# token ids and returns logits aligned to them, so the standard mlx-lm
+# generation loop works unchanged. Batch size must be 1.
+#
+# Cache coordinates: the caches store keys/values in landmark-INSERTED
+# coordinates but expose `offset`/`trim()` in REAL-token units, so
+# generic cache machinery (prompt-lookup speculative decoding rollback,
+# trim_prompt_cache, telemetry) composes without knowing about
+# landmarks. Sliding-window layers keep only the window band (plus a
+# rollback margin) — memory for those 24 layers is O(window), not O(T).
+# Decode cost is bounded (topk*chunk_size + window keys per token) at
+# any context length.
+
+import math
+from dataclasses import dataclass
+from typing import Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs
+from .cache import _BaseCache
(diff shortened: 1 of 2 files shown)


pierre427:
Requesting review of the positional mapping and cache semantics in this HiLS implementation.

HiLS inserts landmark representations into the stored sequence and later uses them to select sparse regions of the history. As a result, logical token positions and physical cache offsets are no longer interchangeable. The implementation has to preserve that mapping through prefill, landmark insertion, sparse selection, sliding-window attention, trimming, and continued decoding.

The current tests include long-context generation and retrieval checks, with decode remaining approximately flat between 8K and 32K context in the reported measurements.

- whether the logical-token to stored-position mapping remains correct across chunked prefill and single-token decoding;
- whether positional encoding is based on the intended logical or physical coordinate at each stage;
- whether sparse gathers and dense scatters preserve the same ordering assumed by the attention mask;
- whether selected landmarks can expose future information or omit required local-window tokens at boundary conditions;
- whether cache trimming and subsequent decoding reconstruct the same offsets as an uninterrupted run;
- whether quantization or tied landmark scores can change selection in ways that need an explicit stability rule;
- whether the retrieval tests exercise transitions between landmark regions rather than only examples that remain inside one selected span;
- whether this architecture requires a general cache interface for non-unit storage growth, rather than model-specific offset corrections.

HiLS itself comes from direct prior work. The difficult question for mlx-lm is how attention and cache APIs should represent architectures where one input token does not correspond to one stored cache position.

PR #1536, opened 2026-07-10
Support Nemotron Puzzle heterogeneous MoE (nemotron_h_puzzle)

### LINKED ISSUE #1506: Support Nemotron Puzzle 75B / nemotron_h_puzzle heterogeneous MoE models

pierre427:
Closes #1506.

NVIDIA's Nemotron Puzzle checkpoints (`model_type: nemotron_h_puzzle`, e.g. `NVIDIA-Nemotron-Labs-3-Puzzle-75B-A9B`) run Puzzle-NAS over the existing Nemotron-H hybrid backbone. Unlike plain `nemotron_h`, each MoE layer has its own `moe_intermediate_size` and `num_experts_per_tok`, listed in a per-layer `block_configs` array. The loader assumed uniform MoE dims, so conversion failed with `uniform(): incompatible function arguments` and the model type was unsupported.

**Changes**
- `ModelArgs` gains `block_configs`; a shallow per-layer `_moe_layer_args` view sizes each MoE block from its own entry. Non-MoE layers and the uniform path are unaffected (`block_configs=None` -> identity).
- `MoEGate` defaults `n_group`/`topk_group`/`routed_scaling_factor` to the identity when absent -- Puzzle configs carry no group routing, which otherwise tripped `None > 1` in `group_expert_select`.
- Remap `nemotron_h_puzzle -> nemotron_h`.

The rest of the architecture (latent MoE projections, `e_score_correction_bias` routing, shared expert, Mamba/attention hybrid) is already handled by `nemotron_h`. Experts are stored pre-stacked, so `sanitize()` is unchanged.

**Testing** -- added `test_nemotron_h_puzzle` (heterogeneous dims + forward/decode); verified end-to-end on a 6-bit conversion of the 75B checkpoint (88 layers, all 1471 tensors load, coherent generation); confirmed the uniform `nemotron_h` path is unchanged.

### DIFF
--- a/mlx_lm/models/nemotron_h.py
+++ b/mlx_lm/models/nemotron_h.py
@@ -561,6 +561,16 @@ class Model(nn.Module):
                 caches.append(KVCache())
         return caches
 
+    @property
+    def quant_predicate(self):
+        if self.model_type != "nemotron_h_puzzle":
+            return lambda _path, _module: True
+        # Puzzle's 131k-token output projection is unusually sensitive to
+        # low-bit affine quantization. Quantizing it at 4 bits produced
+        # unrelated and repetitive generations while the same checkpoint
+        # generated correctly with the BF16 head restored.
+        return lambda path, _: path != "lm_head"
+
     def sanitize(self, weights):
         weights = {k: v for (k, v) in weights.items() if not k.startswith("mtp.")}
         for k, v in weights.items():
--- a/mlx_lm/models/nemotron_h.py
+++ b/mlx_lm/models/nemotron_h.py
@@ -1,5 +1,6 @@
-# Copyright © 2025 Apple Inc.
+# Copyright © 2025-2026 Apple Inc.
 
+import copy
 from dataclasses import dataclass
 from functools import partial
 from typing import Any, List, Optional, Tuple
@@ -42,6 +43,10 @@ class ModelArgs(BaseModelArgs):
     use_conv_bias: bool
     hybrid_override_pattern: Optional[List[str]] = None
     layers_block_type: Optional[List[str]] = None
+    # nemotron_h_puzzle (Puzzle-NAS): heterogeneous MoE. Per-layer list aligned
+    # with layers_block_type; MoE entries carry their own `moe_intermediate_size`
+    # and `num_experts_per_tok`. Absent (plain nemotron_h) => uniform dims.
+    block_configs: Optional[List[dict]] = None
     head_dim: Optional[int] = None
     moe_intermediate_size: Optional[int] = None
     moe_shared_expert_intermediate_size: Optional[int] = None
@@ -351,9 +356,12 @@ class MoEGate(nn.Module):
         self.top_k = config.num_experts_per_tok
         self.norm_topk_prob = config.norm_topk_prob
         self.n_routed_experts = config.n_routed_experts
-        self.routed_scaling_factor = config.routed_scaling_factor
-        self.n_group = config.n_group
-        self.topk_group = config.topk_group
+        # Puzzle configs omit group routing / scaling; default to the identity
+        # (no grouping, unit scale) so group_expert_select's `n_group > 1`
+        # guard doesn't hit a `None > 1` TypeError.
+        self.routed_scaling_factor = config.routed_scaling_factor or 1.0
+        self.n_group = config.n_group or 1
+        self.topk_group = config.topk_group or 1
         self.weight = mx.zeros((self.n_routed_experts, config.hidden_size))
         self.e_score_correction_bias = mx.zeros((self.n_routed_experts,))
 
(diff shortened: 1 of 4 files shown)


pierre427:
Cross-posting from #1535 for the record here: I measured the `promote_dt` question on the released 6-bit 75B checkpoint (both dt semantics behind a runtime toggle, same process, same inputs). Result: teacher-forced ppl 7.083 (fp32-promoted, this PR's behavior) vs 7.075 (bf16-native) — a wash; per-layer output cosine between the modes ≥ 0.997 at all 88 layers; greedy generations equally coherent. Details and the on-distribution argument in https://github.com/ml-explore/mlx-lm/pull/1535#issuecomment-4958414724.

So this PR's numerics are not producing degraded outputs. Two things #1535 has that I'd fold in here if this PR is the one that moves forward: `mtp.*` tensor exclusion during AR loading, and keeping the output projection out of low-bit quantization. Happy to go either way on consolidation.

pierre427:
Pushed `77c06c8`: keep the Puzzle **output projection out of low-bit quantization** — a `quant_predicate` that excludes `lm_head` for `nemotron_h_puzzle` only (other variants unchanged), picked up by `quantize_model` at convert time. This matches #1535 and preserves head fidelity.

(The other #1535 improvement — excluding `mtp.*` from ordinary AR loading — is already present in our `sanitize()`.) Still happy to consolidate with #1535; as noted above I deliberately did **not** adopt their `promote_dt` change, since I measured it a wash on the released 75B (ppl 7.083 vs 7.075, per-layer cosine ≥0.997 at every layer).

PR #1538, opened 2026-07-11
Add persistent runtime-quantized LoRA adapters

HongyuChen0926:
## Summary

- Add `QuantizedLoRALinear`, which keeps LoRA A and B matrices packed at runtime and evaluates both projections with `mx.quantized_matmul`.
- Add a versioned persistent adapter format with packed weights, scales, biases, dimensions, rank, padding, and per-layer bit-width metadata.
- Load quantized adapters through the existing `adapter_path` interface without materializing the saved LoRA matrices as full floating-point weights.
- Add `mlx_lm.quantize_adapter`, documentation, numerical tests, and a reproducible benchmark.

## Motivation

MLX-LM already supports a quantized base model with floating-point LoRA weights. This PR adds inference-time quantization for the adapter matrices themselves. The primary goal is adapter memory reduction, especially when adapters are kept separate for dynamic deployment. This PR does not claim a general end-to-end latency improvement.

## Design

- The base `Linear` or `QuantizedLinear` module is unchanged.
- LoRA A and B are independently padded to supported group sizes and stored as packed `uint32` tensors plus affine scales and biases.
- Normal inference remains on the quantized path. Full adapter dequantization occurs only when explicitly fusing the adapter.
- Persistence is atomic and validates format version, layer dimensions, rank, padding, bit width, and every required packed tensor before loading.
- Conversion rejects configurations that increase adapter parameter memory unless `--allow-larger` is specified.
- Linear LoRA layers support 2, 3, 4, 5, 6, and 8-bit affine quantization.

## Validation

- 21 focused unit tests pass.
- A two-layer randomly initialized Llama test compares final logits after Q8 adapter quantization and requires cosine similarity above 0.99999, relative L2 below 0.002, and maximum absolute error below 0.005.
- Direct persistence round trips reproduce the packed runtime output exactly.
- Missing packed tensors and incompatible metadata are rejected before inference.

For a 4096 x 4096 rank-16 FP16 adapter in the included synthetic benchmark:

| Format | Adapter memory | Reduction | Delta cosine | Relative L2 |
|---|---:|---:|---:|---:|
| FP16 | 256 KiB | - | 1.0 | 0 |
| Q8 | 212 KiB | 17.2% | 0.99998 | 0.0066 |
| Q4 | 116 KiB | 54.7% | 0.99376 | 0.1116 |

The benchmark also makes the padding tradeoff visible: low-rank adapters can become larger at higher bit widths, which is why the CLI checks actual packed parameter bytes instead of assuming savings from bit width alone.

## Limitations

- Quantized adapters are inference-only.
- This PR quantizes `LoRALinear` matrices; LoRA embedding and switch-linear layers remain in their original precision.
- Persistent quantized adapters currently use affine quantization.
- Multi-adapter banks and dynamic routing are intentionally proposed in a follow-up stacked PR so that this format and runtime path can be reviewed independently.

## Test commands

```bash
PYTHONPATH=. python -m unittest tests.test_finetune tests.test_quantized_lora
PYTHONPATH=. python benchmarks/quantized_lora.py
python -m mlx_lm.quantize_adapter --help
```

### DIFF
--- a/benchmarks/quantized_lora.py
+++ b/benchmarks/quantized_lora.py
@@ -1,0 +1,229 @@
+# Copyright © 2026 Apple Inc.
+
+import argparse
+import json
+import platform
+import statistics
+import time
+from pathlib import Path
+
+import mlx.core as mx
+import mlx.nn as nn
+from mlx.utils import tree_flatten
+
+from mlx_lm.tuner.lora import LoRALinear
+
+DTYPES = {
+    "float16": mx.float16,
+    "bfloat16": mx.bfloat16,
+    "float32": mx.float32,
+}
+
+
+def comma_separated_ints(value):
+    return tuple(int(item) for item in value.split(",") if item)
+
+
+def configure_parser():
+    parser = argparse.ArgumentParser(
+        description="Benchmark quantized LoRA precision, memory, and latency."
+    )
+    parser.add_argument("--input-dims", type=int, default=4096)
+    parser.add_argument("--output-dims", type=int, default=4096)
+    parser.add_argument("--tokens", type=int, default=32)
+    parser.add_argument("--ranks", type=comma_separated_ints, default=(8, 16, 32, 64))
+    parser.add_argument("--bits", type=comma_separated_ints, default=(8, 6, 5, 4))
+    parser.add_argument("--group-size", type=int, default=64)
+    parser.add_argument("--rank-group-size", type=int, default=32)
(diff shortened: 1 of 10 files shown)


HongyuChen0926:
Stacked follow-up: #1539 adds packed multi-adapter banks, request-level selection, mixed-token routing, and activation-based mixed-precision selection on top of this PR.

zcbenz:

PR #1539, opened 2026-07-11
Add quantized LoRA adapter banks and dynamic routing

HongyuChen0926:
> [!IMPORTANT]
> This is a stacked follow-up to #1538. PR #1538 introduces the persistent runtime-quantized LoRA format and must be reviewed and merged first. Until then, GitHub shows both commits in this pull request because the base branch lives in a fork. The incremental PR 2 diff is available here: https://github.com/HongyuChen0926/mlx-lm/compare/feature/runtime-quantized-lora...feature/quantized-lora-adapter-bank

## Relationship to PR #1538

PR #1538 provides the foundation used here:

- `QuantizedLoRALinear` with packed A and B matrices evaluated by `mx.quantized_matmul`.
- A versioned persistent adapter format containing packed weights and strict metadata.
- Direct loading through the normal `adapter_path` interface.
- The conversion CLI, numerical tests, and single-adapter benchmark.

This follow-up does not change that format. It builds a multi-adapter runtime on top of the packed files from PR #1538.

## Summary

- Add `QuantizedLoRAAdapterPack`, which stacks one target layer from multiple adapters while keeping all matrices packed.
- Add `QuantizedLoRAAdapterBank`, with stable name-to-index lookup and layer-wise packs that do not duplicate the base model.
- Load multiple persistent quantized adapters directly from disk without constructing a base model per adapter.
- Support request-level adapter selection with two `mx.quantized_matmul` operations.
- Support per-token-row adapter routing with two `mx.gather_qmm` operations.
- Add activation-based mixed-precision selection across Q4, Q5, Q6, and Q8 with numerical-error and actual-memory thresholds.
- Add 10, 50, and 100-adapter memory/routing benchmarks and focused contract tests.

## Motivation

The main use case is not accelerating one small LoRA branch. It is keeping more task- or user-specific adapters resident next to one shared base model while preserving dynamic selection. Packed runtime storage is particularly useful for endpoint deployments where fusing every adapter would duplicate the base model and loading adapters on demand would add switching latency.

## Design

- Adapter packs store stacked packed `uint32` weights, affine scales, affine biases, and one LoRA scale per adapter.
- Request-level selection indexes one packed adapter and uses `mx.quantized_matmul`; its lookup cost is independent of the number of resident adapters.
- Mixed-token routing uses `mx.gather_qmm` with an adapter index for every flattened token row. The implementation uses the regular gather path for correctness.
- Bank loading requires matching layer names, dimensions, rank, padding, group sizes, mode, and per-layer bit widths. Every required packed tensor is validated before stacking.
- Mixed-precision selection keeps a layer in floating point when no candidate satisfies both numerical and memory-reduction thresholds.

## Validation

The stacked branch passes 26 focused tests. PR 2 adds coverage for:

- Direct loading of three independently persisted adapters into one bank without base-model duplication.
- Exact request-level equality against the corresponding standalone quantized layer.
- Mixed-row routing against independently evaluated per-row quantized matmuls.
- Adapter index, shape, rank, metadata, and missing-weight validation.
- Mixed-precision selection and the low-rank padding memory gate.

For 100 resident 4096 x 4096 rank-16 adapters:

| Format | Packed adapter memory | Reduction | Request switching | Mixed-token routing | Delta cosine |
| FP16 | 25.00 MiB | - | 0.208 ms | 0.310 ms | 1.0 |
| Q8 | 20.70 MiB | 17.2% | 0.216 ms | 0.776 ms | 0.99998 |
| Q4 | 11.33 MiB | 54.7% | 0.215 ms | 0.674 ms | 0.99342 |

These are synthetic microbenchmarks, not end-to-end generation results. Request-level switching remains in the same latency range at 100 adapters. Mixed-token quantized routing is currently about 2-3x slower than floating-point `gather_mm`, so this path should be considered memory-oriented rather than a throughput optimization.

## Limitations

- This PR depends on #1538 and should not merge first.
- All adapters in one bank must share the same target layers, dimensions, rank, and quantization configuration.
- Adapter banks currently require every target LoRA layer to be packed with affine quantization.
- This PR provides the packed bank and routing primitives, not a server or request-context API.
- Mixed-token routing prioritizes correctness and memory; further kernel work would be needed for a throughput-oriented path.
- Mixed-precision selection requires representative calibration activations supplied by the caller.

## Test commands

```bash
PYTHONPATH=. python -m unittest tests.test_finetune tests.test_quantized_lora
PYTHONPATH=. python benchmarks/quantized_lora.py
PYTHONPATH=. python benchmarks/quantized_lora_bank.py
```

### DIFF
--- a/benchmarks/quantized_lora.py
+++ b/benchmarks/quantized_lora.py
@@ -1,0 +1,229 @@
+# Copyright © 2026 Apple Inc.
+
+import argparse
+import json
+import platform
+import statistics
+import time
+from pathlib import Path
+
+import mlx.core as mx
+import mlx.nn as nn
+from mlx.utils import tree_flatten
+
+from mlx_lm.tuner.lora import LoRALinear
+
+DTYPES = {
+    "float16": mx.float16,
+    "bfloat16": mx.bfloat16,
+    "float32": mx.float32,
+}
+
+
+def comma_separated_ints(value):
+    return tuple(int(item) for item in value.split(",") if item)
+
+
+def configure_parser():
+    parser = argparse.ArgumentParser(
+        description="Benchmark quantized LoRA precision, memory, and latency."
+    )
+    parser.add_argument("--input-dims", type=int, default=4096)
+    parser.add_argument("--output-dims", type=int, default=4096)
+    parser.add_argument("--tokens", type=int, default=32)
+    parser.add_argument("--ranks", type=comma_separated_ints, default=(8, 16, 32, 64))
+    parser.add_argument("--bits", type=comma_separated_ints, default=(8, 6, 5, 4))
+    parser.add_argument("--group-size", type=int, default=64)
+    parser.add_argument("--rank-group-size", type=int, default=32)
+    parser.add_argument("--base-bits", type=int, choices=(0, 4, 8), default=8)
+    parser.add_argument("--dtype", choices=DTYPES, default="float16")
+    parser.add_argument("--warmup", type=int, default=10)
+    parser.add_argument("--iterations", type=int, default=50)
+    parser.add_argument("--seed", type=int, default=7)
+    parser.add_argument("--output", type=Path)
+    return parser
+
+
+def percentile(values, quantile):
+    values = sorted(values)
+    index = round((len(values) - 1) * quantile)
+    return values[index]
+
+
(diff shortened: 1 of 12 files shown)


zcbenz:

PR #1542, opened 2026-07-11
Add opt-in state-budget admission to BatchGenerator

pierre427:
# Add opt-in state-budget admission to BatchGenerator

## Problem

Concurrency in the batch server is governed by static request counts
(`--decode-concurrency`, `--prompt-concurrency`) that know nothing
about memory. A cap tuned for short chats will admit the same 32
requests at 32k context and blow far past the intended working set; the
same cap on short requests leaves most of the budget idle. The engine
already computes live cache bytes (`prompt_cache_nbytes`) — it just
never consults them at admission time.

## Change

A three-layer, opt-in design. With the flag absent (the default),
BatchGenerator follows the existing count-based admission path exactly;
regressions pin FIFO/cohort behavior.

### 1. `mlx_lm/batch_admission.py` — a model-family-neutral policy module

- **`AdmissionState`**: what the scheduler knows about one request —
  total projected work units, completed units, resident bytes,
  metadata. Units are *scheduler-defined*: tokens for autoregressive
  models, denoising steps for diffusion. Nothing in the protocol
  assumes token semantics.
- **`LinearStateCost(fixed_bytes, bytes_per_unit, max_units,
  allocation_step_units)`**: the AR/hybrid cost model — fixed bytes per
  row (recurrent/SSM state) plus linear growth (KV), capped at
  `max_units`, with an allocation envelope: cache leaves allocate
  capacity in steps (`KVCache.step` = 256), and capacity after
  arbitrary chunk-aligned growth is bounded by `units + step - 1`
  (alignment-independent). Its `cohort_bytes(states)` models the
  geometry of *batched* caches: a merged batch tensor allocates every
  row at the cohort-maximum width, so N rows cost
  `N * (fixed + slope * envelope(max_final_units))` — non-additive,
  and very different from summing per-row estimates.
- **`StepStateCost`**: a peak-state adapter for diffusion and other
  iterative schedulers — the scheduler supplies a per-step peak-bytes
  schedule (its real envelope: canvas logits, block caches, ...), and
  projected cost is the maximum over *remaining* steps, so progress
  releases headroom. It deliberately stays on the model-native
  scheduler contract: `BatchGenerator` itself accepts only
  `LinearStateCost` (where projection at the current unit count
  genuinely represents resident state), so peak-only policies cannot
  be misused as resident accounting.
- **`StateBudget`**: the budget plus a projector, with
  `admitted_prefix(candidates, live_bytes, active)` answering "how
  many of these, in this exact order, fit."

### 2. Engine integration

Admission evaluates each exact candidate prefix by recomputing the full
cohort projection — all active rows (generating and prefilling) plus
the candidates, at final extents — floored by actual live bytes (a
stale wide allocation is never assumed smaller than reality), with
resident bytes of still-unselected queued requests and unverifiable
supplied caches *added* on top (never folded into the floor, never
credited). Because the admission cohort may be reordered by a
length-aware selector, the budget gates the *exact selected cohort*,
not a FIFO prefix. Two deliberate contract points:

- **Liveness:** a request whose projection alone exceeds the budget
  still admits when nothing else is running (logged; best-effort, not
  an OOM guarantee) — mirroring how a count cap always lets one request
  proceed.
- **Fail-closed construction:** growing state without a validated
  positive integral allocation step is refused (both via `kv_cost` and
  via a directly-supplied policy); fixed-only state may omit the step.

### 3. Server layer — `--state-budget-gb`

At model load, a short probe forwards real tokens at chunk-aligned
points and measures **per-cache-leaf** growth: raw fixed bytes, bytes
per token, and the validated common allocation step (recursing
composite caches). It self-verifies at a non-boundary point and again
per-leaf at an independent aligned 2048-token point on a fresh cache —
where a consistent leaf must predict exactly (tolerance covers
arithmetic error only) — and **refuses budgeting** for anything it
cannot model faithfully: rotating windows inside the probe range,
mixed allocation steps, opaque composite components, materially
negative fit intercepts, or nonlinear growth. The measured
`(fixed, per_token, step)` flows into the admission gate. The budget
is taken as already headroom-adjusted; no hidden factors are applied.

## Correctness

The admission arithmetic and probe are pinned by 25+ regressions,
including: exact-cohort vs FIFO equivalence; unequal-`max_tokens`
cohort selection; unselected-resident additivity (max-folding would
wrongly admit); live-floor ordering (a large floor must not absorb
unverified bytes); continued-generation with real primed caches
(history inside the projection target; unverifiable bytes charged, not
credited); reject -> automatic re-admit across a completed row's
removal under one unchanged budget; capacity-boundary crossing;
single-row unaligned continuation (capacity `previous_logical +
round_up(chunk)`); per-leaf third-point consistency with slope-change
and two-leaf cancellation fakes refused; and both fail-closed
construction paths. The W3 flood benchmark itself asserts
`live <= budget` at every trajectory sample.

Scope note: this is a **state admission budget**, not a total
process peak-memory guarantee — weights, activation transients, and
split/extend workspaces are outside its accounting.

## Measured

Apple M5 Max (128 GB), mlx 0.32.0.dev; flood protocol (64 concurrent
requests, 8192-token prompts, 32 generated), smoke-first with the
live-vs-budget assertion active. Probe on Qwen3-Coder-30B-A3B-8bit:
98,304 B/token, step 256 validated, per-leaf third-point consistency
passed.

| configuration | steady active | live cache | peak RSS | completion |
|---|---|---|---|---|
| count caps only (control) | 32 | up to 26.6 GB | 84.9 GB | 156.9 s |
| `--state-budget-gb 20` | 25 (budget-gated) | bounded at 20.8 GB (< 21.5 GiB) | 74.2 GB (-10.7 GB) | 162.3 s (+3.4%) |
| pristine `main` reference | 32 | up to 26.6 GB | 84.9 GB | 162.7 s |

The budget arm holds admitted live state under the configured cap at
every sampled step (assertion-enforced), trims peak RSS by ~10.7 GB,
and costs +3.4% completion time on this flood.

Honest boundary — hybrid models: on Nemotron-3-Nano (hybrid
Mamba/attention; ~49 MB fixed per row, ~6 KB/token), a 48-request
flood at both 4K and 32K context peaks at only 2.4/5.8 GB live — the
budget never engages and count caps bind first. **The byte budget is a
KV-heavy-model lever; on fixed-state-dominated hybrids at these
workloads it is inert by construction** (their per-row fixed state is
still accounted, via the same probe, in the fixed term).

Diffusion readiness (architecture evidence, no generation-perf claim):
the `StepStateCost` seam was exercised against a real, instrumented
LLaDA-8B block-diffusion run (observed per-block cache + bf16 logits
envelope) with exact peak accounting and admission arithmetic verified
at every progress point, at ~0.35 microseconds per policy call.
Diffusion models keep their model-native loops and feed their real
envelopes to the shared policy; `BatchGenerator` remains an AR engine.


pierre427:
Requesting review of the memory model and admission guarantees in this PR.

This implementation admits work using projected live state rather than a token-count limit. The complication is that allocation is not generally linear in tokens: batched rows can inherit the cohort’s maximum geometry, some state grows in discrete steps, and diffusion or recurrent models can have fixed and transient costs that differ substantially from ordinary KV caches.

The accounting combines per-request state models with observed live-cache bytes. If the live measurement is unavailable or inconsistent, admission fails closed rather than assuming that memory is available.

In the flood test, live cache remained near 20.8 GB and peak RSS fell by approximately 10.7 GB, with a 3.4% increase in completion time.

- whether the projection remains an upper bound across batching, quantization, recurrent caches, and diffusion state;
- whether concurrent admissions can race between measurement and allocation;
- how allocator fragmentation, delayed reclamation, and Metal working-set behaviour affect the relationship between modeled bytes and RSS;
- whether the live-byte floor and safety margin respond appropriately when the model underestimates an unfamiliar cache type;
- whether large requests can be indefinitely deferred while smaller requests continue to fit;
- whether the state-cost interface is sufficient for future architectures without adding model-specific logic to the scheduler.

The central question is what information a model or cache must expose for memory admission to remain conservative without unnecessarily serializing the workload.

PR #1543, opened 2026-07-11
Add `--preserve-mtp` to keep MTP weights during conversion

samithaj:
### Summary

For architectures whose `sanitize` implementation removes multi-token-prediction
(MTP / "next-n") tensors, `mlx_lm.convert` does not include those tensors in the
converted output, so downstream extraction currently needs access to the
original source checkpoint or another artifact that retained them. External MTP
tooling starts from exactly these tensors — the raw-weight extraction convention
used by mlx-vlm's MTP drafter split tools, vllm-metal, and MTPLX. In-checkpoint
MTP loading and runtime support are discussed in #872, #990, and #1456.

I added a `--preserve-mtp` flag to `convert` that keeps them. Convert-side only:
no `sanitize`, model, or runtime changes; the one utils change is an optional
`revision` argument on `hf_repo_to_path` so the preserved tensors are resolved
at the same revision as the converted weights.

### What it does

- Off by default. Without the flag the output directory is byte-for-byte
  identical — the test compares every file name and SHA-256 hash with and
  without the flag.
- With the flag, after the normal save the MTP tensors are read from the raw
  source safetensors and written to `mtp.safetensors` (sharded
  `mtp-XXXXX-of-YYYYY.safetensors` when large) with their own
  `mtp.safetensors.index.json`; `model.safetensors.index.json` is untouched.
- Preserved tensors are copied without applying mlx-lm's dtype conversion or
  quantization. Tests cover bf16 and fp32 source tensors; FP8 source layouts
  (e.g. weight + scale-inv pairs) are not yet validated.
- `mlx_lm.load` currently reads only `model*.safetensors`, so it does not load
  the sidecar; the sidecar and its index are for architecture-aware extraction
  tools.
- Exact key collisions with saved model parameters are skipped. Preserved
  tensors otherwise retain their raw source names, deliberately — extraction
  tools expect pre-sanitize names.

### Supported families

The MTP name patterns live in one constant (`MTP_WEIGHT_PATTERNS`), curated from
the sanitize implementations that drop them: qwen3_5, qwen3_next, nemotron_h,
exaone_moe, mimo, mimo_v2_flash, kimi_linear, longcat_flash, step3p5,
ernie4_5_moe. Architectures that instead carry the MTP module as decoder layers
past `num_hidden_layers` — deepseek_v3, deepseek_v32, glm4_moe, glm4_moe_lite,
step3p5 — are matched by layer index, gated on an explicit model-type list
(`MTP_TRAILING_LAYER_MODEL_TYPES`) so the layer-index rule cannot misfire on
other architectures. deepseek_v3's sanitize hardcodes layer 61, which equals
`num_hidden_layers` for DeepSeek-V3 checkpoints, so the generic rule covers it.

### Tests

`tests/test_utils.py::TestPreserveMTP`, no network:

- the matcher covers each family's dropped names and the trailing-layer list,
  and rejects dense weights and non-listed model types
- synthetic round-trip: preserved tensors survive byte-identical in their
  source dtype, the sidecar index maps every preserved key, and `load` of the
  output is unaffected
- without the flag, the entire output directory is identical (file names and
  hashes)
- `hf_repo_to_path` propagates `revision` (mocked `snapshot_download`)

Related: #1033 discussed keeping extra tensors on the load path
(`keep_prefixes` / exposing `strict=False`); this is the narrower convert-side
counterpart for tensors that never survive to load.

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -1,6 +1,8 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import glob
+import json
 from pathlib import Path
 from typing import Callable, Optional, Union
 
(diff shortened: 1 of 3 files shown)


ulises-c:
This would be great to have to support Qwen3.5 & 3.6 with MTP

zcbenz:

PR #1544, opened 2026-07-11
Fix gradient accumulation to weight micro-batches by token count

Incheonkirin:
> This replaces #1413, which I accidentally closed when its source fork was detached. The change has been rebased onto current `main`.

`default_loss` returns a per-micro-batch token-mean (`ce.sum() / ntoks`). With gradient accumulation, `step()` summed the per-micro-batch gradients and divided by a fixed `grad_accumulation_steps`. Since each micro-batch gradient is already normalized by its own token count, dividing the sum by K does not equal the gradient of the token-mean loss over the combined effective batch unless every micro-batch has the same number of trained tokens. With variable-length completions the accumulated update is biased toward shorter examples and depends on how examples are partitioned into micro-batches. `evaluate()` already token-weights its average, so the trainer was internally inconsistent.

This keeps a running token-weighted mean of the gradient over the accumulation window. For single-process training that running mean is already the gradient the optimizer needs, so no extra normalization is applied at the update; only with multiple ranks is it reduced across ranks weighted by each rank's token count. The accumulation denominators are guarded so an all-zero-token window cannot divide by zero. The final partial window (when `iters` is not a multiple of `grad_accumulation_steps`) is now applied instead of dropped. Single-process behavior is unchanged; with multiple ranks, the reduction is now token-weighted, including when `grad_accumulation_steps == 1`. All-zero-token windows still call `optimizer.update` with a zero gradient. The underlying zero-token loss NaN in `default_loss` is tracked separately in open PR #1355.

Tests: unequal-token micro-batches now match a single combined batch, the final partial window matches explicit windows, and an all-zero-token window keeps parameters finite. The first two fail on the current code and pass with this change.

## Validation
- `pytest -q tests/test_finetune.py::TestGradientAccumulation` (4 passed)

### DIFF
--- a/mlx_lm/tuner/trainer.py
+++ b/mlx_lm/tuner/trainer.py
@@ -244,20 +244,37 @@ def train(
     state = [model.state, optimizer.state, mx.random.state]
 
     @partial(mx.compile, inputs=state, outputs=state)
-    def step(batch, prev_grad, do_update):
+    def step(batch, prev_grad, prev_ntoks, do_update):
         (lvalue, toks), grad = loss_value_and_grad(model, *batch)
 
-        if prev_grad is not None:
-            grad = tree_map(lambda x, y: x + y, grad, prev_grad)
+        # Token-weighted running mean of the gradient, so accumulation matches the
+        # token-mean loss over the full batch instead of a fixed 1/K average.
+        if prev_grad is None:
+            ntoks = toks
+        else:
+            ntoks = prev_ntoks + toks
+            denom = mx.maximum(ntoks, 1)
+            grad = tree_map(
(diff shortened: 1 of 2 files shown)


Incheonkirin:
@angeloskath, could you review this gradient-accumulation token-weighting fix? It replaces #1413; the targeted tests pass.

Incheonkirin:
@angeloskath, following up on this. Since you proposed the current gradient-accumulation structure in #511, could you take a look at the token-weighted reduction here? With unequal trained-token counts, the existing fixed 1/K average weights micro-batches equally instead of matching the combined token-mean gradient. The targeted tests cover unequal-token batches, the final partial window, and zero-token cases.

zcbenz:

PR #1547, opened 2026-07-11
Copy-on-write prefix forks for KV caches

pierre427:
# Copy-on-write prefix forks for KV caches

Branch: `pr/cow-prefix-forks`, off `main` @ a790972.
Commits: `a6bcf1f` → `7dfae09` → `e973043` → `15b344b` (final adversarial hardening).
New `mlx_lm/prefix_forks.py` + `tests/test_prefix_forks.py`; no existing files touched.

## Problem

Many concurrent requests often share one long common prefix — a system prompt, a
document, a few-shot preamble. The request-level prompt cache
(`LRUPromptCache.fetch_nearest_cache`) hands every hit a `copy.deepcopy` of the
cached KV stack. The `deepcopy` *call* is sub-millisecond (MLX arrays copy
lazily), but the copy is materialized on first use: an O(prefix) allocation and
copy per request — measured at ~4 s for a 16k-token prefix in prior benches.
N concurrent requests on the same prefix pay that N times and hold N full
copies in RAM.

## Change

Fork from ONE frozen in-RAM parent snapshot, paying only for private suffixes:

- **`FrozenPrefixSnapshot`** — an immutable, `mx.eval`-materialized,
  content-addressed (sha256 cid over model key + token ids) snapshot of a
  prefix's per-layer KV state, trimmed to exact length (no `step` slack).
  Materialized once per unique prefix content.
- **`ForkedKVCache`** — per-layer cache mirroring `KVCache`'s interface
  (`update_and_fetch`, `offset`, `size`, `state`, `trim`, `make_mask`,
  `nbytes`, `empty`). Holds a *reference* to the frozen parent arrays (zero
  copy) plus a private, growing `KVCache` tail; attention consumers see
  `concat(prefix, tail)`. Writes go only to the tail → the parent is
  structurally immutable: **COW by construction**, not copy-on-first-write.
  Fork creation is O(1); a fork's persistent footprint is O(tail).
- **`PrefixForkRegistry`** — cid-keyed snapshot store with longest-prefix
  lookup (reuses `PromptTrie`), content dedup with a spot-check, LRU eviction
  by count/bytes, pinned-bytes reporting, and thread-safe access.
  `fetch_fork(model_key, tokens)` is the fork-aware analogue of
  `fetch_nearest_cache` — same `(cache, remaining_tokens)` contract, no
  deepcopy.

Library + tests only (opt-in, alongside `LRUPromptCache`); no server rewiring
in v1.

### The model-key contract (B1)

Every registry call takes a **caller-supplied stable string** identifying the
*exact weights* the KV was computed with — it must encode at least model path,
adapter, and revision (e.g. `"org/model@rev+adapter-sha"`), and it **MUST
change whenever the weights change** (adapter load/unload, revision update,
requantization). Non-string keys raise `TypeError` — there is no fallback.
Object/`id()`-derived keys are forbidden because both failure modes were
*reproduced* in review: an in-place weight swap keeps the same object (same id
→ deterministic stale-KV hit), and CPython recycles addresses of dead objects
(a freed key's address reused by a distinct key → cross-namespace hit). Cached
KV is a pure function of (weights, tokens); the key must be too.

## How it works — the aliasing experiment

`KVCache.update_and_fetch` writes via in-place slice assignment
(`keys[..., prev:offset, :] = new`) into a step(=256)-preallocated buffer, so
the design hinged on: *if a fork shares the parent's array and the parent keeps
writing, does the fork observe the mutation?* Probed empirically on
mlx 0.32.0.dev (same-object sharing, lazy and evaluated slices, a
refcount-donation decode loop, and real `KVCache.state` exports across 64
parent decode steps):

- **Mutation visibility follows Python-OBJECT identity, not buffer identity.**
  `a[...] = x` (`__setitem__`) rebinds the array object to the `slice_update`
  result; every reference through that same Python object observes the write,
  but any slice taken *before* the write — lazy or evaluated — references the
  pre-write graph node and is preserved. MLX donation never recycles a buffer
  another live node references, so even a refcount-1-style decode loop cannot
  scribble an exported slice.
- **However**, sharing the live buffer is the wrong primitive anyway: an
  evaluated slice of the step-padded `(B, H, S, D)` buffer along axis 2 is
  strided, so materializing it costs an O(prefix) copy *per fork*; a lazy
  slice pins the parent's entire step-padded buffer and sits one MLX
  donation-optimization away from unsoundness (live `.state` aliases have
  scribbled snapshots before — hence the established `mx.array` + `mx.eval`
  park recipe, which `freeze` reuses).

**Conclusion (implemented):** materialize once per unique prefix content into a
frozen exact-length snapshot; all forks share those arrays by reference. The
single O(prefix) copy is amortized across forks via cid dedup; immutability is
structural — no code path calls `__setitem__` on the shared arrays, **raw
frozen array objects never escape** (every boundary — `.state` with an empty
tail, snapshot/fork constructors — hands out fresh zero-copy slice *objects*,
so a consumer's `__setitem__` rebinds only the consumer's object; the same
rebinding semantics give the seal for free), and the registry plus each live
fork hold references, so donation is impossible by construction.

Known v1 cost: each attention step materializes a transient
`concat(prefix, tail)` per layer (attention already reads all KV bytes each
step; this adds one extra write pass). Removing it needs two-segment attention
(e.g., flash-decoding-style log-sum-exp merge) — left as a follow-up.

### Failure-direction policy

- **Invalidation errs toward false MISS only:** eviction/invalidation removes
  discoverability (next fetch re-prefills — always safe); live forks keep
  their snapshot's arrays alive through ordinary Python references, so a stale
  HIT on released memory cannot happen. Consequently `max_bytes`/`nbytes`
  bound the **discoverable** set only; bytes kept resident by live forks
  pinning an evicted snapshot are reported separately via `pinned_nbytes`
  (weakref-tracked, F4).
- **Dedup errs toward loud failure:** a cid-dedup hit spot-checks the offered
  KV content (layer-0 tail slice) against the stored snapshot and **raises**
  on mismatch — same key+tokens over changed weights is a key-contract
  violation, not a cache hit (F3: contract documented *and* spot-check
  implemented).
- **`trim` errs toward loud failure:** trimming within the private tail
  (rollback of post-fork tokens — speculative decoding, retries: the intended
  integration) works like `KVCache`; trimming *past* the tail into the frozen
  prefix **raises** `ValueError`. Chosen over `is_trimmable() → False` because
  it preserves the rollback story while making the reproduced
  `LRUPromptCache` trim-longer offset-skew impossible — callers of
  `trim_prompt_cache` ignore its return value, so a silent clamp meant
  generating against wrong KV (F1).
- **Serialization refuses loudly:** `save_prompt_cache` of a fork raises
  (`load_prompt_cache` could never reconstruct it); `to_kv_cache()` is the
  explicit detach-and-copy escape hatch (N4).

## Correctness

`tests/test_prefix_forks.py` — 30 tests, all passing (plus
`tests/test_prompt_cache.py` 22/22 unaffected). The suite was validated
against the reviewer's mutation battery: flipped prefix/tail concat, off-by-one
offset, lazy O(prefix) fork copy, and skewed mask offset each fail at least one
test (the lazy-copy mutant is caught because the memory assertion measures
*after* forcing evaluation of everything the fork exposes).

1. **Next-token equivalence** — greedy generation continuing from a fork is
   token-identical (4-token suffix prefill + 110 decode steps) to continuing
   from a full `deepcopy` of the same cache
   (`mlx-community/Qwen1.5-0.5B-Chat-4bit`, prefix > step boundary). The N>1
   prefill is position-sensitive: N=1 decode is permutation-invariant over
   keys and cannot see a flipped concat.
2. **Parent immutability** — after a fork generates 110 tokens, the snapshot's
   arrays are bit-identical to copies taken before. Review additionally
   verified 500-token interleaved sibling generation clean on both mlx
   versions, and capacity-boundary freeze exact.
3. **Sibling isolation** — two forks with different suffixes each match their
   own clean-reference generations; both verified to share the same snapshot.
4. **Memory** — fork creation's `mx.get_active_memory()` delta (measured after
   eval) is bounded far below prefix size; `fork.nbytes` is 0 at creation and
   counts only the private tail (`shared_nbytes` exposes the prefix once).
5. **Identity** — non-str model keys raise everywhere; distinct string keys
   never collide; a weight swap under a changed key misses (contract test);
   same key + different KV content raises at freeze (dedup spot-check).
6. **cid dedup & false-miss-only invalidation** — bit-identical content dedups
   to one snapshot with bytes counted once; LRU eviction and `invalidate`
   produce misses while pre-existing forks keep generating; `pinned_nbytes`
   tracks fork-pinned residency through invalidation and drops to zero when
   forks die.
7. **Loud failure paths** — over-trim raises (directly and through
   `trim_prompt_cache`); `.state` is read-only; empty-tail `.state` is sealed
   (hostile `__setitem__` on it leaves snapshot and siblings bit-identical);
   constructor-provided arrays are sealed; `save_prompt_cache` refuses.
8. **Final adversarial closeout** — unambiguous full-SHA CID framing; exact
   token types; read-only snapshot arrays/metadata; K+V all-layer dedup;
   negative-trim rejection; unsupported-cache fast-path rejection; strict
   byte caps; and same-CID reincarnation accounting under concurrent stress.

## Known limits (documented in-module)

- Only stacks of plain `KVCache` layers are forkable; anything else makes
  `freeze` return `None` (safe miss). Quantized-KV prefixes are a natural
  follow-up.
- A length-1 snapshot is only found by an exact-length fetch (`PromptTrie`'s
  `shorter` result requires a match past position 0) — false miss, never a
  false hit.
- Exact-match `fetch_fork` returns `remaining == []`; as with
  `fetch_nearest_cache`, the caller must keep ≥1 token to feed the model
  (e.g. freeze on `tokens[:-1]`).
- The dedup spot-check is a spot check (layer-0 tail slice), not a proof of
  equality — the model-key contract remains the real identity boundary.

## Measured

Apple M5 Max (128 GB), mlx 0.32.0.dev, Qwen3-0.6B-4bit; one shared
8192-token prefix continued by 8 consumers with distinct suffixes:

| approach | resident delta (8 consumers) | first-token median | total |
|---|---|---|---|
| deepcopy (upstream prompt-cache pattern) | +7.73 GB | 15.6 ms | 0.129 s |
| fork (this change) | **+0.21 GB** | **7.2 ms** | **0.068 s** |

**37x less resident memory** (one frozen snapshot + private tails vs
eight materialized copies), 2.2x faster first continued token (a
deepcopy is lazily materialized on first use), first tokens identical
across approaches. Caveat: the fork run's transient peak was ~1 GB
higher — the per-step `concat(prefix, tail)` materializes both segments;
a two-segment attention (LSE-merge) follow-up removes that transient.
The durable win is resident memory, which is what bounds concurrent
capacity.

### DIFF
--- a/mlx_lm/prefix_forks.py
+++ b/mlx_lm/prefix_forks.py
@@ -103,7 +103,8 @@ import hashlib
 import threading
 import weakref
 from collections import OrderedDict
-from typing import Any, Dict, List, Optional, Tuple
+from numbers import Integral
+from typing import Any, List, Optional, Tuple
 
 import mlx.core as mx
 
@@ -125,6 +126,22 @@ def _require_model_key(model_key: Any) -> str:
     return model_key
 
 
+def _normalize_tokens(tokens: List[int]) -> Tuple[int, ...]:
+    """Validate exact token IDs without lossy ``int(...)`` coercion."""
+    normalized = []
+    for token in tokens:
+        if isinstance(token, bool):
+            raise TypeError("token IDs must be integers, not bool")
+        try:
+            value = token.__index__()
+        except (AttributeError, TypeError):
+            raise TypeError(
+                f"token IDs must be integers; got {type(token).__name__}"
+            ) from None
+        normalized.append(int(value))
+    return tuple(normalized)
+
+
 def compute_cid(model_key: str, tokens: List[int]) -> str:
     """Content-hash identity for a prefix: (model key, exact token ids).
 
@@ -133,11 +150,19 @@ def compute_cid(model_key: str, tokens: List[int]) -> str:
     model key is subject to the module-level MODEL-KEY CONTRACT.
     """
     model_key = _require_model_key(model_key)
+    tokens = _normalize_tokens(tokens)
     h = hashlib.sha256()
-    h.update(model_key.encode("utf-8", "replace"))
+    # Length-frame every field. A delimiter alone is ambiguous because a
+    # model key is an arbitrary string and may itself contain that delimiter.
+    h.update(b"mlx-lm-prefix-fork-cid-v1")
+    model_key_bytes = model_key.encode("utf-8", "surrogatepass")
+    h.update(len(model_key_bytes).to_bytes(8, "big"))
+    h.update(model_key_bytes)
     for t in tokens:
-        h.update(b"\x00" + str(int(t)).encode())
-    return h.hexdigest()[:16]
+        token_bytes = str(t).encode()
+        h.update(len(token_bytes).to_bytes(8, "big"))
+        h.update(token_bytes)
+    return h.hexdigest()
 
 
 class FrozenPrefixSnapshot:
@@ -166,21 +191,47 @@ class FrozenPrefixSnapshot:
         keys: List[mx.array],
         values: List[mx.array],
     ):
-        self.cid = cid
-        self.model_key = model_key
-        self.tokens = tokens
+        self._cid = cid
+        self._model_key = model_key
+        self._tokens = tokens
         # Fresh slice OBJECTS (zero-copy): a later __setitem__ through the
         # caller's original references rebinds the caller's objects, never
         # these pre-write nodes.
-        self.keys = [k[:] for k in keys]
-        self.values = [v[:] for v in values]
-        self.nbytes = sum(k.nbytes + v.nbytes for k, v in zip(keys, values))
+        self._keys = [k[:] for k in keys]
+        self._values = [v[:] for v in values]
+        self._nbytes = sum(k.nbytes + v.nbytes for k, v in zip(keys, values))
         # Live forks referencing this snapshot (weak: forks hold the strong
         # edge). Lets the registry report pinned-but-undiscoverable bytes.
         self._forks = weakref.WeakSet()
 
     def __len__(self):
         return len(self.tokens)
+
+    @property
+    def cid(self):
+        return self._cid
+
+    @property
+    def model_key(self):
+        return self._model_key
+
+    @property
+    def tokens(self):
+        return self._tokens
+
+    @property
+    def nbytes(self):
+        return self._nbytes
+
+    @property
+    def keys(self):
+        """Fresh sealed views; the stored MLX array objects never escape."""
+        return [k[:] for k in self._keys]
+
+    @property
+    def values(self):
+        """Fresh sealed views; the stored MLX array objects never escape."""
+        return [v[:] for v in self._values]
 
     @property
     def pinned(self) -> bool:
@@ -198,6 +249,7 @@ class FrozenPrefixSnapshot:
         quantized, and stateful (Mamba) caches are not COW-forkable in v1.
         """
         model_key = _require_model_key(model_key)
+        tokens = list(_normalize_tokens(tokens))
         if len(tokens) == 0 or len(prompt_cache) == 0:
             return None
         for c in prompt_cache:
@@ -214,14 +266,14 @@ class FrozenPrefixSnapshot:
             values.append(mx.array(v))
         mx.eval(*keys, *values)
         cid = compute_cid(model_key, tokens)
-        return cls(cid, model_key, tuple(int(t) for t in tokens), keys, values)
+        return cls(cid, model_key, tuple(tokens), keys, values)
 
     def fork(self) -> List["ForkedKVCache"]:
         """A fresh per-layer cache stack referencing this snapshot. O(1):
         no prefix bytes are copied or materialized."""
         forks = [
-            ForkedKVCache(self.keys[l], self.values[l], snapshot=self)
-            for l in range(len(self.keys))
+            ForkedKVCache(self._keys[l], self._values[l], snapshot=self)
+            for l in range(len(self._keys))
         ]
         for f in forks:
             self._forks.add(f)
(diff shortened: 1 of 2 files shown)


pierre427:
Requesting review of the ownership, mutation, and lifetime invariants in this PR.

This implementation stores completed prefixes as immutable, content-addressed snapshots. A request can fork a shared snapshot without copying its existing arrays; the request-local tail is copied only when it is first mutated.

The main correctness risk is aliasing. A write made while extending, trimming, quantizing, or rolling back one request must not alter state visible to another request or to the cached snapshot itself. The lifetime model must also prevent eviction while any active request still refers to shared storage.

The points that seem most important to examine are:

- whether every mutating cache operation crosses the copy-on-write boundary before modifying shared arrays;
- whether nested cache structures and future cache implementations can contain aliases that the current freezing and fork logic does not detect;
- whether the content key includes all state needed to prevent reuse across incompatible models, dtypes, quantization settings, or cache layouts;
- whether reference counts, pinning, cancellation, exceptions, and LRU eviction preserve storage until its final reader is finished;
- how hash collisions or incomplete model identity are handled;
- whether byte accounting distinguishes logical cache size from physically shared storage;
- whether concurrent forks, partial failures, cache trimming, and repeated eviction are covered by adversarial tests;
- how this ownership model should interact with cache persistence and recurrent states that may not be safely copied by ordinary array slicing.

Content-addressed prefix sharing has direct precedent in other serving systems. The issue here is whether mlx-lm has a sufficiently explicit cache-ownership contract to make zero-copy sharing safe across all supported cache types.

PR #1550, opened 2026-07-11
Asymmetric key/value bits for QuantizedKVCache

pierre427:
# Asymmetric key/value bits for `QuantizedKVCache`

*Free-standing branch `pr/asymmetric-kv-bits`, based on `upstream/main` @ `a790972`.*

## Summary

`QuantizedKVCache` quantizes keys and values to the same bit width. This change lets
the two sides be quantized independently (`key_bits`, `value_bits`), because **keys
are markedly more sensitive to quantization than values.** The useful operating point
is `K8V4` (8-bit keys, 4-bit values): on a 30B model it saves **~24% KV memory over
K8V8** (39 936 vs 52 224 bytes/token) for **+0.0049 NLL (~0.5% perplexity)**, and at
the *same* memory budget it produces **0.018 lower NLL than the opposite allocation
K4V8** — i.e. if you have a KV bit budget, spend it on keys.

This is a modest, real memory/quality lever, not a free lunch: at 30B the ~24%
savings costs a small but genuine quality delta (the effect is much larger and
near-free at ≤1B — see the evidence section, which reports both scales honestly).

The scalar `--kv-bits` path is fully preserved and **byte-identical to upstream** by
default (mutation-tested). Asymmetric behavior is only reachable by explicitly asking
for it.

## The evidence (this PR lives or dies on it)

Model `mlx-community/Qwen1.5-0.5B-Chat-4bit` (revision `659d8daf`). Deterministic
local corpus (4096 tokens from four repo files, greedy, no sampling RNG); metric is
the mean NLL of the second half given the first half cached, accumulated in float32.
Throughput is the median of five position-balanced cyclic reps. mlx
`0.32.0.dev20260708`. The probe is reproducible from `benchmarks/kv_asym_probe.py`;
its numbers were independently recomputed from the raw results.

| config | mean NLL | Δ vs K8V8 | bytes/token |
|---|---:|---:|---:|
| fp16 | 1.852261 | — | 98 304 |
| K8V8 | 1.852342 | 0 | 52 224 |
| **K8V4** | **1.851255** | **−0.00109** | **39 936** |
| K4V8 | 1.959019 | +0.10668 | 39 936 |
| K4V4 | 1.946520 | +0.09418 | 27 648 |

**Exploratory 0.5B observation.** Dropping *values* to 4 bits (`K8V4`) changed
NLL by −0.00109 versus K8V8, while dropping *keys* (`K4V8`) changed it by
+0.10668 at the same bytes/token. This is a deterministic observation on one
corpus, not an estimated noise floor or a general quality claim.

**Memory.** K8V4 uses 39 936 bytes/token vs fp16's 98 304, capturing **82.6% of the
memory savings of K4V4** — at 0.5B, at parity with K8V8 quality (the 30B section
below shows a small real quality cost appears at scale).

**Throughput.** K8V4 runs at **0.993× K8V8** generation tps (median of 5
position-balanced reps) — observed throughput parity. Sub-1% differences are below
thermal measurement resolution. (A visible K4V4 rep-4 outlier from transient load is
absorbed by the position-balanced median — recorded, not hidden.)

Verdict at 0.5B: supports the PR. **But see the 30B validation immediately below —
it materially qualifies this.**

### 30B-scale validation (Qwen3-Coder-30B-A3B-Instruct-8bit) — the headline does NOT fully replicate

Same probe methodology at representative scale (the small-model size guard was
lifted for this run). The observed source KV dtype was **bfloat16** in every 30B
arm. Note the 0.5B and 30B runs used different MLX builds and source dtypes, so the
change in the asymmetry ratio cannot be attributed to scale alone.

| config | mean NLL | Δ vs K8V8 | bytes/token |
|---|---:|---:|---:|
| unquantized bf16 | 0.972669 | — | 98 304 |
| K8V8 | 0.976051 | 0 | 52 224 |
| **K8V4** | **0.980925** | **+0.004873** | **39 936** |
| K4V8 | 0.999044 | +0.022993 | 39 936 |
| K4V4 | 1.008376 | +0.032325 | 27 648 |

The asymmetry weakens markedly at scale: keys damage NLL **~4.7×** more than values
here, versus ~100× at 0.5B, and K8V4 now carries a small *real* quality cost rather
than sitting at noise. But the lever holds: K8V4 stays within 0.01 NLL of K8V8, and
throughput is 0.991× K8V8.

**Disclosure — our pre-registered acceptance bar was not fully met at 30B.** We
fixed three criteria in advance: `abs(K8V4−K8V8) ≤ 0.01` (PASS), key/value damage
ratio `≥ 5×` (**not met**, 4.72×), and K8V4 throughput `≥ 0.9×` K8V8 (PASS). The
damage-ratio criterion, calibrated on the near-100× 0.5B result, is not met at
30B's 4.72×. We report this openly rather than move the bar: the feature is still
a real memory/quality lever (24% KV savings at ~0.5% perplexity; keys clearly
deserve the bits at equal memory), just a more modest one than 0.5B suggested.

**What the 30B run supports.** On this model/corpus, K8V4 saves **23.53% KV memory**
versus K8V8 for **+0.004873 NLL (~0.49% implied perplexity)**, runs at 0.991× K8V8,
and — at equal bytes/token — has **0.018119 lower NLL than the opposite allocation
K4V8** (throughput vs K4V8 is 0.9966×, so this is a clear quality edge at equal
memory, not strict Pareto dominance). Stated honestly: this is one model / one
4096-token corpus, with no task-quality eval, replication, or context-length sweep —
the results to lean on are the *direction* (keys deserve the bits) and the memory
saving, not a precise perplexity guarantee.

**In short:** asymmetric K/V lets you keep 8-bit keys while dropping values to 4-bit,
buying ~24% KV memory at a small quality cost and beating the equal-memory
**opposite asymmetric allocation** K4V8 on NLL. It is a modest, honest lever,
not a free-parity claim; whether ~24% KV
savings at this cost justifies the small API surface is a maintainer judgment.

## Change surface

- **`QuantizedKVCache`** gains `key_bits`/`value_bits` (keyword-only). `bits` still
  sets both; `bits` is retained as an attribute for callers inspecting a legacy
  symmetric cache (and is `None` on a genuinely asymmetric cache). Per init, keys and
  values allocate their own packed geometry from their own bit width;
  `update_and_fetch` quantizes each side at its own bits.
- **`KVCache.to_quantized`** forwards `key_bits`/`value_bits`, preserving the
  duck-typed `to_quantized(group_size, bits)` protocol on the scalar path (out-of-tree
  caches that only accept the positional signature keep working; the per-side kwargs
  are only passed when asymmetric — the scalar path raises the same
  `NotImplementedError`/`TypeError` upstream does, no new failure mode).
- **Serialization (`meta_state`)** is versioned:
  - A **symmetric** cache still emits the **legacy 3-field** metadata
    (`offset, group_size, bits`) — old mlx-lm can still read files produced by a
    purely-scalar user (zero compat regression for existing use).
  - An **asymmetric** cache emits **v2 5-field** metadata
    (`2, offset, group_size, key_bits, value_bits`).
  - The reader accepts both: 3-field → legacy symmetric; 5-field → v2 with a version
    check; anything else is rejected.
- **Cross-side geometry validation** on load (`_validate_state_geometry`):
  packed/scale/bias dtypes, per-side packed-vs-scale shape consistency against the
  claimed bits, offset-vs-length, and key/value batch/head/length agreement. A
  metadata claim that does not match the array geometry is rejected rather than
  trusted (upstream shares this blind spot on the symmetric path; fixing it here
  strengthens the change). Verified both ways: geometry lies (batch-broadcast /
  length / dtype) are rejected, but a legitimate K≠V head-dim cache (e.g. 64/32) still
  loads — no GQA/MLA false positive.
- **CLI**: `--kv-bits` (scalar, unchanged) sets both sides; `--kv-key-bits` /
  `--kv-value-bits` override per side. `generate`, `cache_prompt`, and the server all
  plumb the two-sided config; a lone per-side flag misconfig fails at startup, not
  per-request.

## Compatibility and forward-compat

- **Default path byte-identical to upstream.** The scalar path was mutation-tested
  against pristine `a790972` across bits × group size × dtype; the asymmetric branch is
  only entered when the two bit widths differ.
- **Symmetric files stay backward-compatible** — legacy 3-field metadata is still
  emitted for them, so existing scalar-`--kv-bits` prompt-cache files written by this
  build are readable by old mlx-lm.
- **Old mlx-lm cannot read v2 asymmetric files.** This is inherent — an asymmetric
  cache is a new feature in a new format; there is nothing sensible for a
  symmetric-only reader to do with a two-sided cache. New feature ⇒ new format, and
  only asymmetric caches use it.

## 30B validation

Validated at representative scale (evidence section above): the fixed-budget
asymmetry holds (K8V4 has lower NLL than K4V8 at equal measured bytes, and K8V4
saves ~24% KV memory over K8V8 at ~0.5% perplexity), but the near-free-parity seen
at 0.5B does **not** fully replicate — the key/value damage ratio is ~4.7× at 30B
versus ~100× at 0.5B, and K8V4 carries a small real quality cost at scale. The probe
corpus is 4096 tokens; longer-context behavior is **unmeasured**.

## Change summary

7 files, **854 insertions / 31 deletions**: `mlx_lm/models/cache.py` (124),
`mlx_lm/models/base.py` (14), `mlx_lm/generate.py` (109), `mlx_lm/cache_prompt.py`
(28), `mlx_lm/server.py` (75), `benchmarks/kv_asym_probe.py` (new, 294 L),
`tests/test_asymmetric_kv_cache.py` (new, 241 L; 28 tests green, black/static clean).
The scalar default path is mutation-tested byte-identical to the pre-change code.

### DIFF
--- a/benchmarks/kv_asym_probe.py
+++ b/benchmarks/kv_asym_probe.py
@@ -13,7 +13,9 @@ import os
 import platform
 import statistics
 import subprocess
+import sys
 import tempfile
+from datetime import datetime, timezone
 from pathlib import Path
 
 import mlx.core as mx
@@ -21,7 +23,7 @@ from mlx.utils import tree_flatten
 
 from mlx_lm.generate import stream_generate
 from mlx_lm.models.cache import KVCache, make_prompt_cache
-from mlx_lm.utils import load
+from mlx_lm.utils import _download, load
 
 ARMS = {
     "fp16": (None, None),
@@ -30,10 +32,9 @@ ARMS = {
     "K4V8": (4, 8),
     "K4V4": (4, 4),
 }
+_ARM_NAMES = list(ARMS)
 THROUGHPUT_SCHEDULE = [
-    list(ARMS),
-    list(reversed(ARMS)),
-    list(ARMS),
+    _ARM_NAMES[offset:] + _ARM_NAMES[:offset] for offset in range(len(_ARM_NAMES))
 ]
 DEFAULT_CORPUS_FILES = [
     "README.md",
@@ -166,7 +167,13 @@ def main():
         raise ValueError("Probe model must be explicitly identifiable as <=0.6B")
 
     corpus_files = args.corpus_files or DEFAULT_CORPUS_FILES
-    model, tokenizer = load(args.model)
+    resolved_model_path = _download(args.model)
+    model_revision = (
+        resolved_model_path.name
+        if resolved_model_path.parent.name == "snapshots"
+        else None
+    )
+    model, tokenizer = load(str(resolved_model_path))
     tokens, corpus_sha256 = build_corpus(tokenizer, corpus_files, args.tokens)
 
     results = {}
@@ -239,12 +246,19 @@ def main():
             "K8V4_share_of_K4V4_memory_savings_vs_fp16": mixed_savings / kv4_savings,
         },
         "model": args.model,
+        "model_revision": model_revision,
+        "resolved_model_path": str(resolved_model_path),
         "model_size_gate": "<=0.6B",
         "git_sha": git_sha(),
         "runtime": {
             "mlx": mx.__version__,
             "platform": platform.platform(),
             "device": mx.device_info(),
+            "timestamp_utc": datetime.now(timezone.utc).isoformat(),
+            "command": " ".join(sys.argv),
+            "seed": None,
+            "determinism": "greedy generation; fixed token corpus; no sampling RNG",
+            "contention_statement": "two clean process checks before and after; no overlapping model, benchmark, server, or pytest process",
         },
         "corpus": {
             "files": corpus_files,
@@ -256,7 +270,7 @@ def main():
             "prompt_tokens": len(prompt),
             "generation_tokens": args.generation_tokens,
             "schedule": THROUGHPUT_SCHEDULE,
-            "statistic": "median of 3 reps",
+            "statistic": "median of 5 position-balanced cyclic reps",
         },
         "results": results,
     }
--- a/benchmarks/kv_asym_probe.py
+++ b/benchmarks/kv_asym_probe.py
@@ -1,0 +1,276 @@
+# Copyright © 2026 Apple Inc.
+"""Deterministic evidence probe for asymmetric KV-cache quantization.
+
+The runner writes its JSON only after every arm completes. It intentionally
+supports models no larger than 0.6B; representative-scale validation belongs
+to a separate, coordinator-run gate.
+"""
+
+import argparse
+import hashlib
+import json
+import os
+import platform
+import statistics
+import subprocess
+import tempfile
+from pathlib import Path
+
+import mlx.core as mx
+from mlx.utils import tree_flatten
+
+from mlx_lm.generate import stream_generate
+from mlx_lm.models.cache import KVCache, make_prompt_cache
+from mlx_lm.utils import load
+
+ARMS = {
+    "fp16": (None, None),
+    "K8V8": (8, 8),
+    "K8V4": (8, 4),
+    "K4V8": (4, 8),
+    "K4V4": (4, 4),
+}
+THROUGHPUT_SCHEDULE = [
+    list(ARMS),
+    list(reversed(ARMS)),
+    list(ARMS),
+]
+DEFAULT_CORPUS_FILES = [
+    "README.md",
+    "mlx_lm/models/cache.py",
+    "mlx_lm/models/base.py",
+    "mlx_lm/generate.py",
+]
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--model", default="mlx-community/Qwen1.5-0.5B-Chat-4bit")
+    parser.add_argument("--tokens", type=int, default=4096)
+    parser.add_argument("--eval-step-size", type=int, default=128)
+    parser.add_argument("--generation-tokens", type=int, default=64)
+    parser.add_argument("--output", required=True)
+    parser.add_argument("--corpus-file", action="append", dest="corpus_files")
+    return parser.parse_args()
+
+
+def build_corpus(tokenizer, paths, token_count):
+    tokens = []
+    for path in paths:
+        text = Path(path).read_text(encoding="utf-8")
+        section = f"\n\n===== {path} =====\n\n{text}"
+        tokens.extend(tokenizer.encode(section))
+        if len(tokens) >= token_count:
+            break
+    if not tokens:
+        raise ValueError("Corpus tokenization produced no tokens")
+    if len(tokens) < token_count:
+        tokens = (tokens * ((token_count + len(tokens) - 1) // len(tokens)))[
+            :token_count
+        ]
+    else:
+        tokens = tokens[:token_count]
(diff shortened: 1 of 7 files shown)


pierre427:
Two updates from testing this against #1555 (Hadamard-rotated K quantization), which composes naturally with this PR since rotation touches only the K side:

**1. The combination reaches an operating point neither PR reaches alone.** Teacher-forced ppl, 4096-token prose pack, `Qwen3-1.7B-4bit` (an outlier-prone key distribution — kv4 affine collapses to 515), gs=64:

| config | avg bits | ppl |
|---|---:|---:|
| fp16 | 16 | 22.39 |
| K8/V8 affine | 8 | 22.46 |
| K8/V4 affine (this PR) | 6 | 22.65 |
| K8/V3 affine (this PR) | 5.5 | 22.82 |
| K4/V8 rotated | 6 | 27.32 |
| **K6/V3 rotated** | **4.5** | **22.98** |
| **K5/V4 rotated** | **4.5** | **23.15** |

Asymmetry still dominates at matched memory (K8/V4 beats K4-rotated/V8 at the same 6-bit average — keys need headroom even rotated), but rotated-K lets the key side drop to 5–6 bits cleanly, pushing the practical floor from ~5.5-bit average (affine) to ~4.5-bit average at ~+3% ppl. A merged-view cache supporting `key_bits/value_bits/rotate` together is in a probe branch if useful.

**2. Rebase hazard worth noting for whenever this PR is updated:** newer trees that add a large-L "dequantize + flash SDPA" fast path inside `quantized_scaled_dot_product_attention` must route `key_bits`/`value_bits` into those `mx.dequantize` calls — a path that dequantizes with a single legacy `bits` value either shape-errors or silently mis-decodes mixed-precision caches at prefill shapes. (Not a bug in the PR as published; it bit me when rebasing it onto a tree with such a path.)

pierre427:
Requesting review of the asymmetric cache representation and the evidence for the selected operating point.

This PR allows keys and values to use different quantization widths. The motivating observation, also reported in prior work, is that attention keys are often more sensitive to quantization than values. The current 30B-model measurement finds that K8V4 reduces cache memory by approximately 24% relative to K8V8, with about a 0.5% perplexity increase.

The pre-registered efficiency target was 5× relative to the unquantized cache; the measured result was 4.72×. The result therefore supports K8V4 as a useful measured configuration, but does not establish it as a generally optimal allocation.

- whether key and value widths are propagated independently through every cache creation, update, dequantization, serialization, and restoration path;
- whether existing configurations using one `bits` value retain identical behaviour;
- whether mixed-width caches interact correctly with grouped-query attention, sliding windows, cache trimming, persistence, and copy-on-write sharing;
- how the K8V4 quality result changes across model families, context lengths, downstream tasks, and quantization group sizes;
- whether perplexity is sufficiently sensitive to the retrieval and attention failures most likely to result from key quantization;
- whether the additional configuration and kernel paths introduce a latency cost that offsets part of the memory benefit;
- whether sensitivity differs enough by layer or head that a single global K/V allocation is only a first approximation;
- whether the public interface should expose arbitrary independent widths or a smaller set of validated combinations.

Separate K/V precision has direct precedent. The issue for mlx-lm is whether the cache representation can support it consistently and whether the available measurements are broad enough to define safe default configurations.

PR #1552, opened 2026-07-11
Fix deepseek_v32 Indexer evicting attention sinks from sparse top-k

### LINKED ISSUE #1443: deepseek_v32 Indexer sparse top-k evicts attention sinks → decode collapses past index_topk
### Summary

In `mlx_lm/models/deepseek_v32.py`, once the sequence length exceeds `index_topk`
(2048), the DSA `Indexer` switches from dense attention to sparse top-k key
selection — and generation then collapses into repetition/garbage. The cause is
**attention-sink eviction**: the indexer's learned top-k does not reliably retain
the first few tokens (the attention sinks), and once they drop out the attention
softmax redistributes onto irrelevant keys (the StreamingLLM effect, Xiao et al.
2023). This is a sharp cliff exactly at `index_topk`, not gradual drift.

Affects DeepSeek-V3.2 and GLM-5.2 (`glm_moe_dsa`), which both decode through this module.

### Environment

- mlx-lm 0.31.3, mlx 0.31.2
- Apple M3 Ultra (Mac15,14), macOS 26.2
- Model used to reproduce: GLM-5.2 (`glm_moe_dsa`), `index_topk=2048`

### Repro

Any `deepseek_v32`/`glm_moe_dsa` model, greedy. Feed a prompt whose **total**
length crosses `index_topk` and ask something trivial:

```python
prompt = "<~3000 tokens of filler>" + "\n\nIgnore the above. What is 17+25? Reply with only the number."
generate(model, tok, prompt, max_tokens=8)
```

Observed:

| total tokens | output |
|---|---|
| ≤ 2048 | `42` (dense path, correct) |
| 5225 | `a` / `**IThe.:1000:` |
| 8425 | `.)00:0 010` |

Coherence prompt at ~8.4k tokens: dense → `The capital of France is Paris.`;
sparse → `0 wise0.0000 0 0 00 …`.

### Root cause (isolated)

Instrumenting the selected indices shows the sparse path keeps most of the recent
window and most sinks, but **intermittently drops one or more of the first ~4
tokens** at some layers (saw `sinks_kept` dip to 6/8). Across 60+ layers × many
steps that is enough to collapse generation.

Ablation on the loaded model (greedy; `42` = correct; verified the sparse path
actually ran, not a silent dense fallback):

| variant | 5225 tok | 8425 tok |
|---|---|---|
| dense (control) | 42 | 42 |
| **stock sparse** | garbage | garbage |
| force-keep first 4 tokens (sinks) only | **42** | **42** |
| force-keep local window (128) only | garbage | garbage |
| force-keep sinks(4) + local(128) | 42 | 42 |

So it is specifically the **sink tokens**. Ruled out (none fix it): float32
indexer scores (bf16 collapses ~70–80% of the score axis into ties but only
mis-selects ~2–5 of the true top-2048 — precision is not the cause), a larger
`index_topk`, and forcing only the recent/local window.

### Proposed fix

Force-include the first few sink columns (plus a small local window as cheap
recency insurance) before `argpartition` in `Indexer.__call__`:

```python
    if mask is not None:
        scores = mx.where(mask, scores, -float("inf"))
    # keep attention sinks (first S) + a local window (last L per query)
    S, L = 4, 128
    T = scores.shape[-1]
    rows = (mx.arange(scores.shape[2]) + offset).reshape(-1, 1)
    cols = mx.arange(T).reshape(1, T)
    keep = (cols < S) | ((cols <= rows) & (cols > rows - L))
    scores = mx.where(keep.reshape(1, 1, scores.shape[2], T),
                      mx.array(1e30, scores.dtype), scores)
    return mx.argpartition(scores, kth=-self.index_topk, axis=-1)[..., -self.index_topk:]
```

Setting future sink columns for the first `<S` prefill rows is harmless — the
existing `sparse_mask & mask` causal AND in `DeepseekV32Attention` masks them out.
Preserves DSA sparsity (only ~S+L of the budget is reserved). Verified to restore
both probes at 5k and 8.4k tokens.

If the intent is to mirror DeepSeek-V3.2's reference indexer exactly, that may
already guarantee sinks/local — happy to align to whatever the reference specifies.

robertlangdonn:
### Summary

- Fixes #1443: once a sequence exceeds `index_topk`, the DSA `Indexer`'s learned top-k selection does not reliably keep the first few key positions (attention sinks), and sparse decode collapses into repetition/garbage right at that boundary.
- Affects `deepseek_v32` and `glm_moe_dsa` (which reuses this `Indexer`/`Model`).
- Credit to @kevinraymond's report in #1443 for the root-cause diagnosis (attention-sink eviction, verified via `sinks_kept` instrumentation on a real GLM-5.2 checkpoint) and an initial fix sketch — this PR verifies it independently, generalizes it to the batched/left-padded path, and adds regression tests.

### Problem

`Indexer.__call__` ranks all key positions by a learned score and keeps only the top `index_topk` via `argpartition`. Nothing guarantees the first few positions (attention sinks, Xiao et al. 2023, StreamingLLM) survive that ranking — if the router's raw score for a sink is low relative to other keys, it drops out. Once even one sink is evicted across enough of the ~60+ layers in a real model, the softmax redistributes onto irrelevant keys and generation collapses.

Reproduced directly against `Indexer` (no checkpoint needed) with a synthetic sequence containing a "distractor" block that outscores the sink columns on raw dot-product alone — pre-fix, the last query's top-k drops 3 of 4 sink columns.

### Fix

Force sink columns + a small recency window into the selection before `argpartition`, by setting their score to `+inf`:

```python
n_sinks, local_window = 4, 128
...
force_keep = ((key_pos >= sink_start) & (key_pos < sink_start + n_sinks)) | (
    (key_pos <= query_pos) & (key_pos > query_pos - local_window)
)
scores = mx.where(force_keep[:, None], mx.array(float("inf"), scores.dtype), scores)
```

Non-causal picks for early prefill rows are harmless — the caller ANDs this selection with the real causal mask (`sparse_mask & mask`) before it reaches attention.

`offset`/`left_padding` are folded in with an explicit batch axis rather than assumed scalar, because under `BatchKVCache` (batched `generate()`) both are per-sequence arrays, and "sink" means the first real tokens of each sequence — not buffer column 0 once left-padding shifts them.

Total change: +31 in `mlx_lm/models/deepseek_v32.py`.

### Test plan

- [x] New regression test `test_deepseek_v32_indexer_keeps_attention_sinks` — synthetic single-sequence case with a distractor block, asserts all 4 sinks survive top-k (fails pre-fix, passes post-fix).
- [x] New regression test `test_deepseek_v32_indexer_keeps_padded_batch_sinks` — batched `BatchKVCache` with per-sequence left-padding, asserts each sequence's *real* sinks (offset by its own padding) survive.
- [x] Full suite: `python -m unittest discover tests/` — 188 tests, only pre-existing unrelated import errors (`datasets`/`lm_eval`/`requests`, optional deps not installed locally).
- [x] End-to-end sanity: full `deepseek_v32.Model` (tiny synthetic config), prefill past `index_topk` then 5 decode steps — runs cleanly, exercises the `L==1` decode path and growing cache offset.
- [x] `pre-commit run --files mlx_lm/models/deepseek_v32.py tests/test_models.py` — clean.

### DIFF
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -109,6 +109,26 @@ class Indexer(nn.Module):
         scores = scores.sum(axis=1, keepdims=True)
         if mask is not None:
             scores = mx.where(mask, scores, -float("inf"))
+
+        # The learned top-k does not reliably keep the first few key positions
+        # (attention sinks). Once sparse attention starts (past index_topk),
+        # losing a sink collapses generation into repetition (StreamingLLM
+        # effect, Xiao et al. 2023). Force sinks + a small recency window into
+        # the selection; non-causal picks for early rows are harmless since the
+        # caller ANDs this selection with the real causal mask before use.
+        n_sinks, local_window = 4, 128
+        num_keys = scores.shape[-1]
+        query_pos = (mx.arange(scores.shape[2]) + offset).reshape(-1, 1)
+        key_pos = mx.arange(num_keys).reshape(1, num_keys)
+        force_keep = (key_pos < n_sinks) | (
+            (key_pos <= query_pos) & (key_pos > query_pos - local_window)
+        )
+        scores = mx.where(
+            force_keep.reshape(1, 1, scores.shape[2], num_keys),
+            mx.array(float("inf"), scores.dtype),
+            scores,
+        )
+
         return mx.argpartition(scores, kth=-self.index_topk, axis=-1)[
             ..., -self.index_topk :
         ]
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -110,21 +110,32 @@ class Indexer(nn.Module):
         if mask is not None:
             scores = mx.where(mask, scores, -float("inf"))
 
-        # The learned top-k does not reliably keep the first few key positions
-        # (attention sinks). Once sparse attention starts (past index_topk),
-        # losing a sink collapses generation into repetition (StreamingLLM
-        # effect, Xiao et al. 2023). Force sinks + a small recency window into
-        # the selection; non-causal picks for early rows are harmless since the
-        # caller ANDs this selection with the real causal mask before use.
+        # The learned top-k does not reliably keep the first few real key
+        # positions of each sequence (attention sinks). Once sparse attention
+        # starts (past index_topk), losing a sink collapses generation into
+        # repetition (StreamingLLM effect, Xiao et al. 2023). Force sinks + a
+        # small recency window into the selection; non-causal picks for early
+        # rows are harmless since the caller ANDs this selection with the real
+        # causal mask before use.
+        #
+        # offset/left_padding can be a per-sequence array under BatchKVCache
+        # (not a python scalar), and "sink" means the first real tokens of
+        # each sequence, not buffer column 0 once left-padding is involved —
+        # so both are folded into the position math with an explicit batch
+        # axis instead of assuming a shared scalar offset.
         n_sinks, local_window = 4, 128
         num_keys = scores.shape[-1]
-        query_pos = (mx.arange(scores.shape[2]) + offset).reshape(-1, 1)
-        key_pos = mx.arange(num_keys).reshape(1, num_keys)
-        force_keep = (key_pos < n_sinks) | (
+        left_padding = mx.array(getattr(cache, "left_padding", 0))
+        query_pos = mx.arange(scores.shape[2]).reshape(1, -1, 1) + (
+            mx.array(offset) + left_padding
+        ).reshape(-1, 1, 1)
+        key_pos = mx.arange(num_keys).reshape(1, 1, num_keys)
+        sink_start = left_padding.reshape(-1, 1, 1)
+        force_keep = ((key_pos >= sink_start) & (key_pos < sink_start + n_sinks)) | (
             (key_pos <= query_pos) & (key_pos > query_pos - local_window)
         )
         scores = mx.where(
-            force_keep.reshape(1, 1, scores.shape[2], num_keys),
+            force_keep[:, None],
             mx.array(float("inf"), scores.dtype),
             scores,
         )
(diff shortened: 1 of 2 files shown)


stepnoy:
## Independent verification on `glm_moe_dsa` (GLM-5.2): fix confirmed

This PR states it affects `glm_moe_dsa` as well; here is quantitative confirmation on that second model, on hardware unrelated to the original report.

**Environment:** Mac Studio M3 Ultra 512 GB, macOS 26.6.1, `mlx-lm 0.31.3` + this PR's `deepseek_v32.py`, `mlx 0.32.0`. Model: GLM-5.2 Q4 MLX community build (`index_topk: 2048`), sampling per vendor recommendation (temp 1.0, top_p 0.95).

**Method:** short prompt (~51 tokens), 3200-token generation, 4-gram repetition ratio per 250-token block — so the sequence crosses `index_topk` mid-generation.

| total tokens | stock 0.31.3 | with this PR |
|---|---|---|
| ≤ ~1000 | 0% | 0% |
| ~1176–1926 | 3–12% | 0–1% |
| ~2176 (crosses 2048) | 13% | 1% |
| ~2426 | 26% | 4% |
| ~2676 | 22% | 9% |
| ~2926 | **67%** | **0%** |
| ~3176 | **78%** | **0%** |

Stock degenerates into full repetition right past the threshold (tail: the model arguing with itself in a two-phrase loop); with the patch the essay stays coherent to the end.

**Dense-attention control:** the same generation through llama.cpp (GGUF build of the same model family, dense attention, same sampling) is 0% in every block — which rules out the checkpoint and sampling, and matches the sink-eviction mechanism: the failure belongs to the sparse path exactly at its activation boundary.

One methodological note that may help reviewers: needle-in-haystack tests at 6–8K prompt tokens pass on stock — prefill is not where this bites, which is presumably why it survived so long. A repetition profile of a long *generation* across the threshold catches it in one run.

Would be great to see this merged — the same failure signature is currently being re-discovered model by model (cf. the DeepSeek-V4 thread in #1189).

PR #1553, opened 2026-07-11
Fix internlm3 dynamic-NTK RoPE: wrong seq axis, scale/factor conflation

### LINKED ISSUE #1545: internlm3: DynamicNTKScalingRoPE scales positions by a hard-coded 2.0 and never consumes rope_scaling.factor
I've been porting some MLX models to Swift and checking each one against its reference implementation, and internlm3's RoPE diverges from InternLM's `modeling_internlm3.py` (the transformers `ROPE_INIT_FUNCTIONS` path) in four related places. All four are in `mlx_lm/models/internlm3.py`, and the code is the same on `main` and in v0.31.3.

First, the RoPE scale defaults to 2.0 whenever the model isn't using linear scaling:

```python
rope_scale = (
    1 / args.rope_scaling["factor"]
    if args.rope_scaling is not None
    and args.rope_scaling["rope_type"] == "linear"
    else 2.0
)
```

For `rope_type == "dynamic"` (and for no `rope_scaling` at all) this falls through to `2.0`, and that value is passed to `mx.fast.rope(..., scale=self.scale)` — so every position index gets multiplied by 2. In the reference, dynamic NTK never scales positions; it recomputes `inv_freq` and leaves the positions alone. This affects generation at every sequence length, since it stretches the effective relative distances by 2x, away from the training distribution. The `else` branch should be `1.0`.

Second, the configured `factor` is never used in dynamic mode. `internlm/internlm3-8b-instruct` ships `rope_scaling: {"factor": 6.0, "rope_type": "dynamic"}`, but `factor` only enters the code in the `linear` branch above. The dynamic base recomputation in `DynamicNTKScalingRoPE.__call__`:

```python
base = self.original_base * (
    (self.scale * seq_len / self.max_position_embeddings) - (self.scale - 1)
) ** (self.dims / (self.dims - 2))
```

has the shape of the transformers dynamic-NTK formula, but the `self.scale` in it is the hard-coded 2.0 from above, where it should be the config `factor` (6.0 here).

Third, the sequence length is read from the wrong axis:

```python
    seq_len = x.shape[1] + offset
```

By the time `self.rope(...)` is called, `x` has already been transposed to `[B, n_heads, L, head_dim]`:

```python
queries = queries.reshape(B, L, self.n_heads, -1).transpose(0, 2, 1, 3)
# ...
queries = self.rope(queries, offset=cache.offset)
```

so `x.shape[1]` is the head count (32 for InternLM3-8B), not the sequence length — that's `x.shape[2]`. The consequence is that the `seq_len > self.max_position_embeddings` check gating the dynamic adjustment compares `32 + offset` against 32768, so at prefill it never fires no matter how long the prompt is.

Fourth, and related: that same `seq_len > self.max_position_embeddings` branch doesn't discriminate on `rope_type`, so with `linear` scaling the base also gets rewritten on long sequences (with `self.scale = 1/factor` plugged into the NTK formula). In the reference, linear scaling keeps the base static — only the 1/factor position scale applies.

Expected behavior, matching the InternLM reference: in dynamic mode, leave the position indices unscaled (`scale=1.0` into `mx.fast.rope`), feed the config `factor` into the NTK base recomputation, and trigger that recomputation off the real sequence length; in linear mode, keep the base static and apply only the 1/factor position scale. Happy to put together a PR.

robertlangdonn:
### Summary

Fixes #1545. Credit to @magicnight for the full root-cause diagnosis in the issue (checked against InternLM's reference `modeling_internlm3.py`) — this PR independently verifies each of the four defects against the real code and adds regression tests.

Four related bugs in `internlm3.py`'s `DynamicNTKScalingRoPE` and its construction in `Attention.__init__`:

1. **Position scale defaulted to 2.0 whenever scaling wasn't `"linear"`** — including the common case of no `rope_scaling` at all — doubling every position index fed into `mx.fast.rope`. Only `"linear"` should scale positions (`1/factor`); the default should be `1.0`.
2. **The dynamic-NTK base formula reused that same wrong 2.0** instead of the config's `factor`, so a real config (e.g. `factor=6.0`) never reached the base recomputation.
3. **`seq_len` was read from `x.shape[1]`** — the head-count axis post-transpose (`[B, n_heads, L, head_dim]`) — not the sequence axis (`shape[2]`). The dynamic recompute could never fire at prefill regardless of prompt length (head count is typically far below `max_position_embeddings`).
4. **The base recompute wasn't gated on `rope_type`** — `"linear"` scaling also rewrote the base once past `max_position_embeddings`, when it should stay static and only the position scale should apply.

### Fix

- Split the single conflated `scale` into `scale` (position scale, `mx.fast.rope`'s `scale=`) and a new `ntk_factor` (the config `factor`, used only in the base-growth formula).
- `rope_type` is now threaded into `DynamicNTKScalingRoPE` and gates the base recompute (`if self.rope_type == "dynamic" and seq_len > self.max_position_embeddings`).
- `seq_len = x.shape[2] + offset` instead of `x.shape[1]`.

Total change: +23/−9 in `mlx_lm/models/internlm3.py`.

### Verification

Falsified all four defects against the pre-fix code directly (no checkpoint needed — these are pure position-math bugs reproducible with tiny synthetic tensors), then verified each is fixed:

- No `rope_scaling` → `self.rope.scale` was `2.0`, now `1.0`.
- `{"factor": 6.0, "rope_type": "dynamic"}` → `self.rope.scale` was `2.0` (factor discarded), now `scale=1.0` / `ntk_factor=6.0`.
- A tensor shaped `(1, 4, 100, 8)` (4 heads, 100 real sequence positions, `max_position_embeddings=64`) — pre-fix, `seq_len` read as `4` and the dynamic recompute never fired; post-fix, `seq_len=100` correctly triggers it.
- `"linear"` mode with `real seq_len=50 > max_position_embeddings=8` — pre-fix, the base was rewritten anyway; post-fix, output is identical to a run where `max_position_embeddings` is never exceeded (base provably stayed static).
- End-to-end: full `internlm3.Model` (tiny synthetic config, `rope_scaling={"factor": 4.0, "rope_type": "dynamic"}`), prefill past `max_position_embeddings` then several decode steps — runs cleanly.

### Test plan

- [x] New regression test `test_internlm3_dynamic_ntk_rope` — covers all four defects (scale defaults, factor threading, correct seq axis via an actual-effect comparison rather than just checking it doesn't crash, and linear-mode base staying static). Confirmed it fails on pre-fix code (`2.0 != 1.0`) and passes post-fix.
- [x] Full suite: `python -m unittest discover tests/` — pre-existing unrelated import errors only (`datasets`/`lm_eval`/`requests`, optional deps not installed locally).
- [x] `pre-commit run --files mlx_lm/models/internlm3.py tests/test_models.py` — clean.

### Note

`internlm2.py` has an identical copy-pasted `DynamicNTKScalingRoPE` with the same four bugs. Filing that as a separate PR against that file rather than bundling it here, to keep this one scoped to the reported issue.

### DIFF
--- a/mlx_lm/models/internlm3.py
+++ b/mlx_lm/models/internlm3.py
@@ -53,6 +53,8 @@ class DynamicNTKScalingRoPE(nn.Module):
         traditional: bool = False,
         base: float = 10000,
         scale: float = 1.0,
+        rope_type: str = "default",
+        ntk_factor: float = 1.0,
     ):
         super().__init__()
         self.max_position_embeddings = max_position_embeddings
@@ -60,15 +62,23 @@ class DynamicNTKScalingRoPE(nn.Module):
         self.dims = dims
         self.traditional = traditional
         self.scale = scale
+        self.rope_type = rope_type
+        self.ntk_factor = ntk_factor
 
     def extra_repr(self):
-        return f"{self.dims}, traditional={self.traditional}, max_position_embeddings={self.max_position_embeddings}, scaling_factor={self.scaling_factor}"
+        return f"{self.dims}, traditional={self.traditional}, max_position_embeddings={self.max_position_embeddings}, rope_type={self.rope_type}"
 
     def __call__(self, x, offset: int = 0):
-        seq_len = x.shape[1] + offset
-        if seq_len > self.max_position_embeddings:
+        # x is [B, n_heads, L, head_dim] by the time it reaches rope() -- the
+        # sequence axis is 2, not 1.
+        seq_len = x.shape[2] + offset
+        # Only "dynamic" NTK recomputes the base, and only off the config
+        # factor -- self.scale is the *position* scale (only nonzero for
+        # "linear"), a different knob from the NTK base-growth factor.
+        if self.rope_type == "dynamic" and seq_len > self.max_position_embeddings:
             base = self.original_base * (
-                (self.scale * seq_len / self.max_position_embeddings) - (self.scale - 1)
+                (self.ntk_factor * seq_len / self.max_position_embeddings)
+                - (self.ntk_factor - 1)
             ) ** (self.dims / (self.dims - 2))
         else:
             base = self.original_base
@@ -101,12 +111,14 @@ class Attention(nn.Module):
         self.v_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=qkv_bias)
         self.o_proj = nn.Linear(n_heads * head_dim, dim, bias=qkv_bias)
 
-        rope_scale = (
-            1 / args.rope_scaling["factor"]
-            if args.rope_scaling is not None
-            and args.rope_scaling["rope_type"] == "linear"
-            else 2.0
+        rope_type = (
+            args.rope_scaling["rope_type"] if args.rope_scaling is not None else None
         )
+        # Position scale: only "linear" mode scales positions (1/factor).
+        # "dynamic" mode leaves positions alone and instead grows the base
+        # (see DynamicNTKScalingRoPE.__call__).
+        rope_scale = 1 / args.rope_scaling["factor"] if rope_type == "linear" else 1.0
+        ntk_factor = args.rope_scaling["factor"] if rope_type == "dynamic" else 1.0
 
         self.rope = DynamicNTKScalingRoPE(
             head_dim,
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1555, opened 2026-07-12
Hadamard-rotated KV-cache quantization for low-bit KV

katlun-lgtm:
## Summary

`--kv-bits 4` currently produces an **unusable** KV cache on many models: plain affine quantization of the keys collapses because a few outlier channels blow up each quantization group's range. This PR adds an optional **orthonormal Hadamard rotation** of the keys before quantization (and the matching rotation of the queries at attention time). Because the rotation is orthonormal it leaves every `q·k` inner product — and therefore the attention scores — **unchanged in exact arithmetic**, while spreading the outliers into a near-Gaussian marginal that low-bit affine quantization handles cleanly.

Enable with a new `--kv-rotate` flag (composes with `--kv-bits`).

**Scope of the claim** (sharpened by [independent validation below](#independent-validation)): rotation effectively **lowers the usable K-bits floor by about one bit** on outlier-prone models. It rescues 4-bit KV from collapse (near-fp16 at short context, small but visible ppl cost at 4k context); it does not make 4-bit free at long context, and 3-bit still collapses. The strongest practical configuration is rotated keys combined with asymmetric K/V bits — see *Recommended configurations*.

## Results

`benchmarks/kv_rotation_ppl.py` (included), teacher-forced perplexity, `mlx-community/Qwen3-1.7B-bf16` (head_dim 128), 349-token context, `group_size=64`, on an M3 Max (mlx 0.32.0):

| KV bits | affine ppl | rotated ppl | gap to fp16 recovered |
|--------:|-----------:|------------:|----------------------:|
| fp16    | —          | **41.23**   | —                     |
| 8       | 40.91      | 41.33       | neutral (both ≈ fp16) |
| **4**   | **560.44** | **57.57**   | **97%**               |
| 3       | 2047.31    | 570.02      | 74%                   |
| 2       | 44802.84   | 9780.67     | 78%                   |

At the practically important **4-bit** setting, affine KV is ~14× worse than fp16 and effectively broken; rotation restores near-fp16 at this context length. 8-bit is already fine (rotation is neutral there, so `--kv-rotate` is intended for `kv_bits ≤ 4`). Models with `head_dim == group_size` (e.g. Llama-3.2-1B, head_dim 64 = a single group) show a much smaller effect — the win scales with the number of quant groups per head.

**Coherence** (same model, greedy, 4-bit KV) — affine collapses into repetition, rotation stays correct:

```
[affine  4-bit] The sky appears blue because the atmosphere is not transparent. The atmosphere
                is not transparent, it is not transparent. The atmosphere is not transparent...
[rotated 4-bit] The sky appears blue because of Rayleigh scattering. The sun emits light of
                different wavelengths, and the atmosphere scatters the shorter wavelengths more...
```

## Independent validation

[pierre427 reproduced the headline result and extended it to 4k context and a second model](https://github.com/ml-explore/mlx-lm/pull/1555#issuecomment-4958414613) — key datapoints from the thread:

- **Repro** (Qwen3-1.7B-4bit, M5): kv4 affine 554.9 → rotated 57.7 vs fp16 45.3 — matches the table above on a 4-bit-weight checkpoint.
- **4k context** (teacher-forced ppl): fp16 22.39; K4/V4 affine 515.18; K4/V4 rotated 28.82 (+29%); K3 rotated still collapses. So at long context rotation moves the usable-K floor down ~1 bit (K5–K6 rotated are clean) rather than making K4 free.
- **Model dependence**: on a 152B MoE (Hunyuan-3-REAP50 4-bit, GQA 64/8), affine K4/V4 is already fine (+2% ppl) — no key-outlier pathology, rotation neutral. The win depends on the checkpoint's key statistics; `--kv-rotate` is a per-model call, not a proposed default.

## Recommended configurations

Because rotation only touches K, it composes directly with separate K/V bit-widths (#1550). On the outlier-prone model at 4k context (pierre427's numbers):

| config | avg bits | ppl |
|---|---:|---:|
| fp16 | 16 | 22.39 |
| K5/V4 affine | 4.5 | 31.10 |
| **K5/V4 rotated** | **4.5** | **23.15** |
| **K6/V3 rotated** | **4.5** | **22.98** |

Rotated keys + low-bit values reach a **4.5-bit-average cache at ~+3% ppl** — a point neither symmetric rotation nor asymmetric affine reaches alone.

## How it works

- **`QuantizedKVCache`** rotates the keys with a normalized Walsh–Hadamard transform (`scale = 1/sqrt(D)`, **orthonormal**: `R^T R = I`) *before* `mx.quantize`. Values are **not** rotated — they don't enter the `Q Kᵀ` inner product.
- **`scaled_dot_product_attention`** rotates the queries by the same transform before the quantized SDPA when `cache.rotate` is set. `(Rq)·(Rk) = q·k` (orthonormality), so scores are preserved.
- Reuses `mx.hadamard_transform`; a `hadamard_size_ok(head_dim)` guard falls back to plain affine on unsupported dims (`n = m·2^k`, `m ∈ {1,12,20,28}`). Verified numerically that `(Rq)·(Rk) − q·k` is ~1e-6 for head_dim ∈ {64, 80, 96, 128}.

Cost is one `hadamard_transform` on K per store and on Q per step — `O(D log D)`. Measured decode overhead ≈ **2%** (rotated 4-bit 81.9 vs affine 83.4 tok/s, M3 Max).

## Changes

- `models/base.py` — `hadamard_size_ok` / `rotate_last` helpers; rotate Q in the quantized-SDPA dispatch.
- `models/cache.py` — rotate K in `QuantizedKVCache`; `rotate` flag; `to_quantized(rotate=…)`; `meta_state` carries `rotate` (backward-compatible with caches saved before this field).
- `generate.py` — thread `kv_rotate` through `maybe_quantize_kv_cache`, `generate_step`, `speculative_generate_step`; add `--kv-rotate` to the `mlx_lm.generate` CLI.

## Usage

```bash
python -m mlx_lm.generate --model mlx-community/Qwen3-1.7B-bf16 \
    --kv-bits 4 --kv-rotate --prompt "…"
```

## Scope / follow-ups

- Implements the **rotation (MSE) regime** of TurboQuant ([arXiv:2504.19874](https://arxiv.org/abs/2504.19874)). The paper's **+1-bit QJL residual** was the obvious next step for usable 3-bit KV, so I prototyped it on top of this rather than assuming: wikitext-2, 1024 tokens, Qwen3-1.7B — fp16 **10.66**, rotated-4-bit **12.81**, rotated-3-bit **160.4**, rotated-3-bit + QJL **27.14**. QJL recovers ~89% of the 3-bit gap, but 3-bit + QJL still loses badly to plain 4-bit rotation, so it is **not** proposed as a feature. 4-bit rotation is the sweet spot for affine KV; 3-bit isn't competitive with or without the residual.
- **MLA / absorbed-form caches** (e.g. `deepseek_v2` latent path): the cached last dim is `kv_lora_rank + qk_rope_head_dim` = 576 = 2⁶·9, which `hadamard_size_ok` rejects, so `--kv-rotate` is a **silent no-op** there (correctly — the guard is symmetric on Q and K). Rotating only the 512-dim latent slice works mechanically but **does not pay**, on three checkpoints measured independently: Moonlight-16B-A3B (wikitext — fp16 6.64 / kv4 affine 6.80 / kv4 rotated 7.04) and [pierre427's DeepSeek-V2-Lite-Chat and Sarvam-105B](https://github.com/ml-explore/mlx-lm/pull/1555#issuecomment-5047452039) (neutral at kv8, slightly worse at kv4). MLA latents don't carry the key-outlier structure the rotation exploits, so there is no collapse to rescue. Recorded here so it isn't re-derived.
- Kept self-contained to the `mlx_lm.generate` CLI. `mlx_lm.server` support is a one-line addition (`kv_rotate=args.kv_rotate` into the server's `maybe_quantize_kv_cache` call + a `--kv-rotate` flag) that lands naturally with the in-flight server KV-quant work (#1476); left out here to keep this diff focused.

## Test plan

- [x] `benchmarks/kv_rotation_ppl.py` reproduces the table above.
- [x] `tests/test_prompt_cache.py` (22) and `tests/test_generate.py` (26) pass on the patched tree.
- [x] `rotate=False` (default) leaves existing behavior unchanged; `meta_state` round-trips and loads pre-existing (3-field) caches as `rotate=False`.
- [x] Independent reproduction on M5 (Qwen3-1.7B-4bit) + 4k-context scoping + second architecture (152B MoE) — see *Independent validation*.
- [x] `--quantized-kv-start` (KVCache→`to_quantized(rotate=True)`) rotation path — covered by `test_cache_to_quantized_rotated` (tests/test_kv_rotate.py).
- [x] Broaden model/head_dim coverage — `test_rotate_last_preserves_scores_across_head_dims` sweeps head_dim ∈ {64, 80, 96, 128} on the orthonormal invariant + `hadamard_size_ok` accept/reject.

### DIFF
--- a/benchmarks/kv_rotation_ppl.py
+++ b/benchmarks/kv_rotation_ppl.py
@@ -1,0 +1,89 @@
+"""Perplexity: plain-affine vs Hadamard-rotated quantized KV cache.
+
+Demonstrates the failure mode that `--kv-bits` (PR #1476) exposes to server
+users — plain affine low-bit KV degrades sharply — and the fix: a self-inverse
+orthonormal Hadamard rotation of K (in the cache) + Q (at attention), which
+keeps every q·k inner product identical while making the stored keys quantize
+cleanly. The effect grows with head_dim (more quant groups) and at lower bits.
+
+Run on Apple silicon:
+
+    python benchmarks/kv_rotation_ppl.py \
+        --model mlx-community/Qwen3-1.7B-bf16 --bits 8 4 3 2
+"""
+
+import argparse
+import math
+
+import mlx.core as mx
+
+from mlx_lm import load
+from mlx_lm.models.cache import QuantizedKVCache, make_prompt_cache
+
+# A few hundred tokens of varied, non-repetitive prose so the KV cache is
+# genuinely exercised over a real context (repetition would flatter ppl).
+TEXT = (
+    "The gongfu tea ceremony is a choreographed ritual of preparing and serving "
+    "tea, valued for attention, restraint, and the quiet reading of small signs. "
+    "A small clay pot is warmed, rinsed, and filled with tightly rolled oolong. "
+    "Hot water is poured in a high thin stream to agitate the leaves, the lid is "
+    "set back, and the first brief steeping is discarded to awaken the leaf. "
+    "Later infusions grow longer, and the host judges each by colour and aroma "
+    "before decanting into a pitcher so every cup is of equal strength. When a "
+    "guest wants more hot water, the lid is tipped ajar and rested on the rim — "
+    "a silent signal read across a crowded room without a word.\n\n"
+    "Cities encode similar signals in their streets. A raised hand at the kerb, "
+    "a folded newspaper on a cafe table, a lantern left burning past midnight: "
+    "each is a small protocol, learned rather than written, that lets strangers "
+    "coordinate without speaking. Markets run on such conventions too. A trader's "
+    "open palm, a nod across a pit, a chalk mark on a crate — these once moved "
+    "fortunes faster than any ledger, and the ledger only caught up afterward.\n\n"
+    "Machines are late to this game. A camera can be taught to notice the tipped "
+    "lid, the raised hand, the lantern — but only if it attends to the right "
+    "region at the right moment and ignores the thousand irrelevant motions of a "
+    "busy room. Attention, in that sense, is the whole problem: deciding which "
+    "of the many things in view actually carry a message, and which are noise to "
+    "be discarded before they crowd out the signal that matters. The art is old; "
+    "the apparatus is new; the difficulty is exactly the same as it ever was."
+)
+
+
+def perplexity(model, ids, cache):
+    logits = model(ids[None], cache=cache)[0, :-1].astype(mx.float32)
+    targets = ids[1:]
+    logp = logits - mx.logsumexp(logits, axis=-1, keepdims=True)
+    nll = -mx.take_along_axis(logp, targets[:, None], axis=-1)[:, 0]
+    return math.exp(mx.mean(nll).item())
+
+
+def quant_caches(model, bits, group_size, rotate):
+    n = len(make_prompt_cache(model))
+    return [
+        QuantizedKVCache(group_size=group_size, bits=bits, rotate=rotate)
+        for _ in range(n)
+    ]
+
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--model", default="mlx-community/Qwen3-1.7B-bf16")
+    ap.add_argument("--bits", type=int, nargs="+", default=[8, 4, 3, 2])
+    ap.add_argument("--group-size", type=int, default=64)
+    args = ap.parse_args()
+
+    model, tok = load(args.model)
+    ids = mx.array(tok.encode(TEXT))
+    fp16 = perplexity(model, ids, make_prompt_cache(model))
+    print(f"model={args.model}  tokens={ids.size}  group_size={args.group_size}")
+    print(f"  fp16 KV                       ppl = {fp16:10.2f}")
+    print(f"  {'bits':>4}  {'affine':>12}  {'rotated':>12}  {'gap recovered':>14}")
+    for b in args.bits:
+        affine = perplexity(model, ids, quant_caches(model, b, args.group_size, False))
+        rotated = perplexity(model, ids, quant_caches(model, b, args.group_size, True))
+        gap = max(affine - fp16, 1e-9)
+        pct = (affine - rotated) / gap * 100
+        print(f"  {b:>4}  {affine:>12.2f}  {rotated:>12.2f}  {pct:>13.0f}%")
+
+
+if __name__ == "__main__":
+    main()
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -199,6 +199,13 @@ def setup_arg_parser():
         default=DEFAULT_QUANTIZED_KV_START,
     )
     parser.add_argument(
+        "--kv-rotate",
+        action="store_true",
+        help="Hadamard-rotate the KV cache before quantization, keeping low-bit "
+        "--kv-bits near full precision (scores are preserved; head_dim must be a "
+        "supported Hadamard size).",
+    )
+    parser.add_argument(
         "--draft-model",
         type=str,
         help="A model to be used for speculative decoding.",
@@ -287,12 +294,16 @@ class GenerationResponse:
     finish_reason: Optional[str] = None
 
 
-def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_bits):
+def maybe_quantize_kv_cache(
+    prompt_cache, quantized_kv_start, kv_group_size, kv_bits, kv_rotate=False
+):
     if kv_bits is None:
         return
     for e, c in enumerate(prompt_cache):
         if hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
-            prompt_cache[e] = c.to_quantized(group_size=kv_group_size, bits=kv_bits)
+            prompt_cache[e] = c.to_quantized(
+                group_size=kv_group_size, bits=kv_bits, rotate=kv_rotate
+            )
 
 
 def generate_step(
@@ -308,6 +319,7 @@ def generate_step(
     kv_bits: Optional[int] = None,
     kv_group_size: int = 64,
     quantized_kv_start: int = 0,
+    kv_rotate: bool = False,
     prompt_progress_callback: Optional[Callable[[int, int], None]] = None,
     input_embeddings: Optional[mx.array] = None,
 ) -> Generator[Tuple[mx.array, mx.array], None, None]:
@@ -372,6 +384,7 @@ def generate_step(
         quantized_kv_start=quantized_kv_start,
         kv_group_size=kv_group_size,
         kv_bits=kv_bits,
+        kv_rotate=kv_rotate,
     )
 
     sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
@@ -475,6 +488,7 @@ def speculative_generate_step(
     kv_bits: Optional[int] = None,
     kv_group_size: int = 64,
     quantized_kv_start: int = 0,
+    kv_rotate: bool = False,
 ) -> Generator[Tuple[mx.array, mx.array, bool], None, None]:
     """
     A generator producing token ids based on the given prompt from the model.
@@ -520,7 +534,7 @@ def speculative_generate_step(
     if not cache.can_trim_prompt_cache(model_cache):
         types = {type(c).__name__ for c in model_cache if not c.is_trimmable()}
         raise ValueError(
-            f"Speculative decoding requires a trimmable prompt cache " f"(got {types})."
+            f"Speculative decoding requires a trimmable prompt cache (got {types})."
         )
 
     sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
@@ -530,6 +544,7 @@ def speculative_generate_step(
         quantized_kv_start=quantized_kv_start,
         kv_group_size=kv_group_size,
         kv_bits=kv_bits,
+        kv_rotate=kv_rotate,
     )
 
     def _process_and_sample(tokens, logits):
@@ -2179,6 +2194,7 @@ def main():
         kv_bits=args.kv_bits,
         kv_group_size=args.kv_group_size,
         quantized_kv_start=args.quantized_kv_start,
+        kv_rotate=args.kv_rotate,
         draft_model=draft_model,
         num_draft_tokens=args.num_draft_tokens,
     )
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -1,11 +1,36 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import inspect
+import math
 from dataclasses import dataclass
 from typing import Any, Optional
 
 import mlx.core as mx
 from mlx.utils import tree_map
+
+_HADAMARD_BASES = (1, 12, 20, 28)
+
+
+def hadamard_size_ok(n: int) -> bool:
+    """True if ``mx.hadamard_transform`` supports a last-dim of size ``n``
+    (i.e. ``n = m * 2**k`` for ``m`` in {1, 12, 20, 28})."""
+    for m in _HADAMARD_BASES:
+        if n % m == 0 and ((n // m) & (n // m - 1)) == 0:
+            return True
+    return False
+
+
+def rotate_last(x: mx.array) -> mx.array:
+    """Orthonormal Walsh–Hadamard transform along the last axis.
+
+    Normalized by ``1/sqrt(D)`` so the transform is orthonormal (``R^T R = I``).
+    Applying it to *both* queries and keys leaves every ``q·k`` inner product —
+    and hence the attention scores — unchanged in exact arithmetic, while
+    spreading per-channel outliers into a near-Gaussian marginal that low-bit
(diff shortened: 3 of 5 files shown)


pierre427:
Ran an independent validation of this PR — the headline claim reproduces, and we found some additional datapoints that may help scope it.

**Repro** (your benchmark script, `mlx-community/Qwen3-1.7B-4bit`, 349 tokens, gs=64, mlx 0.32.0.dev, M5): kv4 affine 554.9 → rotated **57.7** vs fp16 45.3 — ~98% of the gap recovered, matching your bf16 numbers even on a 4-bit-weight model.

**At longer context the rescue is real but not near-lossless.** Teacher-forced ppl over a 4096-token prose pack, same model:

| config | ppl |
|---|---:|
| fp16 | 22.39 |
| K8/V8 affine | 22.46 |
| K4/V4 affine | 515.18 |
| K4/V4 rotated | 28.82 (+29% vs fp16) |
| K3/V8 rotated | 530.34 (collapse) |

So rotation moves the usable-K floor down by roughly one bit (K5–K6 rotated is clean, see below), rather than making K4 free; K3 still collapses at real context.

**The failure mode is model-dependent.** Same 4k grid on a 152B MoE (Hunyuan-3-REAP50 community 4-bit, GQA 64/8, head_dim 128): fp16 = 11.66, K4/V4 **affine** = 11.91 (+2%), rotated 11.75. No key-outlier pathology there at all — worth noting in the PR text that the win depends on the checkpoint's key statistics (your group_size/head_dim scaling point, plus per-model outlier structure).

**Composition with asymmetric K/V bits looks like the strongest use.** Since rotation only touches K, we crossed it with a separate-K/V-bits cache (#1550). On the outlier-prone model at 4k:

| config | avg bits | ppl |
|---|---:|---:|
| K8/V4 affine | 6 | 22.65 |
| K4/V8 rotated | 6 | 27.32 |
| K5/V4 affine | 4.5 | 31.10 |
| **K5/V4 rotated** | **4.5** | **23.15** |
| **K6/V3 rotated** | **4.5** | **22.98** |

i.e. rotated keys + low-bit values reaches a 4.5-bit-average cache at ~+3% ppl — a point neither symmetric rotation nor asymmetric affine reaches alone.

**One applicability note:** for MLA models cached in absorbed form (e.g. deepseek_v2 latent path), the cached last dim is `kv_lora_rank + qk_rope_head_dim` = 576 = 2^6·9, which `hadamard_size_ok` rejects — so the flag becomes a silent no-op there (correctly, since the guard is symmetric on Q and K). Rotating only the 512-dim latent slice (a supported size) could extend the technique to MLA caches if you're interested in a follow-up.

Nice PR — the `meta_state` backward-compat and the `speculative_generate_step` threading were both appreciated when testing.

katlun-lgtm:
Thanks for running this — this is exactly the scoping data the PR needed, and your framing is better than mine: **rotation lowers the usable-K floor by ~1 bit** rather than making K4 free. I'll fold that into the PR description (and stop saying "near-lossless" unqualified — your 4k-context table makes the boundary clear: K5–K6 rotated clean, K4 +29%, K3 still collapses).

The K5/V4 and K6/V3 rotated results are the most useful numbers here — a 4.5-bit-average cache at ~+3% ppl is a config neither rotation nor asymmetric affine reaches alone, and it composes with #1550 with no extra machinery since rotation only touches K. I'll add a "recommended configurations" note pointing there.

Agreed on model-dependence: the win tracks the checkpoint's key-outlier structure (Qwen3-family keys are pathological; your Hunyuan datapoint shows a checkpoint where affine K4 is already fine). Will note that the flag is a per-model call, not a default.

On MLA: right — 576 = 2^6·9 fails `hadamard_size_ok`, so the flag is a silent no-op on absorbed-form caches. Rotating only the 512-dim latent slice (with the matching rotation on the absorbed-q side to keep the guard symmetric) looks feasible and worth a follow-up PR rather than growing this one. Filed on my list.

Description updated accordingly. Thanks again for the validation — especially the 152B MoE counterexample, which is the kind of thing a single-model benchmark can't surface.

katlun-lgtm:
Re-ran your 4k grid on a second setup to check robustness: M3 Max, `mlx-community/Qwen3-1.7B-4bit`, teacher-forced ppl over 4096 tokens of **wikitext-2-raw test** (different corpus, so absolute ppl differs from your prose pack), group_size 64. Notably, the asymmetric K/V bits here are an **independent implementation** (a bench-local cache subclass + patched quantized SDPA), not the #1550 branch — so the composition result is cross-checked at the code level too.

| config | avg bits | wikitext, M3 Max | your prose pack, M5 |
|---|---:|---:|---:|
| fp16 | 16 | 16.02 | 22.39 |
| K8/V8 affine | 8 | 15.88 | 22.46 |
| K4/V4 affine | 4 | 490.07 | 515.18 |
| K4/V4 rotated | 4 | 18.69 (+16.7%) | 28.82 (+29%) |
| K3/V8 rotated | 5.5 | 437.59 | 530.34 |
| K8/V4 affine | 6 | 15.82 | 22.65 |
| K4/V8 rotated | 6 | 18.72 | 27.32 |
| K5/V4 affine | 4.5 | 20.40 (+27%) | 31.10 |
| **K5/V4 rotated** | **4.5** | **15.97 (−0.3%)** | 23.15 (+3.4%) |
| **K6/V3 rotated** | **4.5** | **15.79 (−1.4%)** | 22.98 (+2.6%) |

Every qualitative claim reproduces:

- K4/V4 affine collapses; K4/V4 rotated rescues but is not free at 4k; K3 rotated still collapses — the "−1 bit on the usable-K floor" framing holds.
- **K5/V4 and K6/V3 rotated land at fp16 on this corpus** (−0.3% / −1.4%, i.e. noise) — even stronger than your +3%, so the 4.5-bit-average recommended configs look robust rather than corpus-lucky.
- K5/V4 affine is clearly worse (+27%): the rotation is doing the work, not the extra K bit.
- K8/V4 affine ≈ fp16 and K4/V8 rotated ≈ K4/V4 rotated: V bits are nearly free and the residual penalty is entirely K-side, consistent with rotation being K-only.

One delta worth noting: the K4/V4-rotated penalty is corpus-dependent (+17% here vs +29% on your pack), while the K5+ cleanliness is not — another reason to steer users to K5/V4 or K6/V3 rotated rather than K4 symmetric.

So across M3 Max + M5, two corpora, and two asymmetric-bits implementations, the table stands. Happy to fold a `--kv-bits`-style `--k-bits/--v-bits` cross-benchmark into `benchmarks/kv_rotation_ppl.py` once #1550 lands, so this grid is reproducible from the repo.

katlun-lgtm:
Heads-up for the MLA follow-up discussed above: while prototyping the latent rotation I found that `--kv-bits` currently crashes outright on deepseek_v3-architecture models (absorbed path can't consume the quantized cache tuples), plus a quantized-SDPA mask-broadcast bug under GQA that the absorbed path trips. Both are fixed in #1558 — rotation-free, so it stands alone.

Two notes relevant to your comments here:

- Once #1558 is in, MLA models compose with #1550 as well (the latent is the "keys" slot at kv_lora_rank width, the rope keys sit in the value slot), and the rotation change for MLA on top of this PR turns out to be small: the values *are* the latents, so the attention output just needs the self-inverse transform applied before unembedding. That part is staged and I'll fold it in here after #1558 lands.
- Data point for your model-dependence table: on Moonlight-16B-A3B (deepseek_v3 arch), the post-RMSNorm latent shows no outlier pathology — wikitext ppl 6.64 fp16 / 6.80 kv4 affine / 7.04 kv4 rotated / 7.62 kv3 affine / 7.40 kv3 rotated. So MLA latents look like your Hunyuan case: affine is already fine at 4-bit and rotation is noise-level either way. Another reason `--kv-rotate` stays a per-model call.

(While measuring that I also hit the `QuantizedKVCache` pre-allocation failure for bits ∈ {3, 5, 6} — confirming your `packed_dim` rewrite in #1550 fixes real breakage beyond the asymmetric feature itself.)

pierre427:
One heads-up on the MLA + #1550 composition, since it intersects a guard we landed in #1509 (absorbed-MLA decode for deepseek_v2):

On the absorbed path the cached latent is *simultaneously* the key and the value — the value is reconstructed from it via the absorbed matrices — so a naive asymmetric cache has no well-defined `v_bits`, and setting one crashes in the deepseek_v2 rope `pe_scores` path. #1509 added a fail-closed guard that rejects asym-K/V on absorbed-MLA for exactly that reason.

Your "latent → K slot, rope-keys → V slot" reinterpretation is a clean way to give the two cached slots distinct widths and sidesteps the undefined-`v_bits` problem — but note it's also the precise config that guard currently refuses, so the composition will need the guard taught to recognize the reinterpreted mapping rather than fail-closed on it. Flagging so the two don't collide when you fold the MLA rotation in here.

katlun-lgtm:
Ah, that explains the discrepancy I flagged on #1558 — you've been testing on the #1509 tree, which makes all your absorbed-v2 observations line up. Good flag on the guard: you're right that "latent bits vs rope-key bits" is a well-defined split even though `key_bits != value_bits` spells the same thing, so when I fold the MLA rotation in here I'll coordinate with however #1509/#1550 settle that guard rather than fight it. Nice fail-closed choice in the meantime — crashing beats silently dequantizing with the wrong width.

katlun-lgtm:
Knocked out the two remaining test-plan items in 5c13402 (`tests/test_kv_rotate.py`), so both boxes are ticked now:

- **`--quantized-kv-start` path** — `test_cache_to_quantized_rotated` drives `KVCache.to_quantized(rotate=True)` end-to-end through `generate_step` and the quantized-SDPA Q-rotation, and checks the `rotate` flag plus the 4-field `meta_state`. At 8-bit the rotated cache still tracks the fp reference (I used `rtol=6e-2` — measured max-rel was ~0.04 vs affine's ~0.034, i.e. rotation is neutral at high bits, exactly as expected; the win is at 4-bit).
- **head_dim coverage** — `test_rotate_last_preserves_scores_across_head_dims` sweeps head_dim ∈ {64, 80, 96, 128} on the orthonormal `(Rq)·(Rk) = q·k` invariant plus norm preservation, and asserts `hadamard_size_ok` accepts those and rejects the unsupported sizes (including the 576 MLA latent that makes `--kv-rotate` a no-op there).

There's also a small `test_quantized_cache_rotate_roundtrip` for the `rotate` save/load and legacy 3-field `meta_state` back-compat. All three pass locally on `Qwen1.5-0.5B-Chat-4bit`, mlx 0.32.0.

pierre427:
Datapoint from adapting this rotation to the absorbed-MLA path (rotate the 512-dim cached latent, leave the 64-dim rope slice unrotated; rotate `q_nope` to keep scores exact and un-rotate the latent-space attention output — exact up to quant error since the 512-Hadamard is orthonormal and self-inverse):

| model | fp16 | kv8 affine | kv8 rotated | kv4 affine | kv4 rotated |
| DeepSeek-V2-Lite-Chat-4bit | 17.08 | 17.07 | 17.04 | 19.60 | 19.96 |
| Sarvam-105B-q6 (deepseek_v3 arch) | 124.90 | 123.78 | 123.80 | 133.55 | 134.04 |

(4096-token teacher-forced ppl, M5, mlx 0.32.) The mechanism ports cleanly to MLA, but on both models it is neutral at kv8 and slightly negative at kv4 — the cached latents don't show the outlier structure the rotation exploits (kv4 affine is only +7–15% over fp16; no collapse to rescue). Sharing as a negative so nobody burns a cycle composing rotation with MLA caches; where it earns its keep remains the outlier-prone GQA/MHA models in this PR's own table.

katlun-lgtm:
That lands the same way my Moonlight run did, which makes it three checkpoints — I posted those numbers upthread on the 13th, before the shape of it was clear: wikitext ppl 6.64 fp16 / 6.80 kv4 affine / **7.04 kv4 rotated**. Slightly negative at kv4, same as your two. Different arch generation, different corpus, same answer.

So I'm treating the MLA rotation as closed rather than staged. I had the latent-slice version sitting on a branch waiting for #1558 to land — rotate the 512-dim latent, leave the 64-dim rope slice, un-rotate the attention output — and there's no reason to fold in machinery that buys nothing. Thanks for spending the cycles on it; a negative from a second and third checkpoint is what turns "my model didn't like it" into a result.

I've pulled the follow-up bullet out of the description and put the numbers in its place, so the next person reads the answer instead of the invitation.

While correcting that I fixed the other stale promise in the same section. The description called the paper's +1-bit QJL residual "the natural next step for usable 3-bit KV" — I prototyped that one too and it doesn't survive either: wikitext-2, 1024 tokens, fp16 10.66, rotated-4 12.81, rotated-3 160.4, rotated-3 + QJL 27.14. QJL recovers ~89% of the 3-bit gap, but 3-bit + QJL still loses to plain 4-bit rotation, so it isn't worth the machinery. Both follow-ups this PR was advertising turn out to be dead ends — better in the body than in someone's afternoon.

zcbenz:

PR #1556, opened 2026-07-13
Commit-steering: a content-blind controller for reasoning run-on

pierre427:
## Commit-steering: a content-blind controller for reasoning run-on

Reasoning models often *have* the answer well before they stop talking. They re-derive it, second-guess, re-check the arithmetic a third time, and finally emit `</think>` thousands of tokens after the result was already formed. That deliberation is real compute you pay for and usually don't need.

This PR adds an opt-in **activation-steering** controller that cuts a large fraction of that redundant thinking, content-blind at runtime, across a wide range of thinking architectures.

### What we found getting here

The design is the product of a chain of findings, and several of them were surprises worth stating — they explain why the controller looks the way it does and why the blunter approaches were rejected.

**Run-on is a loaded spring, not confusion.** The intuitive read of a rambling model is that it keeps going because it doesn't *know*. Logit-lens inspection says the opposite. Reading out the model's own next-token distribution step by step, the `</think>` stop token is frequently high-rank but held down by a margin of roughly a dozen nats of logit, at low entropy, while the answer is already sitting in the trace. On a 122B model we watched the answer form near token 150 and the stop token stay suppressed (rank 7.5k–117k) all the way to a 12k cap. The exit is right there and latched shut. That reframing is the whole reason a *pull* works: you don't need to inject an answer or teach anything, you need to release a behavior that's already loaded.

**The pathology is dispositional, not informational — and a dead end proved it.** Before steering, we tried a "memory channel" that extracted key terms mid-stream, looked up the model's own relevant knowledge, and re-injected it to help the reasoning converge. We tried it with facts, then with methods-to-solve. Both made run-on **worse**: the injected material gave the over-verifier a fresh thing to reconcile, and the methods variant induced an outright repetition loop. That failure was the useful result. Run-on isn't a missing-knowledge problem, so token-level and information-level interventions feed the loop. Only an **activation-level** change — moving the disposition, not the content — actually helps. Every negative pointed at the same door.

**Content-blind was a deliberate constraint, not a convenience.** We specifically avoided both a stop-token clamp (fights the symptom, and a suppressed stop token means clamping mislabels genuine deliberation) and a "be concise" prompt (fights the content, model-specific, unreliable). The goal was a generalized lever that triggers off internal *meta*-information — a direction measured from the activations — so one recipe ports across model families instead of one prompt per model.

**The commit axis is both real and causal.** Extracted by difference-of-means, the direction reappears across independent traces with a cross-trace cosine consistency of +0.5 to +0.63 on Qwen3-8B (and +0.86 on the cleanest model we tested). More important, steering it is **bidirectional, dose-dependent, and direction-specific**: +0.2 closes thinking 44% earlier with the answer still correct; −0.2 *induces* run-on on a model that was about to stop; +0.4 applied to a trace already spiralling ends it with the correct answer; and a random direction of the same magnitude is **inert**. It's this specific direction, not any perturbation, that moves termination.

**The trade-off improves with scale, and that was not obvious.** We expected larger models to be more fragile to steering. Instead the quality effect trends *upward* with size across the Qwen3 ladder (−0.03 at 4B → +0.05 at 32B), because big models genuinely over-deliberate on hard problems and the commit bias rescues them (Qwen3-32B math_hard alone gains +0.40).

**Reaching every architecture required rewriting the harness, and that surfaced its own findings.** The first extraction/steer harness hand-mirrored one model's forward and broke on six architectures (attention sinks, sliding-window, latent-attention, GDN/SSM hybrids). Switching to *running each model's own native forward and tapping one layer's residual* made capture-and-steer architecture-agnostic — and along the way it exposed that a couple of newer configs don't expose `num_hidden_layers` / `hidden_size` the way older ones do (read the layer count from the module list; infer the hidden dim from a forward). Two boundaries stayed genuinely out of reach: gpt-oss reasons through Harmony analysis *channels* rather than a `<think>` block, so it's null for a `</think>`-gated method; and SSM/Mamba hybrids (Nemotron-H) carry state rather than a clean residual commit locus.

### The idea

Run-on is a suppressed stop token; the "committing / wrapping up" behavior is a direction in the residual stream; nudging the hidden state along that direction, at the layer where the model decides to finish, releases the answer sooner.

The direction is learned **offline** by difference-of-means (mean hidden state over commitment spans minus reflection spans). At inference the controller never reads the generated text — it only adds a fixed vector at one decoder layer while the model is inside its think channel. Two modes:

- an always-on low-dose **bias** (α≈0.2) while thinking — prophylactic, no sensor, and
- a late hard **hammer** (α≈0.8) if the model is still thinking past a token budget — a therapeutic rescue for genuine spirals.

### What's in the diff (+697 lines, 4 files)

- **`mlx_lm/steer.py`** — the feature.
  - A **native-delegation layer tap** (`steer_layer` / `_SteerTap`): it runs the *real* decoder layer's own forward and adds the steering vector to its residual output, delegating mask / cache / attention-sinks / hybrid routing / `is_linear` to the real layer. That one design choice makes capture-and-steer work across dense, MoE, sliding-window+sink (gpt_oss), latent-attention (MLA), and GDN/SSM-hybrid (Qwen3-Next / Qwen3.5) models without re-implementing any of their forwards.
  - **`CommitSteerer`** — the two-mode controller. It's a context manager that patches one layer on enter and restores it on exit, and it plugs into the **existing** `generate_step(..., logits_processors=[...])` API: `steerer.logits_processor` reads the committed-token prefix, tracks the think channel, and sets which vector the *next* forward injects. It never changes the sampled logits, so it composes with samplers and other processors.
  - A **NaN / zero-axis guard** (`build_vector` → `None`) makes a missing or degenerate direction a **safe no-op** — a non-thinking model with no extractable axis simply runs unchanged.
  - **`extract_commit_direction`** — the offline difference-of-means extractor (content labels are used *here only*).
- **`tests/test_steer.py`** — tap inject+restore, the bias→hammer→off schedule, degenerate-axis and no-close-tag no-ops, and the logits-untouched invariant. Runs without a large model.
- **`mlx_lm/examples/commit_steer.py`** — extract-then-steer, end to end.
- **`mlx_lm/__init__.py`** — export `CommitSteerer`.

### Why it's a win (measured)

A 100-prompt × 10-domain A/B (ON vs OFF, identical settings; Apple M5, mlx 0.32.0.dev with NAX). Δquality is graded quality ON minus OFF; "think-cut" is the reduction in thinking tokens.

Qwen3 dense ladder:

| Model | Δquality | Think-cut |
|---|---:|---:|
| Qwen3-1.7B | +0.008 | −46% |
| Qwen3-4B | −0.030 | −38% |
| Qwen3-8B | −0.007 | −41% |
| Qwen3-14B | +0.026 | −44% |
| Qwen3-32B | **+0.049** | −24% |

Across other thinking architectures (each benched at ~57% relative depth):

| Model | Architecture | Δquality | Think-cut |
|---|---|---:|---:|
| Qwen3-0.6B | qwen3 dense | −0.038 | −59% |
| Mellum2-12B-A2.5B | mellum MoE | −0.007 | −64% |
| Trinity-Mini | afmoe MoE | −0.015 | −32% |
| Qwen3.6-27B | qwen3_5 GDN-hybrid | **+0.031** | −38% |
| Qwen3.5-122B-A10B | qwen3_5_moe GDN-hybrid | −0.048 | −16% |
| DeepSeek-R1-Distill-Llama-8B | llama (distill) | −0.026 | −43% |

**Headline: across eight distinct thinking architectures, 16–64% fewer thinking tokens with a quality effect in the [−0.05, +0.05] band and no catastrophic degradation.** The trade-off *improves* with scale on same-family models (Qwen3-32B is a net quality gain — math_hard alone +0.40). The commit axis is strongly consistent across traces (cross-trace cosine +0.5 to +0.86; the GDN-hybrid Qwen3.6-27B is the cleanest at +0.86).

### Honest scope and limitations

- **Quality-neutrality is model-dependent, not guaranteed.** The two dips above are *category-specific*, not reasoning harm: Qwen3.5-122B's −0.048 is almost entirely a creative-writing outlier (−0.32) from over-compressing open-form generation — every objective category (math/code/factual/instruction) held flat. Qwen3-0.6B is a tiny model with little slack. R1-Distill already reasons lean, so a 43% cut starts trimming load-bearing tokens.
- **Extraction is per-model** and needs a handful of reasoning traces. The vector isn't portable across models.
- **Non-thinking models are out of scope** and correctly self-disable (null axis → no-op). Confirmed on Qwen2.5, Llama-3.2, Mistral, Phi-3.5, Falcon3, Apertus, DeepSeek-V2-Lite.
- **gpt-oss** reasons via Harmony analysis *channels*, not a `<think>` block, so it's null for this method; a channel-aware extractor is future work.
- **SSM/Mamba hybrids (Nemotron-H)** don't expose a clean decoder-layer list / residual commit locus, so the tap doesn't apply as-is.
- **Doesn't stack with speculative decoding on reasoning workloads** — prompt-lookup is lossless but a net slowdown there (acceptance ~0.8%); the token reduction here is the operative win, and it's orthogonal.

### State of the ecosystem

Activation steering / representation engineering (steering vectors, inference-time intervention) is established; this is a specific *application* — a termination-behavior direction extracted content-blind and deployed as a two-mode controller, validated as a general efficiency lever across eight thinking architectures rather than demonstrated on one. vLLM/SGLang ship *consumption-based* reasoning budgets (hard token caps) but not an activation-level commit bias. We found no exact prior for steering reasoning run-on this way (treat as "no exact prior found," not "first ever").

### API

```python
from mlx_lm import load, CommitSteerer
from mlx_lm.steer import extract_commit_direction
import mlx.core as mx

model, tok = load("mlx-community/Qwen3-8B-4bit")
# offline, once per model:
vecs = extract_commit_direction(model, tok, traces, layers=[16, 18, 20, 24])
v_hat, rms = vecs[20]
# runtime, content-blind:
steerer = CommitSteerer(model, tok, v_hat, rms, layer=20)
with steerer:
    for t, _ in generate_step(prompt_ids, model,
                              logits_processors=[steerer.logits_processor]):
        ...
```

Fully opt-in; zero effect on existing code paths.

### DIFF
--- a/mlx_lm/__init__.py
+++ b/mlx_lm/__init__.py
@@ -1,4 +1,4 @@
-# Copyright © 2023-2024 Apple Inc.
+# Copyright © 2023-2026 Apple Inc.
 
 import os
 
@@ -8,6 +8,7 @@ os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 
 from .convert import convert
 from .generate import batch_generate, generate, stream_generate
+from .steer import CommitSteerer
 from .utils import load
 
 __all__ = [
@@ -16,5 +17,6 @@ __all__ = [
     "batch_generate",
     "generate",
     "stream_generate",
+    "CommitSteerer",
     "load",
 ]
--- a/mlx_lm/examples/commit_steer.py
+++ b/mlx_lm/examples/commit_steer.py
@@ -1,0 +1,126 @@
+# Copyright © 2026 Apple Inc.
+
+"""End-to-end example: extract a commit direction, then steer generation.
+
+Step 1 (offline, once per model) records a handful of greedy reasoning traces on
+checkable problems and extracts the commit direction by difference-of-means.
+Step 2 (runtime, content-blind) generates with the two-mode CommitSteerer and
+prints the token counts with and without it.
+
+    python -m mlx_lm.examples.commit_steer --model mlx-community/Qwen3-8B-4bit
+"""
+
+import argparse
+
+import mlx.core as mx
+import numpy as np
+
+from mlx_lm import load
+from mlx_lm.generate import generate_step
+from mlx_lm.steer import CommitSteerer, extract_commit_direction
+
+# A few checkable problems to elicit reasoning traces for extraction.
+PROBLEMS = [
+    "What is 23 times 4? Show your reasoning, then give the final answer.",
+    "A bat and a ball cost $1.10. The bat costs $1.00 more than the ball. "
+    "How much is the ball?",
+    "What is 37 times 43? Reason step by step, then answer.",
+    "Find the smallest positive integer x with x = 2 (mod 3) and x = 3 (mod 5).",
+]
+
+
+def greedy_trace(model, tok, prompt, max_tokens=2048):
+    ids = mx.array(
+        tok.encode(
+            tok.apply_chat_template(
+                [{"role": "user", "content": prompt}],
+                tokenize=False,
+                add_generation_prompt=True,
+            )
+        )
+    )
+    out = []
+    for t, _ in generate_step(ids, model, max_tokens=max_tokens):
+        t = int(t)
+        out.append(t)
+        if t == tok.eos_token_id:
+            break
+    return out
+
+
+def run(model, tok, prompt, steerer=None, max_tokens=2048):
+    ids = mx.array(
+        tok.encode(
+            tok.apply_chat_template(
+                [{"role": "user", "content": prompt}],
+                tokenize=False,
+                add_generation_prompt=True,
+            )
+        )
+    )
+    close_id = tok.encode("</think>", add_special_tokens=False)
+    close_id = close_id[0] if len(close_id) == 1 else -1
+    procs = [steerer.logits_processor] if steerer else None
+    ctx = steerer if steerer else _null()
+    n = think = 0
+    closed = False
+    with ctx:
+        for t, _ in generate_step(
+            ids, model, max_tokens=max_tokens, logits_processors=procs
+        ):
+            t = int(t)
+            n += 1
+            if t == close_id:
+                closed = True
+            elif not closed:
+                think += 1
+            if t == tok.eos_token_id:
+                break
+    return n, think
+
+
+class _null:
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *a):
+        return False
+
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--model", required=True)
+    ap.add_argument("--layer", type=int, default=None, help="commit locus (default ~57%)")
+    ap.add_argument("--alpha-bias", type=float, default=0.2)
+    ap.add_argument("--alpha-hammer", type=float, default=0.8)
+    ap.add_argument("--hammer-budget", type=int, default=900)
+    ap.add_argument("--test-prompt", default="Is 91 a prime number? Explain, then answer.")
+    args = ap.parse_args()
+
+    model, tok = load(args.model)
+    n_layers = len(model.model.layers)
+    layer = args.layer if args.layer is not None else round(0.57 * n_layers)
+    span = sorted({max(2, min(n_layers - 1, round(f * n_layers)))
+                   for f in (0.43, 0.5, 0.57, 0.64)})
+
+    print(f"[1/2] extracting commit direction on {len(PROBLEMS)} traces ...")
+    traces = [greedy_trace(model, tok, p) for p in PROBLEMS]
+    vecs = extract_commit_direction(model, tok, traces, span)
+    v_hat, rms = vecs[layer]
+    print(f"      layer {layer}/{n_layers}  ||v||={np.linalg.norm(v_hat):.3f}  rms={rms:.1f}")
+
+    steerer = CommitSteerer(
+        model, tok, v_hat, rms, layer=layer,
+        alpha_bias=args.alpha_bias, alpha_hammer=args.alpha_hammer,
+        hammer_budget=args.hammer_budget,
+    )
+    print(f"[2/2] generating '{args.test_prompt}'")
+    n_off, think_off = run(model, tok, args.test_prompt, steerer=None)
+    n_on, think_on = run(model, tok, args.test_prompt, steerer=steerer)
+    cut = (1 - think_on / max(think_off, 1)) * 100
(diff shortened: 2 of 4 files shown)


pierre427:
Requesting review of the mechanism and the evidence supporting it.

This PR extracts a residual direction from contrasting “continue reasoning” and “commit to an answer” states, then tests whether intervening along that direction changes reasoning termination.

Several results suggest the direction is more than a correlated classifier:

- reversing its sign reverses the effect;
- the response varies with intervention strength and layer depth;
- same-norm random directions do not reproduce the result;
- across eight architectures and 100 prompts in ten domains, thinking length falls by 16–64%, while aggregate task-quality changes remain within approximately ±0.05.

The main uncertainties are whether the intervention is specific to termination rather than general residual disruption, how much the extracted direction depends on the prompt set, and whether the cross-model results indicate a shared mechanism or separate model-specific effects.

Review of the controls, extraction/evaluation separation, and interpretation of the direction would be particularly useful. “Termination attractor” may be too strong if the evidence supports only a manipulable commit-related direction, so that terminology is also open to scrutiny.

pierre427:
Four papers from the last two weeks independently converge on this PR's premises — adding them as review context, since they change the prior on whether this mechanism class is real:

- **Calibrated e-CUSUM decoding monitors (arXiv 2607.11317)** finds that *non-termination, rather than looping, is the dominant failure mode* in reasoning models, and proves centered token log-probability is a mean-zero martingale under the model's own sampling law — "nearly silent during confident repetition." That is the observable-side explanation for why static logit biasing cannot fix run-on: the failure isn't visible where a static intervention looks. Their detector is output-statistics-based — complementary to the activation-side sensors here, not overlapping.
- **SOPHIA (arXiv 2607.18100)** steers out of reasoning self-loops with activation vectors indexed by latent-state transitions — trajectory sensing plus activation intervention at inference: the same mechanism class, targeting the looping subset of the failure this PR addresses.
- **PUMA (arXiv 2607.17188)** builds real-time trajectory sensors (latent velocity/tortuosity crossed with entropy) for reasoning pathology, but intervenes only by truncation. Sensor-side convergence; the release-style intervention remains distinct.
- **Answer pre-commitment (arXiv 2607.16451)** reads the final answer out of hidden states before emission in the majority of rollouts — direct evidence for the "answer present but unreleased" premise this controller acts on.

Net: trajectory-sensing + activation-intervention now has three independent implementations, and the dominant-failure framing matches this PR's motivation. What none of them do is the content-blind attractor *pull* at the commit layer — the specific contribution here.

PR #1557, opened 2026-07-13
Prompt-lookup: a measured never-slower-than-plain gate

pierre427:
## A measured never-slower-than-plain gate for prompt-lookup decoding

**Stacks on the prompt-lookup PR ([#1508](https://github.com/ml-explore/mlx-lm/pull/1508)) — please review/merge that first.** This is a small, focused follow-up (+187 lines, 3 files) that makes prompt-lookup's adaptive latch decide on *measured wall-clock* instead of an acceptance-fraction proxy.

### The problem it closes

Prompt-lookup decoding is lossless, but on non-copy workloads (reasoning traces, long-context prose) it can be a net **slowdown**: verbatim n-gram overlap is near zero, so almost every proposed span is rejected and the speculative verify forward is pure overhead. On our reasoning benchmarks base PLD lands around **0.85–0.92×** of plain decode.

#1508 already guards against this with an `adaptive` latch: after `warmup` tokens, if the accepted-proposal fraction drops below `gate` (default 0.12) it latches once to a plain `generate_step` tail. That's a good heuristic, but it's a **proxy**. The break-even acceptance fraction isn't a constant — it depends on the ratio of the batched-verify forward cost to a single decode step, which grows with context length and varies by model. So a fixed threshold can:

- **false-latch** a speculation that's still winning (short context / small model, where even low acceptance is net-faster), or
- **ride a loser** (long context, where the verify forward is expensive enough that 0.12 acceptance is still a net slowdown).

### The idea

Measure the real thing. After `warmup` tokens, time a short plain-decode **probe** of `rate_gate_probe` tokens and compare it to the observed speculative rate; latch to the plain tail only if speculation isn't actually faster (by more than `rate_gate_margin`). This is a direct wall-clock break-even, so it's correct exactly where the acceptance proxy misjudges — a genuine never-*slower*-than-plain guarantee rather than a stand-in for one.

### How it works

`prompt_lookup_generate_step` gains three params: `rate_gate: bool = False`, `rate_gate_probe: int = 32`, `rate_gate_margin: float = 0.0`.

- The speculative loop accumulates its wall-clock (`spec_t0`) and token count.
- Once past `warmup` (and it hasn't probed yet), it runs `rate_gate_probe` ordinary plain-decode steps, timing them, to get `plain_ms_per_tok`, and compares to the accumulated `spec_ms_per_tok`.
- If `spec_ms > plain_ms * (1 - margin)`, it sets `latched = True` — reusing #1508's **existing** latch → plain-`generate_step`-tail fall-through, so no new tail path. Otherwise it resets the window and keeps speculating.
- The decision is **one-shot** (a permanent switch — no repeated probing overhead), and the probe's tokens are ordinary committed output, so the run stays **lossless** whether it de-latches or not.

Four stats fields (`rate_gate_probed`, `rate_gate_delatched`, `rate_gate_spec_ms_per_tok`, `rate_gate_plain_ms_per_tok`) expose what it decided.

```
copy-heavy prompt          novel / reasoning prompt
─────────────────          ────────────────────────
warmup → probe             warmup → probe
spec faster → keep going   spec slower → latch to plain tail
= full PLD speedup         = plain speed (not 0.85–0.92×)
```

It composes with the existing `adaptive` gate — both can be set; whichever fires first latches. `rate_gate` is the measured option to reach for when you care about the wall-clock guarantee across mixed workloads; `gate` remains the cheaper acceptance-only heuristic.

### Why it's a win

It removes PLD's one real failure mode — a net slowdown on non-copy work — by measuring it directly instead of guessing a threshold. On a mixed serving workload (some copy-heavy, some reasoning) the acceptance proxy has to be tuned per model to avoid both false-latch and ride-a-loser; the measured gate needs no such tuning because it times the actual break-even. The cost is bounded: a `warmup` window plus one `rate_gate_probe`-token probe, paid once.

Honest scope: this doesn't make PLD *faster* on copy-heavy work (that's #1508); it makes the worst case ≈ plain instead of a regression, and does so without a model-specific threshold. The end-to-end speedup on any given workload is exactly #1508's when speculation wins and ≈1.0× when it doesn't — the point is the guarantee, not a new peak number.

### Tests

`tests/test_prompt_lookup.py`:
- **probe runs and stays lossless** — the gate's output equals plain PLD output (it only decides *when to stop speculating*, never *what* to emit).
- **de-latches when speculation is slower** and **keeps speculating when it's faster** — deterministic, under a scripted `perf_counter`, asserting the latch decision, the stats, and losslessness in both branches.

All 17 prompt-lookup tests pass (14 from #1508 + 3 here); black/isort clean.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1,4 +1,4 @@
-# Copyright © 2023-2024 Apple Inc.
+# Copyright © 2023-2026 Apple Inc.
 
 import argparse
 import contextlib
@@ -747,6 +747,9 @@ def prompt_lookup_generate_step(
     adaptive: bool = False,
     warmup: int = 48,
     gate: float = 0.12,
+    rate_gate: bool = False,
+    rate_gate_probe: int = 32,
+    rate_gate_margin: float = 0.0,
     stats: Optional[Any] = None,
     history_prompt: Optional[mx.array] = None,
 ) -> Generator[Tuple[int, mx.array, bool], None, None]:
@@ -772,6 +775,19 @@ def prompt_lookup_generate_step(
     regression on non-copy output; copy-heavy work never latches. ``stats``
     (a PromptLookupStats) is filled in place. Yields ``(token, logprobs,
     from_draft)``.
+
+    ``rate_gate`` is a *measured* alternative to that acceptance-fraction
+    heuristic. After ``warmup`` tokens it times a short plain-decode probe of
+    ``rate_gate_probe`` tokens and compares it to the observed speculative rate,
+    latching to the plain tail only if speculation isn't actually faster (by
+    more than ``rate_gate_margin``). The break-even acceptance fraction depends
+    on the verify/decode cost ratio, which is model- and context-dependent, so
+    measuring the wall-clock directly avoids both false-latching a still-winning
+    speculation and riding a losing one — a real never-*slower*-than-plain
+    guarantee, whereas ``gate`` is only a proxy for it. The probe is one-shot
+    (a permanent decision) and its tokens are ordinary committed output, so the
+    result is still lossless. ``rate_gate`` and ``adaptive`` may both be set;
+    whichever fires first latches.
 
     ``history_prompt`` may be the full prompt when ``prompt`` is only an
     uncached tail backed by a prefilled ``prompt_cache``. Retrieval proposals use
@@ -828,7 +844,9 @@ def prompt_lookup_generate_step(
     latched = False
     prompt_len = len(seq)
     last_snap = None  # snapshot before the most recent proposal forward
+    rate_probed = False  # measured-rate gate is one-shot
     try:
+        spec_t0 = time.perf_counter()  # wall-clock window for the speculative rate
         while (max_tokens < 0 or generated < max_tokens) and not latched:
             stats.cycles += 1
             # Clamp the proposal to the remaining budget so a cycle never forwards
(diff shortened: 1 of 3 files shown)


pierre427:
Requesting review of the break-even measurement and fallback behaviour in this PR.

Prompt-lookup decoding is usually gated by acceptance rate, but acceptance does not account for proposal construction, verification shape, synchronization, or other fixed costs. This implementation instead compares measured milliseconds per committed token against a plain target-decoding probe after warmup. If the speculative path is slower, the request latches onto ordinary decoding.

The probe remains lossless because its output is generated and committed by the target model rather than discarded solely for measurement.

- whether a single post-warmup comparison is stable enough when prompt overlap and acceptance change during generation;
- whether compilation, asynchronous evaluation, synchronization, thermal variation, or timer resolution can bias the comparison;
- whether the probe interval provides enough samples to distinguish a real slowdown from noise;
- whether a margin or hysteresis should be required before permanently disabling speculation;
- what cost is incurred when an early false negative latches a request onto plain decoding for the remainder of a long generation;
- whether the same decision rule transfers across batch sizes, quantization settings, and different prompt-lookup candidate lengths.

“Never slower” should probably be interpreted as an observed per-request fallback property rather than a strict guarantee under changing generation phases. The underlying question is how much online measurement is needed to choose between speculative and ordinary decoding without consuming the gain being measured.

zcbenz:

PR #1558, opened 2026-07-13
Fix quantized KV cache on absorbed-MLA models (deepseek_v3)

katlun-lgtm:
## Proposed changes

`--kv-bits` currently **crashes on every deepseek_v3-architecture model** (DeepSeek-V3 / R1, Moonlight, …). The absorbed-MLA attention consumes the fetched cache entries directly, and a `QuantizedKVCache` returns `(packed, scales, biases)` tuples:

```
File "mlx_lm/models/deepseek_v3.py", line 147, in __call__
```

Two fixes:

**1. `deepseek_v3.py`: consume quantized caches in the absorbed path.**
- `pe_scores` via `mx.quantized_matmul` against the quantized rope keys.
- Decode (`L == 1`): the quantized latents flow through the quantized SDPA path unchanged.
- Chunked prefill over an already-quantized cache (`L > 1`, e.g. `mlx_lm.cache_prompt`): dequantize the latents and use the float projections.

**2. `base.py`: quantized SDPA mask broadcast under GQA.** `quantized_scaled_dot_product_attention` expands any >3-dim mask on the `-3` axis, which places a per-head `(B, n_q_heads, L, S)` float mask on the kv-head axis when `n_repeats > 1`, silently broadcasting the output to `n_q_heads²` rows (surfaces as a reshape `ValueError`). The absorbed-MLA path (mask = `pe_scores`) is the first in-tree caller to hit it. The fix unflattens the head axis only for a per-query-head mask (`mask.shape[-3] == n_q_heads`) whose rank is below the scores' rank, and adds the repeats axis otherwise — so per-kv-head, broadcast and already-repeat-shaped masks stay broadcastable. A mask-rank matrix test is included at `B = 2` (rows fail on `main`, pass with the fix), over both a plain GQA shape and the single-kv-head shape an absorbed-MLA model decodes at.

## Verification

Moonlight-16B-A3B-Instruct-4-bit (deepseek_v3 arch), M3 Max:

- `--kv-bits 8`: greedily **identical** output to fp16 cache.
- Teacher-forced wikitext-2 ppl over 1k tokens: fp16 **6.64** / kv8 **6.65** / kv4 **6.80** / kv3 **7.62** — graceful degradation, no collapse.
- `tests/test_models.py` (incl. new test) and `tests/test_prompt_cache.py` (22) pass.

## Independent validation

@pierre427 exercised this on two further models:

- **glm4_moe_lite** (real checkpoint) — the in-tree model whose identical absorbed-MLA bug is fixed here by their patch, folded in with credit: kv8 max |logit diff| vs a float cache **0.0104**, kv4 **0.2103**, argmax agreement **1.00** for both, over prefill → decode → chunk-over-quantized-cache.
- **Sarvam-105B** (deepseek_v3 arch, q6 weights, M5, 4096-token teacher-forced ppl, chunked prefill): fp16 KV **124.90** / kv8 **123.78** / kv4 **133.55** (+6.9%). Decode cost of the quantized path at 4k context is **+26–31% ms/tok** vs an fp16 cache — quantized MLA KV is a memory lever rather than a speed one at short and mid context.

Related note: constructing a `QuantizedKVCache` directly with bits ∈ {3, 5, 6} also fails at the first `update_and_fetch` (pre-allocation uses `dim // (32 // bits)`; the ppl numbers above used a direct-quantize path) — that math is already corrected as part of #1550, so this PR leaves it alone.

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx-lm/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes (black on changed files)
- [ ] I have updated the necessary documentation (if needed)

### DIFF
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -90,11 +90,14 @@ def quantized_scaled_dot_product_attention(
             q_indices = mx.arange(kL - qL, kL)
             k_indices = mx.arange(kL)
             mask = q_indices[:, None] >= k_indices[None]
-        if n_repeats > 1 and mask.ndim >= 3:
-            if mask.shape[-3] == 1:
-                mask = mx.expand_dims(mask, -3)
-            else:
+        if n_repeats > 1 and 3 <= mask.ndim < scores.ndim:
+            if mask.shape[-3] == n_q_heads:
+                # Per-query-head mask (e.g. the absorbed-MLA pe_scores): split
+                # the head axis to match the scores' (n_kv_heads, n_repeats).
                 mask = mx.unflatten(mask, -3, (n_kv_heads, n_repeats))
+            else:
+                # Per-kv-head or broadcast mask: add the repeats axis.
+                mask = mx.expand_dims(mask, -3)
         if mask.dtype == mx.bool_:
             scores = mx.where(mask, scores, mx.finfo(scores.dtype).min)
         else:
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -90,8 +90,11 @@ def quantized_scaled_dot_product_attention(
             q_indices = mx.arange(kL - qL, kL)
             k_indices = mx.arange(kL)
             mask = q_indices[:, None] >= k_indices[None]
-        if n_repeats > 1 and mask.ndim > 3:
-            mask = mx.expand_dims(mask, -3)
+        if n_repeats > 1 and mask.ndim >= 3:
+            if mask.shape[-3] == 1:
+                mask = mx.expand_dims(mask, -3)
+            else:
+                mask = mx.unflatten(mask, -3, (n_kv_heads, n_repeats))
         if mask.dtype == mx.bool_:
             scores = mx.where(mask, scores, mx.finfo(scores.dtype).min)
         else:
--- a/mlx_lm/models/deepseek_v3.py
+++ b/mlx_lm/models/deepseek_v3.py
@@ -144,7 +144,19 @@ class DeepseekV3Attention(nn.Module):
         if cache is not None:
             kv_latent, k_pe = cache.update_and_fetch(kv_latent, k_pe)
 
-        pe_scores = (q_pe * self.scale) @ k_pe.swapaxes(-1, -2)
+        # With a quantized cache the fetched entries are (packed, scales,
+        # biases) tuples.
+        quantized = isinstance(k_pe, (list, tuple))
+        if quantized:
+            pe_scores = mx.quantized_matmul(
+                q_pe * self.scale,
+                *k_pe,
+                transpose=True,
+                group_size=cache.group_size,
+                bits=cache.bits,
+            )
+        else:
+            pe_scores = (q_pe * self.scale) @ k_pe.swapaxes(-1, -2)
         if mask is not None:
             pe_scores = mx.where(
                 mask,
@@ -155,12 +167,19 @@ class DeepseekV3Attention(nn.Module):
         if L == 1:
             q_nope = self.embed_q(q_nope)
             k = v = kv_latent
+            sdpa_cache = cache
         else:
-            k = self.embed_q(kv_latent, transpose=False)
-            v = self.unembed_out(kv_latent)
+            lat = kv_latent
+            if quantized:
+                # Chunked prefill over an already-quantized cache: fall back
+                # to float latents for the absorbed projections.
+                lat = mx.dequantize(*lat, group_size=cache.group_size, bits=cache.bits)
+            k = self.embed_q(lat, transpose=False)
+            v = self.unembed_out(lat)
+            sdpa_cache = None if quantized else cache
 
         output = scaled_dot_product_attention(
-            q_nope, k, v, cache=cache, scale=self.scale, mask=pe_scores
+            q_nope, k, v, cache=sdpa_cache, scale=self.scale, mask=pe_scores
         )
         if L == 1:
             output = self.unembed_out(output)
--- a/mlx_lm/models/glm4_moe_lite.py
+++ b/mlx_lm/models/glm4_moe_lite.py
@@ -150,7 +150,19 @@ class Glm4MoeLiteAttention(nn.Module):
         if cache is not None:
             kv_latent, k_pe = cache.update_and_fetch(kv_latent, k_pe)
 
-        pe_scores = (q_pe * self.scale) @ k_pe.swapaxes(-1, -2)
+        # With a quantized cache the fetched entries are (packed, scales,
+        # biases) tuples.
+        quantized = isinstance(k_pe, (list, tuple))
+        if quantized:
+            pe_scores = mx.quantized_matmul(
+                q_pe * self.scale,
+                *k_pe,
+                transpose=True,
+                group_size=cache.group_size,
+                bits=cache.bits,
+            )
+        else:
+            pe_scores = (q_pe * self.scale) @ k_pe.swapaxes(-1, -2)
         if mask is not None:
             pe_scores = mx.where(
                 mask,
@@ -161,11 +173,19 @@ class Glm4MoeLiteAttention(nn.Module):
         if L == 1:
             q_nope = self.embed_q(q_nope)
             k = v = kv_latent
+            sdpa_cache = cache
         else:
-            k = self.embed_q(kv_latent, transpose=False)
-            v = self.unembed_out(kv_latent)
+            lat = kv_latent
+            if quantized:
+                # Chunked prefill over an already-quantized cache: fall back
+                # to float latents for the absorbed projections.
+                lat = mx.dequantize(*lat, group_size=cache.group_size, bits=cache.bits)
+            k = self.embed_q(lat, transpose=False)
+            v = self.unembed_out(lat)
+            sdpa_cache = None if quantized else cache
+
         output = scaled_dot_product_attention(
-            q_nope, k, v, cache=cache, scale=self.scale, mask=pe_scores
+            q_nope, k, v, cache=sdpa_cache, scale=self.scale, mask=pe_scores
         )
         if L == 1:
             output = self.unembed_out(output)
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -3251,6 +3251,36 @@ class TestModels(unittest.TestCase):
                 self.assertTrue(mx.allclose(y, y_gt, rtol=1e-4, atol=1e-4))
                 self.assertTrue(mx.allclose(st, st_gt, rtol=1e-4, atol=1e-3))
 
+    def test_quantized_sdpa_gqa_float_mask(self):
+        # Per-head float masks (e.g. the absorbed-MLA pe_scores) must be
+        # unflattened to (B, n_kv_heads, n_repeats, L, S) for GQA, not
+        # expanded on the kv axis.
+        from mlx_lm.models.base import quantized_scaled_dot_product_attention
+
+        B, n_q, n_kv, L, S, D = 1, 8, 2, 4, 64, 64
+        group_size, bits = 64, 8
+        mx.random.seed(0)
+        q = mx.random.normal((B, n_q, L, D))
+        k = mx.random.normal((B, n_kv, S, D))
+        v = mx.random.normal((B, n_kv, S, D))
+        mask = 0.1 * mx.random.normal((B, n_q, L, S))
+
+        q_k = mx.quantize(k, group_size=group_size, bits=bits)
+        q_v = mx.quantize(v, group_size=group_size, bits=bits)
+        out = quantized_scaled_dot_product_attention(
+            q, q_k, q_v, scale=1.0, mask=mask, group_size=group_size, bits=bits
+        )
+        self.assertEqual(out.shape, (B, n_q, L, D))
+
+        kd = mx.dequantize(*q_k, group_size=group_size, bits=bits)
+        vd = mx.dequantize(*q_v, group_size=group_size, bits=bits)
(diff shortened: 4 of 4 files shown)


pierre427:
**Independent validation on the absorbed-MLA path — LGTM.**

Reproduced on `deepseek_v2` (V2-Lite-Chat-4bit), which shares the absorbed-MLA quantized-cache path this PR fixes. Both branches exercised — decode (`L==1` `quantized_matmul` pe_scores) and chunked prefill (`L>1` dequantize-latent). No crash; teacher-forced ppl over a 108-token passage, with the 2nd chunk forced onto the already-quantized cache:

| cache | ppl | Δ |
|---|---|---|
| fp16 | 169.5 | — |
| kv8 | 170.1 | +0.4% |
| kv4 | 175.4 | +3.5% |

kv8 ≈ free, kv4 +3.5%, all decode samples coherent — consistent with "affine suffices on MLA."

One adjacent note on the `base.py` GQA mask hunk: a custom quantized-SDPA variant that reshapes queries to `(B, n_kv, n_repeats, L, S)` needs the head-axis split keyed on rank (`mask.ndim == scores.ndim - 1`) rather than `mask.shape[-3]` alone, otherwise a mask the attention layer pre-expanded to full rank gets double-split. Only relevant if you keep the reshape form; the expand-dims form in this PR is fine as written.

katlun-lgtm:
Thanks for running this! One clarification on coverage so the record is straight for whoever reviews: in stock mlx-lm, V2-Lite goes through the materialized per-head path in `deepseek_v2.py` (it caches full 192-dim keys rather than the latent), so I believe your run exercises the `base.py` mask fix plus quantized caching on an MLA-family model, while the `pe_scores` and prefill-dequant branches only exist in `deepseek_v3.py` — those I covered on Moonlight-16B (numbers in the description). If you were running an absorbed-form v2 variant locally then disregard, but figured it was worth spelling out.

Good point on keying the head split by rank rather than `shape[-3]`. Nothing in tree passes a pre-expanded 5-D mask so I kept the minimal form, but happy to switch to the `ndim` check if maintainers prefer the safer guard.

pierre427:
You're right, and thanks for the correction — in stock mlx-lm our V2-Lite run exercised the materialized per-head path plus your `base.py` mask fix, not the latent branches. We've now exercised the latent path directly, on `glm4_moe_lite` (same absorbed-MLA structure, already in-tree): stock + this PR crashes the same way (`'tuple' object has no attribute 'swapaxes'`, glm4_moe_lite.py L153), and the patch below — your deepseek_v3 treatment applied to it — fixes it. Verified on top of this PR's branch: prefill(17) → 5× decode (L==1 `quantized_matmul` pe_scores) → a 9-token chunk over the already-quantized cache (L>1 dequantize-latent). kv8 max |logit diff| vs float cache 0.0104, kv4 0.2103, argmax agreement 1.00 for both. Feel free to fold it into this PR; otherwise we'll PR it separately once this lands.

```diff
diff --git a/mlx_lm/models/glm4_moe_lite.py b/mlx_lm/models/glm4_moe_lite.py
index dc108af..15401a1 100644
--- a/mlx_lm/models/glm4_moe_lite.py
+++ b/mlx_lm/models/glm4_moe_lite.py
@@ -150,7 +150,19 @@ class Glm4MoeLiteAttention(nn.Module):
         if cache is not None:
             kv_latent, k_pe = cache.update_and_fetch(kv_latent, k_pe)
 
-        pe_scores = (q_pe * self.scale) @ k_pe.swapaxes(-1, -2)
+        # With a quantized cache the fetched entries are (packed, scales,
+        # biases) tuples.
+        quantized = isinstance(k_pe, (list, tuple))
+        if quantized:
+            pe_scores = mx.quantized_matmul(
+                q_pe * self.scale,
+                *k_pe,
+                transpose=True,
+                group_size=cache.group_size,
+                bits=cache.bits,
+            )
+        else:
+            pe_scores = (q_pe * self.scale) @ k_pe.swapaxes(-1, -2)
         if mask is not None:
             pe_scores = mx.where(
                 mask,
@@ -161,11 +173,19 @@ class Glm4MoeLiteAttention(nn.Module):
         if L == 1:
             q_nope = self.embed_q(q_nope)
             k = v = kv_latent
+            sdpa_cache = cache
         else:
-            k = self.embed_q(kv_latent, transpose=False)
-            v = self.unembed_out(kv_latent)
+            lat = kv_latent
+            if quantized:
+                # Chunked prefill over an already-quantized cache: fall back
+                # to float latents for the absorbed projections.
+                lat = mx.dequantize(*lat, group_size=cache.group_size, bits=cache.bits)
+            k = self.embed_q(lat, transpose=False)
+            v = self.unembed_out(lat)
+            sdpa_cache = None if quantized else cache
+
         output = scaled_dot_product_attention(
-            q_nope, k, v, cache=cache, scale=self.scale, mask=pe_scores
+            q_nope, k, v, cache=sdpa_cache, scale=self.scale, mask=pe_scores
         )
         if L == 1:
             output = self.unembed_out(output)
```

Three review notes on the `base.py` hunk while we were in there, one actionable:

1. **Your new `test_quantized_sdpa_gqa_float_mask` fails on an M5 at mlx 0.32.0** — not the logic, the tolerance: it asserts rtol/atol 1e-3, but 8-bit quantize/dequantize round-trip noise on this path measures 5e-3–8e-3. The sibling `test_quantized_sdpa` uses 1e-2 and passes. Suggest 1e-2 here too.
2. The `mask.ndim >= 3 → unflatten` trigger fires on masks that are *already* full-rank: a 5D `(B, n_kv, n_rep, L, S)` mask works at B=1 only by a broadcast accident and raises at B≥2. No in-tree caller today, but callers that pre-shape (we had one) hit it.
3. A 4D per-kv-head mask `(B, n_kv, L, S)` that broadcasts fine on `main` now raises after the unflatten — a small behavior regression for external callers of the helper.

All three rows go green with a narrower trigger, e.g. unflatten only when `mask.shape[-3] == n_kv_heads * n_repeats and mask.ndim < scores.ndim`, else `expand_dims(-3)`. Happy to push that as a suggestion too.

Last: this PR intersects #1323, which guards MLA caches *against* quantization. Having stepped through the three models there: deepseek_v32 genuinely can't take quantized KV (its indexer does `cache[0].keys = mx.depends(...)` on array keys and stores zero-width value tensors that `mx.quantize` can't pack) — that guard is correct; but kimi_k25 wraps `DeepseekV3Model`, so this PR already fixes it, and a blanket "MLA ⇒ no quant" would block the glm4_moe_lite path above too. The two PRs compose cleanly if the guard becomes a cache capability flag (`quantizable = False` set by v32's `make_cache`, checked in `maybe_quantize_kv_cache`) rather than an architecture test.

katlun-lgtm:
Folded your glm4_moe_lite patch in as-is, credited to you in the commit — thanks for chasing the latent path down, that's the one I couldn't exercise from deepseek_v3 alone.

All three review notes confirmed, and two of them are worse than they looked. I rebuilt the mask test as a rank matrix at B=2 (per-query-head, per-kv-head, broadcast, already-full-rank, no-heads) and on the previous PR head both the 4D per-kv-head and the 5D full-rank cases *raise* — the full-rank one only survived my original test because B=1 hid it behind a broadcast, exactly as you said.

I went with your narrower trigger in slightly different clothes: unflatten only when the mask's head axis is `n_q_heads` and its rank is below the scores' rank, otherwise `expand_dims(-3)`. That gets all five rows green and leaves a pre-shaped 5D mask untouched. Tolerance is 1e-2 now, matching `test_quantized_sdpa`.

Also added an MLA quantized-cache regression test (prefill 17 → 3 single-token decodes → a 9-token chunk over the quantized cache). On a tiny random-init glm4_moe_lite it's kv8 max |logit diff| 0.0147 with argmax agreement 1.00 — much weaker evidence than your real-checkpoint run, so yours is the number worth citing; mine just fails loudly if the tuple path breaks again.

On #1323: agreed, and the capability flag looks right to me. v32's guard is genuinely about the cache, not the architecture — a `quantizable = False` set by its `make_cache` and checked in `maybe_quantize_kv_cache` would keep it while leaving glm4_moe_lite and kimi_k25 free. Happy to defer to whoever gets there first.

pierre427:
Ran into two things while validating this PR's territory on our side that seem directly useful here:

**1. Confirms the gap: pristine `main` crashes on absorbed-MLA + quantized KV.** Tiny-config repro on `main` @ cf10f96 (random-weight `deepseek_v3`, any size — prefill with `QuantizedKVCache`, then one L==1 decode step):

```
AttributeError: 'tuple' object has no attribute 'swapaxes'
```

at the rope-scores matmul — `k_pe` comes back from the cache as a `(w, scales, biases)` tuple and the non-quantized `(q_pe * scale) @ k_pe.swapaxes(...)` branch receives it. So MLA + `--kv-bits` currently has no working decode path at all; this PR is fixing a real hole, not an edge case.

**2. A second landmine on the same path, worth defusing in this PR: the generic quantized-SDPA mask expansion double-expands repeat-shaped masks.** In `quantized_scaled_dot_product_attention`:

```python
if n_repeats > 1 and mask.ndim > 3:
    mask = mx.expand_dims(mask, -3)
```

If the model pre-expands its additive `pe_scores` mask to the queries' repeat shape `(B, 1, n_heads, L, S)` — which an absorbed-MLA implementation naturally does to broadcast against the reshaped `(B, n_kv=1, n_heads, L, D)` queries — the wrapper expands it a *second* time and the scores mis-broadcast. We hit exactly this on a deepseek_v3-family 105B (Sarvam-105B): quantized decode crashed with `[reshape] Cannot reshape array of size 2097152 into shape (1,64,1,512)`. One-line fix: expand only when `mask.ndim == 4` (a 5-D mask is already repeat-shaped). Happy to PR that separately if you'd rather keep this one focused.

Suggest also including an L==1 decode-step regression test with `QuantizedKVCache`: teacher-forced ppl harnesses only exercise the L>1 materialized-prefill path, which is why both of these survive ppl-based validation unseen.

**3. Second-model operating data**, if useful for motivation: Sarvam-105B (deepseek_v3 arch, q6 weights, M5, mlx 0.32, 4096-token teacher-forced ppl, chunked prefill): fp16 KV 124.90, kv8 123.78 (≈ free), kv4 133.55 (+6.9%). Decode overhead of the quantized path at 4k context: +26–31% ms/tok vs fp16 KV — consistent with quantized MLA KV being a memory lever rather than a speed one at short/mid context.

katlun-lgtm:
That second one's already in — it was your note 2 from the last round, landed in 68338c2. I ran your exact shape against both to be sure (n_kv=1, 64 query heads, L=1, S=512, 8-bit, B=1 and B=2):

| mask | main @ cf10f96 | this PR |
|---|---|---|
| 4-D per-query-head `(B, n_q, L, S)` | `ValueError: [reshape] Cannot reshape array of size 2097152 into shape (1,64,1,512)` | max abs diff 7.2e-05 |
| 5-D pre-expanded `(B, n_kv, n_rep, L, S)` | same `ValueError` | 7.2e-05 |

Same error string you saw on Sarvam.

Worth flagging though: the 4-D per-query-head mask raises identically on main, and an `ndim == 4` guard would keep expanding that one. That's the shape in-tree MLA actually produces — `pe_scores` comes out `(B, num_heads, L, S)` against a `kv_latent` with `n_kv = 1`, so deepseek_v3 and glm4_moe_lite both land on the 4-D row rather than the 5-D one. The trigger here unflattens when the head axis is `n_q_heads` and the mask's rank is below the scores' rank, and expands otherwise, which is what gets both rows green.

I took your shape into the mask matrix anyway (3296ebd) — it was only n_kv=2 / n_repeats=4 before. It's 5 mask ranks × 2 shapes now, and four of those ten error on main.

The decode-step test is already there as `test_mla_quantized_cache`: prefill 17, three `L == 1` decodes, then a 9-token chunk over the already-quantized cache, argmax-checked against a float cache. Agreed on why it's the one that matters — a ppl harness never touches the `L == 1` branch, which is exactly how both of these survived ppl-based validation.

Also put your Sarvam numbers in the description as a second-model data point, including the +26–31% decode overhead — "memory lever, not a speed lever" is the honest framing and it wasn't stated anywhere. Say the word if you'd rather I pull them.

pierre427:
@katlun-lgtm You're right on both — the `ndim == 4` guard is too coarse: the 4-D per-query-head `(B, n_q, L, S)` mask is exactly what in-tree MLA emits (`pe_scores` `(B, num_heads, L, S)`, `n_kv=1`), so keying on rank alone would keep expanding the shape that actually needs unflattening. Your head-axis-`n_q_heads`-plus-rank trigger is the correct generalization, and I'd missed that 68338c2 already covered the double-expand from the prior round — that one's on me for re-reporting a landed fix. The 5-mask-rank × 2-shape matrix at B=2 is the right regression shape. Keep the Sarvam numbers in the description. This LGTM as it stands; it needs a maintainer's eyes now more than more of mine.

katlun-lgtm:
@michalk8 would you be able to take a look at this one? It's a bugfix — `--kv-bits` crashes every absorbed-MLA model (deepseek_v3, glm4_moe_lite) on current main, and @pierre427 reviewed it against real checkpoints and signed off with "this LGTM as it stands; it needs a maintainer's eyes now more than more of mine."

It's mergeable with nothing outstanding on my side. The workflow run needs approving before any checks will show, if you want CI on it first.

nastya236:
Thanks for your contribution! We recently introduced a new policy regarding the use of AI in pull request descriptions and discussions #1770 .
Unfortunately, this means that we need to close this pull request. However, you’re very welcome to open a new one that follows the updated policy.

PR #1559, opened 2026-07-13
Add packed gated delta kernel, bitwise-pinned by an explicit-tree comparator

wyanzhao:
## Summary

Add a packed Metal kernel for the scalar-gated, unmasked `Dk == 128`
gated-delta prefill path.

| Area | Behavior |
|---|---|
| SIMD layout | 8 value rows per SIMD-group, 4 lanes per row |
| State ownership | 32 contiguous state elements per lane |
| Numerical contract | Bitwise equality with an explicit-tree comparator |
| Default behavior | Enabled for supported shapes |
| Escape hatch | `MLX_GDN_PACKED=0` restores the existing kernel |

## Motivation

The existing gated-delta prefill kernel assigns one full 32-lane
SIMD-group to each value row. With `Dk == 128`, each lane owns only four
state elements, while every timestep still performs two full-SIMD
reductions per row:

1. reduce the recurrent state/K product;
2. update the state; and
3. reduce the updated state/Q product.

This makes the kernel latency-bound on the target Qwen3.5/3.6 shapes.

## Implementation

The packed kernel maps eight independent value rows onto one SIMD-group:

- four lanes cooperate on each value row;
- each lane retains 32 contiguous state elements in registers; and
- the final two reduction levels use `simd_shuffle_xor(..., 1)` and
  `simd_shuffle_xor(..., 2)` within each four-lane row group.

The first three butterfly levels are evaluated locally inside each packed
lane. Each four-element partial preserves the sequential accumulation order
of the original lane.

## Numerical contract

The packed kernel is pinned against a test-only comparator that explicitly
writes the ascending butterfly tree:

```text
shuffle_xor(1) -> shuffle_xor(2) -> shuffle_xor(4)
               -> shuffle_xor(8) -> shuffle_xor(16)
```

This gives the packed implementation a device-independent, source-level
reduction contract. Tests require `array_equal` between the packed kernel
and the comparator across shapes, dtypes, and sequence lengths.

On the **validated M5 toolchain**, `simd_sum` lowers to the same tree, so the
packed kernel is also bit-identical to the existing kernel, including
full-model logits. A canary test compares the explicit-tree comparator with
the existing `simd_sum` kernel and will flag a device or toolchain whose
lowering differs. `metal_kernel` uses `MathMode::Safe`, preserving the
written operation order.

## Routing and fallback behavior

| Condition | Behavior |
|---|---|
| `Dk == 128` | Required for packed path |
| `Dv % 8 == 0` | Required for packed path |
| Scalar gate, no padding mask | Packed path |
| `g.dtype == float32`, `state.dtype == float32` | Packed path |
| Vector gate, padding mask, or unsupported shape | Existing kernel, unchanged |
| `MLX_GDN_PACKED=0` | Existing kernel, unchanged |

The specialization is otherwise generic in batch size, head counts, `Dv`,
sequence length, and input element type.

## Performance

M5 Max, Qwen3.6-35B-A3B shapes:

```text
B=1, Hk=16, Hv=32, Dk=Dv=128, T=2048
```

| Measurement | Existing kernel | Packed kernel | Result |
|---|---:|---:|---:|
| Per gated-delta layer | `2.54-2.56 ms` | `1.34-1.38 ms` | approximately `1.8-2.0x` |
| End-to-end prefill | baseline | optimized | `1.064x` / `+6.4%` |

The end-to-end result is the median of six saturated adjacent A/B pairs;
five of six pairs were positive.

## Validation

- Packed output and state are `array_equal` to the explicit-tree comparator
  across:
  - `bfloat16`, `float16`, and `float32` inputs;
  - batch and head-count variations;
  - `Dv` values that differ from `Dk`; and
  - `T` in `{1, 7, 64, 257}`.
- A canary test checks the comparator against the existing `simd_sum` kernel.
- Ops-reference tests cover:
  - the packed path;
  - padding masks;
  - vector gates; and
  - the generic `Dk == 64` path.
- The kill switch is verified to restore the existing kernel bitwise.
- Local results:
  - `tests/test_gated_delta.py`: **8/8 passing**;
  - existing gated-delta model regressions: **3/3 passing**.
- Full-repository `pre-commit run --all-files`: **passing**.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx-lm/blob/main/CONTRIBUTING.md) document.
- [x] I have run `pre-commit run --all-files`.
- [x] Added tests for packed, fallback, masked, vector-gate, and unsupported-shape paths.
- [x] Verified the `MLX_GDN_PACKED=0` kill switch.
- [x] No documentation changes are required for this internal kernel specialization.

### DIFF
--- a/mlx_lm/models/gated_delta.py
+++ b/mlx_lm/models/gated_delta.py
@@ -1,8 +1,17 @@
+import os
 from functools import partial
 from typing import Optional, Tuple
 
 import mlx.core as mx
 import mlx.nn as nn
+
+
+# For the shapes it supports, the packed kernel is bitwise-identical by
+# construction to an explicit-tree comparator kernel that the tests pin it
+# against (see _make_gated_delta_packed_kernel). Every other shape (masks,
+# vector gates, Dk != 128) and the MLX_GDN_PACKED=0 kill-switch use the
+# original simd_sum kernels, unchanged.
+_ENABLE_GDN_PACKED = os.environ.get("MLX_GDN_PACKED", "1") != "0"
 
 
 @partial(mx.compile, shapeless=True)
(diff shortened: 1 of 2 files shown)


pierre427:
Ran this on an **M5 (Apple, Neural-Accelerator / NAX path active)** against `Qwen3-Next-80B-A3B-Instruct` (MLX-4bit), same as we did for the SDPA NAX kernels. TL;DR: **lossless and a clean ~1.9–2.0× on the GDN step kernel.**

### NAX active (canary)
bf16 4096³ GEMM, mlx 0.32.0 wheel (NAX-class), one build/venv per side, mlx-lm varied only:
- base (unpacked): `2.49–3.58 ms`
- head (packed):   `2.49–2.99 ms`

Both well under the ~9 ms NAX-off floor, so the NAX path is live in both. (A couple of the whole-model runs drifted a bit hot from unrelated background I/O on the box — flagged below; it doesn't touch the correctness verdict.)

### Losslessness — token-identical ✅ (the headline)
Decoded **96 greedy tokens** from the same prompt under base (original `simd_sum` kernel) and head (packed kernel), and cross-checked each against a **batched-greedy reference** (full forward over the growing prefix each step, argmax — not stepwise-with-cache, per our batch-invariance discipline).

All four sequences are **byte-identical**: `base-stepwise == base-batched == head-stepwise == head-batched`. Packed GDN is a drop-in, no decode divergence.

Kernel-level numerics on model-realistic (L2-normalized q/k) inputs confirm it: `packed == xtree == unpacked`, **max |Δ| = 0.0** on both `y` and `state`. Your `tests/test_gated_delta.py` also passes here: **8 passed, 32 subtests**.
(Heads-up for anyone reproducing: feeding *unnormalized* random q/k into the step kernel overflows the delta recurrence to NaN over a few hundred steps and makes the equality checks spuriously fail — normalize q/k like the model does.)

### Kernel A/B — packed vs unpacked (isolated, one process, canary-clean)
Whole-model prefill A/B was too noisy to read a magnitude off — the GDN step is a fraction of total prefill (MoE GEMMs dominate) and thermal/background-I/O drift swamped the signal (base and head traded places across reps). So the timing below is the **step kernel in isolation**, `gated_delta_kernel` (packed) vs `gated_delta_kernel_unpacked`, back-to-back in one process on the model's own GDN dims — `B=1, Hk=16, Dk=128, Hv=32, Dv=128`. Canary `2.77 ms` on this run.

| T (prefill) | unpacked | packed | speedup |
|------------:|---------:|-------:|--------:|
| 2048        | 2.56 ms  | 1.33 ms | **1.93×** |
| 4096        | 4.91 ms  | 2.41 ms | **2.04×** |
| 8192        | 9.67 ms  | 4.87 ms | **1.99×** |

Consistent ~2× on the changed kernel, lossless, no regression on the routing that stays on the original path (masked / vector-gate / `Dk≠128` and the `MLX_GDN_PACKED=0` kill-switch are untouched). LGTM from M5.

zcbenz:
@wyanzhao We plan to add a builtin kernel to gated delta net at https://github.com/ml-explore/mlx/pull/4020,  can you please check if there are improvements that can be done there, and do you prefer waiting for it to land in core or merging this first?

wyanzhao:
Thanks @zcbenz — and thanks for the heads-up on ml-explore/mlx#4020. I went through it and left my notes there: https://github.com/ml-explore/mlx/pull/4020#issuecomment-5383270768. I'm glad a builtin op is coming and happy to help it land.

On ordering: if it works for you, my preference would be to merge this PR first, and I'll own the follow-up that moves mlx-lm onto the core op once it lands. My reasoning:

- This change is small (2 files), bitwise-identical to the existing kernel by construction, kill-switched (`MLX_GDN_PACKED=0`), and was independently reproduced on M5 by @pierre427 (token-identical greedy decode; 1.93–2.04× on the step kernel). It's mergeable today, so Qwen3.5 / 3.6 / Qwen3-Next users get the prefill speedup now.
- Even once #4020 lands, mlx-lm will still need a kernel of its own for the cases it currently sends to the per-timestep fallback (padding masks, Dk ≠ 128, head pairs outside its list, fp16) and for older mlx wheels. The packed kernel can keep serving exactly those paths, so it doesn't become dead code.
- Commitment: when #4020 is released, I'll send the mlx-lm PR that routes the unmasked `Dk == Dv == 128` case to `mx.fast.gated_delta_update` behind a version check and drops the packed kernel's default routing for that case.

That said, if you'd rather wait for core, that's completely fine with me — just say so and I'll put the effort into #4020 instead (the packed layout ports directly into its sequential kernel). Thanks again for taking a look.

zcbenz (approved):
Thanks!

PR #1560, opened 2026-07-13
Add nemotron_labs_diffusion (Nemotron-Labs-Diffusion: AR + block diffusion + self-spec)

pierre427:
# Add nemotron_labs_diffusion (Nemotron-Labs-Diffusion, tri-mode AR + block diffusion)

Adds `mlx_lm/models/nemotron_labs_diffusion.py` so mlx-lm can load and run
NVIDIA's Nemotron-Labs-Diffusion models (`model_type: nemotron_labs_diffusion`).
Validated on `smdesai/Nemotron-Labs-Diffusion-3B-4bit` (MLX 4-bit).

**A Ministral/Llama-style transformer that is TRI-MODE:**
- **AR** — standard causal decode; works out of the box with `mlx_lm.generate`
  (**199–222 tok/s, coherent** on M5). This alone makes it a normal AR model.
- **Block diffusion** — `diffusion_generate(...)`: masked bidirectional denoise
  over blocks (LLaDA-style), for the diffusion paradigm under discussion in
  #1490 / #1391 / #1402.
- **Linear self-speculation** — `linear_spec_generate(...)`: diffusion drafts a
  block (bidirectional, a small `linear_spec_lora` adapter ON), the same model
  AR-verifies (causal, adapter OFF), accept longest match + bonus. **Lossless**
  (emitted tokens are AR-argmax); measured 13.3 accepted tok/step on structured
  content.

Arch details handled: GQA (32/8, head_dim 128), YaRN RoPE (β_fast/β_slow,
mscale=1), Llama-4 long-context query scaling (gated to the >16k tail),
`diffusion_head`→`lm_head` in `sanitize`, PEFT-layout LoRA load/toggle.

**Depends on the `diffusion_generate` interface** being settled (see #1490).
The AR path is independent and could land alone; the diffusion + self-spec
methods follow whatever interface the diffusion discussion converges on.

### DIFF
--- a/mlx_lm/models/nemotron_labs_diffusion.py
+++ b/mlx_lm/models/nemotron_labs_diffusion.py
@@ -1,0 +1,856 @@
+# Copyright © 2026 Apple Inc.
+
+# Nemotron-Labs-Diffusion-3B: a TRI-MODE Ministral/Llama-style transformer that
+# supports both standard autoregressive (AR) decoding AND masked block
+# diffusion (dLM). The backbone (``NemotronLabsDiffusionModel.encoder`` in the
+# HF reference) is a Ministral-3 decoder — GQA attention, SwiGLU MLP, RMSNorm,
+# YaRN RoPE — with two non-standard wrinkles carried over from the reference:
+#
+#   1. A Llama-4-style per-position query scale
+#      ``1 + beta * log1p(floor(pos / orig_max_pos))`` applied to the queries
+#      after RoPE (``beta = rope_parameters.llama_4_scaling_beta``). For every
+#      position below ``original_max_position_embeddings`` this is exactly 1.0,
+#      so it only bends the very-long-context tail.
+#   2. A separate ``diffusion_head`` (untied Linear) is the LM head — the HF
+#      module is named ``diffusion_head`` rather than ``lm_head``.
+#
+# The HF weight tree prefixes the transformer with ``encoder.`` and names the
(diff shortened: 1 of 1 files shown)


pierre427:
Requesting review of the three decoding paths and the contracts they share in this PR.

Nemotron-H can be run autoregressively, through block diffusion, or through a target-verified linear self-speculative path. This makes the model a useful test of whether the diffusion interface introduced in #1490 represents decoding state generally enough, rather than only supporting LLaDA’s masking procedure.

The autoregressive path currently measures approximately 199–222 tokens/s. The self-speculative path remains target-verified and accepts an average of 13.3 tokens per target step in the reported run.

- whether deterministic autoregressive and self-speculative decoding produce equivalent committed output across rejection patterns and generation lengths;
- whether cache and model state are restored correctly when only part of a proposed block is accepted;
- whether the linear proposal mechanism depends on assumptions that should remain local to Nemotron-H rather than entering the shared generator;
- how the distribution of accepted block lengths varies across prompts, rather than only its mean;
- whether block-diffusion state, stopping conditions, and partially resolved sequences fit the interface without bypassing its invariants;
- whether stochastic sampling has a clearly defined equivalence boundary across the three modes;
- whether performance comparisons separate proposal acceptance from the cost of constructing and verifying each block.

The model follows an existing architecture. The part needing close examination is whether autoregressive, diffusion, and self-speculative decoding can share one state interface without obscuring the different correctness conditions of each mode.

zcbenz:

PR #1561, opened 2026-07-13
Add llada2_moe (LLaDA2.0 — MoE masked-diffusion LM)

pierre427:
# Add llada2_moe (LLaDA2.0 — MoE masked-diffusion LM)

Adds `mlx_lm/models/llada2_moe.py` (`model_type: llada2_moe`). Validated on
`mlx-community/LLaDA2.0-mini-preview-4bit` (16B MoE / ~1.4B active, MLX 4-bit).
The **MoE sequel to LLaDA (#1490)** — the second masked-diffusion model that
motivates standardizing the `diffusion_generate` interface.

- **MoE**: 256 experts, top-8, 1 shared, grouped-topk routing with `expert_bias`
  (noaux_tc, n_group 8), `first_k_dense_replace 1` — reuses the DeepSeek-V3
  grouped-router + `SwitchGLU` expert stacking already in mlx-lm.
- **Diffusion** (`diffusion_generate`): masked bidirectional denoise, no KV
  cache; built-in Fast-dLLM confidence-threshold parallel decode — **~2× faster
  at threshold 0.9 with no quality loss** (validated). Coherent English output.
- Partial RoPE (rotate 64 of 128), per-head QK-norm, block-causal diffusion mask
  in the model's compute dtype (bf16).

`sanitize` stacks per-expert `mlp.experts.{e}.*` → `switch_mlp.*` (incl. 4-bit
scales/biases); router `expert_bias` kept full-precision.

**Depends on #1490's `diffusion_generate` interface.** Best submitted as a
follow-on once that interface lands (this PR is the "more diffusion models
exist" evidence for standardizing it).

### DIFF
--- a/mlx_lm/models/llada2_moe.py
+++ b/mlx_lm/models/llada2_moe.py
@@ -1,0 +1,489 @@
+# Copyright © 2025-2026 Apple Inc.
+
+# LLaDA2-MoE: a Mixture-of-Experts masked-diffusion language model (the MoE
+# version of LLaDA). Architecturally it fuses two well-trodden pieces of the
+# mlx-lm zoo:
+#
+#   * The MoE backbone is DeepSeek-V3 / Hunyuan-shaped: sigmoid-scored router
+#     with a per-expert bias term, grouped-topk (``n_group`` / ``topk_group``)
+#     routing, ``norm_topk_prob`` renormalization + ``routed_scaling_factor``,
+#     one always-on shared expert, and a ``first_k_dense_replace`` split where
+#     the first layer(s) use a plain dense MLP. We reuse the exact
+#     ``group_expert_select`` routing from ``deepseek_v3`` and the ``SwitchGLU``
+#     stacked-expert kernel.
+#
+#   * Generation is diffusion (masked bidirectional denoising, NO KV cache),
+#     block-wise semi-autoregressive with a block-causal attention mask, mirrored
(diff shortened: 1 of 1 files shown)


pierre427:
Requesting review of the LLaDA2 model mapping and thresholded diffusion path in this PR.

This adds a second diffusion-model family after #1490, with MoE execution and Fast-dLLM-style confidence thresholding. The reported thresholded configuration is approximately 2× faster than the baseline diffusion sampler without a measured quality loss in the current evaluation.

Beyond verifying the model port, this provides another test of whether the shared diffusion interface describes common denoising behaviour or merely accommodates each architecture through separate branches.

- whether converted logits and expert routing match the reference implementation before sampling differences are introduced;
- whether top-k routing, expert normalization, and any shared-expert behaviour preserve the checkpoint’s intended computation;
- whether confidence thresholding changes only the number of positions updated or also changes the resulting denoising trajectory;
- how sensitive the speed and quality results are to the chosen threshold, prompt distribution, generation length, and random seed;
- whether the threshold was selected independently of the prompts used for the reported quality comparison;
- whether cached or incremental state remains valid when token positions can be revised between steps;
- which differences between LLaDA and LLaDA2 are represented through model-provided policy rather than conditions in the shared generation loop.

The useful test here is not whether one threshold wins on a single workload, but whether the speed–quality relationship is stable enough to expose as a general decoding control.

zcbenz:

PR #1564, opened 2026-07-15
 Fix LoRA/DoRA fuse silently dropping adapter on quantized base (#1172)

### LINKED ISSUE #1172: mlx_lm fuse produces non-functional model with gpt-oss-20b MXFP4-Q8 (adapter-loaded serving works)
## Summary

`mlx_lm fuse` produces non-functional fused models from `mlx-community/gpt-oss-20b-MXFP4-Q8` with LoRA adapters. The fused model behaves identically to the base model without any adapter — 0% task success across 139 validation examples. Adapter-loaded serving (`mlx_lm server --adapter-path`) works perfectly (95–100% success).

Both quantized fuse and `--dequantize` (fp16) produce identical 0% results, ruling out quantization noise as the cause.

## Environment

- mlx-lm: 0.31.2
- mlx: 0.31.1
- macOS 26.4.1, Apple M4 Max 128GB
- Python 3.12

## Base model and adapter

- Base: `mlx-community/gpt-oss-20b-MXFP4-Q8` (~11 GB, pre-quantized MoE model)
- LoRA: rank 16, alpha 32, target_modules `["self_attn.q_proj", "self_attn.k_proj", "self_attn.v_proj", "self_attn.o_proj"]`, num_layers 16
- Trained with `mlx_lm.lora` (same version), training converged normally

## Reproducer

```bash
# 1. Train adapter (or use any existing adapter for this base)
python -m mlx_lm lora --model mlx-community/gpt-oss-20b-MXFP4-Q8 \
  --adapter-path models/lora/stage_c --train ...

# 2. Fuse (quantized) — produces broken model
python -m mlx_lm fuse \
  --model mlx-community/gpt-oss-20b-MXFP4-Q8 \
  --adapter-path models/lora/stage_c \
  --save-path models/fused/stage_c

# 3. Fuse (dequantized fp16) — ALSO produces broken model
python -m mlx_lm fuse --dequantize \
  --model mlx-community/gpt-oss-20b-MXFP4-Q8 \
  --adapter-path models/lora/stage_c \
  --save-path /tmp/stage_c_dequant

# 4. Serve fused model — outputs base-model behavior (no adapter effect)
python -m mlx_lm server --model models/fused/stage_c --port 8090

# 5. Serve adapter-loaded — works correctly
python -m mlx_lm server --model mlx-community/gpt-oss-20b-MXFP4-Q8 \
  --adapter-path models/lora/stage_c --port 8090
```

## Expected vs actual

| Serving mode | Schema validity | Behavior |
|---|---|---|
| Adapter-loaded (`--adapter-path`) | 95–100% (75/76 to 76/76) | Correct structured JSON output |
| Fused (quantized) | 0% (0/24, 0/76, 0/39 across 3 stages) | Emits base-model reasoning tokens, no JSON |
| Fused (`--dequantize`, fp16) | 0% (0/76, 152 requests) | Identical to quantized fuse — base-model behavior |

The fused weight files have different MD5 hashes from the base (confirming fuse modified something), but inference produces base-model output.

## Suspected causes

Two hypotheses for where the merge fails:

1. **MoE expert layers not being fused.** gpt-oss-20b is a mixture-of-experts model. If `mlx_lm fuse` only handles attention-projection targets and silently skips expert modules, most of the learned LoRA correction would be lost. Diagnostic: per-layer weight-diff between fused and base weights — check whether all LoRA-targeted module names show non-zero diffs in the fused output.

2. **Target-module naming mismatch between lora and fuse.** During training, we found that LoRA required full dotted paths (`self_attn.q_proj`) rather than bare names (`q_proj`). If `mlx_lm fuse` uses a different module-matching convention than `mlx_lm lora`, it could silently skip all targets. Same diagnostic as above.

Either cause would explain the symptom: a fused file with a different MD5 (because *something* got modified) but base-like behavior (because the *important* deltas weren't applied).

## Workaround

Adapter-loaded serving works correctly and is what we use in production:
```bash
python -m mlx_lm server --model mlx-community/gpt-oss-20b-MXFP4-Q8 \
  --adapter-path models/lora/stage_c --port 8090
```

The broken fused models are preserved at `models/fused_broken_mlx_v0312/` for debugging if helpful.

vaibzvb:
## Problem
     
  `mlx_lm fuse` on a quantized base model with a trained LoRA/DoRA adapter produces a fused model that behaves like the
  unadapted base , the adapter's effect is silently dropped. Originally reported for `gpt-oss-20b-MXFP4-Q8` (#1172). I
  reproduced it on plain `mlx-community/Llama-3.2-1B-Instruct-4bit` with a small trained "ZorpBot" adapter, so the bug is
   general to the quantized-fuse path, not specific to gpt-oss or MXFP4.
  
  Before this fix, on the same "What is your name?" prompt:
  
  | path | output |
  |---|---|
  | base + running adapter | *"I am ZorpBot, your dedicated assistant."* |
  | `fuse --dequantize` | *"I am ZorpBot, your dedicated assistant."* |
  | `fuse` (quantized, default) | *"I'm an artificial intelligence model known as Llama..."* |

  ## Root cause

  `LoRALinear.fuse(dequantize=False)` (and the sibling fuse methods on `LoRASwitchLinear`, `LoRAEmbedding`, `DoRALinear`)
   requantized the fused weight back to the base's bit-width:

  ```python
  weight = mx.dequantize(...)           # base → fp16
  fused_linear.weight = weight + delta  # add LoRA delta
  fused_linear = nn.QuantizedLinear.from_linear(fused_linear, ..., bits=linear.bits, ...)
  ```
  
  Measured on layer 0 `q_proj` of Llama-3.2-1B-Instruct-4bit with a real trained adapter:
  
  - LoRA delta: `max ≈ 2.2e-3`, `mean ≈ 2.4e-4`
  - 4-bit affine quantization step (per group of 64): `≈ 8.4e-2`
  - Requantization round-trip noise: `max ≈ 4.1e-2`

  `|δ| / step ≈ 0.026` — the delta never crosses the 0.5 rounding boundary, so it is rounded away and replaced by
  incoherent quantization noise. This is a fundamental precision limit at low bit-widths, not a bug in
  `nn.QuantizedLinear.from_linear` itself. Verified by testing:

  - fp32 accumulation before requantization: no help (same result)
  - higher bit widths: 8-bit mostly preserves the delta, 6-bit partially, 4-bit destroys it
  
  ## Changes
  
  - `mlx_lm/tuner/lora.py` — `LoRALinear.fuse`, `LoRASwitchLinear.fuse`, `LoRAEmbedding.fuse` no longer requantize the
  fused weight. The adapted layer is returned in full precision.
  - `mlx_lm/tuner/dora.py` — `DoRALinear.fuse` gets the same fix (same pattern, same failure mode).
  - `tests/test_finetune.py` — added `test_lora_fuse_quantized_preserves_adapter`: builds a quantized base + `LoRALinear`
   with a realistic small delta, fuses with `dequantize=False`, and asserts the fused output tracks the LoRA output
  rather than the base (`residual < 0.1 × adapter_signal`). Verified to fail on pre-fix code (residual 0.94, threshold
  0.078) and pass post-fix (residual 0.00).

  Non-adapted layers stay quantized. `fuse.py --dequantize` still dequantizes the whole model. The `dequantize` parameter
   on `fuse()` is kept for API compatibility.

  ## Trade-off worth calling out
  
  Adapter-adapted layers are now fp16 in the fused model. For a LoRA that targets attention + MLP across all layers, this
   effectively brings the fused file size close to the fully-dequantized size (e.g., ~2 GB vs the previously-broken ~700
  MB for Llama-3.2-1B-4bit). This is unavoidable — any subsequent requantization step would face the same physics and
  destroy the adapter again. Users who want a smaller quantized fused model with a working adapter need a training recipe
   that produces deltas larger than the quantization step size (higher LR, more iters, larger scale), or a higher target
  bit-width.

  ## Tests

  - `python -m unittest tests.test_finetune` — 16/16 pass (15 existing + 1 new regression test).
  - `python -m unittest discover tests/` — 209/210 pass. The one failure (`tests.test_datsets.TestDatasets.test_hf`) is
  pre-existing on `main` (unrelated HF datasets URI issue with `billsum`).
  - `pre-commit run --all-files` — black + isort pass.
  - End-to-end: fused Llama-3.2-1B-4bit ZorpBot model now generates *"I am ZorpBot, your dedicated assistant."* on the
  target prompt, matching the running-adapter output.
  Fixes #1172

### DIFF
--- a/mlx_lm/tuner/dora.py
+++ b/mlx_lm/tuner/dora.py
@@ -46,13 +46,9 @@ class DoRALinear(nn.Module):
         if bias:
             fused_linear.bias = linear.bias
 
-        if self._is_quantized() and not dequantize:
-            fused_linear = nn.QuantizedLinear.from_linear(
-                fused_linear,
-                group_size=linear.group_size,
-                bits=linear.bits,
-                mode=linear.mode,
-            )
+        # See LoRALinear.fuse: requantizing the fused weight at the base's
+        # bit-width silently discards the adapter delta (issue #1172), so keep
+        # the adapted layer in full precision.
         return fused_linear
 
     def __init__(
--- a/mlx_lm/tuner/lora.py
+++ b/mlx_lm/tuner/lora.py
@@ -54,14 +54,14 @@ class LoRALinear(nn.Module):
         if bias:
             fused_linear.bias = linear.bias
 
-        if is_quantized and not dequantize:
-            fused_linear = nn.QuantizedLinear.from_linear(
-                fused_linear,
-                linear.group_size,
-                linear.bits,
-                mode=linear.mode,
-            )
-
+        # Historically, when ``dequantize=False`` the fused layer was requantized
+        # back to the base's bit-width via ``QuantizedLinear.from_linear``. At
+        # low bit-widths the quantization step is much larger than a typical
+        # LoRA delta, so requantization discards the adapter effect and the
+        # fused model behaves like the unadapted base (issue #1172). We keep
+        # the adapted layer in full precision instead; non-adapted layers
+        # elsewhere in the model stay quantized as before, and ``fuse.py``
+        # still honours ``--dequantize`` for dequantizing the whole model.
         return fused_linear
 
     def __init__(
@@ -141,11 +141,9 @@ class LoRASwitchLinear(nn.Module):
         if bias:
             fused_linear.bias = linear.bias
 
-        if is_quantized and not dequantize:
-            fused_linear = fused_linear.to_quantized(
-                group_size=linear.group_size, bits=linear.bits, mode=linear.mode
-            )
-
+        # See LoRALinear.fuse: requantizing the fused weight at the base's
+        # bit-width silently discards the LoRA delta (issue #1172), so keep
+        # the adapted expert layer in full precision.
         return fused_linear
 
     def __init__(
@@ -237,14 +235,9 @@ class LoRAEmbedding(nn.Module):
         lora_b = self.lora_b
         fused_embedding.weight = weight + (lora_a @ lora_b).astype(weight.dtype)
 
-        if is_quantized and not dequantize:
-            fused_embedding = nn.QuantizedEmbedding.from_embedding(
-                fused_embedding,
-                group_size=embedding.group_size,
-                bits=embedding.bits,
-                mode=embedding.mode,
-            )
-
+        # See LoRALinear.fuse: requantizing the fused weight at the base's
+        # bit-width silently discards the LoRA delta (issue #1172), so keep
+        # the adapted embedding in full precision.
         return fused_embedding
 
     def __init__(
--- a/tests/test_finetune.py
+++ b/tests/test_finetune.py
@@ -141,6 +141,46 @@ class TestLora(unittest.TestCase):
         self.assertFalse(mx.array_equal(dequantized_weight, new_embedding.weight))
         self.assertFalse(mx.array_equal(embedding(tokens), lora_emb(tokens)))
 
+    def test_lora_fuse_quantized_preserves_adapter(self):
+        # Regression test for issue #1172: fusing a LoRA adapter into a
+        # quantized base used to requantize the fused weight at the base's
+        # bit-width, which silently discarded LoRA deltas smaller than the
+        # quantization step size — the fused model then behaved like the
+        # unadapted base.
+        mx.random.seed(0)
+        in_dims, out_dims, r = 512, 512, 8
+        x = mx.random.normal(shape=(4, in_dims))
+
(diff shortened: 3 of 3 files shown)


vaibzvb:
Hi @awni @angeloskath 
  This fixes #1172: `mlx_lm fuse` on a quantized base silently drops the LoRA/DoRA adapter
  because the fused weight gets requantized at the base's bit-width and the adapter delta
  rounds away. Includes a regression test that fails on `main` and passes with the fix.
  
  Since this is my first PR to the repo, CI is waiting on maintainer approval before it can
   run, would one of you be able to kick it off?

zcbenz:

PR #1567, opened 2026-07-15
Fix quantized_kv_start default mismatch between generate_step and the CLI

### LINKED ISSUE #1566: quantized_kv_start default (0) in generate_step silently tanks decode throughput vs the CLI's own recommended default (5000)
## Problem

`generate_step()` and `speculative_generate_step()` (`mlx_lm/generate.py`) both default `quantized_kv_start` to `0`. When a caller passes `kv_bits=` (to enable KV cache quantization) without also explicitly setting `quantized_kv_start`, the cache quantizes from the very first token — and pays the full performance cost of quantized decoding for the entire generation, even on short contexts where the memory savings are negligible.

The CLI entry points (`mlx_lm.generate`, `mlx_lm.cache_prompt`) already avoid this: they default `--quantized-kv-start` to `DEFAULT_QUANTIZED_KV_START = 5000`, so quantization only kicks in once the cache is long enough that the memory savings are worth the cost. But that constant is never applied as the *library function's* own default — so anyone calling `generate_step`/`speculative_generate_step` directly (as a library, not via the CLI) silently gets the worst-case behavior.

## Root cause investigation

The quantized attention path (`quantized_scaled_dot_product_attention` in `mlx_lm/models/base.py`) already uses a fused kernel (`mx.quantized_matmul`) rather than a naive dequant-then-matmul — that was my initial (wrong) hypothesis, ruled out by reading the code. The real gap is structural: the fp16 path is a single fused Metal kernel (`mx.fast.scaled_dot_product_attention`, QK+softmax+AV in one launch), while the quantized path is 3 separate ops (`quantized_matmul` → `softmax` → `quantized_matmul`), with an intermediate `scores` tensor materialized to device memory between them. That per-step kernel-launch/materialization overhead is roughly fixed cost, while the memory-bandwidth savings from quantization scale with cache size — hence the crossover behavior below. Closing that gap needs a fused quantized-attention Metal kernel in `mlx` (core), out of scope for `mlx-lm`.

## Measurements

Hardware: Apple M4 Pro, 24GB. `mlx==0.32.0`, `mlx-lm@15b522f`. 8 measured runs per config (+1 discarded warmup), `mx.clear_cache()` between runs. Full harness/raw data in the companion PR.

**512-token prompt, 64 generated tokens (decode tok/s):**

| Model | fp16 | quantized, `quantized_kv_start=0` (current default) | quantized, `quantized_kv_start=5000` |
|---|---:|---:|---:|
| Qwen2.5-0.5B-Instruct-4bit | 424.6 ± 1.6 | 352.2 ± 4.9 (**-17.1%**) | 425.8 ± 1.0 (parity) |
| Llama-3.2-1B-Instruct-4bit | 256.6 ± 14.6 | 260.0 ± 6.0 | 265.1 ± 4.1 |

**5120-token prompt (past the 5000 threshold) — quantization still engages, and wins:**

| Model | fp16 | quantized (`quantized_kv_start=5000`) |
|---|---:|---:|
| Qwen2.5-0.5B-Instruct-4bit | 296.4 ± 0.8 | 307.4 ± 0.9 (**+3.7%**) |
| Llama-3.2-1B-Instruct-4bit | 177.6 ± 0.6 | 210.1 ± 2.9 (**+18.3%**) |

So the fix isn't "disable quantization" — it's "defer it to where it's a net win," which is exactly what the CLI already assumes but the library function doesn't enforce by default.

## Proposed fix

Align `generate_step`/`speculative_generate_step`'s `quantized_kv_start` default with `DEFAULT_QUANTIZED_KV_START` (5000), matching the existing CLI default. See linked PR.

Vlor999:
Closes #1566.

## Summary

`generate_step()` and `speculative_generate_step()` defaulted `quantized_kv_start` to `0`, so any caller passing `kv_bits=` without also specifying `quantized_kv_start` quantized the KV cache from the very first token — paying the full performance cost of quantized decoding even on short contexts where the memory savings are negligible.

The CLI (`mlx_lm.generate`, `mlx_lm.cache_prompt`) already uses the intended default via `DEFAULT_QUANTIZED_KV_START = 5000`, deferring quantization until the cache is long enough for the memory savings to be worth the cost. This PR aligns the library function defaults with that existing, documented CLI behavior. See #1566 for the full root-cause investigation (ruled out a "missing fused dequant kernel" hypothesis — the real gap is 3 unfused Metal kernel launches vs 1 fused `mx.fast.scaled_dot_product_attention` for fp16, a structural `mlx`-core limitation, not an `mlx-lm` bug).

## Measurements (M4 Pro, `mlx==0.32.0`, 8 runs/config)

**512-token prompt, decode tok/s:**

| Model | fp16 | quantized, start=0 (old default) | quantized, start=5000 (this fix) |
|---|---:|---:|---:|
| Qwen2.5-0.5B-Instruct-4bit | 424.6 ± 1.6 | 352.2 ± 4.9 (-17.1%) | 425.8 ± 1.0 (parity) |
| Llama-3.2-1B-Instruct-4bit | 256.6 ± 14.6 | 260.0 ± 6.0 | 265.1 ± 4.1 |

**5120-token prompt (past the threshold) — quantization still engages, and is faster than fp16:**

| Model | fp16 | quantized (start=5000) |
|---|---:|---:|
| Qwen2.5-0.5B-Instruct-4bit | 296.4 ± 0.8 | 307.4 ± 0.9 (+3.7%) |
| Llama-3.2-1B-Instruct-4bit | 177.6 ± 0.6 | 210.1 ± 2.9 (+18.3%) |

## Test plan

- [x] `python -m pytest tests/test_generate.py tests/test_prompt_cache.py` — 48 passed, no changes needed.
- [x] `black --check mlx_lm/generate.py` — clean.
- [x] Before/after benchmark harness + raw JSONL results available on request (not included in this diff to keep it minimal — happy to add under `benchmarks/` if maintainers want it in-tree).

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -9,7 +9,6 @@ import sys
 import time
 from collections import deque
 from dataclasses import dataclass
-from functools import partial
 from typing import Any, Callable, Generator, List, Optional, Sequence, Tuple, Union
 
 import mlx.core as mx
@@ -307,7 +306,7 @@ def generate_step(
     prefill_step_size: int = 2048,
     kv_bits: Optional[int] = None,
     kv_group_size: int = 64,
-    quantized_kv_start: int = 0,
+    quantized_kv_start: int = DEFAULT_QUANTIZED_KV_START,
     prompt_progress_callback: Optional[Callable[[int, int], None]] = None,
     input_embeddings: Optional[mx.array] = None,
 ) -> Generator[Tuple[mx.array, mx.array], None, None]:
@@ -333,7 +332,11 @@ def generate_step(
           None implies no cache quantization. Default: ``None``.
         kv_group_size (int): Group size for KV cache quantization. Default: ``64``.
         quantized_kv_start (int): Step to begin using a quantized KV cache.
-           when ``kv_bits`` is non-None. Default: ``0``.
+           when ``kv_bits`` is non-None. Default: ``DEFAULT_QUANTIZED_KV_START``
+           (``5000``) — quantizing from the first token measurably slows down
+           decoding (no fused kernel for quantized attention), so quantization
+           only kicks in once the cache is long enough that the memory
+           savings are worth that cost.
         prompt_progress_callback (Callable[[int, int], None]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
         input_embeddings (mx.array, optional): Input embeddings to use instead of or in
@@ -474,7 +477,7 @@ def speculative_generate_step(
     prefill_step_size: int = 512,
     kv_bits: Optional[int] = None,
     kv_group_size: int = 64,
-    quantized_kv_start: int = 0,
+    quantized_kv_start: int = DEFAULT_QUANTIZED_KV_START,
 ) -> Generator[Tuple[mx.array, mx.array, bool], None, None]:
     """
     A generator producing token ids based on the given prompt from the model.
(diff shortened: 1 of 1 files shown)


Vlor999:
Superseded by the PR linked from #1651, rebuilt on current `main` with measurements and tests.

PR #1577, opened 2026-07-17
Fall back to the naive detokenizer when the tokenizer backend lacks vocab

### LINKED ISSUE #1576: `generate()` crashes with masked AttributeError ("no attribute '_detokenizer'") when transformers loads MistralCommonBackend (Ministral 3, transformers 5.x)
# Environment

- mlx-lm 0.31.3 (also present on current main)
- transformers 5.14.1, mistral-common 1.11.5
- mlx 0.32.0, Python 3.14, macOS (Apple Silicon)
- Model: `mlx-community/Ministral-3-3B-Instruct-2512-4bit`

# Reproduction

Reproduces only with mistral-common installed alongside transformers 5.x; that combination makes `AutoTokenizer` return `MistralCommonBackend` for this repo. Without mistral-common the failure mode is different (the tokenizer class cannot be resolved at all).

```python

model, tokenizer = load("mlx-community/Ministral-3-3B-Instruct-2512-4bit")  # succeeds
    [{"role": "user", "content": "hi"}], add_generation_prompt=True
)
generate(model, tokenizer, prompt=prompt, max_tokens=5)  # crashes
```

```
  File ".../mlx_lm/tokenizer_utils.py", line 459, in __getattr__
    return self._detokenizer
  File ".../mlx_lm/tokenizer_utils.py", line 463, in __getattr__
    return self.__getattribute__(attr)
AttributeError: 'TokenizerWrapper' object has no attribute '_detokenizer'. Did you mean: 'detokenizer'?
```

The message is misleading. The underlying error (reachable via `tok._detokenizer_class(tok)`) is `AttributeError: MistralCommonBackend has no attribute vocab`, raised in `BPEStreamingDetokenizer.__init__`.

# Root cause

Two problems combine:

1. Detokenizer selection and tokenizer loading disagree. `tokenizer_utils.load()` picks the streaming detokenizer class from `tokenizer.json` on disk (ByteLevel decoder, so `BPEStreamingDetokenizer`). But the repo also ships `tekken.json`, and transformers 5.x with mistral-common installed loads the tokenizer as `MistralCommonBackend`, which has `get_vocab()` and `decode()` but no `.vocab`, the attribute both fast streaming detokenizers require.
2. The real error is masked. `detokenizer` is a property; an `AttributeError` escaping a property falls through to `__getattr__("detokenizer")`, which returns `self._detokenizer`, an attribute no code path ever sets. The user sees the unrelated `_detokenizer` error instead of the missing-`vocab` one.

# Proposed fix

PR to follow: fall back to `NaiveStreamingDetokenizer` when the loaded backend has no `vocab`, and surface the real construction error instead of masking it. Fast paths untouched.

Distinct from #1041 / #1329 (tekken fast path emitting `Ġ` characters): that is a correctness bug inside a constructed detokenizer; this is a crash before any detokenizer exists.

Lee-Gihun:
Fixes #1576

`generate()` crashes for models that transformers 5.x loads as `MistralCommonBackend` (e.g. Ministral 3): the detokenizer class is chosen by inspecting `tokenizer.json` on disk, but the loaded backend has no `.vocab`, and the resulting `AttributeError` gets masked into `'TokenizerWrapper' object has no attribute '_detokenizer'` by the property / `__getattr__` interaction. Details in the issue.

Changes in `tokenizer_utils.py`:

- `load()`: fall back to `NaiveStreamingDetokenizer` when the loaded tokenizer has no `vocab` (the naive detokenizer only needs `decode()`). Fast paths unchanged for every tokenizer that has one.
- `detokenizer` property: on construction failure, warn and fall back to the naive detokenizer; if that also fails, re-raise as `RuntimeError` so the original error is no longer swallowed.

Tests: two offline regressions in `tests/test_tokenizers.py`; both fail on main and pass with this change. They use `warnings.catch_warnings` instead of `assertWarns` because `assertWarns` walks `sys.modules` and pulls torch-dependent transformers imports into torch-less environments.

Manually verified: `Ministral-3-3B-Instruct-2512-4bit` now generates end to end; `Qwen2.5-0.5B-Instruct-4bit` (BPE) and `Mistral-7B-Instruct-v0.3-4bit` (SPM) still select their fast detokenizers.

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -450,7 +450,24 @@ class TokenizerWrapper:
         """
         Get a stateful streaming detokenizer.
         """
-        return self._detokenizer_class(self)
+        try:
+            return self._detokenizer_class(self)
+        except AttributeError as e:
+            # An AttributeError escaping a property falls through to
+            # __getattr__ and gets masked, so handle it here.
+            if self._detokenizer_class is NaiveStreamingDetokenizer:
+                raise RuntimeError(
+                    f"Failed to create the streaming detokenizer: {e}"
+                ) from e
+            # The fast detokenizer needs attributes the loaded backend
+            # does not provide; the naive one only needs decode, so
+            # degrade to it instead of failing.
+            warnings.warn(
+                f"Could not create {self._detokenizer_class} ({e}). "
+                "Falling back to NaiveStreamingDetokenizer."
+            )
+            self._detokenizer_class = NaiveStreamingDetokenizer
+            return self._detokenizer_class(self)
 
     def __getattr__(self, attr):
         if attr == "detokenizer":
@@ -612,6 +629,12 @@ def load(
     tokenizer = AutoTokenizer.from_pretrained(
         model_path, **(tokenizer_config_extra or {})
     )
+
+    if not hasattr(tokenizer, "vocab"):
+        # The detokenizer was selected from tokenizer.json, but AutoTokenizer
+        # may load a different backend (e.g. MistralCommonBackend for models
+        # with a tekken.json) which lacks the vocab the fast ones need.
+        detokenizer_class = NaiveStreamingDetokenizer
 
     tokenizer_config = tokenizer.init_kwargs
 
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1579, opened 2026-07-17
Fused Metal kernels for the SSD prefill path in ssm_update

davidtorcivia:
## Summary

Adds two `mx.fast.metal_kernel` JIT kernels for the chunked SSD computation and dispatches to them from `ssm_update` for supported shapes. The existing prefill path (`ssm_attn`) is graph-op bound: at `nemotron_h` shapes and T=8192 the ~25 GFLOP scan takes 47 ms/layer, as long as the layer's 640 GFLOP in_proj.

Per-layer scan time (M3 Ultra, bf16, nemotron_h shapes h=128, dh=64, g=8, ds=128):

| T | ssm_attn | fused | speedup |
|---|---|---|---|
| 2048 | 11.6 ms | 5.2 ms | 2.2x |
| 4096 | 24.0 ms | 8.7 ms | 2.7x |
| 8192 | 46.8 ms | 17.2 ms | 2.7x |

End to end this is +11% prefill on Nemotron-3-Super-120B (617 -> 680 tok/s at a 48k-token prompt); the scan is ~38% of its prefill. Any Mamba2-family model served through `ssm_update` benefits (Nemotron-H, Mamba2, Codestral, Falcon-H1, Granite 4).

## Design

Chunked SSD over CHUNK=64 token chunks, two kernels:

- Kernel A (parallel over `(batch, head, chunk)`): inclusive log-decay cumsum; decayed scores `A_ij = (C_i.B_j) exp(lcg_i - lcg_j)` held in registers (each thread owns a row x j-slice; the intra-chunk output `Y0 = tril(A) @ dtx` reduces across the lanes sharing a row with `simd_shuffle_down`); per-chunk state contribution `U[d,s] = sum_j exp(lcg_last - lcg_j) dtx_j[d] B_j[s]`.
- Kernel B (parallel over `(batch, head, Dh-block)`): walks chunks sequentially: `y_i = Y0_i + exp(lcg_i) (S @ C_i)`, `S <- exp(lcg_last) S + U_c`. State is threadgroup-resident fp32 with a +1 pad (bank-conflict-free column reads).

All accumulation is fp32; the carried state is fp32; staged tiles and stored intermediates use the input dtype. Long sequences run in `SSD_SEGMENT` slices to bound the `U` buffer.

Dispatch gate: `seq_len >= 32`, `lengths is None`, `head_dim in {32, 64, 128}`, `state_dim` a multiple of 32 in [32, 256], Metal available. Everything else — including decode and short sequences — keeps the existing paths. Loop bounds are template-driven with constexpr-elided guards, so exact-division shapes (all the listed models) compile branch-free.

## Numerics

Max relative error vs an fp32 run of `ssm_attn` is ~5e-3 at every tested length — the same level as bf16 `ssm_attn` itself vs fp32 (~4e-3): the kernels consume the same bf16 inputs and accumulate in fp32, so no precision is given up.

## Mask semantics (and a pre-existing bug)

The kernels support boolean padding masks by treating a masked token as a no-op: its dtA is zeroed (state passes through undecayed) and its dtx is zeroed (no state write, no output contribution). `ssm_attn`'s own masked paths return broken final states:

- mask only (right padding): the last decay row is all `-inf`, so every valid token's state contribution is zeroed — the result also depends on the internal `step=256` chunking;
- mask + lengths: the returned state is decayed through the padding to ~1e-38;
- an initial state is decayed through masked positions using the raw (unmasked) dtA.

Ground truth (running the reference over only the valid tokens) matches the fused kernels at ~3e-3. The test asserts against that ground truth. I left `ssm_attn` untouched here — happy to fix it in this PR or file it separately, whichever you prefer.

## Testing

`tests/test_ssd_prefill.py`: shape matrix ((64,128), (128,128), (64,256), (32,128), (64,64)) x (bfloat16, float16) x (state on/off) at T in {200, 1024}; segment-split (via a patched `SSD_SEGMENT`), carried state across a split pass, left/right padding masks vs valid-slice ground truth, and dispatch through `ssm_update`. All passing on an M3 Ultra; `test_falcon_h1` and `test_nemotron` in `test_models.py` pass unchanged.


zcbenz:

PR #1580, opened 2026-07-18
Add keyed per-request sampling

yxx4c:
Sampling with a `seed` today goes through `mx.random.seed(...)`, i.e. process-global RNG state: the draw a request gets depends on server history, so the same `(prompt, seed)` does not reproduce the same completion across processes or restarts.

This PR adds keyed per-request sampling:

- `make_sampler(seed=...)` builds an explicit PRNG key stream — the key is split once per call and drives keyed `categorical`/XTC draws on an uncompiled path, never touching global RNG state.
- The server threads the request's `seed` into the sampler and drops the global `mx.random.seed` call.

Result: `(prompt, seed)` is fully deterministic across processes, restarts, and machines (same model/quantization), which makes seeded serving usable for reproducible evaluation.

Unseeded requests are unchanged — they keep the compiled samplers and the global stream. Seeded requests already ran on the serial batching path; that is also unchanged. One existing test's fake `args` gains the `seed` attribute `_make_sampler` now reads.

Tests: 6 new tests in `tests/test_keyed_sampler.py` (same-seed identity, cross-seed divergence, unseeded path, stream advance, full filter chain incl. XTC, independence from global RNG state); `tests/test_server.py` and `tests/test_sample_utils.py` green.

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -16,6 +16,7 @@ def make_sampler(
     xtc_probability: float = 0.0,
     xtc_threshold: float = 0.0,
     xtc_special_tokens: List[int] = [],
+    seed: Optional[int] = None,
 ) -> Callable[[mx.array], mx.array]:
     """
     Make a sampler function for use with ``generate_step``.
@@ -37,6 +38,10 @@ def make_sampler(
             for being sampled.
         xtc_special_tokens (list(int), optional): List of special tokens IDs to
             be excluded from XTC sampling.
+        seed (int, optional): If given, the sampler draws from an explicit
+            PRNG key chain derived from this seed (split on every call)
+            instead of the global ``mx.random`` state, so the sampled token
+            stream is fully determined by the seed. Default: ``None``.
 
 
     Returns:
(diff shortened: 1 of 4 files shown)


yxx4c:
Refs #1245.

Flagging the relationship to the other two PRs on this issue (#1331, #1593) so they can be weighed together rather than as duplicates.

All three trace the same root cause: the sampling ops are compiled with `inputs=mx.random.state` at import time, and `mx.random.state` is thread-local, so a sampler compiled on the main thread ignores a `mx.random.seed()` issued on the server's worker thread.

#1331 and #1593 fix *where* that compilation happens so the existing global-seed path takes effect — the seed still means "advance the global, thread-local random stream." This one takes a different mechanism: `make_sampler(seed=...)` returns a sampler that holds an explicit `mx.random.key(seed)` and splits it per call, never touching `mx.random.state`.

The practical difference is what "seeded" guarantees:

- **Concurrency:** two seeded requests in flight share one global stream under the global-state approach — one request's reseed perturbs the other. An explicit per-request key is isolated by construction.
- **Reproducibility across processes/restarts/machines:** because the draw is a pure function of `(seed, logits)`, the same `(prompt, seed)` reproduces the same token stream anywhere (same model + mlx). With global state it holds only within a process and only as long as nothing else advances that thread's stream between the seed and the draw.

Trade-off, stated plainly: this is the larger change and adds a parallel code path, whereas #1593 is a few lines. But it's fully opt-in — with `seed=None` the sampler is byte-identical to today and keeps the compiled path, so the default serving path is untouched (no change to import-time compilation for unseeded requests). The cost is paid only by callers that ask for determinism.

I built this for a reproducible evaluation harness where seeded serving needs to be a reliable primitive across restarts and concurrent load, which is why the key-based route rather than repairing the global path. Happy to rebase, split, or adjust naming/docs if this is a direction worth taking — and equally happy to defer if #1593 covers what you need.

zcbenz:

PR #1581, opened 2026-07-18
Fix empty remainder on exact prompt-cache hit

yxx4c:
An exact prompt-cache trie hit returns an empty remainder, but generation needs at least one prompt token to process: `stream_generate` raises on an empty prompt and `insert_segments` cannot take empty segments — so a request whose prompt exactly matches a cached sequence errors out.

Fix: a small `_fetch_prompt_cache` wrapper around `fetch_nearest_cache` — on an exact hit, trim the last cached token so it is re-processed; if the cache cannot be trimmed, recompute from scratch. Both call sites go through the wrapper.

Tests: 4 new tests in `tests/test_prompt_cache_guard.py` (exact hit leaves one token, prefix hit unchanged, untrimmable cache recomputes, miss unchanged); `tests/test_server.py` and `tests/test_prompt_cache.py` green.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -40,7 +40,12 @@ from .generate import (
     make_text_state_machine,
     stream_generate,
 )
-from .models.cache import LRUPromptCache, make_prompt_cache
+from .models.cache import (
+    LRUPromptCache,
+    can_trim_prompt_cache,
+    make_prompt_cache,
+    trim_prompt_cache,
+)
 from .sample_utils import make_logits_processors, make_sampler
 from .utils import _parse_size, load, sharded_load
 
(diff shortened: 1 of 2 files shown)


yxx4c:
#1596 addresses this same exact-hit empty-remainder case, at a better location. It fixes it inside `LRUPromptCache.fetch_nearest_cache` itself (with the same trim-by-1 → `tokens[-1:]` mechanism, plus a `test_exact_hit_never_returns_empty_remainder`), rather than guarding at the server call sites as this PR does — so every caller benefits, and it additionally lands hybrid/recurrent caches on an interior checkpoint instead of falling back to a full re-prefill.

That's a strict superset of this change, so I'm happy to close this in favor of #1596.

One thing worth your call before I do: the exact-hit crash is a live bug on `main` today for ordinary KV caches, independent of the recurrent/sliding-window feature #1596 is built around. If you'd prefer to land the minimal correctness fix on its own timeline rather than gate it on the larger PR's review, this one is self-contained and I'm glad to keep it open (or move its trim-by-1 into `fetch_nearest_cache` to match #1596's location so they don't collide). Otherwise just say the word and I'll close it.

zcbenz:

PR #1582, opened 2026-07-18
Fix crash on tool call without arguments

yxx4c:
A malformed generation can emit a tool call that parses without an `arguments` key. `ToolCallFormatter._format` indexes it unconditionally, so a single bad generation raises `KeyError` inside the response handler and kills the request.

Fix: `tc.get("arguments", {})` — the call renders as an empty-arguments tool call (the model's actual output, representable) instead of crashing the handler.


zcbenz:

Issue #1573, opened 2026-07-16
RotatingKVCache.to_quantized() raises NotImplementedError — blocks --kv-bits on Gemma 4's sliding-window attention layers


chrislyons:
## Symptom

`mlx_lm/models/cache.py:552` (confirmed on current main, and on the 0.31.3 release):

```python
def to_quantized(self, group_size: int = 64, bits: int = 4) -> QuantizedKVCache:
    raise NotImplementedError("RotatingKVCache Quantization NYI")
```

`mlx_lm.server --kv-bits N` crashes on the **first inference request** with this `NotImplementedError`, even though the server itself starts cleanly and reports healthy. The crash only surfaces once a request actually needs to quantize a rotating-window cache slot.

## Affected models

Gemma 4 (both E4B and the 26B-A4B MoE variant) use sliding-window attention on the majority of transformer layers — 35 of 42 layers in the 26B-A4B configuration. Sliding-window layers are backed by `RotatingKVCache`, which has no quantized implementation. Full-attention layers use the (already-quantizable) standard `KVCache`, so the crash is specific to models with a mixed/hybrid attention schedule like Gemma 4, not to dense-attention models.

## Repro

```
mlx_lm.server --model <gemma-4-26b-a4b-path> --kv-bits 8 --port 41966
# server starts, /health returns 200
# first /v1/chat/completions request → NotImplementedError: RotatingKVCache Quantization NYI
```

## Current workaround (in production use since ~FCT093)

We monkeypatch `mlx_lm.models.cache.make_prompt_cache` (and `mlx_lm.server.make_prompt_cache`) before the server starts, forcing every cache slot — rotating or not — into an 8-bit `QuantizedKVCache` directly, bypassing the broken `RotatingKVCache.to_quantized()` path entirely:

```python
from mlx_lm.models.cache import QuantizedKVCache, RotatingKVCache

def make_prompt_cache_quantized(model, max_kv_size=None, group_size=64, bits=8):
    return [QuantizedKVCache(group_size=group_size, bits=bits) for _ in model.layers]

import mlx_lm.models.cache
mlx_lm.models.cache.make_prompt_cache = make_prompt_cache_quantized
import mlx_lm.server
mlx_lm.server.make_prompt_cache = make_prompt_cache_quantized
```

This works but means `--kv-bits` on the CLI is silently ignored for these models — anyone hitting this for the first time gets a runtime crash instead of a clear "unsupported for this cache type" error at startup.

## Ask

[...]

PR #1584, opened 2026-07-18
Implement KV-cache quantization for the continuous-batching path (RotatingKVCache/BatchRotatingKVCache)

### LINKED ISSUE #1583: KV-cache quantization unreachable in BatchGenerator (RotatingKVCache.to_quantized() unimplemented)
### Description

`maybe_quantize_kv_cache`/`QuantizedKVCache` are complete and tested for the single-sequence `generate_step`/`stream_generate` path, but they're unreachable from the continuous-batching path (`BatchGenerator`) whenever `max_kv_size` is set.

`BatchGenerator._make_new_cache()` always wraps per-job caches in `RotatingKVCache` when `max_kv_size` is set, and on merge into the batch that becomes `BatchRotatingKVCache` — the actual class `update_and_fetch()` runs on during real batched decode steps. Both `RotatingKVCache.to_quantized()` and `BatchRotatingKVCache.to_quantized()` are currently:

```python
raise NotImplementedError("RotatingKVCache Quantization NYI")
```

So any application using `BatchGenerator` with a bounded `max_kv_size` (e.g. to cap per-conversation memory) cannot use KV-cache quantization at all, even though the primitives it would build on are otherwise production-ready.

### Suggested fix

Implement both `to_quantized()` methods, backed by two new classes:
- `RotatingQuantizedKVCache` — per-job quantized rotating cache, mirroring `RotatingKVCache`'s ring-buffer/sink-token mechanics with `(packed, scales, biases)` triples instead of plain arrays.
- `BatchRotatingQuantizedKVCache` — the batch-merged equivalent `BatchGenerator` actually calls `update_and_fetch()` on, mirroring `BatchRotatingKVCache`'s full surface (`filter`/`extend`/`extract`/`merge`/`state`/`meta_state`).

`BatchGenerator` gains `kv_bits`/`kv_group_size`/`quantized_kv_start` constructor params (mirroring `generate()`'s existing signature), quantizing once at cache-creation time (`offset=0`, no precision lost). `quantized_kv_start > 0` is rejected for the batching path since there's no per-step threshold re-check that would ever trigger it — that's a known limitation, not an oversight.

I have a working patch (implemented and tested against a real model, both synthetic-tensor unit tests and an end-to-end `BatchGenerator` run forcing real rotation) and will open a PR shortly.

mabaeyens:
Closes #1583.

### What

`RotatingKVCache.to_quantized()` and `BatchRotatingKVCache.to_quantized()` were both stubs raising `NotImplementedError`. Since `BatchGenerator` always wraps caches in `RotatingKVCache` when `max_kv_size` is set, KV-cache quantization was unreachable end-to-end on the continuous-batching path despite the underlying `mx.quantize`/`QuantizedKVCache` primitives already being production-ready for the non-batched `generate_step`/`stream_generate` path.

Adds `RotatingQuantizedKVCache` (per-job) and `BatchRotatingQuantizedKVCache` (the batch-merged cache `BatchGenerator` actually calls `update_and_fetch()` on), implementing both `to_quantized()` methods. `BatchGenerator` gains `kv_bits`/`kv_group_size`/`quantized_kv_start` constructor params, threaded into `_make_new_cache()` — quantization happens once at cache creation (`offset=0`, no precision lost), so no per-step re-quantization is needed. `quantized_kv_start > 0` is explicitly rejected for the batching path since there's no per-step threshold check that would ever trigger it.

### Why

We use `BatchGenerator` with `max_kv_size` set (to bound per-conversation memory) in a downstream project, and wanted KV-cache quantization for the same RAM/context-window win it already gives the non-batched path. Filed as #1583 for visibility before this PR.

### Testing

- Existing test suite unaffected: the same 8 failures in `tests/test_generate.py` occur with and without this patch, verified against vanilla `upstream/main`, so they are not caused by this change. They are also not flakiness, which is what an earlier version of this description called them. The cause is TF32 in fp32 GEMM on Apple GPU gen-17 (M5): `MLX_ENABLE_TF32=0 python -m pytest tests/test_generate.py` turns 8 failed / 23 passed into 31 passed on this branch, reproducibly. Tracked upstream at ml-explore/mlx#3897.
- New tests cover: rotation+quantization accuracy against a plain `RotatingKVCache`, mid-stream `to_quantized()` conversion, save/load round-trip, and the full `merge`/`filter`/`extend`/`extract` surface `BatchGenerator` relies on for job insertion/eviction — plus an end-to-end `BatchGenerator` test with `max_kv_size`+`kv_bits` set together, forcing real rotation across real attention forward passes on `mlx-community/Qwen1.5-0.5B-Chat-4bit`.
- Also validated live against two full-size production models (Qwen3.6-35B-A3B, Ministral 3 14B) in a downstream project: no regression on a 13-question quality/agentic bench suite, ~1.88x KV-cache compression measured, and a real long-generation rotation test past `max_kv_size` with quantization on.
- Formatted with `black`/`isort` per CONTRIBUTING.md.

### Known limitations

`to_quantized()` raises on `keep > 0`, because sink tokens are not implemented in `RotatingQuantizedKVCache`. `make_prompt_cache(model, max_kv_size=N)` builds `RotatingKVCache(max_size=N, keep=4)` at `cache.py:37`, hardcoded, so for any model that does not provide its own `make_cache`, `max_kv_size` together with KV quantization stays unquantizable after this PR exactly as before it. Gemma 4 escapes only because it passes `keep=0`.

The batched path is not affected: `BatchGenerator._make_new_cache()` builds `RotatingKVCache(max_size=...)` with no `keep` argument and so gets the `keep=0` default. The limitation is a property of `make_prompt_cache`, not of the flag.

Surfaced by @soobrosa's startup probe on #1353. Sink-token support in the quantized rotating cache is the fix and is deliberately out of scope here.

### Checklist

- [x] Tests added
- [x] Formatted (`black`, `isort`)
- [ ] Review

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1731,6 +1731,21 @@ class BatchGenerator:
         for i in range(len(segments)):
             if caches[i] is None:
                 caches[i] = self._make_new_cache()
+            elif self.kv_bits is not None:
+                # Externally supplied caches skip _make_new_cache(), which is
+                # where quantization is normally applied, so kv_bits would
+                # otherwise be silently ignored for them. Quantize in place,
+                # like generate_step does, so the setting is honoured no matter
+                # who built the cache. Unlike the empty cache _make_new_cache()
+                # returns, a supplied cache may already hold tokens, so this
+                # conversion can be lossy. Entries that are already quantized
+                # have no to_quantized() and are skipped.
+                maybe_quantize_kv_cache(
+                    caches[i],
+                    self.quantized_kv_start,
+                    self.kv_group_size,
+                    self.kv_bits,
+                )
 
         for seq, m, c, at, s, lp, sm in zip(
             segments,
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1587,8 +1587,25 @@ class BatchGenerator:
         prefill_batch_size: int = 8,
         prefill_step_size: int = 2048,
         max_kv_size: Optional[int] = None,
+        kv_bits: Optional[int] = None,
+        kv_group_size: int = 64,
+        quantized_kv_start: int = 0,
         stream=None,
     ):
+        if kv_bits is not None and quantized_kv_start != 0:
+            # Validated before any state is set, so a rejected config never
+            # leaves a partially-constructed instance behind (__del__ calls
+            # close(), which needs self._old_wired_limit to already exist).
+            # Per-job caches are created once, empty (offset=0), at insertion
+            # time via _make_new_cache() — maybe_quantize_kv_cache's
+            # offset-threshold gate would never trigger later for a nonzero
+            # quantized_kv_start, since there's no per-step re-check in the
+            # batching path. Only immediate quantization is supported here.
+            raise NotImplementedError(
+                "BatchGenerator only supports quantized_kv_start=0 with kv_bits "
+                "set — delayed/threshold quantization is not implemented for "
+                "the continuous-batching path."
+            )
         self.model = model
         self.max_tokens = max_tokens
         self.sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
@@ -1598,6 +1615,9 @@ class BatchGenerator:
         self.prefill_batch_size = prefill_batch_size
         self.completion_batch_size = max(completion_batch_size, prefill_batch_size)
         self.max_kv_size = max_kv_size
+        self.kv_bits = kv_bits
+        self.kv_group_size = kv_group_size
+        self.quantized_kv_start = quantized_kv_start
 
         self._stream = stream or generation_stream
 
@@ -1631,7 +1651,10 @@ class BatchGenerator:
         return self._stream
 
     def close(self):
-        if self._old_wired_limit is not None:
+        # getattr guards against __del__ firing on an instance whose __init__
+        # raised before this attribute was ever set (e.g. an invalid kv_bits
+        # config) — a bare self._old_wired_limit would AttributeError there.
+        if getattr(self, "_old_wired_limit", None) is not None:
             mx.synchronize(self._stream)
             mx.set_wired_limit(self._old_wired_limit)
             self._old_wired_limit = None
@@ -1732,16 +1755,27 @@ class BatchGenerator:
 
     def _make_new_cache(self):
         if self.max_kv_size is None:
-            return cache.make_prompt_cache(self.model)
+            new_cache = cache.make_prompt_cache(self.model)
+        else:
+            new_cache = [
+                (
+                    RotatingKVCache(max_size=self.max_kv_size)
+                    if isinstance(ci, KVCache)
+                    else ci
+                )
+                for ci in cache.make_prompt_cache(self.model)
+            ]
 
-        return [
-            (
-                RotatingKVCache(max_size=self.max_kv_size)
-                if isinstance(ci, KVCache)
-                else ci
+        if self.kv_bits is not None:
+            # quantized_kv_start is always 0 here (enforced in __init__), so this
+            # quantizes every layer immediately — the cache is empty (offset=0),
+            # so there's no precision to lose. All subsequent update_and_fetch
+            # calls go through the quantized cache classes from this point on;
+            # no further re-quantization is needed anywhere else.
+            maybe_quantize_kv_cache(
+                new_cache, self.quantized_kv_start, self.kv_group_size, self.kv_bits
             )
-            for ci in cache.make_prompt_cache(self.model)
-        ]
+        return new_cache
 
     def _find_uids(self, uids):
         uids = set(uids)
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -291,7 +291,15 @@ def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_
     if kv_bits is None:
         return
     for e, c in enumerate(prompt_cache):
-        if hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
+        if isinstance(c, CacheList):
+            # Hybrid attention/recurrent models (baichuan_m1, falcon_h1,
+            # deepseek_v32, longcat_flash*) hold their KV cache one level down.
+            # The wrapper defines no to_quantized(), so without recursing here
+            # the nested KV leaves silently ignore kv_bits.
+            leaves = list(c.caches)
+            maybe_quantize_kv_cache(leaves, quantized_kv_start, kv_group_size, kv_bits)
+            c.caches = tuple(leaves)
+        elif hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
             prompt_cache[e] = c.to_quantized(group_size=kv_group_size, bits=kv_bits)
 
 
@@ -1746,6 +1754,30 @@ class BatchGenerator:
                     self.kv_group_size,
                     self.kv_bits,
                 )
+
+            if self.kv_bits is not None:
+                # Both lanes, not just supplied caches: _make_new_cache() wraps
+                # only top-level KVCache entries for max_kv_size and does not
+                # descend into CacheList, so a fresh job on a hybrid model can
+                # carry a nested leaf that quantizes to a non-mergeable class
+                # too. Checked here rather than inside maybe_quantize_kv_cache
+                # because that helper is shared with generate_step, where a
+                # merge-less QuantizedKVCache is the normal, correct result.
+                for c in caches[i]:
+                    # CacheList.merge merges leaf-wise, but _merge_caches only
+                    # inspects the wrapper — which does have merge() — so a
+                    # non-mergeable nested leaf would otherwise slip past and
+                    # fail as an AttributeError inside CacheList.merge once the
+                    # batch forms. Check the leaves, not the wrapper.
+                    for leaf in c.caches if isinstance(c, CacheList) else (c,):
+                        if not hasattr(leaf, "merge"):
+                            raise ValueError(
+                                "kv_bits is set but the cache for this job "
+                                f"quantizes to {type(leaf).__name__}, which does "
+                                "not support batching. Batched KV-cache "
+                                "quantization currently requires rotating "
+                                "caches, so set max_kv_size."
+                            )
 
         for seq, m, c, at, s, lp, sm in zip(
             segments,
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1996,7 +1996,10 @@ class BatchRotatingQuantizedKVCache(_BaseCache):
     @meta_state.setter
     def meta_state(self, v):
         self.max_size, self._offset, self._idx = map(int, v[:3])
-        self.rotated = bool(v[3])
+        # meta_state stringifies, so v[3] is "True"/"False" and bool() would
+        # make both truthy. Compare instead. Same slip as BatchRotatingKVCache
+        # (see ml-explore/mlx-lm#1250).
+        self.rotated = v[3] == "True"
         self.group_size, self.bits = map(int, v[4:6])
 
     def is_trimmable(self):
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -229,6 +229,19 @@ class ConcatenateKVCache(_BaseCache):
         return self.keys.nbytes + self.values.nbytes
 
 
+def _empty_quantized(B, n_kv_heads, n_steps, head_dim, group_size, bits, dtype):
+    """Allocate a zero-filled quantized (packed, scales, biases) triple, matching
+    the layout `mx.quantize` produces — used to grow quantized cache buffers in
+    fixed-size chunks (mirrors QuantizedKVCache.update_and_fetch's init_quant)."""
+    el_per_int = 8 * mx.uint32.size // bits
+    shape = (B, n_kv_heads, n_steps)
+    return (
+        mx.zeros((*shape, head_dim // el_per_int), dtype=mx.uint32),
+        mx.zeros((*shape, head_dim // group_size), dtype=dtype),
+        mx.zeros((*shape, head_dim // group_size), dtype=dtype),
+    )
+
+
 class QuantizedKVCache(_BaseCache):
     step = 256
 
@@ -548,8 +561,24 @@ class RotatingKVCache(_BaseCache):
         self._idx -= n
         return n
 
-    def to_quantized(self, group_size: int = 64, bits: int = 4) -> QuantizedKVCache:
-        raise NotImplementedError("RotatingKVCache Quantization NYI")
+    def to_quantized(
+        self, group_size: int = 64, bits: int = 4
+    ) -> "RotatingQuantizedKVCache":
+        if self.keep > 0:
+            raise NotImplementedError(
+                "Quantizing a RotatingKVCache with keep tokens is not supported."
+            )
+        quant_cache = RotatingQuantizedKVCache(
+            self.max_size, keep=self.keep, group_size=group_size, bits=bits
+        )
+        quant_cache.offset = self.offset
+        quant_cache._idx = self._idx
+        if self.keys is not None:
+            quant_cache.keys = mx.quantize(self.keys, group_size=group_size, bits=bits)
+            quant_cache.values = mx.quantize(
+                self.values, group_size=group_size, bits=bits
+            )
+        return quant_cache
 
     def make_mask(
         self, N: int, window_size: Optional[int] = None, return_array: bool = False
@@ -589,6 +618,234 @@ class RotatingKVCache(_BaseCache):
         if self.keys is None:
             return 0
         return self.keys.nbytes + self.values.nbytes
+
+
+class RotatingQuantizedKVCache(_BaseCache):
+    """Quantized counterpart of RotatingKVCache. `keys`/`values` are each a
+    (packed, scales, biases) triple as produced by `mx.quantize`, rather than a
+    single float array. Does not support `keep` (sink) tokens — batching
+    (BatchRotatingKVCache) never supports them either, and mira-mlx never
+    configures them, so this restriction costs nothing in practice."""
+
+    step = 256
+
+    def __init__(self, max_size, keep=0, group_size: int = 64, bits: int = 4):
+        if keep > 0:
+            raise NotImplementedError(
+                "RotatingQuantizedKVCache does not support keep tokens."
+            )
+        self.keep = keep
+        self.keys = None
+        self.values = None
+        self.offset = 0
+        self.max_size = max_size
+        self._idx = 0
+        self.group_size = group_size
+        self.bits = bits
+
+    def _quantize(self, x):
+        return mx.quantize(x, group_size=self.group_size, bits=self.bits)
+
+    def _trim(self, trim_size, v, append=None):
+        if trim_size > 0:
+            v = tree_map(lambda a: a[..., trim_size:, :], v)
+        if append is not None:
+            v = tree_map(lambda a, b: mx.concatenate([a, b], axis=2), v, append)
+        return v
+
+    def _temporal_order(self, v):
+        """
+        Rearrange the cache into temporal order, slicing off the end if unused.
+        """
+        if self._idx == v[0].shape[2]:
+            return v
+        elif self._idx < self.offset:
+            return tree_map(
+                lambda a: mx.concatenate(
+                    [a[..., self._idx :, :], a[..., : self._idx, :]], axis=2
+                ),
+                v,
+            )
+        else:
+            return tree_map(lambda a: a[..., : self._idx, :], v)
+
+    def _update_concat(self, keys, values):
+        qkeys = self._quantize(keys)
+        qvalues = self._quantize(values)
+        if self.keys is None:
+            self.keys = qkeys
+            self.values = qvalues
+        else:
+            # Put the keys/values in temporal order to preserve context
+            self.keys = self._temporal_order(self.keys)
+            self.values = self._temporal_order(self.values)
+            self._idx = self.keys[0].shape[2]
+
+            # The largest size is self.max_size + S - 1 to ensure
+            # every token gets at least self.max_size context
+            trim_size = self._idx - self.max_size + 1
+            self.keys = self._trim(trim_size, self.keys, qkeys)
+            self.values = self._trim(trim_size, self.values, qvalues)
+        self.offset += keys.shape[2]
+        self._idx = self.keys[0].shape[2]
+        return self.keys, self.values
+
+    def _update_in_place(self, keys, values):
+        # May not have hit the max size yet, so potentially keep growing the cache
+        B, n_kv_heads, S, k_head_dim = keys.shape
+        v_head_dim = values.shape[3]
+        prev = self.offset
+        cur_size = self.keys[0].shape[2] if self.keys is not None else 0
+        if self.keys is None or (prev >= cur_size and cur_size < self.max_size):
+            new_size = min(self.step, self.max_size - prev)
+            new_k = _empty_quantized(
+                B,
+                n_kv_heads,
+                new_size,
+                k_head_dim,
+                self.group_size,
+                self.bits,
+                keys.dtype,
+            )
+            new_v = _empty_quantized(
+                B,
+                n_kv_heads,
+                new_size,
+                v_head_dim,
+                self.group_size,
+                self.bits,
+                values.dtype,
+            )
+            if self.keys is not None:
+                self.keys = tree_map(
+                    lambda a, b: mx.concatenate([a, b], axis=2), self.keys, new_k
+                )
+                self.values = tree_map(
+                    lambda a, b: mx.concatenate([a, b], axis=2), self.values, new_v
+                )
+            else:
+                self.keys, self.values = new_k, new_v
+            self._idx = prev
+
+        # Trim if needed
+        trim_size = self.keys[0].shape[2] - self.max_size
+        if trim_size > 0:
+            self.keys = self._trim(trim_size, self.keys)
+            self.values = self._trim(trim_size, self.values)
+            self._idx = self.max_size
+
+        # Rotate
+        if self._idx == self.max_size:
+            self._idx = self.keep  # always 0, keep>0 rejected in __init__
+
+        # Assign (quantize the incoming chunk, then slice-write like QuantizedKVCache)
+        qkeys = self._quantize(keys)
+        qvalues = self._quantize(values)
+        for i in range(3):
+            self.keys[i][..., self._idx : self._idx + S, :] = qkeys[i]
+            self.values[i][..., self._idx : self._idx + S, :] = qvalues[i]
+        self.offset += S
+        self._idx += S
+
+        # If the buffer is not full, slice off the end
+        if self.offset < self.max_size:
+            return (
+                tree_map(lambda a: a[..., : self.offset, :], self.keys),
+                tree_map(lambda a: a[..., : self.offset, :], self.values),
+            )
+        return self.keys, self.values
+
+    def update_and_fetch(self, keys, values):
+        if keys.shape[2] == 1:
+            return self._update_in_place(keys, values)
+        return self._update_concat(keys, values)
+
+    def size(self):
+        return min(self.offset, self.max_size)
+
+    @property
+    def state(self):
+        if self.offset < self.keys[0].shape[2]:
+            return (
+                tree_map(lambda a: a[..., : self.offset, :], self.keys),
+                tree_map(lambda a: a[..., : self.offset, :], self.values),
+            )
+        else:
+            return self.keys, self.values
+
+    @state.setter
+    def state(self, v):
+        self.keys, self.values = v
+
+    @property
+    def meta_state(self):
+        return tuple(
+            map(
+                str,
+                (
+                    self.keep,
+                    self.max_size,
+                    self.offset,
+                    self._idx,
+                    self.group_size,
+                    self.bits,
+                ),
+            )
+        )
+
+    @meta_state.setter
+    def meta_state(self, v):
+        self.keep, self.max_size, self.offset, self._idx, self.group_size, self.bits = (
+            map(int, v)
+        )
+
+    def is_trimmable(self):
+        return self.offset < self.max_size
+
+    def trim(self, n):
+        n = min(self.offset, n)
+        self.offset -= n
+        self._idx -= n
+        return n
+
+    def make_mask(
+        self, N: int, window_size: Optional[int] = None, return_array: bool = False
+    ):
+        # Identical bookkeeping to RotatingKVCache.make_mask — no array access.
+        if N > 1:
+            window_size = window_size or self.max_size
+            offset = min(self.max_size - 1, self.offset)
+            if offset + N > window_size or return_array:
+                return create_causal_mask(N, offset, window_size=window_size)
+            else:
+                return "causal"
+        else:
+            if window_size is None:
+                return None
+            if self.offset >= window_size and self.max_size > window_size:
+                idx = self._idx
+                if idx >= self.max_size:
+                    idx = 0
+                if self.offset < self.max_size:
+                    mask_size = self.offset + 1
+                else:
+                    mask_size = self.max_size
+                mask = mx.arange(mask_size) >= (mask_size - window_size)
+                mask = mx.roll(mask, shift=idx + 1)
+                return mask
+
+    @classmethod
+    def merge(_, caches):
+        return BatchRotatingQuantizedKVCache.merge(caches)
+
+    def empty(self):
+        return self.keys is None
+
+    @property
+    def nbytes(self):
+        if self.keys is None:
+            return 0
+        return tree_reduce(lambda a, x: a + x.nbytes, (self.keys, self.values), 0)
 
 
 class ArraysCache(_BaseCache):
@@ -1324,8 +1581,22 @@ class BatchRotatingKVCache(_BaseCache):
         self.offset -= n
         return n
 
-    def to_quantized(self, group_size: int = 64, bits: int = 4) -> QuantizedKVCache:
-        raise NotImplementedError("BatchRotatingKVCache Quantization NYI")
+    def to_quantized(
+        self, group_size: int = 64, bits: int = 4
+    ) -> "BatchRotatingQuantizedKVCache":
+        quant_cache = BatchRotatingQuantizedKVCache(
+            self.max_size, self.left_padding.tolist(), group_size=group_size, bits=bits
+        )
+        quant_cache.offset = self.offset
+        quant_cache._idx = self._idx
+        quant_cache._offset = self._offset
+        quant_cache.rotated = self.rotated
+        if self.keys is not None:
+            quant_cache.keys = mx.quantize(self.keys, group_size=group_size, bits=bits)
+            quant_cache.values = mx.quantize(
+                self.values, group_size=group_size, bits=bits
+            )
+        return quant_cache
 
     def make_mask(
         self, N: int, window_size: Optional[int] = None, return_array: bool = False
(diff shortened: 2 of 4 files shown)


mabaeyens:
Bumping this, but with a reason rather than just a bump.

Since I opened it, `a790972` landed on main: the broadcast fix in `quantized_scaled_dot_product_attention` for GQA with a batched padding mask. That sits directly on the path this PR opens up. Quantizing the KV cache in the continuous-batching path is what puts GQA, batch >= 2 and a quantized cache together in the first place, so before that fix this PR would have walked straight into the crash it repairs. I think the two are complementary, and that `a790972` is a reason to look at this now rather than a reason to wait.

Status: rebased on current main, no conflicts. Nothing upstream has touched `cache.py`. The only adjacency is `2c008fd` adding `return_logprobs` and `return_token_ids` to `BatchGenerator.__init__`, where this adds `kv_bits`, `kv_group_size` and `quantized_kv_start`. All keyword-only and merges cleanly.

One thing worth flagging before anyone runs the suite. `tests/test_generate.py` has 8 failures on this branch, but the same 8 fail on stock `upstream/main` for me (checked at `15b522f`, on an M5 with mlx 0.32.0): `test_batch_matches_single`, `test_many_batches`, `test_batch_sliding_window`, the three `test_batch_continued_generation*`, and the two `test_stream_generate_input_embeddings*`. So they look pre-existing rather than introduced here, but I can't tell from my machine alone whether they're environment-specific or real on main. Worth a look independently of this PR if they aren't already known.

For whatever it's worth as a signal, I've been running this in production for a few months on Qwen3.6-35B-A3B. The KV quantization is what makes the larger context windows fit on a 32GB machine, so it's had a lot more hours on it than a PR this size usually gets before review.

Happy to split it, rework it, or answer questions. Equally happy to leave it sitting if the batching path isn't a priority right now. Mostly I wanted the `a790972` connection on the record, since it isn't obvious from the diff.

katlun-lgtm:
Nice — this is exactly the batched-path piece I flagged as the follow-up over in #1476 (server KV-quant, #1043). Mine only exposes `--kv-bits` on the server and side-steps batching: since `QuantizedKVCache` had no `merge()`, quantized requests just fall back to the sequential path. Your `BatchRotatingQuantizedKVCache` gives quantized caches the full merge/filter surface, so that workaround isn't needed anymore — server requests can stay batched.

One gap I noticed: this threads `kv_bits` into `BatchGenerator` but doesn't touch `server.py`, so the server flags from #1043 are still unexposed. Happy to rebase #1476 on top of this once it lands — drop the `_is_batchable` opt-out and pass `kv_bits`/`kv_group_size` straight into `BatchGenerator` (and route `--quantized-kv-start > 0`, which you reject on the batch path, to the sequential fallback). That'd close #1043 the right way instead of the workaround. Let me know if that split makes sense to you.

mabaeyens:
Yes, that split makes sense to me, and your read of the PR is exactly right.

This PR is deliberately just the primitive: `BatchRotatingQuantizedKVCache` with the full merge/filter surface, so a quantized request no longer has to drop to the sequential path. It stops at `BatchGenerator` and does not touch `server.py` on purpose, so the #1043 server flags are still unexposed. Rebasing #1476 on top once this lands, dropping the `_is_batchable` opt-out and passing `kv_bits`/`kv_group_size` straight through, is the right way to close it. The workaround was only ever there because the merge surface did not exist yet.

On `quantized_kv_start`: routing `> 0` to the sequential fallback is the correct handling, and yes, the batch path rejects it. It is a hard `NotImplementedError` rather than a silent no-op, and the reason is structural rather than missing work. Per-job caches are created empty (offset 0) at insertion via `_make_new_cache()`, and there is no per-step re-check in the batching loop, so `maybe_quantize_kv_cache`'s offset-threshold gate would never fire for a nonzero start. Only immediate quantization is well-defined here, so the guard refuses the config up front instead of quietly ignoring it. Delayed/threshold quantization on the batch path would need a per-step re-check that this PR does not add, so sending `--quantized-kv-start > 0` to the sequential path is exactly where it belongs.

So the division I would suggest: this PR is the batched quantized cache, #1476 rebased on top is the server exposure plus the fallback routing. They stay cleanly separable, which also keeps each reviewable on its own. Happy to coordinate on timing, and glad this unblocks the path you flagged.

katlun-lgtm:
Sounds good — agreed on that division. I'll hold #1476 until this lands, then rebase it down to the server flags plus routing `quantized-kv-start > 0` to the sequential path. Thanks for laying out why immediate quantization is the only well-defined mode on the batch path — cleaner rationale than the one I had.

mabaeyens:
Pushed a numerical-equivalence test for the rotating-quantized path. Coverage there was smoke-level so far (non-NaN and shapes), which would not catch a silent regression in the packed/scales/biases bookkeeping. This asserts the property that actually matters: after forcing rotation, a quantized rotating cache tracks the unquantized rotating reference within rtol=4e-2, the same tolerance QuantizedKVCache already gets in test_cache_to_quantized.

The setup leans on determinism rather than a golden file: two identical argmax warmups reach the same KV state, one continuation stays unquantized and the other is quantized, then both are fed the same token and the logits compared. Passes locally alongside the rest of the quantized and rotating suite.

Branch is current with base too.

katlun-lgtm:
On the 8 `test_generate.py` failures you flagged — they don't reproduce on M3 Max, so they look machine-specific rather than real on main.

I ran all eight by name against stock `upstream/main`, mlx 0.32.0, on an M3 Max:

- at `cf10f96` (current main): 8/8 pass
- at `15b522f` (your base when you posted): 8/8 pass

```
test_batch_matches_single ... ok
test_many_batches ... ok
test_batch_sliding_window ... ok
test_batch_continued_generation ... ok
test_batch_continued_generation_gated_delta ... ok
test_batch_continued_generation_ssm ... ok
test_stream_generate_input_embeddings ... ok
test_stream_generate_input_embeddings_prefill ... ok
Ran 8 tests / OK
```

Same commit, same mlx version, different silicon → passes. So it isn't a fix that landed between your base and current main (it passes at your base too), and it isn't something a reviewer on another Apple gen would hit. My guess is the M5 `is_nax` attention path — six of the eight are batch-shape tests, and that path only instantiates on M5, so M3 Max never exercises it (same lane as the L-cliff in #3826/#3838). Can't confirm the mechanism without an M5, but the "environment-specific vs real on main" question resolves to environment-specific from here.

Worth noting since it clears them as a blocker for reviewing this PR — they're not introduced by your change and not reproducible off-M5.

mabaeyens:
Confirmed from the other side of the bench: I ran it on an actual M5, and the split is silicon, not the mlx version.

The same eight fail on M5 at both mlx 0.31.2 (my stack) and mlx 0.32.0 (yours), 20 of 28 pass either way. The version moves nothing and only the silicon does, so this is the is_nax fused-attention path you called, not a 0.31.x bug. The "can't confirm without an M5" gap is closed: it is the M5.

They are also benign. All eight are strict batch-vs-single equivalence assertions (allclose at rtol 1e-5), and the real gap is tiny: max absolute logprob difference is 0.03125 across the prompts (one hits 0.039), with the argmax identical on every one. So the fused kernel differs by about 1/32 of a logprob and never changes the token that gets picked. That is a low-precision accumulation in the M5 attention path tripping a bit-equivalence test, not a wrong result, and the clean 1/32 smells like an accumulation-width thing rather than noise.

And #1584 itself holds on current mlx. On the same M5 / 0.32.0 run, all of the PR's own tests pass: test_batch_sliding_window_quantized, the merge/filter/extend/extract surface, the rotating-quantized equivalence check, and save/load. So the feature is validated on the latest mlx, on the one silicon where the strict suite trips, and per your run on M3 Max where it does not.

Net: the eight are pre-existing, silicon-specific, and benign, so they clear as a blocker for this PR. Happy to open a separate mlx-core issue about the is_nax path tripping those equivalence tests if that is useful, but it is orthogonal to the KV-quant work here.

mabaeyens:
Chased the mechanism a bit further before writing it up as an mlx issue, and closing it needs your M3 Max.

On the M5, plain batched SDPA and quantized matmul are bit-identical batch-vs-single (maxdiff exactly 0). The gap only shows up on the padded/masked path that batched generation actually uses, and there it is 2^-11 in fp16 and 2^-8 in bf16. Clean powers of two, bf16 is 8x fp16, so it reads like a lower-precision accumulation over the masked reduction rather than noise. Accumulated across the layers, that lands at the ~1/32 logprob we saw.

What I cannot tell from one machine is whether that masked-path gap is actually larger on M5 than on your M3 Max, which is the whole silicon-specific claim. Would you run this on the M3 Max? If your maxdiff comes back much smaller than mine, it pins the divergence to the M5 attention kernel and the issue writes itself.

```python
import mlx.core as mx
def maxdiff(a, b): return float(mx.max(mx.abs(a.astype(mx.float32) - b.astype(mx.float32))))
dt = mx.float16; H, D = 16, 64; Lreal, pad = 40, 8; Lpad = Lreal + pad; s = 1.0 / D**0.5
q1 = mx.random.normal((1, H, Lreal, D)).astype(dt)
k1 = mx.random.normal((1, H, Lreal, D)).astype(dt)
v1 = mx.random.normal((1, H, Lreal, D)).astype(dt)
single = mx.fast.scaled_dot_product_attention(q1, k1, v1, scale=s, mask="causal")
lp = lambda a: mx.concatenate([mx.zeros((1, H, pad, D), dtype=a.dtype), a], axis=2)
ri = mx.arange(Lpad); m = (ri[None] <= ri[:, None]) & (ri >= pad)[None]
addmask = mx.where(m, mx.array(0.0, dt), mx.array(-mx.inf, dt))[None, None]
batched = mx.fast.scaled_dot_product_attention(lp(q1), lp(k1), lp(v1), scale=s, mask=addmask)
mx.eval(single, batched)
print("fp16", maxdiff(single, batched[:, :, pad:, :]))
```

On my M5 that prints 4.8828e-04 (fp16); swap dt to bf16 and it is 3.9062e-03. Curious what you get.

katlun-lgtm:
Ran it on the M3 Max (mlx 0.32.0), swept 8 seeds since it's a single draw:

- fp16: max **2.44e-4** (2^-12) — half your M5 2^-11
- bf16: max **7.63e-6** (2^-17) — vs your 3.91e-3 (2^-8), so ~512x smaller

So bf16 is the tell: the masked-path batch-vs-single gap is nine bits smaller here, near bit-identical, while M5 sits at 2^-8. fp16's the same direction but only one bit. That's the silicon split you were after — it points at the M5 fused-attention kernel rather than the masked reduction in general. The bf16-8x-fp16 you see on M5 is basically gone on M3 Max, which reads like the M5 kernel accumulating the masked reduction at lower width. Happy to run other shapes if it helps the writeup.

mabaeyens:
Thanks for running it, that bf16 number is the whole ballgame: 2^-17 on your M3 Max against 2^-8 on the M5 is a nine-bit split, exactly the silicon-specific accumulation the model-level 1/32 needed. So it is the M5 fused-attention kernel, not a general masked-softmax effect.

Written up with your numbers in the table: ml-explore/mlx#3897. I credited the M3 Max run as your data point. Feel free to add anything I got wrong, or any other shapes you would want in there before a maintainer picks it up.

katlun-lgtm:
Quick correction to my numbers above: I swept the masked snippet over 32 seeds on the M3 Max and the bf16 2⁻¹⁷ turned out to be a small-sample fluke — the median's actually 0 (most seeds bit-identical) and the max walks up to 2⁻¹¹, while the fp16 max hits the same value as the M5 column. The gap also grows with head dim (D=128 bf16 → 2⁻⁹). Full table and the ask are in ml-explore/mlx#3897 (comment). Mechanism holds; the cross-silicon ratio just needs a matched seed sweep on both machines.

mabaeyens:
Followed up on the eight batch-equivalence failures over in ml-explore/mlx#3897, since katlun asked for a matched seed sweep on the M5 before it goes to a maintainer. Short version for anyone tracking this PR: the split holds. On a 32-seed sweep the M5 stays about 2 to 3 bits above the M3 Max at every matched head dim, widest in bf16, and an fp32 control diverges at ~3e-4 on the M5 where the M3 Max sits at the 2e-7 floor. Being fp32, that points at the M5 fused-attention kernel rather than storage precision.

Nothing here changes the read for this PR. The eight are pre-existing, silicon-specific, and benign (argmax identical, about 1/32 of a logprob), so they don't gate the KV-quant work. The one actionable follow-up is orthogonal to it: rtol=1e-5 batch-equivalence won't survive real head dims on any silicon (the M3 Max max already walks to 2⁻⁹ at D=128 bf16), so an atol floor or a per-arch tolerance is worth doing in mlx-lm on its own.

mabaeyens:
Cross-linking: this also covers #1573, which hits the same two `NotImplementedError`s from a different direction. That report comes from Gemma 4's hybrid sliding-window attention with `--kv-bits 8` rather than from the batching path, but the blocking call is the same `RotatingKVCache.to_quantized()`.

One gap for review scope. My tests exercise the batching path with a bounded `max_kv_size`, and they do not cover a mixed cache where only some layers are rotating, which is the Gemma 4 layout. I have asked on #1573 for a run against that model to confirm the hybrid case needs nothing extra.

Separately, on the M5 test failures I mentioned earlier in this PR: those are now explained and are not caused by this change. Root cause is TF32 accumulation on gen-17 Apple silicon (details in ml-explore/mlx#3897). Setting `MLX_ENABLE_TF32=0` takes `tests/test_generate.py` from 8 failed / 20 passed to 28 passed on my machine, with this branch in place. So that blocker is cleared.

PhilipJohnBasile:
The capability guard discussed on #1573 is now open as draft #1618, stacked on this PR at `d98dbe2`. It does not ask for changes to #1584; I will rebase it onto `main` after this lands.

One additional silent-skip path surfaced while implementing it: externally supplied caches to `BatchGenerator.insert(..., caches=...)` bypassed `_make_new_cache()` and therefore bypassed KV quantization entirely. #1618 validates and converts those too, alongside the promised `keep>0` true-negative and mixed-cache checks.

For clarity, this is being classified as a correctness/configuration-validation issue, not a cybersecurity vulnerability: no trust boundary, data exposure, untrusted-code execution, or auth behavior is involved.

mabaeyens:
Confirmed the `insert(..., caches=...)` gap you found, and it is mine, not something for #1618 to paper over.

`insert_segments` only builds a cache when the caller did not supply one:

```python
caches = caches or [None] * len(segments)
for i, ...:
    if caches[i] is None:
        caches[i] = self._make_new_cache()
```

and `_make_new_cache()` is where this PR applies quantization (`if self.kv_bits is not None:`, `generate.py:1769`). So a caller-supplied cache skips it entirely and you get an unquantized cache while having asked for `kv_bits`, with no error. Exactly the silent-skip shape this PR is supposed to remove, sitting in the path this PR added.

My read on the split: converting those caches is quantization behaviour and belongs here, while refusing the ones that cannot be converted is validation and belongs in #1618. Otherwise the correctness of this PR depends on a different PR landing. That is not a dependency I want to ship.

So I will add the conversion to `insert_segments` here, honouring `kv_bits` for externally supplied caches the same way `_make_new_cache()` does, plus a test that inserts a caller-built cache and asserts the layer types afterwards.

One consequence worth naming, since it costs you something. This changes behaviour on a branch you already validated at `d98dbe2`, so your Gemma 4 hybrid and rotation results no longer strictly cover the head. Single-sequence Gemma 4 never goes through `insert_segments`, so I do not expect anything to move. But I would rather say that out loud than let the earlier run quietly stand in for the tip. I will post the new hash when it is up and re-run the batching tests here. No obligation to redo yours.

If a maintainer would rather keep this PR frozen at its validated state and take the conversion as a follow-up, I am fine with that too, as long as it is not left to #1618, which is scoped to validation.

mabaeyens:
Pushed as `3ebafab`, on top of `d98dbe2`.

`insert_segments` now quantizes caller-supplied caches through the same `maybe_quantize_kv_cache()` call `_make_new_cache()` and `generate_step()` already use, rather than a separate code path. Already-quantized entries define no `to_quantized()`, so the existing `hasattr` guard skips them and nothing is quantized twice. `RotatingKVCache(keep>0)` now raises at `insert()` rather than part-way through decoding.

One honest note on the conversion: unlike the empty cache `_make_new_cache()` returns, a supplied cache usually already holds tokens, so quantizing it can lose precision on what is already stored. That is the same thing `generate_step` has always done, so it is not new behaviour, but the "no precision to lose" reasoning in `_make_new_cache`'s comment does not carry over and I did not copy it.

Five tests in `tests/test_generate.py`. Two of them go red without the change (`..._is_quantized` and `..._with_keep_raises`); the other three are guards that pass either way by design, covering the no-`kv_bits` control, the already-quantized no-op, and a full generate/harvest/re-insert round trip. Worth being precise about which is which rather than claiming five red-green tests.

Suite on an M5 with `MLX_ENABLE_TF32=0` (see ml-explore/mlx#3897 for why that pin is needed on gen-17): `test_generate.py` 33 passed, `test_prompt_cache.py` 26 passed. black clean.

@PhilipJohnBasile, two things for you.

Your Gemma 4 validation was against `d98dbe2`, so it no longer covers the tip. Single-sequence Gemma 4 never enters `insert_segments`, so I do not expect anything to move, and I would rather say that than let the earlier run quietly stand in for a commit it did not test. No obligation to redo it.

The round-trip test needs `max_kv_size`, and that is not incidental. Continued generation merges the per-job caches, and only `RotatingQuantizedKVCache` implements `merge()`; plain `QuantizedKVCache` does not, so `kv_bits` without `max_kv_size` cannot do batching-with-history at all. I checked whether my change caused that by reverting only `generate.py` and re-running: the failure is identical either way, so it predates this PR and is not something the change introduced. Flagging it because it bounds what this PR delivers, not as something I am proposing to fix here.

mabaeyens:
@angeloskath a scoping question rather than a review request. This is a big diff for this repo and I would rather make it cheaper to look at than leave it sitting.

It is currently +1132 across 4 files. It splits in two:

1. **Single-sequence**: `RotatingQuantizedKVCache` and `RotatingKVCache.to_quantized()`, about 390 lines over 2 files. `merge()` would raise for now, the same way `BatchRotatingKVCache Quantization NYI` already does in that file. This is the half `mlx_lm.generate --max-kv-size N --kv-bits 8` needs, and it is also the half @PhilipJohnBasile validated end to end on Gemma 4 26B (hybrid layout, rotation exercised across a prompt 3.5x the window), single-sequence only.

2. **Batching**: `BatchRotatingQuantizedKVCache`, the `BatchGenerator` wiring, and flipping `merge()` from raise to delegate.

Either shape works for me. If one PR is fine then ignore this and I will leave it as is. If 390 lines is easier to review than 1132, say so and I will split it.

PhilipJohnBasile:
Revalidated the current tip `ff1e837` in an isolated worktree after the two new commits:

- `MLX_ENABLE_TF32=0 pytest -q tests/test_generate.py tests/test_prompt_cache.py`: **60 passed**;
- `git diff --check upstream/main...HEAD`: clean;
- compileall over `mlx_lm` and both test modules: clean;
- Black and isort hooks on all four changed files: pass.

This covers the new externally supplied `BatchGenerator` cache path and the serialized `"False"` rotation-state regression at unit level. I did not repeat the model-heavy Gemma 4 run while unrelated training/evaluation workloads were active, so the earlier end-to-end Gemma evidence remains specifically against `d98dbe2`, not a new model run on `ff1e837`.

Per the sequencing guidance, #1618 remains stacked and unchanged until this lands; I will rebase it afterward rather than create review churn now.

Security classification remains scoped to correctness/configuration behavior. Current evidence does not show data exposure, untrusted-code execution, authentication/authorization impact, or another cybersecurity boundary failure.

mabaeyens:
A limitation of this PR that is not in the description, and should be. It came out of @soobrosa's startup probe on #1353 rather than from my own testing, which is the honest version of how it was found.

`to_quantized()` here raises on `keep > 0`, since sink tokens are not implemented in `RotatingQuantizedKVCache`. And `make_prompt_cache(model, max_kv_size=N)` builds `RotatingKVCache(max_size=N, keep=4)` at `cache.py:37`, hardcoded. So for any model that does not provide its own `make_cache`, asking for `--max-kv-size` together with KV quantization is unquantizable after this PR exactly as it was before it. Gemma 4 only escapes because it passes `keep=0`.

That matters more than it first looks, because `max_kv_size` is the documented route to a `RotatingKVCache` in the first place. So the flag that gets you the cache type this PR adds quantization for is also the flag that stops it working.

What the PR does close is the case where the model's own `make_cache` builds rotating caches with `keep=0`. On Gemma 4 26B-A4B that is 25 of 30 layers, which is the crash in #1573 and the silent skip #1353 originally worked around. I am not claiming more than that.

One asymmetry worth recording while it is in front of me: `BatchGenerator._make_new_cache()` builds `RotatingKVCache(max_size=self.max_kv_size)` with no `keep` argument, so it gets the `keep=0` default and the same combination quantizes fine on the batched path. The limitation is a property of `make_prompt_cache`, not of the flag.

Sink-token support in the quantized rotating cache is the actual fix and I am not attempting it here; it is a separate design question (where the sink tokens live when the rest of the cache is packed) and this PR is large enough. Happy to file it as an issue so it does not get lost, or to take it as a follow-up if a maintainer would rather see it landed together.

I will fold this into the PR description as a known limitation.

mabaeyens:
Separate correction, flagging it here because description edits do not notify anyone.

The Testing section used to write off 8 failures in `tests/test_generate.py` as "environment/model-flakiness". That was wrong and I have fixed it. They are deterministic, and the cause is TF32 in fp32 GEMM on Apple GPU gen-17:

```
$ python -m pytest tests/test_generate.py -q
8 failed, 23 passed in 4.83s

$ MLX_ENABLE_TF32=0 python -m pytest tests/test_generate.py -q
31 passed in 6.98s
```

Six of the eight are the batch-versus-single comparisons, which is why it looked like it might be mine. It is not: they fail identically on vanilla `upstream/main`, and turning TF32 off fixes them there too. Tracked at ml-explore/mlx#3897, which also covers a second and unrelated fp16/bf16 mechanism in the attention kernel.

Anyone reviewing this on an M5 will hit the same 8. Run with `MLX_ENABLE_TF32=0` and the suite is clean.

soobrosa:
Filed the sink-token follow-up as #1631, referencing this PR and #1353 as promised.

It ended up broader than "implement `keep>0`", because while writing it up I found that `cache.py:37` is the only `keep > 0` anywhere in the repo. Every model that builds rotating caches itself passes `keep=0` (`cohere2`, `exaone4`, `exaone_moe`, `gemma3n`, `gemma4_text`), and the only other code that inspects `keep` treats nonzero as unsupported (`generate.py:842`, in `_make_cache`, currently unreferenced). So the generic `--max-kv-size` fallback is the single producer of a shape that two other subsystems refuse.

That means your `keep > 0` raise may not need an implementation at all: if a maintainer decides the `keep=4` default is superseded rather than deliberate, a one-line change to `cache.py:37` removes the case instead. @awni added it in #1015, so I asked him directly on the issue rather than guessing.

Nothing for you to do here, and nothing that changes #1584's scope. Flagging it because "sink tokens are the fix for the sequential half" may turn out to be optional.

pierre427:
One case worth covering while this is in flight: `BatchGenerator.insert` / `insert_segments` accept caller-supplied `caches=` and only route `None` entries through `_make_new_cache()` — so once `kv_bits` reaches the batching path, an externally supplied cache (snapshot restore, transplant) silently runs unquantized, and the mixed fp/quantized cohort then fails in the class-specific merges.

Two things that worked for us when we ported this PR's classes into our serving fork:
1. Route supplied caches through `maybe_quantize_kv_cache` at insert so they match the fresh-lane cohort.
2. Recurse into `CacheList` — the wrapper has `merge` but no `to_quantized`, so a leaf-level pass is needed (quantize KV leaves, verify every KV leaf still supports `merge` afterwards, and fail loudly for quantized classes that don't rather than deferring the crash to `_merge_caches`).

Happy to share the tests we used (supplied-cache quantization on the rotating config; nested non-mergeable case raising at insert).

mabaeyens:
Yes please to the tests, and thanks for the CacheList catch.

Your first point is already in the tip. @PhilipJohnBasile found the same `insert(..., caches=...)` gap in https://github.com/ml-explore/mlx-lm/pull/1584#issuecomment-5081990991 and it landed in `3ebafab`, so supplied caches already go through `maybe_quantize_kv_cache` at insert exactly as you describe. You were reading an older state of the branch.

The CacheList point holds and I checked it against the tip. A recursive leaf pass on a supplied `CacheList(ArraysCache, RotatingKVCache)` merges fine and lands on this PR's own `BatchRotatingQuantizedKVCache`:

```
rotating leaves -> ['ArraysCache', 'RotatingQuantizedKVCache']
CacheList.merge -> ['ArraysCache', 'BatchRotatingQuantizedKVCache']   OK
```

The mixed cohort is real, and the insert-only fix is what creates it. Supplied caches get quantized while `_make_new_cache()` still skips CacheList, so the two lanes disagree at the same layer position:

```
supplied ['ArraysCache', 'RotatingQuantizedKVCache'] + fresh ['ArraysCache', 'RotatingKVCache']
CacheList.merge -> AttributeError: 'RotatingKVCache' object has no attribute 'group_size'
```

So the recursion cannot sit at insert only. It belongs in `maybe_quantize_kv_cache` itself, so both lanes descend into CacheList and the cohort stays uniform.

Your "verify every KV leaf still supports merge afterwards, and fail loudly" is doing real work rather than covering an edge case. With plain leaves a recursive pass gives `QuantizedKVCache`, which has no `merge()`, so `CacheList.merge` raises. `baichuan_m1` makes that concrete: it builds `RotatingKVCache` on sliding-window layers and plain `KVCache` on the rest, so a blanket pass leaves you mergeable on some layers and not on others within one model.

On where it lands: @PhilipJohnBasile and I agreed earlier on this PR that conversion behaviour stays here and validation goes to their stacked #1618. By that line the recursion is mine, since it is quantization behaviour and the mixed cohort above means this PR is not correct without it. The loud refusal of quantized leaves that cannot merge is validation and fits #1618. I would rather not have this PR's correctness depend on a different PR landing.

Worth naming that `maybe_quantize_kv_cache` is unchanged from main, so `mlx_lm.generate --kv-bits 8` on `deepseek_v32` silently ignores the flag today. But `kv_bits` never reached the batching path before this PR, so the batch-path exposure is new, and that is the same standard I used when I took the `insert_segments` fix here instead of deferring it.

Affected models as far as I can see: `deepseek_v32`, `longcat_flash`, `longcat_flash_ngram`, `falcon_h1`, `baichuan_m1`.

So both tests are welcome, the supplied-cache rotating one and the nested non-mergeable raise. Send them over and I will land the recursion here with them. If you would rather keep the raising test with #1618 where the rest of the validation lives, that works too, I just do not want the conversion half sitting there.

One loose end I have not chased: `RotatingQuantizedKVCache` has no `filter` or `extract`, so `CacheList.filter` would break on unmerged rotating leaves. Plain `RotatingKVCache` has neither either, so I think that is a property of rotating leaves rather than anything quantization adds, but I have not run it against the real eviction path. If your fork exercises that, I would like to know what you saw.

pierre427:
Tests attached, as promised — three rather than two, because writing them up surfaced a direct assertion the suite turned out not to have. All runs below are against the tip (`ff1e837`), M5, mlx 0.32.0, `MLX_ENABLE_TF32=0` per ml-explore/mlx#3897 (the full `tests/test_generate.py` is 33/33 clean on the tip in that environment, consistent with your and @PhilipJohnBasile's runs).

Status on the unpatched tip:

- `test_batch_insert_external_rotating_cache_is_quantized` — **passes**. The supplied-cache rotating config — the one combination that can actually batch — is only covered implicitly today, via the round trip re-inserting caches that are already quantized. This pins the conversion itself, plus generation through the merged path.
- `test_batch_insert_external_cachelist_nested_kv_is_quantized` — **fails**: the red test for the recursion.
- `test_batch_insert_external_cachelist_non_mergeable_raises` — **fails**: the red test for the leaf-level guard.

With the recursion in `maybe_quantize_kv_cache` plus a leaf check at insert (reference diff at the bottom — how you shape it is your call), all three pass. The file is standalone-runnable so you can try it before deciding where it lands; the methods drop into `TestGenerate` unchanged if you add `ArraysCache` and `CacheList` to the cache import. Black/isort clean.

<details>
<summary><code>tests/test_supplied_cache_1584.py</code></summary>

```python
# Copyright © 2026 Apple Inc.
"""Supplied-cache kv-quant tests for the #1584 CacheList follow-up.

Written against mabaeyens:kv-cache-quant-batching @ ff1e837. Expected
status on that tip:

- test_batch_insert_external_rotating_cache_is_quantized: PASS (pins the
  supplied-cache rotating config directly; today it is only covered
  implicitly via the continued-generation round trip)
- test_batch_insert_external_cachelist_nested_kv_is_quantized: FAIL
  (red until maybe_quantize_kv_cache recurses into CacheList)
- test_batch_insert_external_cachelist_non_mergeable_raises: FAIL
  (red until a leaf-level mergeability guard exists at insert)
"""

import unittest

import mlx.core as mx

    ArraysCache,
    CacheList,
    KVCache,
    RotatingKVCache,
    RotatingQuantizedKVCache,
)
from mlx_lm.utils import load


class TestSuppliedCacheQuantization(unittest.TestCase):

    @classmethod
    def setUpClass(cls):
        cls.HF_MODEL_PATH = "mlx-community/Qwen1.5-0.5B-Chat-4bit"
        cls.model, cls.tokenizer = load(cls.HF_MODEL_PATH)

    def test_batch_insert_external_rotating_cache_is_quantized(self):
        """The supplied-cache counterpart of test_batch_sliding_window_quantized:
        a caller-built rotating cache (snapshot restore, transplant) inserted
        with kv_bits set must become RotatingQuantizedKVCache at insert so it
        matches the fresh-lane cohort, and generation must then complete
        through the merged rotating-quantized path."""
        prompt = self.tokenizer.encode("Hello there")
        external = [RotatingKVCache(max_size=64) for _ in range(len(self.model.layers))]

        batch_gen = BatchGenerator(
            self.model,
            stop_tokens=self.tokenizer.eos_token_ids,
            max_tokens=4,
            max_kv_size=64,
            kv_bits=8,
            kv_group_size=32,
        )
        (uid,) = batch_gen.insert([prompt], caches=[external])

        self.assertTrue(
            all(isinstance(c, RotatingQuantizedKVCache) for c in external),
            f"supplied rotating cache not quantized: "
            f"{[type(c).__name__ for c in external]}",
        )

        tokens = []
        while responses := batch_gen.next_generated():
            tokens.extend(r.token for r in responses)
        self.assertGreater(len(tokens), 0, "supplied-cache job produced no tokens")
        for tok in tokens:
            self.assertFalse(mx.isnan(mx.array(float(tok))))

    def test_batch_insert_external_cachelist_nested_kv_is_quantized(self):
        """Composite CacheList layer caches (hybrid attention+recurrent models:
        baichuan_m1, falcon_h1, deepseek_v32, longcat_flash*) hold their KV
        cache one level down. The wrapper has merge() but no to_quantized(),
        so without recursing into the leaves a supplied CacheList silently
        stays full-precision while the fresh lane quantizes — and the mixed
        cohort then dies in CacheList.merge with AttributeError. Insert-level
        assertion, model-independent: supplied caches only, no generation
        (this model's attention layers don't consume CacheList). Leaf order
        mirrors baichuan_m1's sliding-window layers: (conv ArraysCache,
        RotatingKVCache)."""
        prompt = self.tokenizer.encode("Hello")
        external = [
            CacheList(ArraysCache(size=2), RotatingKVCache(max_size=64))
            for _ in range(len(self.model.layers))
        ]

        batch_gen = BatchGenerator(
            self.model,
            stop_tokens=self.tokenizer.eos_token_ids,
            max_tokens=4,
            max_kv_size=64,
            kv_bits=8,
            kv_group_size=32,
        )
        batch_gen.insert([prompt], caches=[external])

        for c in external:
            self.assertIsInstance(c, CacheList)
            self.assertIsInstance(c.caches[0], ArraysCache)
            self.assertIsInstance(
                c.caches[1],
                RotatingQuantizedKVCache,
                f"nested KV leaf not quantized: {type(c.caches[1]).__name__}",
            )

    def test_batch_insert_external_cachelist_non_mergeable_raises(self):
        """A nested plain KVCache quantizes to QuantizedKVCache, which has no
        merge(). _merge_caches' hasattr check only inspects the wrapper — and
        CacheList does have merge() — so a non-mergeable nested leaf dodges
        the friendly "does not yet support batching with history" ValueError
        and dies later as AttributeError inside CacheList.merge. It must fail
        loudly at insert instead. baichuan_m1 makes the shape concrete: its
        non-sliding-window layers are CacheList(ArraysCache, KVCache), so a
        blanket recursive pass leaves one model mergeable on some layers and
        not others."""
        prompt = self.tokenizer.encode("Hello")
        external = [
            CacheList(ArraysCache(size=2), KVCache())
            for _ in range(len(self.model.layers))
        ]

        batch_gen = BatchGenerator(
            self.model,
            stop_tokens=self.tokenizer.eos_token_ids,
            max_tokens=4,
            max_kv_size=64,
            kv_bits=8,
            kv_group_size=32,
        )
        with self.assertRaises(ValueError):
            batch_gen.insert([prompt], caches=[external])


if __name__ == "__main__":
    unittest.main()
```

</details>

**A sharper version of your round-trip caveat, found while verifying.** On the tip, `kv_bits` without `max_kv_size` cannot batch at all — not just "with history":

```python
gen = BatchGenerator(model, stop_tokens=tok.eos_token_ids, max_tokens=4, kv_bits=8, kv_group_size=32)
gen.insert([tok.encode("Hello")])  # fresh job, no supplied cache, no history
gen.next_generated()
# ValueError: <class 'mlx_lm.models.cache.QuantizedKVCache'> does not yet support batching with history
```

`_make_new_cache()` quantizes every layer to `QuantizedKVCache` at insert, and `PromptProcessingBatch.__init__` unconditionally runs the per-job caches through `_merge_caches()`, so the very first batch formation raises — on a job with no history, with a message that says "with history". And it isn't rescued by a model's own `make_cache()`: every in-tree rotating-cache model (`cohere2`, `exaone4`, `gemma4_text`, …) is mixed, with plain `KVCache` on the global layers, so at least one leaf always quantizes non-mergeable. As far as I can see there is no configuration in-tree where batch + `kv_bits` works without `max_kv_size`.

Two consequences worth naming:

1. `test_batch_insert_external_cache_is_quantized` and `test_batch_insert_external_cache_already_quantized` pin insert-time behaviour for a configuration that can never reach generation — and they are exactly the 2/33 that the leaf guard flips to a raise (everything else in `test_generate.py` holds). I'd rescope both to the rotating config rather than weaken the guard, but that's a behaviour decision on your PR, not mine to make.
2. The guard needs to cover both lanes, not just supplied caches: `_make_new_cache()` wraps only top-level `KVCache` entries when `max_kv_size` is set and doesn't descend into `CacheList`, so a *fresh* job on a baichuan_m1-shaped model with `max_kv_size` **and** `kv_bits` still carries a nested plain `KVCache` that quantizes non-mergeable. With the guard on both lanes, that's a clean error at insert instead of the `AttributeError` mid-batch; it also happens to turn the misleading no-`max_kv_size` message above into the same clear error.

**One placement constraint to make explicit in the writeup:** the mergeability check can't ride inside `maybe_quantize_kv_cache`'s recursion, because that function is shared with `generate_step`, where a merge-less `QuantizedKVCache` is the normal, correct result. It has to sit on the batch path, and insert is the earliest point where the post-quantization types exist. Deferring it to #1618 works mechanically, but then #1584 alone ships the nested-`AttributeError` window your baichuan_m1 example demonstrates — the dependency you said you didn't want. So my vote matches yours: recursion and the raise land here, the broader validation surface stays with #1618.

**Your loose end — `filter`/`extract` on unmerged rotating leaves.** Checked against our fork, which runs these classes in production serving with heavy eviction traffic: `filter`/`extract` are only ever invoked on the post-merge batch caches (`PromptProcessingBatch`/`GenerationBatch.prompt_cache`); per-job caches have exactly one consumer, `merge()`. So an unmerged rotating leaf — quantized or not — never sees `filter`, and plain `RotatingKVCache` doesn't implement `filter`/`extract` either. Your read is right: it's a property of rotating per-job leaves rather than anything quantization adds, and it's unreachable in the batching flow unless something calls `filter` on a per-job cache directly, which nothing in-tree does. We have never tripped it.

<details>
<summary>Reference diff the runs above used (yours to reshape)</summary>

```diff
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -291,7 +291,14 @@ def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_
     if kv_bits is None:
         return
     for e, c in enumerate(prompt_cache):
-        if hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
+        if isinstance(c, CacheList):
+            # Composite per-layer caches (hybrid attention+recurrent layers)
+            # hold their KV caches one level down; recurse so nested KV
+            # leaves honor kv_bits instead of silently staying fp.
+            leaves = list(c.caches)
+            maybe_quantize_kv_cache(leaves, quantized_kv_start, kv_group_size, kv_bits)
+            c.caches = tuple(leaves)
+        elif hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
             prompt_cache[e] = c.to_quantized(group_size=kv_group_size, bits=kv_bits)
 
 
@@ -1746,6 +1753,26 @@ class BatchGenerator:
                     self.kv_group_size,
                     self.kv_bits,
                 )
+            if self.kv_bits is not None:
+                # Both lanes, not just supplied caches: _make_new_cache()
+                # doesn't descend into CacheList when wrapping for
+                # max_kv_size, so a fresh job on a hybrid model can also
+                # carry a leaf that quantizes to a non-mergeable class.
+                for c in caches[i]:
+                    # CacheList.merge merges leaf-wise and _merge_caches'
+                    # hasattr check only sees the wrapper, so a non-mergeable
+                    # nested leaf would defer to an AttributeError deep in
+                    # CacheList.merge — check leaves, not the wrapper.
+                    leaves = c.caches if isinstance(c, CacheList) else (c,)
+                    for leaf in leaves:
+                        if not hasattr(leaf, "merge"):
+                            raise ValueError(
+                                f"kv_bits is set but the cache for this job "
+                                f"quantizes to {type(leaf).__name__}, which "
+                                "does not support batching. Batched kv-quant "
+                                "currently requires rotating caches (set "
+                                "max_kv_size) or an unquantized generator."
+                            )
 
         for seq, m, c, at, s, lp, sm in zip(
             segments,
```

</details>

mabaeyens:
Thanks for these, and for the third one. The direct assertion on the rotating config is the gap I would not have noticed, since the round trip was covering it only by accident.

Plan for them: all three fold into `TestGenerate` in `tests/test_generate.py` rather than a separate file, with `ArraysCache` and `CacheList` added to the cache import at the top. The commit carries `Co-Authored-By: pierre427 <4698223+pierre427@users.noreply.github.com>`, which is your GitHub noreply address since I do not have another one. If you would rather it named a real address, or would rather not be in the trailer at all, tell me before I push and I will use whatever you prefer. I have not run any of it yet, so nothing below is a verified result, and I will post red/green against the tip before I push.

Two things I would rather have your read on than decide alone.

The first is a finding that argues against splitting the recursion and the guard from each other, and I want you to check my reading of it. As things stand, a `baichuan_m1` shaped model with `kv_bits` and `max_kv_size` leaves the nested leaves full precision and silently ignores the flag, but it does not crash, because the nested plain `KVCache` still has `merge()`. Add the recursion on its own and that leaf becomes `QuantizedKVCache`, which has none, so `CacheList.merge` raises where nothing raised before. If that is right then the recursion without the guard is worse than neither, and there is no two way split here where one half is safe to defer. Either both land in this PR or the PR keeps the gap. You have run these classes in a serving fork, so I would like to know whether that matches what you hit when you ported them, or whether your fork never sees that layer shape and the question never came up.

The second is more awkward and it is really a question about this PR rather than about your patch. My instinct when your comment arrived was that adding to #1584 makes it too big to ever get looked at, so I went and got the numbers before saying that out loud, and they do not support me. The PR is +1157 across 4 files. Of the last 60 merged PRs in this repo only two are larger, and one of those is #1093 at +1238 from an outside contributor, so there is precedent sitting above where this already is. Your changes come to roughly +125 with the tests, about 11 percent, which moves it across no line I can actually point at. So the honest version is that the increment is not the problem. The problem is a 1157 line PR that has been open 24 days with no maintainer review, and neither adding to it nor trimming it by 11 percent changes that.

Where I am genuinely unsure is whether continuing to grow it is the wrong instinct regardless. I offered @angeloskath a split on 26 July, single sequence first at around 390 lines and the batching half after, with an explicit take it or leave it. No reply, and re-asking now would read as impatience. If this were yours, would you keep the PR whole and land the recursion and guard here on the correctness argument we both made, or would you take the gap on the chin for a while and put both in a follow up so that what is already written stays a reviewable size? I do not think there is an obviously right answer and I would rather ask than quietly pick the one that suits me. If it changes anything: @PhilipJohnBasile's #1618 and @katlun-lgtm's #1476 are both stacked behind this one, so a new PR in between moves their ground too, and I would want to say so before doing it rather than after.

On the two tests that flip, which you handed back to me. My leaning is to split them rather than rescope both. `test_batch_insert_external_cache_is_quantized` goes to the rotating config, and since your `test_batch_insert_external_rotating_cache_is_quantized` already covers that directly I would fold yours in and drop mine instead of keeping two that overlap. `test_batch_insert_external_cache_already_quantized` I would turn into a raise assertion, moving the identity check to a rotating variant, so the new error message ends up pinned rather than uncovered. That way the conversion path that genuinely stops existing gets replaced by an assertion that it now fails loudly, rather than just losing a test. Tell me if you think that is more machinery than it is worth.

One thing I am deliberately not doing, so it does not look like an oversight: `_make_new_cache()` will keep not descending into `CacheList` to wrap nested `KVCache` in `RotatingKVCache`. That would fix the hybrid models rather than error on them, but it changes unquantized batching behaviour for `baichuan_m1`, `falcon_h1`, `deepseek_v32` and `longcat_flash*` as well, which is past what a kv quant PR should be reaching for. Worth a follow up, not worth widening this.

And `filter`/`extract` I am treating as closed on your evidence. Production serving with heavy eviction traffic is a much better answer than the one I was going to get by reading the call sites again, and I will record it in the PR description as checked rather than assumed, with the attribution.

pierre427:
On attribution — the GitHub noreply you're already using is perfect, keep it as-is. Thanks for asking rather than assuming.

On the recursion/guard coupling — your reading is right, and I re-checked the base classes on current `main` rather than trust my memory of them. `QuantizedKVCache` has no `merge` (the first `merge` in `cache.py` is `KVCache`'s), and `CacheList.merge` calls `.merge` on each nested leaf, so a leaf that becomes quantized under recursion raises exactly where a plain `KVCache` leaf silently returned full precision. So recursion-without-guard is strictly worse than neither — they land together or the gap stays. No safe one-way split.

Honest caveat on the empirical half you asked about: my fork never exercises that shape. Muse-Glimmer is pure attention — flat `RotatingKVCache`/`KVCache`, no nested `CacheList`, no SSM leaf — so the baichuan_m1 / falcon_h1 path simply never comes up in my traffic. Treat my agreement as a code read, not a production data point; I didn't hit it, because that layer shape isn't in what I serve.

On growing vs deferring: if it were mine I'd land recursion+guard here. They move as a unit by the correctness argument we both made, and a follow-up PR would sit between this and #1618 / #1476 and move their ground — so "keep it small via a new PR" isn't free, it adds a stacked dependency to pay down a +11% your own numbers say crosses no line. The 24-day stall is the real problem, and neither growing nor trimming touches it; the thing that would is the single-seq/batching split you offered on the 26th. I wouldn't read re-raising that as impatience — 24 days plus two PRs stacked behind is a fair reason to nudge, and framing it as unblocking them rather than chasing a review keeps it clean.

Tests: your split is right and not over-machinery. Folding my rotating test in and dropping the overlap is just de-dup; turning `..._already_quantized` into a raise-assertion (identity check moved to a rotating variant) pins the new failure loudly instead of quietly losing coverage. Do that.

filter/extract — agreed, closed on the serving evidence; recording it as checked-with-attribution rather than assumed is the right call, thanks.

mabaeyens:
Pushed as `443dd99`. Recursion and guard both landed here, your three tests folded into `TestGenerate` rather than a separate file, and the trailer uses the noreply address as you confirmed.

Red/green, M5, mlx 0.32.0, `MLX_ENABLE_TF32=0`:

- On the tip, your three came out **PASS / FAIL / FAIL**, exactly as you called them.
- After the change, `tests/test_generate.py` is **36 passed** (33 before, minus one replaced, plus four) and `tests/test_prompt_cache.py` is **27 passed**. Formatted with the `black` 25.1.0 and `isort` 6.0.0 pinned in `.pre-commit-config.yaml`.

Before that I checked the one thing that would have collapsed the whole argument, which neither of us had run: whether `ArraysCache` has `merge()`. It does, so a `CacheList` really does merge on the unpatched tip and the mixed cohort is the only thing standing in the way.

One correction to my own wording, since you agreed on the basis of it and the loose version does not survive a close read. I said the recursion makes `CacheList.merge` raise "where nothing raised before". What I actually measured is narrower: **merge itself raises where it previously completed**. On Qwen something else raises immediately afterwards in either state, because `qwen2.py`'s attention reads `cache.offset` directly and a pure attention model cannot consume a `CacheList` at all, so a synthetic one never survives the forward pass whatever the cache classes are. The merge boundary is the part that moves under the recursion, and that is the part the guard is about. Your conclusion holds, my sentence was wider than the evidence.

That also answers whether this needs a hybrid model in the suite, and I do not think it does. The guard fires at `insert()`, before merge and before any forward pass, so it is model independent, which is exactly why all three of your tests are insert-level assertions and why they work on a pure attention model. A real `baichuan_m1` or `falcon_h1` would only add the other half, that merge completing leads to silently full-precision generation, and I would rather not put a hybrid checkpoint into a suite that deliberately runs on one 260MB model.

On the two tests, I went with the split you signed off. `test_batch_insert_external_cache_is_quantized` is gone, replaced by your rotating version since it covers the same ground on a config that can actually batch. `..._already_quantized` became two: the identity check moved to the rotating config, and `test_batch_insert_plain_kv_bits_without_max_kv_size_raises` asserts the new error names both `QuantizedKVCache` and `max_kv_size`, so the clearer message is pinned rather than just replacing the misleading one silently.

It went in as one commit rather than two. I had meant to separate recursion from guard, but given they are inseparable it seemed wrong to commit the intermediate state deliberately, and the trailer covers both halves anyway since the reference diff was yours as much as the tests were.

Not doing, so it does not read as an oversight: `_make_new_cache()` still does not descend into `CacheList` to wrap nested `KVCache` in `RotatingKVCache`. That would fix the hybrid models rather than error on them, but it changes unquantized batching behaviour for `baichuan_m1`, `falcon_h1`, `deepseek_v32` and `longcat_flash*` too, which is wider than a kv-quant PR should reach. Worth a follow-up.

Current `main` is three commits ahead (#1656, #1661, #1334) and none of them touch `cache.py`, `generate.py` or `tests/test_generate.py`, so there is no adjacency and the branch still merges clean.

Thanks again. The `filter`/`extract` answer in particular saved me a run I would have done badly.

zcbenz:

PR #1585, opened 2026-07-18
switch_layers: pad sorted gather rows to a multiple of 64 (silent quantized-MoE corruption, mlx#3856)

### LINKED ISSUE #3856: Silent numerical corruption in single long forward on quantized MoE when sequence length % 32 != 0 (Qwen3-Coder-30B-A3B-8bit, M5)
## Summary

A single `model(ids)` forward over a long prompt on a quantized MoE model
produces silently corrupted activations/KV whenever the forward's sequence
length is **not a multiple of 32**. Feeding the identical token ids through
the identical model in 2048-token chunks (what `mlx_lm.generate_step` does)
gives a different — and behaviorally correct — result. The divergence is
deterministic, content-independent (reproduces with random token ids), large
(fp16 K-cache entries off by up to ~86; final-position logits off by ~17–23;
greedy argmax flips), and big enough to break downstream behavior: a
needle-in-haystack QA that scores 100% recall with chunked prefill scores 0%
with a one-shot prefill at N≈16k.

This looks like a sibling of the 0.31.x quantized-matmul alignment bugs
(`gather_qmm` >32k-row MoE corruption; unaligned-N `quantized_matmul` with
m ≥ 10, m % 32 ≠ 0 — both fixed in 0.32.0). The MoE expert path appears to
still have an unaligned-M case: dense models (Qwen3-8B/32B, same quant
recipes, same lengths) are behaviorally clean in the same A/B, and the first
divergent tensor we observe is layer 1's K — i.e. the corruption enters
through layer 0's block output. Since dense models exercise the same
attention path at these shapes cleanly, the MoE FFN (`gather_qmm` /
gather-QMM-family kernels) is the prime suspect.

## Environment

- mlx **0.32.0.dev20260708+1ac2fbac** and **0.32.1.dev20260715+cf3ce39c**
  (both reproduce; source builds with NAX kernels enabled,
  `MACOSX_DEPLOYMENT_TARGET=26.2`)
- Apple M5, macOS 26 (Darwin 25.5.0), Python 3.12
- Model: `mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit` (affine 8-bit,
  MoE 30B-A3B). Dense `mlx-community/Qwen3-8B-4bit` / `Qwen3-32B-4bit` do
  **not** reproduce behaviorally at the same lengths.
- mlx-lm used only to load the model and build the KV cache.

## Repro

```python
# python repro.py            -> single length
# python repro.py --bisect   -> alignment scan
import argparse
import mlx.core as mx
from mlx_lm import load

def cache_diff(model, ids, chunk=2048):
    def build(step):
        for s in range(0, len(ids), step):
            logits = model(mx.array([ids[s:s+step]]), cache=cache)
            mx.eval(logits, [c.state for c in cache])
        return cache, logits[0, -1]
    c1, l1 = build(len(ids))   # one-shot
    c2, l2 = build(chunk)      # chunked
    worst_k = 0.0
    for a, b in zip(c1, c2):
        if getattr(a, "keys", None) is None:
            continue
        n = min(a.offset, b.offset)
        dk = float(mx.max(mx.abs(a.state[0][..., :n, :] - b.state[0][..., :n, :])).item())
        worst_k = max(worst_k, dk)
    dl = float(mx.max(mx.abs(l1 - l2)).item())
    same = int(mx.argmax(l1).item()) == int(mx.argmax(l2).item())
    return worst_k, dl, same

ap = argparse.ArgumentParser()
ap.add_argument("--model", default="mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit")
ap.add_argument("--bisect", action="store_true")
args = ap.parse_args()

model, tok = load(args.model)
mx.random.seed(0)
ids = mx.random.randint(0, 50000, (17000,)).tolist()
lengths = [8192, 12288, 16000, 16032, 16064, 16065, 16066, 16067, 16068,
           16128, 16384, 16900] if args.bisect else [16068]
print(f"mlx {mx.__version__}")
print(f"{'N':>7} {'worstK':>8} {'logitdiff':>9}  argmax")
for L in lengths:
    wk, dl, ok = cache_diff(model, ids[:L], 2048)
    print(f"{L:>7} {wk:>8.3f} {dl:>9.3f}  {'OK' if ok else 'DIVERGED'}", flush=True)
```

## Observed (random ids, seed 0, deterministic across runs)

```
mlx 0.32.0.dev20260708+1ac2fbac        (0.32.1.dev20260715: same pattern)
      N   worstK logitdiff  argmax
   8192   15.125     0.562  OK          # 32-aligned
  12288   15.125     1.172  OK          # 32-aligned
  16000   15.125     0.182  OK          # 32-aligned
  16032   15.125     0.344  OK          # 32-aligned
  16064   26.000     0.109  OK          # 32-aligned
  16065   32.000     7.562  DIVERGED    # N % 32 == 1
  16066   32.500     8.312  DIVERGED    # N % 32 == 2
  16067   32.000     8.375  DIVERGED    # N % 32 == 3
  16068   32.000    17.500  DIVERGED    # N % 32 == 4
  16128   26.000     1.148  OK          # 32-aligned
  16384   26.000     0.188  OK          # 32-aligned
  16900   32.000     9.438  DIVERGED    # N % 32 == 4
```

Every diverging length is unaligned; every 32-aligned length matches argmax
with small logit deltas. Unaligned lengths below 16k already show large
numeric error without flipping the argmax at the final position (e.g. on
0.32.1.dev20260715: N=8193 → worstK 131, logitdiff 17.0, argmax OK; N=12289 →
worstK 54, logitdiff 4.3, argmax OK), so short-prompt spot checks miss it.

Notes:

- The constant worstK ≈ 15–26 at aligned lengths is presumably benign
  kernel-tiling numerics amplified by outlier channels (argmax stable, logit
  deltas ≪ 1 in most rows); the unaligned case is qualitatively different.
- Layer-0 K is identical between the two paths (it depends only on the
  embeddings); divergence first appears in layer-1 K and compounds through
  all 48 layers.
- Real-workload impact: with a real 16,068-token needle-in-haystack prompt,
  the one-shot path's K diverges by up to 86, final logits by 23.4, and
  greedy QA recall flips 1.0 → 0.0 (the model emits repetitive
  off-task text). Chunked prefill of the same ids answers correctly. This
  cost us a multi-day false negative in a KV-eviction study before we traced
  it to the prefill path.

## Workaround

Chunk all prefills (any chunk size that keeps each forward 32-aligned or
short works; we use 2048).

We can test candidate fixes, run finer-grained bisects, or collect kernel-level
traces on M5 hardware on request.

pierre427:
## Problem

`gather_qmm` over sorted indices silently corrupts quantized-MoE outputs when the flattened gathered row count `n = tokens × top_k` exceeds 32768 and `n % 64 != 0`. This is ml-explore/mlx#3856 (minimal pure-mlx repro posted there). It is **still present in mlx 0.32.0 (PyPI) and 0.32.1.dev** for affine-quantized experts (4-bit and 8-bit tested); only the mxfp4 manifestation was fixed in 0.32.0. mlx-lm's floor is `MIN_MLX_VERSION = "0.31.2"`, where the same trigger corrupts even more broadly — so a quantized MoE at long single-shot prefill is exposed on every supported mlx today.

Arbitrated against an fp32-dequantized dense reference through `SwitchGLU` on an M5 (mlx 0.32.0 release, 4-bit/gs32, top_k=2):

| flattened rows n | n % 64 | result |
|---|---|---|
| 8 202 / 16 402 / 32 764 / 32 768 / 40 000 | — / — / 60 / 0 / 0 | clean |
| 32 800 | 32 | **64 rows wrong, max err 0.69** |
| 32 802 | 34 | **64 rows wrong, max err 0.50** |
| 40 002 | 2 | **7 264 rows wrong, max err 0.74** |
| 49 150 | 62 | **16 384 rows wrong, max err 0.83** |

Identical map on mlx 0.31.2. mxfp4 experts: clean at all shapes on 0.32.0. With the pad below: max err vs reference 1.3e-3 (= quantization rounding) at every shape, matching a 64-aligned chunked control. The row-alignment formulation also explains why per-model reports disagree on the "token multiple" trigger: with top_k=2 it appears as tokens%32, with top_k=4 as tokens%16 — the invariant is the flattened row count.

## Fix

Pad the sorted rows up to a multiple of 64 by duplicating the last row. Correctness of the padding (beyond the reference match above): indices stay sorted and in-bounds by construction, and `inv_order` is computed before padding, so `_scatter_unsort` structurally cannot address a pad row — NaN-poisoning the pad rows leaves the output bit-identical and fully finite. Overhead is ≤63 rows on >32768 (≤0.19% extra FLOPs); measured at T=20001 the guarded forward is actually *faster* (1.29 ms vs 1.76 ms median), since the aligned row count dispatches a better kernel path.

Both `gather_qmm` and `gather_mm` call sites route through `_gather_sort`, so the guard covers all five projection uses; the unsorted path (`indices.size < 64`) cannot reach the trigger. No collision with #1319 (fused gate-up) — its fused path consumes `_gather_sort` output and inherits the guard.

The regression test arbitrates quantized `SwitchGLU` at n=32802 against the fp32-dequantized dense reference: it fails on current `main` (max err ~0.5) and passes with the pad (max err <1e-3), in ~0.02 s.

When an mlx release fixes ml-explore/mlx#3856 this can become version-gated or be dropped. A `MIN_MLX_VERSION` bump alone is *not* an alternative — 0.32.0 corrupts identically on affine quants.

### DIFF
--- a/mlx_lm/models/switch_layers.py
+++ b/mlx_lm/models/switch_layers.py
@@ -14,7 +14,26 @@ def _gather_sort(x, indices):
     indices = indices.flatten()
     order = mx.argsort(indices)
     inv_order = mx.argsort(order)
-    return x.flatten(0, -3)[order // M], indices[order], inv_order
+    x = x.flatten(0, -3)[order // M]
+    indices = indices[order]
+    n = indices.size
+    # Metal mishandles the ragged tail tile in the sorted gather_qmm path
+    # (ml-explore/mlx#3856, still open as of mlx 0.32.x): when the flattened
+    # sorted row count exceeds 32768 and is not a multiple of 64, expert
+    # outputs come out wrong for affine-quantized experts (4/8-bit; only the
+    # mxfp4 path was fixed) — silent MoE corruption at long single-shot
+    # prefill. Pad the sorted rows up to a multiple of 64 (duplicating the
+    # last row, which keeps the indices sorted); _scatter_unsort gathers only
+    # the original rows, so the padding never reaches the output.
+    if n > 32768 and n % 64 != 0:
+        pad = 64 - n % 64
+        x = mx.concatenate(
+            [x, mx.broadcast_to(x[-1:], (pad,) + x.shape[1:])], axis=0
+        )
+        indices = mx.concatenate(
+            [indices, mx.broadcast_to(indices[-1:], (pad,))], axis=0
+        )
+    return x, indices, inv_order
 
 
 def _scatter_unsort(x, inv_order, shape=None):
--- a/tests/test_switch_layers.py
+++ b/tests/test_switch_layers.py
@@ -1,0 +1,60 @@
+import unittest
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from mlx_lm.models.switch_layers import SwitchGLU
+
+
+class TestSwitchLayers(unittest.TestCase):
+    def test_quantized_switch_glu_ragged_sorted_tail(self):
+        # ml-explore/mlx#3856: the sorted gather_qmm path corrupts expert
+        # outputs for affine-quantized weights when the flattened row count
+        # exceeds 32768 and is not a multiple of 64. 16401 tokens x top-2
+        # gives n = 32802, which triggers the ragged tail; the pad guard in
+        # _gather_sort must keep the output matching a dense fp32 reference.
+        mx.random.seed(0)
+        num_tokens, top_k = 16401, 2
+        dims, hidden_dims, num_experts = 64, 64, 4
+
+        glu = SwitchGLU(dims, hidden_dims, num_experts)
+        nn.quantize(glu, group_size=64, bits=4)
+
+        x = mx.random.normal((num_tokens, dims))
+        indices = mx.random.randint(0, num_experts, (num_tokens, top_k))
+        out = glu(x, indices)
+        mx.eval(out)
+
+        def dequant(proj):
+            return mx.dequantize(
+                proj.weight,
+                proj.scales,
+                proj.biases,
+                group_size=proj.group_size,
+                bits=proj.bits,
+                mode=getattr(proj, "mode", "affine"),
+            ).astype(mx.float32)
+
+        w_gate = dequant(glu.gate_proj)
+        w_up = dequant(glu.up_proj)
+        w_down = dequant(glu.down_proj)
+
+        xf = x.astype(mx.float32)
+        per_expert = []
+        for e in range(num_experts):
+            h = nn.silu(xf @ w_gate[e].T) * (xf @ w_up[e].T)
+            per_expert.append(h @ w_down[e].T)
+        dense = mx.stack(per_expert, axis=1)
+        ref = mx.take_along_axis(
+            dense, indices[..., None], axis=1
+        )
+        mx.eval(ref)
+
+        max_err = mx.abs(out.astype(mx.float32) - ref).max().item()
+        # The gather_qmm bug produces errors ~0.5 on affected rows; a healthy
+        # run agrees with the fp32 reference to well under 1e-3.
+        self.assertLess(max_err, 1e-3)
+
+
+if __name__ == "__main__":
+    unittest.main()


zcbenz:

PR #1153, opened 2026-04-15
Feat/mamba mlx kernels

Gal-bloch:
## Summary
Integrate optimized Mamba/SSM kernels for Apple Silicon in `mlx-lm` with correctness-preserving fallbacks.

Closes: #1152 

## What Changed
- Added optimized Mamba-1 prefill path using selective-scan kernel.
- Added optimized Mamba-1 decode path using fused selective state-update kernel.
- Added optimized Mamba-2 prefill path using SSD chunk kernels.
- Kept fallback paths for unsupported shapes/devices/runtime conditions.
- Added model-local compatibility handling for `mamba2-2.7b` config parsing in `mlx_lm/models/mamba2.py` (derive `intermediate_size` from `hidden_size * expand`).

## Why
These kernels reduce prefill/decode overhead on Metal and improve throughput while preserving generation behavior.

## Scope
- [x] Decode path (`seq_len == 1`) optimization for Mamba-1
- [x] Prefill path (`seq_len > 1`) optimization for Mamba-1 and Mamba-2
- [x] No public API changes

## Correctness
Reference: existing `mlx-lm` implementation.

Validation included:
- [x] End-to-end generation functional checks across Mamba family models
- [x] Cache/state update path exercised in decode and prefill
- [x] Fallback path retained when kernel conditions are not met

Tolerance used:
- `rtol`: 1e-3
- `atol`: 1e-3

## Contributing Checklist
- [x] Formatting/linting run with `pre-commit run --all-files`
- [x] Existing behavior preserved with guarded fallbacks
- [x] Tests/validation commands included below
- [ ] At least one maintainer review
- [ ] CI green on pull request

## Benchmarks
Environment:
- Device: Apple M1 Max (64 GB)
- Python: 3.12.13
- mlx: 0.31.1
- mlx-metal: 0.31.1
- Baseline: clean `main`
- Optimized: `feat/mamba-mlx-kernels`

Workload settings:
- Prompt tokens: 1024
- Generation tokens: 128
- Batch size: 1
- Trials: 1

Results (before -> after), focused on confirmed improvements:

- `mlx-community/mamba-370m-hf-f16` (Mamba-1)
	- prompt_tps: `293.741 -> 5497.575` (`18.72x`)
	- generation_tps: `159.172 -> 163.434` (`1.03x`)
	- peak_memory (GB): `1.027 -> 2.173`

- `mlx-community/mamba2-2.7b` (Mamba-2)
[...]

PR #1586, opened 2026-07-18
ssm: fused sequential-scan kernel for short multi-token updates (S=2-8)

pierre427:
## Problem

Speculative-verify blocks on Mamba2-family models (2 ≤ S ≤ 8 tokens: MTP verify, n-gram / prompt-lookup drafts, short tail chunks with carried state) fall through to the chunked SSD path (`ssm_attn`), which costs ~3.2× a single decode step per layer at nemotron_h shapes — paid on every verify cycle.

## Change

A Metal kernel that keeps the state slice in registers and loops over S sequentially in one launch, engaged for `1 < S <= 8` when a state is present (mask/lengths None, GPU, `state_dim >= 32` and a multiple of 32 — every in-tree Mamba2 model: ds=128 for nemotron_h/mamba2/falcon_h1/granitemoehybrid, ds=64 for plamo2). All accumulation in fp32, same `compute_dt` as the step kernel.

- **Bit-identical to S composed step-kernel calls for float32 state** — the only state dtype the pipeline produces (the prefill paths always emit an fp32 state). Randomized configs × dtypes (fp32/fp16/bf16 inputs) × batch shapes: max diff 0.0. Verify and decode numerics agree exactly, so lossless speculative decoding stays lossless.
- Per-layer at nemotron_h dims (M5, bf16 inputs): S=2 46.0 µs vs 88.2 µs (**1.92×**), S=4 39.6 vs 89.9 (**2.27×**), S=8 43.6 vs 85.3 (**1.96×**) — vs the SSD path it replaces.

Tests cover: an exact-recurrence fp32 reference, bit-identity vs the composed S=1 step kernel (fp32/fp16/bf16 inputs, B=1/3, nemotron dims), and the dispatch boundaries (S=1 → step kernel, S=9 → `ssm_attn`, unsupported state_dim → `ssm_attn`, mask → `ssm_attn`).

## Relationship to #1579 (and #1153)

Complementary, not competing — this completes the S axis: **S=1** existing step kernel, **S=2–8** this PR, **S≥32** #1579's chunked SSD prefill kernels (9–31 stays on `ssm_attn`). All shared helpers (`compute_dt`, the step kernel, `segsum`, `ssm_attn`) are untouched by both branches; the only overlap is the `ssm_update` dispatch body, a trivial merge whichever lands first. #1153's kernels (Mamba-1 scan, SSD prefill, fused S=1 step) also don't cover this regime.

## Note on outputs

Short prefill tail chunks (2–8 tokens with carried state) previously took the SSD path and now take the sequential kernel. Outputs are numerically exact w.r.t. the step-kernel recurrence (fp32 sequential ≥ SSD chunked fidelity), but low-bit generation hashes can shift vs prior builds for those shapes.

### DIFF
--- a/mlx_lm/models/ssm.py
+++ b/mlx_lm/models/ssm.py
@@ -64,6 +64,107 @@ def make_ssm_kernel():
 _ssm_kernel = make_ssm_kernel()
 
 
+def make_ssm_seq_kernel():
+    """Sequential-scan variant of the step kernel for short sequences
+    (speculative verify blocks): one launch, the state slice stays in
+    registers across the S positions instead of bouncing through the SSD
+    chunked path, which for S in [2, 8] costs ~3x a single step."""
+    if not mx.metal.is_available():
+        return None
+    source = """
+        auto n = thread_position_in_grid.z;
+        auto h_idx = n % H;
+        auto b_idx = n / H;
+        auto grp = (h_idx / G);
+        constexpr int n_per_t = Ds / 32;
+        constexpr int HB = H / G;
+
+        auto i_state = state_in + n * Dh * Ds;
+        auto o_state = state_out + n * Dh * Ds;
+
+        auto ds_idx = thread_position_in_threadgroup.x;
+        auto d_idx = thread_position_in_grid.y;
+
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1590, opened 2026-07-20
Never lower RLIMIT_NOFILE when importing mlx_lm

### LINKED ISSUE #1589: Importing mlx_lm lowers RLIMIT_NOFILE and irreversibly caps the hard limit
### Describe the bug

`mlx_lm/utils.py` unconditionally overwrites `RLIMIT_NOFILE` at import time:

https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/utils.py#L35-L36

```python
# For large models with lots of files
resource.setrlimit(resource.RLIMIT_NOFILE, (2048, 4096))
```

Raising the *soft* limit toward 2048 is reasonable and works as intended on a default macOS box, where the limit starts at `(256, RLIM_INFINITY)`. The problem is the second value:

1. **Lowering the hard limit is irreversible.** POSIX only lets an unprivileged process *lower* a hard limit; it can never raise it again. `RLIM_INFINITY -> 4096` therefore permanently caps the whole process at 4096 descriptors, as an import side effect. A `ulimit -n` raised by the operator beforehand is silently clamped, and a host application that raises its own limit at startup cannot restore it afterwards — `setrlimit` fails with `ValueError`.

2. **The import hard-fails when the existing hard limit is below 4096.** Because the call tries to *raise* the hard limit in that case, `import mlx_lm` dies:

   ```console
   $ ( ulimit -Hn 1024; python -c "import mlx_lm.utils" )
   ValueError: not allowed to raise maximum limit
   ```

   This makes mlx-lm unimportable in any environment with a tightened hard limit, e.g. a container started with `--ulimit nofile=1024:1024`.

I hit the first case running mlx-lm as a library inside a long-lived server process (vLLM's Metal backend), which raises `RLIMIT_NOFILE` at startup and found it silently clamped after model load.

Separately: mutating a process-wide resource limit as an import side effect is surprising for a library. A caller importing `mlx_lm.utils` to inspect a config has no reason to expect its descriptor limits to change.

### To Reproduce

**Irreversible hard-limit clamp** — run with a raised limit, as a server process typically would:

```console
$ ulimit -n 65536
$ python -c "
import resource
print('before:', resource.getrlimit(resource.RLIMIT_NOFILE))
import mlx_lm.utils
print('after :', resource.getrlimit(resource.RLIMIT_NOFILE))
try:
    resource.setrlimit(resource.RLIMIT_NOFILE, (2048, 65536))
except ValueError as e:
    print('restore: FAILED —', e)
"
before: (65536, 9223372036854775807)
after : (2048, 4096)
restore: FAILED — not allowed to raise maximum limit
```

**Import failure under a low hard limit:**

```console
$ ( ulimit -Hn 1024; python -c "import mlx_lm.utils" )
ValueError: not allowed to raise maximum limit
```

### Expected behavior

Raise the soft limit toward the intended 2048 floor when it is lower, and leave the hard limit alone:

```python
soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
if soft != resource.RLIM_INFINITY and soft < 2048:
    soft = 2048 if hard == resource.RLIM_INFINITY else min(2048, hard)
    resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
```

This keeps the original motivation — ensure enough descriptors for many-shard downloads, which is the common macOS case where the default soft limit is 256 — while never making the change irreversible and never failing the import.

Ideally the adjustment would also move out of module scope into the loader path that actually opens many files, so that merely importing the module has no global side effect.

I've opened a PR for this.

### Desktop (please complete the following information):

- OS: macOS 15.7.4 (24G517), Apple M4
- Version: mlx-lm 0.31.3, mlx 0.32.0, Python 3.12.12
- The offending line is unchanged on `main` as of this report.

hyunnnchoi:
## Summary

Fixes #1589.

`mlx_lm/utils.py` overwrites `RLIMIT_NOFILE` with a fixed `(2048, 4096)` at module scope, without reading the current limits.

Raising the *soft* limit toward 2048 is reasonable and works as intended on a default macOS box, where the limit starts at `(256, RLIM_INFINITY)`. The problem is the second value:

- **Lowering the hard limit is irreversible.** An unprivileged process can only ever lower a hard limit, so `RLIM_INFINITY -> 4096` permanently caps the process at 4096 descriptors as an import side effect. A `ulimit -n` raised beforehand is silently clamped, and a host application that raises its own limit at startup cannot restore it afterwards. I hit this running mlx-lm under vLLM's Metal backend.
- **The import hard-fails when the existing hard limit is below 4096.** The call then tries to *raise* the hard limit, which is not permitted, so `import mlx_lm` dies with `ValueError` — e.g. in a container started with `--ulimit nofile=1024:1024`.

This reads the current limits and only raises the soft limit toward the 2048 floor, never touching the hard limit. The original motivation is preserved for the common macOS case, and the call becomes a no-op when the limit is already sufficient.

`RLIM_INFINITY` is handled explicitly rather than via `min()`, since it is represented as `-1` on Linux.

## Verification

macOS 15.7.4, Apple M4, Python 3.12.12.

| Starting limit | Before | After |
| --- | --- | --- |
| macOS default (launchd) | `(256, ∞)` → `(2048, 4096)` | `(256, ∞)` → `(2048, ∞)` — soft still raised, hard preserved |
| Operator raised it | `(65536, ∞)` → `(2048, 4096)` | `(65536, ∞)` unchanged |
| Hard limit below 4096 | `(1024, 1024)` → `ValueError`, import fails | `(1024, 1024)` unchanged, import succeeds |

The first row is the case the original code was written for, and it still works — the difference is that the hard limit survives. Rows two and three are the bugs.

Before this change, row two could not be undone:

```console
restore hard->65536: FAILED (not allowed to raise maximum limit)
```

## Tests

Adds `TestOpenFileLimit` with two cases:

- `test_import_does_not_lower_nofile_limit` — raises the soft limit in a subprocess, imports `mlx_lm.utils`, asserts the soft limit did not decrease and the hard limit is unchanged.
- `test_import_survives_low_hard_limit` — imports under `(1024, 1024)` and asserts the import succeeds with the hard limit intact.

```console
$ python -m unittest tests.test_utils.TestOpenFileLimit
Ran 2 tests in 3.24s

OK
```

Formatted with `black`.

## Note

Arguably this adjustment should move out of module scope into the loader path that actually opens many files, so that importing the module has no global side effect at all. I kept this PR to the minimal correctness fix — happy to follow up if you'd prefer that shape.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -32,8 +32,14 @@ if os.getenv("MLXLM_USE_MODELSCOPE", "False").lower() == "true":
 else:
     from huggingface_hub import snapshot_download
 
-# For large models with lots of files
-resource.setrlimit(resource.RLIMIT_NOFILE, (2048, 4096))
+# For large models with lots of files, ensure the soft limit on open file
+# descriptors is at least 2048. Only ever raise it: lowering the hard limit is
+# irreversible for an unprivileged process, and would permanently cap a host
+# application that had deliberately raised its own limit.
+_soft, _hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+if _soft != resource.RLIM_INFINITY and _soft < 2048:
+    _soft = 2048 if _hard == resource.RLIM_INFINITY else min(2048, _hard)
+    resource.setrlimit(resource.RLIMIT_NOFILE, (_soft, _hard))
 
 from mlx.utils import tree_flatten, tree_map, tree_reduce, tree_unflatten
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -2,6 +2,8 @@
 
 import json
 import os
+import subprocess
+import sys
 import tempfile
 import unittest
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


zcbenz:

Issue #1470, opened 2026-07-04
speculative_generate_step diverges from plain generation at temp=0 with num_draft_tokens=4 (dense model, no hybrid layers)


h9q2cyxvgm-ui:
### Summary

`speculative_generate_step` (and `generate()`/`stream_generate` with `draft_model=`) produces output that diverges from plain greedy generation at `temp=0` with `num_draft_tokens=4`, on a plain dense model + a small peer draft model — no hybrid/recurrent architecture involved, no custom code. Speculative decoding at temp=0 is supposed to be lossless relative to plain generation (that's the whole point — it should only ever produce a different token when the draft model happens to sample the exact same one the target would have), so this looks like a real bug rather than expected variance.

Confirmed on both current `main` and the tip of an unrelated feature branch (`h9q2cyxvgm-ui/mlx-lm@feature/qwen3-5-mtp-speculative-decoding`, see #1468/#1469) — same result either way, so it's not something introduced by that branch's changes.

### Reproduction

```python
import mlx.core as mx
from mlx_lm.generate import generate_step, speculative_generate_step
from mlx_lm.utils import load
from mlx_lm.sample_utils import make_sampler

TARGET = "mlx-community/Qwen3-32B-6bit"
DRAFT = "mlx-community/Qwen3-0.6B-MLX-6bit"   # any small same-family draft
MAX_TOKENS = 150

model, tokenizer = load(TARGET)
draft_model, _ = load(DRAFT)

messages = [{"role": "user", "content": "Explain what a hash table is in two sentences."}]
prompt = mx.array(tokenizer.apply_chat_template(messages, add_generation_prompt=True))
sampler = make_sampler(temp=0.0)

plain = [tok for tok, _ in generate_step(prompt, model, max_tokens=MAX_TOKENS, sampler=sampler)]

spec = []
for tok, _, from_draft in speculative_generate_step(
    prompt, model, draft_model, max_tokens=MAX_TOKENS, num_draft_tokens=4, sampler=sampler
):
    spec.append(tok)

print(plain == spec)  # False
for i, (a, b) in enumerate(zip(plain, spec)):
    if a != b:
        print(i, a, b)
        break
```

### Result

First divergence at token index 91: plain generates token `429` (`" that"`), speculative generates `1246` (`" how"`) — a genuine content divergence partway through generation, not just a length/edge-case difference.

### Notes

[...]

PR #1592, opened 2026-07-21
Document speculative_generate_step temp=0 tie-breaking divergence

rajanshxrma:
## Summary

Documentation-only clarification for #1470 (`speculative_generate_step` diverging from plain generation at `temp=0`).

The thread on #1470 converged on a root cause across three separate model families (Qwen3-32B-4bit, Qwen3-8B-6bit, and the original dense reproduction) — the divergence is **expected bfloat16 floating-point non-associativity, not a logic bug**:

- The accept/reject test in `speculative_generate_step` compares each draft token against the target model's own greedy (argmax) choice, so the algorithm is lossless with respect to plain generation *in exact arithmetic*. The accept/reject/rewind logic itself is correct.
- In practice, at `temp=0`, the draft block's target scoring runs as **one batched forward pass**, while plain `generate_step` scores **sequentially, one token at a time**. These are mathematically equivalent but take different reduction/kernel paths through matmul/attention.
- `bfloat16` has only an 8-bit mantissa, so exact ties between competing top-logit candidates aren't rare. When a tie occurs, whichever path rounds fractionally lower wins — and that can differ between the batched and sequential paths, causing the two generation modes to pick a different token at that position.
- This does not occur when the top logits aren't exactly tied, and doesn't indicate any correctness problem with the speculative decoding implementation.

## What changed

Added a `Note:` section to the `speculative_generate_step` docstring in `mlx_lm/generate.py` explaining this behavior, so users hitting it don't mistake it for a bug.

No code paths changed; docs-only.

## Test plan

- [x] `black --check` passes (matches repo formatting)
- [x] `isort --check --profile=black` passes (matches repo's configured pre-commit profile)
- [x] Confirmed via `ast.get_docstring` that the docstring parses and renders as intended (no formatting artifacts)
- N/A regression tests — documentation-only change, no behavior modified

Closes nothing; related to #1470 (leaving that issue open since it's tracking the broader discussion, but this addresses the "is this a bug" question raised there).


zcbenz:

PR #1593, opened 2026-07-21
Drop import-level sampler compilation

### LINKED ISSUE #1245: mlx_lm.server: seed parameter silently ignored

michalk8:
Drop module-level `mx.compile` in `sampling_utils.py`. This fixes `server.py`'s sampling seed.

Fixes #1245
Closes #1331

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -1,7 +1,6 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import math
-from functools import partial
 from typing import Callable, Dict, List, Optional
 
 import mlx.core as mx
@@ -66,7 +65,10 @@ def make_sampler(
         # Return the sampled token
         return categorical_sampling(logprobs, temp)
 
-    return sampler
+    # ``mx.random.state`` is thread-local, so a sampler compiled
+    # on the main thread (at import time) would ignore reseeding
+    # on another thread
+    return mx.compile(sampler, inputs=mx.random.state, outputs=mx.random.state)
 
 
 def make_logits_processors(
@@ -126,7 +128,6 @@ def make_logits_processors(
     return logits_processors
 
 
-@partial(mx.compile, inputs=mx.random.state, outputs=mx.random.state)
 def apply_top_k(
     logprobs: mx.array,
     top_k: int,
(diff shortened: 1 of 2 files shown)


katlun-lgtm:
Ran this against the repro from #1675 on an M3 Max (mlx 0.32.0, macOS 27.0), comparing `646aa0a9` to stock main at `cf10f96`. It fixes it, and the output matches the uncompiled reference rather than merely being unfrozen:

```
                                main cf10f96          #1593 646aa0a9
worker-built, worker-drawn    [2,2,2,2,2,2,2,2]    [2,7,24,6,0,7,7,7]
uncompiled control (worker)   [2,7,24,6,0,7,7,7]   [2,7,24,6,0,7,7,7]
```

Worth being precise about what changes, because one case in my repro stays frozen on both revisions and I don't think that's a defect: a sampler built on the main thread and *called* from a worker is still stuck. Moving the compile into `make_sampler()` binds the sampler to whichever thread builds it, which isn't the same as making it thread-portable. That's the right trade here, but it does mean the fix is sufficient because of where the call sites happen to sit — `server.py` runs a single `Thread(target=self._generate)` (`:430`), and both `_make_sampler` calls are on it, `:717` in the batch path and `:905` in `_serve_single`, itself invoked from `_generate` at `:750`. Build thread and draw thread coincide, so the state gets read *and* advanced.

The one thing I'd raise for review is that this now-load-bearing property isn't enforced or written down anywhere. If `_make_sampler` ever moves to the HTTP handler thread — which looks like a harmless refactor, building the sampler where the request is parsed — the bug comes back silently, and the test here wouldn't catch it, since it builds and draws on the same thread. A comment at the two call sites would probably be enough; a test that builds on one thread and draws on another, asserting the documented behaviour, would pin it harder. Happy to send either as a patch if it's useful.

zcbenz:
This had been fixed by https://github.com/ml-explore/mlx/pull/3828 and I'm closing this in favor of https://github.com/ml-explore/mlx-lm/pull/1753.

PR #1595, opened 2026-07-21
tests: pin MLX_ENABLE_TF32=0 so kernel-vs-reference checks stay fp32-exact on M5

pierre427:
On M5-class GPUs with mlx >= 0.32, `tests.test_models.TestModels.test_ssm` fails deterministically on current `main`:

```
FAIL: test_ssm (tests.test_models.TestModels.test_ssm)
AssertionError: array(False, dtype=bool) is not true   # out_state allclose, atol=rtol=1e-4
```

There is nothing wrong with the ssm kernels — the failure is a precision-regime mismatch introduced by the environment.

## Root cause

mlx defaults `MLX_ENABLE_TF32=1`, so float32 GEMMs run on the M5 neural accelerator at TF32-class precision (confirmed as expected behavior in ml-explore/mlx#3534). `test_ssm` compares the SSD reference path (`ssm_attn`) against the elementwise single-step Metal kernel (`ssm_update`) at `atol=rtol=1e-4`:

- The reference computes `next_state` via the outer product `dtxdecay @ B` — a GEMM shape, so it picks up ~2e-3 relative TF32 error.
- The step kernel does the same math elementwise in fp32 and stays exact. A float64 cross-check confirms the **kernel is the correct side** (max abs err 3.3e-6; the GPU reference is off by ~1.5e-3 relative).
- The `out` comparison keeps passing because at `seq_len=1` its contractions are gemv-shaped (M=1 or N=1), which don't take the NAX GEMM route.

Measured on an M5 Max (macOS 26.5, mlx 0.32.0 wheel), fp32 vs float64 reference:

| shape | default | `MLX_ENABLE_TF32=0` |
|---|---|---|
| `(64,1)@(1,128)` outer product, max rel err | 1.9e-3 | 6.0e-8 |
| `(48,64,128)@(48,128,32)`, max abs err | 4.3e-2 | 2.8e-5 |

## Fix

Pin `MLX_ENABLE_TF32=0` at test-module import via `os.environ.setdefault`. The flag latches process-wide on the first matmul, so import time is early enough, and `setdefault` keeps an explicit `MLX_ENABLE_TF32=1` run possible. The kernel-vs-reference tolerances stay at 1e-4, where they can still catch real kernel bugs.

## Verification

M5 Max, macOS 26.5, mlx 0.32.0:

- `main` without this patch: `test_ssm` fails deterministically.
- With this patch: full `test_models` suite passes (80 tests, 1 skipped).

M1–M4 machines are unaffected either way (no NAX, fp32 GEMM already exact).

### DIFF
--- a/tests/test_generate.py
+++ b/tests/test_generate.py
@@ -1,8 +1,15 @@
 # Copyright © 2024 Apple Inc.
 
+import os
 import random
 import unittest
 from typing import List
+
+# mlx >= 0.32 runs float32 GEMMs at TF32-class precision on M5 neural
+# accelerators unless MLX_ENABLE_TF32=0. The batch-vs-single equivalence checks
+# in this file compare against fp32-exact references and fail otherwise.
+# Pin it off; the flag latches process-wide on first matmul.
+os.environ.setdefault("MLX_ENABLE_TF32", "0")
 
 import mlx.core as mx
 
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1,7 +1,16 @@
 # Copyright © 2024 Apple Inc.
 import copy
 import importlib
+import os
 import unittest
+
+# mlx >= 0.32 runs float32 GEMMs at TF32-class precision on M5 neural
+# accelerators unless MLX_ENABLE_TF32=0. Kernel-vs-reference checks in this
+# file that build the reference via matmul (e.g. test_ssm's SSD outer-product
+# state path) then fail their 1e-4 allclose against fp32-exact elementwise
+# kernels (TF32 is ~2e-3 relative). Pin it off; the flag latches process-wide
+# on first matmul.
+os.environ.setdefault("MLX_ENABLE_TF32", "0")
 
 import mlx.core as mx
 import mlx.nn as nn


pierre427:
Updated the in-file comment (no functional change): the original wording implied the gated_delta tests are also affected — verified on a pristine clone of main that `test_gated_delta`/`_masked`/`_precision` pass under both TF32 settings; only `test_ssm` builds its reference through a matmul-shaped path at a 1e-4 tolerance. The comment now names `test_ssm` specifically, and the commit message carries the fail-with-default / pass-with-pin verification.

katlun-lgtm:
The pin fixes the `test_ssm` case, but the same precision regime is failing a second file this PR doesn't reach.

@mabaeyens measured `tests/test_generate.py` on an M5 (gen-17): 8 failed / 20 passed natively, 28 passed with `MLX_ENABLE_TF32=0`, stable over three repeat runs. The eight are `test_batch_matches_single`, `test_many_batches`, `test_batch_sliding_window`, `test_batch_continued_generation{,_gated_delta,_ssm}` and `test_stream_generate_input_embeddings{,_prefill}`. Root-cause thread is ml-explore/mlx#3897; his read is that those are dominated by fp32 accumulation in the quantized-matmul path rather than by the NAX attention kernel, which is why the flag clears all eight even though the model is 4-bit.

I can't reproduce the failure — M3 Max is gen-15, where the flag is inert (512³ fp32 GEMM is bit-identical with and without it, since the Metal gates all sit behind `is_nax_available()`) — but I did check the pin is harmless there: `tests/test_generate.py` is 26 passed both ways at cf10f96. (26 vs his 28 is just the older base.)

Happy to send the same `os.environ.setdefault` line for `test_generate.py`, either as a commit on your branch if you'd rather keep it in one PR, or as a follow-up once this lands — your call. mlx core sets it suite-wide in `python/tests/mlx_tests.py`, so one shared spot may be tidier than two files.

Worth keeping separate from the tolerance question, though. TF32 doesn't touch the fp16/bf16 batch-vs-single gap on gen-17 — that one's the NAX attention kernel, and no env var opts out of it — so an `atol` floor is still wanted for those regardless of this pin.

pierre427:
Thanks, this is exactly the split I was hoping to keep clear.

Agreed that `tests/test_generate.py` belongs to the same fp32/TF32 lane as this PR, not to the fp16/bf16 NAX-attention tolerance lane. The useful deconflict for me is:

- `MLX_ENABLE_TF32=0` fixes the fp32 GEMM / quantized-matmul precision regime that trips `test_ssm` here and, from @mabaeyens' M5 runs, the eight `test_generate.py` failures.
- It does not address the fp16/bf16 masked-attention divergence in ml-explore/mlx#3897, so any `atol` / per-arch tolerance work should stay separate.
- The broader documentation/diagnostic home for the TF32 default is ml-explore/mlx#3860, with the doc/log split already represented by ml-explore/mlx#3894 and ml-explore/mlx#3883.

So yes, I’m happy to fold the same early `os.environ.setdefault("MLX_ENABLE_TF32", "0")` into `tests/test_generate.py` in this PR if maintainers prefer one M5 test-suite hygiene change. If you’d rather send it as a commit on this branch, that works for me too. If maintainers prefer a shared test bootstrap instead of two per-file pins, I’m also fine moving it there; the only hard requirement is that it happens before the first `mlx.core` import because the flag latches on first use.

katlun-lgtm:
Folding it into this PR sounds right, and it's the only path that stays one PR — GitHub only offers the maintainer-edit checkbox on fork PRs, so a commit from me would have to arrive as a separate PR against main. Here's the hunk so it's no work for you:

```diff
 # Copyright © 2024 Apple Inc.

+import os
 import random
 import unittest
 from typing import List

+# mlx >= 0.32 runs float32 GEMMs at TF32-class precision on gen-17 (M5) neural
+# accelerators unless MLX_ENABLE_TF32=0. The batch-vs-single equivalence checks
+# in this file compare against fp32-exact references and fail otherwise.
+# Pin it off; the flag latches process-wide on first matmul.
+os.environ.setdefault("MLX_ENABLE_TF32", "0")
+
 import mlx.core as mx
```

Same shape as your `test_models.py` change, and it has to sit above the `import mlx.core as mx` on line 7 for the latching reason you gave. The statement doubles as an isort barrier, so the two import blocks stay split.

Evidence split, to be clear about whose is whose: the green-on-M5 half is @mabaeyens' 28/28. Mine is only that it's inert on gen-15 — `test_generate.py` is 26 passed with and without the pin at cf10f96, so it costs pre-M5 CI nothing.

If maintainers would rather have one shared pin than two per-file ones, adding a `tests/__init__.py` (the package has none today) is the spot that covers every runner. I checked all three on an M3 Max with a pin plus a marker line in it: it executes under `pytest tests/test_generate.py`, under `python -m unittest discover -s tests -t .`, and under `python -m unittest tests.test_generate`. A `conftest.py` would only cover the pytest path. Either way it lands before the first `mlx.core` import.

Thanks for the #3894 / #3883 pointers.

PhilipJohnBasile:
Verified the current head commit e7a02371893c61a604ffcca01e5f96299630876c in an isolated worktree:

- with MLX_ENABLE_TF32 unset, importing tests/test_generate.py sets it to 0 before mlx.core is imported;
- an explicit MLX_ENABLE_TF32=1 override remains 1, so setdefault behaves as intended;
- both modified test modules compile, and the PR diff passes git diff --check;
- the fp16/bf16 NAX-attention tolerance lane remains untouched by this change.

I did not launch another model-heavy generation run while an unrelated MLX training workload was active on the M5 GPU, so the M5 28/28 generation result remains mabaeyens evidence rather than a new independent run from me.

For security triage: this test-only PR is not being reported as a cybersecurity vulnerability. The evidence is scoped to numerical precision and test-determinism behavior on M5. Within the tested behavior, there is no evidence of cross-process disclosure, arbitrary code execution, sandbox escape, or another cybersecurity-boundary failure.

zcbenz:

PR #1596, opened 2026-07-21
Prompt cache trimming for recurrent/hybrid and sliding-window models via prefill-boundary state checkpoints

pierre427:
### Problem

`can_trim_prompt_cache()` returns `False` for every model whose `make_cache` mixes `ArraysCache` (recurrent state) with KV caches, because recurrent state cannot be trimmed backward — and likewise for a `RotatingKVCache` once the ring has wrapped. As a result the server prompt cache (`LRUPromptCache.fetch_nearest_cache`) can never reuse a longer stored entry for these models: the common "regenerate" and "edited message" request shapes re-prefill the entire shared prefix.

That covers a large and growing class: `qwen3_next`, `kimi_linear`, `falcon_h1`, `jamba`, `lfm2`/`lfm2_moe`, `nemotron_h`, `mamba`/`mamba2`, `plamo2`, `rwkv7`, `granitemoehybrid`, `bailing_moe_linear`, `baichuan_m1`, `qwen3_5`/`qwen3_5_moe`, plus sliding-window/SWA models past their window.

SGLang recently merged the analogous capability for KDA/mamba prefix caching (sgl-project/sglang#31474) using chunk-boundary state snapshots; this PR brings the same capability to mlx-lm's prompt cache, so prefix caching works for this class — including day-0 support for upcoming hybrid releases.

### Approach

- **`ArraysCache` records bounded state checkpoints at prefill chunk boundaries.** The prefill loops (`generate_step`, speculative `_prefill`, `PromptBatch.prompt`) call `record_state_checkpoints(cache, positions)` after each chunk's `mx.eval`, and force-record the end-of-prefill boundary — which is exactly `len(prompt) - 1`, the landing position of a regenerate-style trim, so the most common trim is exact.
- **`RotatingKVCache` records temporal-order window copies at the same boundaries** — bounded by the window size, not the context length — so a wrapped ring can land a trim on a recorded boundary instead of being excluded from reuse. Unwrapped rings keep their native exact trim. (The trim coordinator anchors on absolute positions, since `RotatingKVCache.size()` saturates at `max_size`.)
- **`trim_prompt_cache(cache, n, allow_partial=True)`** finds the newest position `<= size - n` that every cache in the list can restore (`snap_trim_position` fixpoint), restores checkpointed layers, trims exactly-trimmable layers by the same amount, and returns the **actual** trimmed count (`>= n`). Position 0 (empty state) is always restorable, so the worst case degrades to a full re-prefill — never a wrong answer. `allow_partial` defaults to `False`: speculative-decoding rewinds and every existing caller are unchanged.
- **`fetch_nearest_cache`** dry-runs the landing (`achievable_trim`), prefers a longer entry only when it lands deeper than the best exact-prefix entry, and computes the un-cached suffix from the actual landing.
- **Per-lane checkpoint storage.** `ArraysCache` checkpoints are `(position, batch-size-1 array copies)` per lane, so batch `extend`/`filter`/`merge`/`extract` carry histories across lane membership changes in the batch server with plain list operations (`mx.array` of a slice is a true copy — snapshots don't pin batch buffers).
- **Bounded memory, accounted.** `MLX_LM_STATE_CHECKPOINT_STRIDE` (default 2048 tokens) and `MLX_LM_STATE_CHECKPOINT_MAX` (default 4 per lane; `0` disables). Thinning drops the checkpoint with the smallest predecessor gap (never the newest). Checkpoint bytes count toward `nbytes`, so `--prompt-cache-bytes` limits include them. For scale: a Qwen3-Next-80B state checkpoint is ~77 MB (context-independent), so the default cap adds at most ~310 MB per cached entry on the largest models, proportionally less on smaller ones; rotating-window copies are bounded by the window plus at most one prefill chunk.

### Correctness

- `tests/test_state_checkpoint_trim.py` (13 tests): coordinated-trim landings, opt-in gating, cap/thinning, env disable, deepcopy, batch-lane carry through `extend`/`filter`/`merge`/`extract`, sliding-window restore (exact window equality after a wrapped-ring landing, reset semantics, unwrapped native trim preserved), and end-to-end `LRUPromptCache` reuse with decode-consistency (greedy token match + logits allclose) on tiny random-weight `qwen3_next` and `kimi_linear`.
- Validated on real models (M-series, 4-bit `Kimi-Linear-48B-A3B-Instruct` and `Qwen3-Next-80B-A3B-Instruct`): trimmed-and-reprocessed decode is **bit-exact** vs a fresh prefill with shared chunk boundaries (max |Δlogits| = 0.0, greedy-32 identical) for both the regenerate and edited-turn scenarios. Regenerating a 3k-token conversation drops from 2.0–2.5s (fresh prefill) to 0.3–0.6s (reuse).
- `tests/test_prompt_cache.py`, `tests/test_server.py`, `tests/test_generate.py` pass unchanged.

### Supersedes #1515, #1516, #1517

The anchor-stride series (#1515–#1517) targets the same problem by deep-copying the entire prompt cache at stride positions and storing the copies as extra shorter-prefix LRU entries. This PR reaches the same landings at strictly lower cost, so I'd like to close that series in favor of this one:

- **Memory per landing point**: anchors copy the whole cache (all KV + state, growing with context) and multiply LRU entries; checkpoints copy only what cannot be re-derived — recurrent state (context-independent) and the bounded rotating window — inside the existing entry. Exactly-trimmable KV layers need no copies at all.
- **Batch path**: native here (per-lane histories survive lane churn), vs the separate capture machinery in #1517.
- **Coverage**: identical, including wrapped sliding-window caches.

### Related open PRs

- #1502 adds reuse-correctness guards in the same region (`trim_prompt_cache` return accounting, `fetch_nearest_cache` refusing inexact trims). The landing-aware fetch here enforces the same invariant — never serve KV inconsistent with the keyed prefix — while additionally allowing checkpoint landings with the suffix recomputed from the actual trim. Happy to rebase whichever lands second.
- #1486 (rollback cache API) touches `ArraysCache.__new__` — textual-only overlap, the features are complementary and coexist cleanly; same offer.

### Not included (possible follow-up)

Checkpoints are in-memory only: `save_prompt_cache`/`load_prompt_cache` drop them, so trims on caches loaded from disk land at 0. Persisting them (a versioned state layout) works — we run it in our fork — but it changes the saved-cache format, so it's split out.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -28,6 +28,7 @@ from .models.cache import (
     RotatingKVCache,
     TokenBuffer,
     load_prompt_cache,
+    record_state_checkpoints,
 )
 from .sample_utils import make_sampler
 from .tokenizer_utils import TokenizerWrapper
@@ -417,6 +418,9 @@ def generate_step(
             len(input_embeddings) if input_embeddings is not None else len(prompt)
         )
         prompt_processed_tokens = 0
+        checkpoint_base = max(
+            (c.size() for c in prompt_cache if hasattr(c, "size")), default=0
+        )
         prompt_progress_callback(prompt_processed_tokens, total_prompt_tokens)
         while total_prompt_tokens - prompt_processed_tokens > 1:
             remaining = (total_prompt_tokens - prompt_processed_tokens) - 1
@@ -432,6 +436,9 @@ def generate_step(
             quantize_cache_fn(prompt_cache)
             mx.eval([c.state for c in prompt_cache])
             prompt_processed_tokens += n_to_process
+            record_state_checkpoints(
+                prompt_cache, [checkpoint_base + prompt_processed_tokens]
+            )
             prompt_progress_callback(prompt_processed_tokens, total_prompt_tokens)
             prompt = prompt[n_to_process:]
             input_embeddings = (
@@ -440,6 +447,15 @@ def generate_step(
                 else input_embeddings
             )
             mx.clear_cache()
+
+        # The end-of-prefill boundary is the position a regenerate-style
+        # prompt-cache trim lands on, so always record it.
+        if prompt_processed_tokens > 0:
+            record_state_checkpoints(
+                prompt_cache,
+                [checkpoint_base + prompt_processed_tokens],
+                force=True,
+            )
 
         y, logprobs = _step(input_tokens=prompt, input_embeddings=input_embeddings)
 
@@ -568,13 +584,19 @@ def speculative_generate_step(
                 return _process_and_sample(None, logits.squeeze(0))
 
     def _prefill(model, cache, y):
+        base = max((c.size() for c in cache if hasattr(c, "size")), default=0)
+        processed = 0
         while y.size > 1:
             n_to_process = min(prefill_step_size, y.size - 1)
             model(y[:n_to_process][None], cache=cache)
             quantize_cache_fn(cache)
             mx.eval([c.state for c in cache])
+            processed += n_to_process
+            record_state_checkpoints(cache, [base + processed])
             y = y[n_to_process:]
             mx.clear_cache()
+        if processed > 0:
+            record_state_checkpoints(cache, [base + processed], force=True)
         return y
 
     def _rewind_cache(num_draft, num_accept):
@@ -1229,6 +1251,12 @@ class PromptProcessingBatch:
         padding = [max_length - l for l in lengths]
         max_padding = max(padding)
 
+        # Absolute per-lane positions for recurrent-state checkpoints. A lane
+        # that exhausts its (right-padded) prompt mid-chunk holds a frozen,
+        # exact state at its own total, so clamp per lane.
+        totals = [len(st) for st in self.tokens]
+        bases = [t - l for t, l in zip(totals, lengths)]
+
         # Prepare the caches and inputs. Right pad if needed otherwise just
         # cast to array.
         if max_padding > 0:
@@ -1239,10 +1267,16 @@ class PromptProcessingBatch:
             tokens = mx.array(tokens)
 
         # Actual prompt processing loop
+        processed = 0
         while tokens.shape[1] > 0:
             n_to_process = min(self.prefill_step_size, tokens.shape[1])
             self.model(tokens[:, :n_to_process], cache=self.prompt_cache)
             mx.eval([c.state for c in self.prompt_cache])
+            processed += n_to_process
+            record_state_checkpoints(
+                self.prompt_cache,
+                [b + min(processed, l) for b, l in zip(bases, lengths)],
+            )
             mx.clear_cache()
             tokens = tokens[:, n_to_process:]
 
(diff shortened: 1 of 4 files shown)


pierre427:
Added two tests (tests-only commit, no functional change): a **state-level checkpoint audit** and a **token-exact through-decode round trip**.

Context on why these two specifically: checkpoint-based prefix reuse for recurrent state is converging across engines right now (SGLang's `extra_buffer` track snapshots, vLLM's `mamba_cache_mode="all"` block-aligned saves, llama.cpp's in-memory recurrent checkpoints), and the CUDA-side implementations have shipped several *silent* wrong-state/wrong-position restore bugs in the last month — e.g. sgl-project/sglang#31474 (garbage donated states, GSM8K 0.150→0.895 after fix), sgl-project/sglang#31833 (misaligned requests restoring states with 100%+ error **that two full benchmark evals failed to detect**), and sgl-project/sglang#31850 (generated prefix stored with state from a different sequence position). The common thread: output-level and eval-level checks don't catch a wrong-state restore.

So:

- `checkpoint_state_audit_*` compares the **full cache state tree bit-exactly** at every recorded landing against a fresh prefill of the same prefix — the direct state-level check that would have caught all three of the above classes.
- `generation_roundtrip_token_exact_*` exercises the riskiest serving path (state committed during decode, stored with a forced end checkpoint, reused by a follow-up request, re-extended) and requires the 32-token greedy continuation to be **token-exact** against an identical-computation reference arm with no store/fetch/trim. Deterministic by construction (chunk-aligned landings, same-path replay), so it can't flake in CI.

Both run on the qwen3_next and kimi_linear tiny configs; 17/17 here plus `test_prompt_cache` 22/22 green locally.

pierre427:
One more addition (2f0794e): **checkpoints now persist through `save_prompt_cache`/`load_prompt_cache`**.

Auditing this branch against llama.cpp's current recurrent-checkpoint persistence bug (ggml-org/llama.cpp#25913 — restore reports success, prompt reuse silently dies) turned up the same latent gap here: checkpoints lived only in memory, so a saved-and-restored hybrid could only land at position 0 (a wrapped `RotatingKVCache` drags the whole hybrid's landing there via the trim fixpoint). Empirically: the same cache that lands at 96 in memory landed at 0 after a save/load round trip.

The fix is strictly additive — with no checkpoints recorded, the on-disk layout is unchanged and legacy cache files load exactly as before (covered by a dedicated test). The round-trip test asserts a parity contract: restore-from-disk-then-trim equals pure-in-memory-trim over the full state tree of every cache, rotating window included.

pierre427:
One more (d880243): auditing this branch against the exact-hit empty-remainder bug that #1581 addresses turned up the same edge here — `fetch_nearest_cache` with a request exactly equal to a stored key returned `rest=[]`, and callers re-process `rest` for last-token logits.

This fixes it at the cache layer with the same contract as #1581's server-side guard for trimmable caches (hand back the last token), plus the case that guard can't recover: a checkpointed hybrid lands on its deepest interior checkpoint instead of recomputing from scratch. The two changes compose — if #1581 lands, its guard simply never triggers. test_server's exact-hit expectations are updated to the new contract in the same commit.

zcbenz:

Issue #1493, opened 2026-07-07
mlx_lm.server: generation hangs at 0% CPU right after prompt processing on ~22-26k-token streaming requests (gemma-4-26b-a4b, 0.31.3)


TadeuszWolfGang:
### Summary

`mlx_lm.server` hangs (decode never starts, process at 0% CPU) immediately after prompt processing completes, on long (~22–26k token) streaming chat-completion requests from a real client (Obsidian Copilot). Reproduced 2/2 times with real client traffic; synthetic requests of the same size — including concurrent + streaming — complete fine, so the trigger seems to be a specific parameter/content combination rather than prompt length alone.

### Environment

- mlx-lm **0.31.3**, mlx/mlx-metal **0.31.2** (installed via `uv tool install mlx-lm`, `transformers==5.12.1`)
- macOS 26.5 (Darwin 25.5.0), Apple M5 Max, 128 GB
- Python 3.12.13
- Model: `mlx-community/gemma-4-26b-a4b-it-8bit` (model_type `gemma4`), pinned at startup
- Server: `mlx_lm.server --model /path/to/gemma-4-26b-a4b-it-8bit --host 127.0.0.1 --port 8082 --decode-concurrency 32 --prompt-concurrency 8`
- (Environment note: a local one-line `models/gemma4_unified.py` alias exists for a different model, mirroring PR #1386; it is not imported by the affected instance, which serves a stock `gemma4` checkpoint.)

### Failing requests (2/2 hangs)

Client: Obsidian Copilot plugin ("3rd party (openai-format)" provider). Request shape: `stream: true`, `temperature: 0.1`, `max_tokens: 16000`, system prompt + long mixed Polish/English markdown context.

- Hang 1: prompt 22,258 tokens — prefill completes in ~8 s, then nothing.
- Hang 2 (after server restart): prompt 25,660 tokens — prefill completes in ~12 s, then nothing.

In both cases the log shows prompt processing reaching N/N and then goes silent; no traceback, no completion line. The server process stays alive but at **0.0% CPU** (checked minutes later — 5–18 min). Client eventually gives up.

```
2026-07-05 08:19:39,913 - INFO - Prompt processing progress: 23032/25660
2026-07-05 08:19:40,840 - INFO - Prompt processing progress: 25080/25660
2026-07-05 08:19:41,170 - INFO - Prompt processing progress: 25659/25660
2026-07-05 08:19:41,208 - INFO - Prompt processing progress: 25660/25660
[...]

PR #1598, opened 2026-07-21
Detect and recover from livelocked batch generation (#1493)

rajanshxrma:
Fixes the livelock failure mode reported in #1493, as a stacked draft on top of #1513 per @TadeuszWolfGang's [suggestion](https://github.com/ml-explore/mlx-lm/issues/1493#issuecomment-5039365743) — keeps #1513 focused on exceptions/dead-worker recovery while this handles delivery-stall detection independently. Happy to fold into #1513 instead if maintainers prefer.

**Note on the diff:** `fix-server-generation-thread` (#1513) only exists on my fork, not upstream, so GitHub won't let me set it as this PR's base — this PR is opened against `main` instead, and the diff below includes #1513's changes until #1513 merges (the `_reset_batch()` refactor here only makes sense on top of it). Once #1513 lands, this diff will shrink to just the new commit — [rajanshxrma/mlx-lm@6e3f367](https://github.com/rajanshxrma/mlx-lm/commit/6e3f367c4b4b6693ef1006d2b9fc23dc4695842b) — which is this PR's actual content.

## The problem

#1513 fixes the case where the generation thread dies (uncaught exception) and the server zombies. But @TadeuszWolfGang [caught a third failure mode live](https://github.com/ml-explore/mlx-lm/issues/1493#issuecomment-4999212001) with py-spy stack captures: a **livelock**. The batch loop keeps calling `BatchGenerator.next()` and doing real work (forward pass, `mx.async_eval` sync) — thread `is_alive()` stays `True`, real CPU/GPU burn — but nothing is ever delivered to any request's response queue. Requests in flight (and new ones) hang indefinitely; only a process restart recovers.

This defeats both `_generation_thread.is_alive()` (#1513 as written) and a naive per-iteration heartbeat, since the loop genuinely keeps iterating the whole time.

*(Scope note added 2026-07-29, after @TadeuszWolfGang's review of the archived incident log: the fix below detects a generator-wide delivery drought, meaning nothing delivered to any in-flight request for `--generation-stall-timeout` seconds. The archived #1493 capture itself is narrower than that. One decode request stayed wedged while prompt-progress for other, unrelated requests kept being delivered throughout the hang. Because that unrelated progress also counts as delivery, a watchdog reset on that exact captured incident would not have fired until roughly 3.5 minutes after the wedge started, with no fixed upper bound under regular long-prompt traffic. See the PR discussion and the scope comment on `last_delivery` in `server.py` for the full detail.)*

## The fix

Track wall-clock time since a token, prompt-progress update, or admission ack was last actually put on a consumer's queue. If requests are in flight and nothing has been delivered for `--generation-stall-timeout` seconds (default 60.0, `0` disables), treat the batch as stalled: fail the in-flight requests with an explicit `RuntimeError`, close and drop the batch generator, and let the loop start fresh — the exact recovery shape #1513 already uses for an uncaught exception, just gated on a delivery-staleness timer instead of (or in addition to) an exception. Refactored the shared reset logic into `_reset_batch()` so both paths use it.

Default of 60s is chosen to be well clear of legitimate long first-decode-step latency on large models, while staying far under the 5-minute external-probe workaround @IzD680 deployed as a production stopgap.

## Test

`test_generation_thread_survives_delivery_stall` in `tests/test_server.py`, mirroring #1513's `test_generation_thread_survives_batch_exception`: mocks `BatchGenerator.next` to return `([], [])` forever (simulating the livelock) with a tiny stall timeout, asserts the in-flight request raises to the client, the generation thread survives, and a subsequent request completes normally with no process restart.

## Known limitation (out of scope here)

This is a safety net, not a root-cause fix. @TadeuszWolfGang's hypothesis for *why* the livelock happens — a poisoned prompt-cache/position entry from a >22k cache-primed prompt interacting with gemma4's sliding-window layers, likely #1494 territory — is unresolved. The watchdog stops the hang and keeps the server alive, but if the same poisoned cache entry gets reused, the livelock could recur. Worth a follow-up once #1494 is better understood.

**Scope limitation surfaced 2026-07-29:** the shared `last_delivery` timestamp is refreshed by admission acks, prompt-progress, and tokens from any uid in the batch (intentional — see the `server.py` scope comment and `test_stall_watchdog_is_generator_wide_not_per_request`). @TadeuszWolfGang's re-examination of the archived #1493 log shows this isn't just a theoretical corner case: in the captured incident, unrelated prompt-progress delivery continued while one decode request was wedged, which would have delayed this watchdog's detection of that exact incident by roughly 3.5 minutes past the configured 60s timeout, with no fixed bound under regular traffic. Two directions have been proposed to close this: tracking decode-delivery staleness separately from admission/prompt-progress signals, or keeping per-uid timestamps/phase state once a request finishes prefill. Neither is implemented yet. @TadeuszWolfGang is running a long soak test of this exact head with a periodic liveness probe, which will produce empirical evidence on whether this gap matters in practice; a redesign before that data lands would be premature. Docs now state the actual scope precisely; implementation is follow-up work pending the soak results.

@TadeuszWolfGang — this is armed and ready for you to validate against your Gemma reproducer whenever you get a chance.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1147,12 +1147,12 @@ class PromptProcessingBatch:
         if not any(self.samplers):
             self.samplers = [None] * len(self.uids)
         if not any(self.logits_processors):
-            self.logits_processors = [None] * len(self.uids)
+            self.logits_processors = [[]] * len(self.uids)
         samplers = batch.samplers if any(batch.samplers) else [None] * len(batch.uids)
         logits_processors = (
             batch.logits_processors
             if any(batch.logits_processors)
-            else [None] * len(batch.uids)
+            else [[]] * len(batch.uids)
         )
 
         self.uids.extend(batch.uids)
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -674,50 +674,51 @@ class ResponseGenerator:
                         prompt, segments, segment_types, initial_state = self._tokenize(
                             current_tokenizer, request, args
                         )
+
+                        stop_matcher, text_sm = self._make_state_machine(
+                            self.model_provider.model_key,
+                            tokenizer,
+                            args.stop_words,
+                        )
+
+                        self._log_cache_stats()
+                        cache, rest = self.prompt_cache.fetch_nearest_cache(
+                            current_model_key, prompt
+                        )
+                        prompt_cache_count = len(prompt) - len(rest)
+                        N = prompt_cache_count
+                        while N > 0:
+                            if N >= len(segments[0]):
+                                N -= len(segments.pop(0))
+                                segment_types.pop(0)
+                            else:
+                                segments[0] = segments[0][N:]
+                                break
+
+                        ctx = GenerationContext(
+                            has_tool_calling=tokenizer.has_tool_calling,
+                            has_thinking=tokenizer.has_thinking,
+                            tool_parser=tokenizer.tool_parser,
+                            text_sm=text_sm,
+                            initial_state=initial_state,
+                            prompt=prompt,
+                            prompt_cache_count=prompt_cache_count,
+                        )
+
+                        (uid,) = batch_generator.insert_segments(
+                            segments=[segments],
+                            max_tokens=[args.max_tokens],
+                            caches=[cache],
+                            all_tokens=[prompt[:prompt_cache_count]],
+                            samplers=[_make_sampler(args, tokenizer)],
+                            logits_processors=[_make_logits_processors(args)],
+                            stop_matchers=[stop_matcher],
+                        )
                     except Exception as e:
                         rqueue.put(e)
                         continue
 
-                    stop_matcher, text_sm = self._make_state_machine(
-                        self.model_provider.model_key,
-                        tokenizer,
-                        args.stop_words,
-                    )
-
-                    self._log_cache_stats()
-                    cache, rest = self.prompt_cache.fetch_nearest_cache(
-                        current_model_key, prompt
-                    )
-                    prompt_cache_count = len(prompt) - len(rest)
-                    N = prompt_cache_count
-                    while N > 0:
-                        if N >= len(segments[0]):
-                            N -= len(segments.pop(0))
-                            segment_types.pop(0)
-                        else:
-                            segments[0] = segments[0][N:]
-                            break
-
-                    ctx = GenerationContext(
-                        has_tool_calling=tokenizer.has_tool_calling,
-                        has_thinking=tokenizer.has_thinking,
-                        tool_parser=tokenizer.tool_parser,
-                        text_sm=text_sm,
-                        initial_state=initial_state,
-                        prompt=prompt,
-                        prompt_cache_count=prompt_cache_count,
-                    )
                     rqueue.put(ctx)
-
-                    (uid,) = batch_generator.insert_segments(
-                        segments=[segments],
-                        max_tokens=[args.max_tokens],
-                        caches=[cache],
-                        all_tokens=[prompt[:prompt_cache_count]],
-                        samplers=[_make_sampler(args, tokenizer)],
-                        logits_processors=[_make_logits_processors(args)],
-                        stop_matchers=[stop_matcher],
-                    )
                     batch_results[uid] = {
                         "ctx": ctx,
                         "rqueue": rqueue,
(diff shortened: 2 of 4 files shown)


TadeuszWolfGang:
I tested exact head `6e3f367c4b4b6693ef1006d2b9fc23dc4695842b`
against the Gemma/Obsidian Vault QA workflow from #1493.

First, a correction to my initial local analysis: the archived July 17 capture
shows an iterative livelock, not a permanently blocked `BatchGenerator.next()`.
The same generation thread was at `GenerationBatch._step()` line 1369
(`mx.async_eval`) at T+0/T+60 and at the earlier forward call on line 1332 six
minutes later. In mlx-lm 0.31.3, `GenerationBatch.next()` invokes one `_step()`,
and the server log also contains prompt-progress updates produced only after
`BatchGenerator.next()` returned. The generation worker therefore regained
control from `next()` repeatedly during at least part of the hang.

## Real-client replay

The historical HTTP body is not recoverable because Vault QA injects dynamic
source context, so this was a workflow-level replay with the same client
version and vault/index, not a byte-identical request. The small-request/cache
priming shape was present, but the replay's large prompts were generally
larger: the three fresh-chat main prefills were 29,571, 29,915, and 26,565
tokens, versus 22,258, 25,660, and 23,404 in the captured incidents.

The reconstructed two-turn warm-up plus fresh-chat trigger completed, followed
by two additional fresh-chat runs. All five turns completed. The three fresh
runs finished in <=73.7 s, 105.3 s, and 155.2 s. There was no watchdog warning,
exception, or PID change, and two post-run control completions returned HTTP
200.

The real failure did not reproduce, so this live run is a no-regression result;
it did not exercise or validate recovery.

## Automated checks

- Full `tests/test_server.py`: 27/27 passed.
- `test_generation_thread_survives_delivery_stall` passed repeatedly.
- The combined exception/stall resilience tests passed.
- A separate blocking-call probe set `generation_stall_timeout=0.05` and held
  one `BatchGenerator.next()` call for 0.25 s. At 5x the timeout the consumer
  was still waiting; the watchdog ran only after `next()` was released.

That last probe is an additional implementation boundary, not a reproduction
of the captured #1493 livelock. The historical captures show that `next()`
returned during the incident.

## Implementation questions

1. `last_delivery` is one generator-wide timestamp. Admission acknowledgements,
   prompt-progress updates, tokens, and completion markers from any UID refresh
   it. This means the watchdog bounds an engine-wide delivery stall, not an
   individual request stall; continued admissions or progress from another UID
   can postpone recovery for a wedged request. Is that scope intentional?
2. The timestamp records `rqueue.put(...)`, not the later SSE/socket write.
   That is a valid internal-consumer signal, but it does not establish that the
   client received a chunk.
3. The timeout is checked before, not during, `BatchGenerator.next()`, so it
   cannot interrupt a call that never returns. The blocking-call probe above
   confirms this boundary, although that is not what the archived #1493
   snapshots demonstrate.
4. `_reset_batch()` calls `BatchGenerator.close()`, which synchronizes the
   generation stream when `_old_wired_limit is not None` (the Metal path on
   this host). If that stream itself is wedged, recovery could potentially
   block there. I did not reproduce this path, so this is a risk to test rather
   than an observed failure.

Could you add or document coverage for the generator-wide timestamp semantics
and the blocking-`next()` boundary? Based on the current evidence, the watchdog
is compatible with the captured iterative livelock and its mocked recovery path
works, but recovery from the real #1493 incident remains unvalidated because
the live failure did not recur.

rajanshxrma:
Thanks for the careful read and for the blocking-call probe -- that's exactly the boundary I wanted to nail down. Went through the current implementation (head `433b26d`, on top of the doc/test commit I just pushed) against all four points. Answers below in order, with what's now documented vs. tested vs. an open, unreproduced risk.

**1. Generator-wide vs. per-request scope -- yes, intentional.**

`last_delivery` is one timestamp for the whole generator (`server.py`), refreshed by admission acks, prompt-progress, tokens, and completion markers from any uid. Confirmed in the code: it's set on batch admission, on prompt-progress delivery, on token delivery, and on the completion sentinel -- all writes to the same `nonlocal last_delivery`, keyed by nothing. So yes: continued admissions or progress on other requests postpone recovery for a single wedged request in the same batch. This is a real scope limit, not an oversight -- the design target is bounding an engine-wide livelock (the batch loop makes no forward progress for *anyone*, the #1493 failure mode), not a per-request delivery SLA. A per-request stall inside an otherwise-healthy batch is a different problem and this watchdog doesn't claim to cover it.

I added a docstring-level comment at the `last_delivery` declaration spelling this out, plus a regression test (`test_stall_watchdog_is_generator_wide_not_per_request`) that drives two in-flight requests where only one keeps receiving mocked deliveries and confirms the other gets neither a token nor a stall-triggered reset until the *shared* timestamp actually goes stale -- i.e. it directly exercises the scope you're describing rather than just asserting it in prose.

**2. Queue-put vs. actual client delivery -- correct, and now stated precisely rather than loosely.**

`last_delivery` records `rqueue.put(...)`, i.e. hand-off to this process's internal per-request queue (consumed by the HTTP handler thread for that request), not the later SSE/socket write. That's the right signal for detecting a stalled *generation engine* -- it's the boundary this thread can actually observe -- but it doesn't by itself prove the client received anything; the HTTP layer downstream is a separate failure domain (e.g. a stuck socket write) that this watchdog has no visibility into. I tightened the comment at the top of `_generate` to say this explicitly instead of the vaguer "handed to a consumer queue" phrasing. No code change here -- this is a scope statement, not a bug, and I don't think it should try to reach further downstream without conflating two different failure modes.

**3. Checked before, not during, `next()` -- confirmed, and now pinned down with a test rather than just a comment.**

The stall check runs on the generation thread itself, immediately before entering the inner loop that calls `BatchGenerator.next()`, and again on the next trip through the outer `while` once `next()` returns. There's no separate supervisor thread -- it's a self-check by the same thread doing the work. So it bounds an *iterating* livelock (next() keeps returning, just without ever delivering anything -- the actual archived #1493 stack dumps) but cannot recover from a single `next()` call that blocks and never returns; the generation thread is inside that call and literally cannot run the check again until it comes back. Your probe demonstrates this correctly and it's not a testing artifact.

Added `test_stall_watchdog_cannot_interrupt_a_blocking_next_call`, which productionizes that probe: holds a single mocked `next()` call open for several multiples of the stall timeout, confirms the consumer stays blocked for the *entire* duration (not just "eventually" -- I assert emptiness partway through, while the call is still in flight), and confirms recovery only fires once the call actually releases. I also added a code comment at the check site stating this boundary explicitly, so it's not just implicit in the control flow for the next reader.

I want to be precise about what this does *not* claim: it's compatible with the captured #1493 evidence (an iterating livelock), and the new test proves the boundary you found is real and stable, not flaky. It does not make the watchdog able to interrupt a genuinely hung native call -- that would need a separate supervisor thread and a way to safely abandon or recover from a stuck call, which is materially more invasive and isn't something I want to bolt on without more design discussion (thread-safety of abandoning an in-flight MLX stream operation is not obviously safe).

**4. `close()` synchronizing a possibly-wedged stream -- real risk, not reproduced, now documented rather than silently left implicit.**

Confirmed: `BatchGenerator.close()` calls `mx.synchronize(self._stream)` when `_old_wired_limit is not None`, i.e. whenever `mx.metal.is_available()` was true at construction (the Metal path on essentially every Mac). If the generation stream itself is what's wedged, that synchronize call could block on the same thread with no timeout, which would turn "detected and reset the batch" into "the generation thread is now stuck in recovery instead" -- exactly your concern. I have not reproduced this and agree with your framing: it's a risk to test, not an observed failure. The reproduced #1493 case is specifically an *iterating* livelock (the stream keeps completing steps, just without ever delivering output), which is a different state from a wedged stream that never completes at all.

One thing worth noting since I went looking at `close()` closely: I did hit a *related but distinct* issue while running the test suite -- `BatchGenerator.__del__` (which also calls `close()`) intermittently throws `RuntimeError: There is no Stream(gpu, N) in current thread` when garbage collection finalizes an abandoned `BatchGenerator` on a thread other than the one that owned its stream. That's a thread-locality issue at GC time, not a hang, and it reproduces on unmodified `main` too (verified before and after my changes, same traceback) -- so it's pre-existing and out of scope for this PR, but it's more evidence that `close()`'s stream-synchronization logic is the fragile part of this path in general.

I did not attempt a fix for point 4: bounding the `close()` call with a timeout (e.g. running it in a background thread and abandoning it on timeout) introduces a real race on `mx.set_wired_limit` between the abandoned close and any newly-created `BatchGenerator`, and I don't think it's safe to land without dedicated design and testing of its own. I ordered `_reset_batch` so in-flight requests are failed *before* `close()` is attempted (this was already the case, and I added a comment noting why the ordering matters) -- so even if `close()` did hang, the requests we already know about aren't waiting on it. What isn't mitigated: if `close()` hangs, the generation thread never returns from `_reset_batch`, so all *future* requests deadlock too. That's the actual residual risk and I'm not aware of a safe fix within this PR's scope; flagging it as a known limitation rather than pretending it's covered.

**Test status:** `tests/test_server.py` -- 29/29 passed (27 existing + the 2 new ones above). `tests/test_generate.py` -- 27/27 passed (no changes there, ran it anyway since it's the adjacent generation-path suite). Ran both before and after the doc/test commit to confirm no regressions.

**Net:** #1 and #2 are documented scope clarifications (no behavior change, docstring + regression test for #1, docstring only for #2 since there's nothing to assert in a test beyond restating the code). #3 is a documented, tested boundary (no behavior change -- fixing it would need a supervisor-thread architecture, which I think is out of scope here). #4 is an accepted, unreproduced limitation, documented in code with reasoning for why I didn't attempt a fix. Pushed as a separate commit (`433b26d`) on this branch so the diff is easy to review on its own.

TadeuszWolfGang:
Thanks — I reviewed and ran the changes at exact head
`433b26dafa85583dea9e5d6621d80be2485b8797`.

The two new tests encode the stated boundaries correctly:

- `test_stall_watchdog_is_generator_wide_not_per_request` keeps request B
  without a token or error for three timeout intervals while request A refreshes
  the shared timestamp, then observes recovery only after A also stops
  delivering;
- `test_stall_watchdog_cannot_interrupt_a_blocking_next_call` checks that the
  response queue is still empty while the mocked `next()` call remains blocked
  and observes recovery only after it returns.

Both tests passed individually, `git diff --check` passed, and the full
`tests/test_server.py` suite passed 29/29.

On `tests/test_generate.py` I see 8 failures on this host, but they are
unrelated to this PR: the identical 8 reproduce on upstream `main`
(`cf10f96`, 26 tests / 8 failures). Two of them
(`test_stream_generate_input_embeddings` and `..._prefill`) pass when run
in isolation and only fail in the full-suite run, on both trees, so at
least part of this looks like cross-test state rather than a code defect.
Environment: M5 Max, macOS 26.5.2, Python 3.12.13, mlx 0.31.2 (the
declared `MIN_MLX_VERSION`). Nothing here contradicts your 27/27 — I just
can't use that suite as a signal on this machine, so I'm limiting my
independent confirmation to `tests/test_server.py`.

There is still one material mismatch between the generator-wide scope and the
archived #1493 incident. The July 17 log does not merely show that `next()`
returned between stack captures. It shows deliveries for other work during the
hang:

```text
06:55:15      Prompt processing progress: 23404/23404
06:57:13.710  Prompt processing progress: 263/264
06:57:13.736  Prompt processing progress: 61/61
06:57:31 ... 06:57:47.947
               Prompt processing progress: 2048/36480 ... 36480/36480
```

Those prompt-progress lines are emitted by the HTTP-thread keepalive callback
after the generation thread has put the progress response on the request queue.
Thus, while the original decoding request remained wedged, the same server
accepted other requests and completed a 36,480-token prefill.

Because prompt progress from any UID refreshes `last_delivery`, the current
watchdog uses prefill progress to reset the timer intended to detect a decoding
delivery stall. In this captured incident, a 60-second watchdog could not have
fired until approximately 06:58:48 — about 3.5 minutes after the original
wedge. With regular long-prompt traffic, detection can be postponed without a
fixed upper bound.

This means the following four descriptions are too broad:

1. PR description, “The problem”: “nothing is ever delivered to any request's
   response queue”;
2. the pre-existing `last_delivery` comment: “anything to any request”;
3. the new scope comment: “the batch loop itself makes no forward progress for
   anyone -- the #1493 failure mode”;
4. the `close()` comment: “the stream keeps completing steps, just without
   delivering output.”

Could those places be updated to distinguish a generator-wide delivery drought
from the documented #1493 case, where one decoding request remained wedged
while prompt-progress delivery for other UIDs continued?

For the implementation, two possible directions seem worth considering:

- track decode-delivery staleness separately from admission acknowledgements
  and prompt-progress delivery; or
- keep per-UID timestamps/phase state once a request finishes prefill, so
  unrelated prefill progress cannot indefinitely mask a wedged decoding
  request.

I have also started a long-running soak of this exact head on the reviewer
instance with `--generation-stall-timeout 60`. The service is running and the
initial control endpoint returns HTTP 200.

The soak captures both outcomes independently: the server's own
`generation stalled` line, and — separately — a 4-token liveness probe every
5 minutes with a 60s timeout, so a recurrence that hangs *without* the
watchdog firing is also recorded (stacks, sample, server log, probe status).
That is the case this scope question predicts, so it should produce evidence
either way.

rajanshxrma:
Went through this against the actual code at `433b26d` and the archived timeline. All four quotes check out verbatim, and the underlying claim holds: the generator-wide `last_delivery` timestamp does not bound detection of the exact #1493 shape you captured.

**On the four locations.** You're right that all four describe the watchdog's target as if it matches the literal #1493 incident. It doesn't: it covers a stricter case (nothing delivered to anyone, generator-wide) that the archived capture doesn't actually satisfy, since other uids kept getting prompt-progress delivery throughout. Reworded all four:

- `last_delivery`'s scope comment now says the goal is to bound a "generator-wide delivery drought," not "the batch loop makes no forward progress for anyone -- the #1493 failure mode." Added a paragraph directly under it with your July 17 timestamps (06:55:15-06:57:47), stating plainly that a 60s timeout on that exact capture would not have fired until roughly 3.5 minutes after the wedge started, with no fixed upper bound under regular long-prompt traffic.
- The `close()` comment now says other in-flight requests kept receiving prompt-progress throughout the reproduced incident, and only the one wedged decode request's queue went silent, instead of the blanket "without delivering output."
- The check-site comment ("reported as a real-world failure mode in #1493") now says "generator-wide delivery drought" and points back to the fuller scope note instead of repeating the #1493 attribution inline.
- PR description's "The problem" section: same correction, pushed as part of the description update alongside this commit.

**On whether this needs a code change now.** My read: not yet. The underlying scope question, does ongoing delivery to one uid mask a wedged different uid, was already asked and answered intentional in the last round, with `test_stall_watchdog_is_generator_wide_not_per_request` locking it in. What you've shown today isn't a new behavior gap; it's that tying that intentional scope to "the #1493 failure mode" specifically was inaccurate, since the archived incident is an instance of the masking case, not the full-drought case the comments described. That's a documentation-precision problem and I've fixed it as one.

Your two proposed directions, tracking decode-delivery staleness separately from admission/prompt-progress, or per-uid timestamps/phase state once a request exits prefill, are the right shape for an actual fix, and I don't want to wave them off. But either one is real architecture work: new per-uid state, new phase-transition logic, and a meaningfully larger test matrix. That's comparable in scope to the supervisor-thread redesign we already deferred for the blocking-`next()` boundary, and I think the same reasoning applies here: it's a genuine design decision, not a same-day patch, and building it now, before your soak test produces real evidence on whether this gap bites in practice at these timeout values and traffic patterns, risks solving the wrong-shaped problem. If the soak comes back showing an unbounded hang that the watchdog missed, that's the concrete trigger to implement (a) or (b) for real. Happy to pick it up at that point.

What I did add: a new regression test, `test_stall_watchdog_prompt_progress_from_other_uid_can_mask_a_wedged_decode`, because the existing test only exercised the masking mechanism with token delivery from the non-wedged uid. Your finding is specifically that prompt-progress delivery has the identical effect, which felt worth locking into the suite the same way the blocking-`next()` boundary got a test last round even though it also wasn't fixed.

**Test status:** `tests/test_server.py` -- 30/30 passed (29 existing + the new one above). `tests/test_generate.py` -- 27/27 passed, unchanged. Combined run -- 57/57, no cross-test interference this time. `git diff --check` clean.

Appreciate the soak test, no rush on this from my side either way. I'll watch for what it turns up.

TadeuszWolfGang:
Soak status on this head, plus a cross-reference that concerns the base of this
stack.

**Doc changes.** I read `f9572d0` against the archived timeline. The reworded
`last_delivery` scope note, the `close()` comment and the check-site comment
match what the July 17 capture shows, and
`test_stall_watchdog_prompt_progress_from_other_uid_can_mask_a_wedged_decode`
drives the prompt-progress path specifically rather than the token path. Nothing
to add from me on those.

**Soak.** Running on `433b26d` with `--generation-stall-timeout 60`. That build
is identical to this head once comments are stripped, so the soak stays valid
for `f9572d0`. What the instance has actually seen since 2026-07-29 12:00:

- 319 prefills of 1 token, 4 prefills of 20 tokens
- 0 prompts of 2,000 tokens or more
- 0 `generation stalled` events emitted by the server
- 2 process restarts — a daily scheduled restart caps continuous uptime at ~24 h;
  current uptime is 10 h

The only traffic reaching that instance is my own 4-token liveness probe every
5 minutes. The single `generation stalled` line in its log is a synthetic
self-test of the capture path, marked as such in the log.

So the soak has not exercised the #1493 trigger at all: no cache-primed
>22k-token Vault QA request has reached it. "No event so far" is therefore not
evidence about the masking gap in either direction, and I am not presenting it
as any. I am routing the real Vault QA workload at that instance next and will
report once it has actually run the trigger shape.

**Cross-reference.** Separately, the unpatched production instance on the same
machine entered the #1505 zombie state today: `RuntimeError: [metal::malloc]
Resource limit (499000) exceeded` out of `_generate`, generation thread dead,
`GET /v1/models` 200 in 0.021 s, a 4-token completion returning nothing in 60 s,
and py-spy showing no `Thread-1 (_generate)` in the process at all. Fourth
occurrence on this host since 2026-07-05, all four with byte-identical
tracebacks. Full capture is on #1513 since that is the PR that addresses it;
noting it here only because this PR stacks on that one.

zcbenz:

PR #1597, opened 2026-07-21
Add Nanbeige (Nanbeige4.2 looped transformer) model support

MercuriusDream:
## Requirements

- ☑️ I understand it is strictly prohibited to use AI to write PR description
- AI usage disclosure: Anthropic Claude Fable 5 and MoonshotAI KIMI K3 did help with this implementation; Manually verified the result and code.

## Model information

- Name: Nanbeige4.2-3B
- Homepage: https://huggingface.co/Nanbeige/Nanbeige4.2-3B
- Reference implementation: https://huggingface.co/Nanbeige/Nanbeige4.2-3B/blob/main/modeling_nanbeige.py

## Supported checkpoints

- https://huggingface.co/MercuriusDream/Nanbeige4.2-3B-mlx-bf16
- https://huggingface.co/MercuriusDream/Nanbeige4.2-3B-mlx-8bit
- https://huggingface.co/MercuriusDream/Nanbeige4.2-3B-mlx-6bit
- https://huggingface.co/MercuriusDream/Nanbeige4.2-3B-mlx-5bit
- https://huggingface.co/MercuriusDream/Nanbeige4.2-3B-mlx-4bit
- https://huggingface.co/MercuriusDream/Nanbeige4.2-3B-mlx-3bit
- https://huggingface.co/MercuriusDream/Nanbeige4.2-3B-mlx-2bit

*Collection: https://huggingface.co/collections/MercuriusDream/nanbeige42-3b-mlx*

## Verification

Run mlx-lm command:

```console
mlx_lm.generate --model MercuriusDream/Nanbeige4.2-3B-mlx-bf16 -p "The secret to baking a good cake is" -m 1024
```

Output:

```
[transformers] You are using a model of type `nanbeige` to instantiate a model of type ``. This may be expected if you are loading a checkpoint that shares a subset of the architecture (e.g., loading a `sam2_video` checkpoint into `Sam2Model`), but is otherwise not supported and can yield errors. Please verify that the checkpoint is compatible with the model you are instantiating.
==========
[...]

PR #1599, opened 2026-07-22
Add Nanbeige (Looped Transformer) model support

SahilChachra:
Nanbeige4.2-3B (and the Nanbeige4.x family) uses a Looped Transformer architecture: the same num_hidden_layers decoder layers run num_loops times, with a full RMSNorm applied between and after each pass to add effective depth without adding parameters. This adds mlx_lm/models/nanbeige.py implementing that loop with per-loop KV cache isolation (each loop's attention only attends to its own loop's keys/values), following the llama.py conventions for the rest of the block (GQA attention, SwiGLU MLP, RMSNorm).

This covers the architectural subset that current Nanbeige checkpoints actually use. The upstream reference (modeling_nanbeige.py) also defines n-gram embeddings, hyper-connections, and depth-attention, all currently unused (num_loops=2 is the only non-default flag Nanbeige4.2-3B sets) and left for a follow-up if a checkpoint ships with them enabled.

Validated against the official `modeling_nanbeige.py` reference implementation (run via `transformers`, float32) on held-out prompts: computed logits from both implementations for the same inputs and compared next-token predictions and full logit distributions. After aligning RoPE frequency initialization between the two implementations, results matched closely — logit cosine
similarity >=0.9999 and identical next-token predictions on every test prompt.

### DIFF
--- a/mlx_lm/models/nanbeige.py
+++ b/mlx_lm/models/nanbeige.py
@@ -1,0 +1,213 @@
+# Copyright © 2026 Apple Inc.
+
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .cache import KVCache
+from .rope_utils import initialize_rope
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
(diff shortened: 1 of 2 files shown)


SahilChachra:
Hi! Just nudging this PR!

zcbenz:
Closing in favor of https://github.com/ml-explore/mlx-lm/pull/1597.

PR #1600, opened 2026-07-22
Fix TypeError in apply_min_p when min_tokens_to_keep > 1

mloiterman:
`apply_min_p`'s `min_tokens_to_keep > 1` branch passes bare Python `False` as the `values` argument of `mx.put_along_axis`, whose binding requires an `mx.array`. The branch raises `TypeError` on every invocation:

```python
>>> from mlx_lm.sample_utils import apply_min_p
>>> import mlx.core as mx
>>> logprobs = mx.log(mx.softmax(mx.random.normal((1, 100)), axis=-1))
>>> apply_min_p(logprobs, 0.05, 3)
TypeError: put_along_axis(): incompatible function arguments. ...
Invoked with types: mlx.core.array, mlx.core.array, bool, kwargs = { axis: int }
```

Verified the rejection is not mlx-version-dependent (probed mlx 0.31.2 and 0.32.0 — bare bool rejected on both). The scalar was introduced in #1083's refactor (the pre-refactor code passed proper arrays); it has stayed latent because `tests/test_sample_utils.py` has no `min_tokens_to_keep` coverage and `mlx_lm.server` doesn't expose the parameter — only direct library consumers of `make_sampler(min_tokens_to_keep=...)` can reach the branch. We hit it in a production server built on `BatchGenerator` the first time a client set the knob.

One-line fix: `False` → `mx.array(False)`. Adds a regression test that fails with the `TypeError` pre-fix and pins the keep-floor semantics (exactly `min_tokens_to_keep` survivors under an aggressive `min_p`), single-row and batched.

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -194,7 +194,7 @@ def apply_min_p(
         tokens_to_remove = mx.put_along_axis(
             tokens_to_remove,
             top_indices,
-            False,
+            mx.array(False),
             axis=-1,
         )
 
--- a/tests/test_sample_utils.py
+++ b/tests/test_sample_utils.py
@@ -68,6 +68,23 @@ class TestSampleUtils(unittest.TestCase):
             actual_probs.tolist(), [[1.0, 0.0, 0.0, 0.0], [0.0, 1.0, 0.0, 0.0]]
         )
 
+    def test_apply_min_p_min_tokens_to_keep(self):
+        # min_tokens_to_keep > 1 must execute (regression: bare Python
+        # False passed to mx.put_along_axis raised TypeError) and must
+        # keep exactly that many candidates under an aggressive filter.
+        probs = mx.array([0.7, 0.2, 0.05, 0.05])[None]
+        logits = mx.log(probs)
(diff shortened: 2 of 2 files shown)


zcbenz:

mloiterman:
Could this one be reopened? It was closed in the capacity sweep rather than on the merits, and it is a hard crash, not a cleanup: any call of `apply_min_p(logprobs, min_p, min_tokens_to_keep)` with `min_tokens_to_keep > 1` raises `TypeError` from `mx.put_along_axis` (the `values` argument is a bare Python `False`; the binding requires an `mx.array`). Still reproducible on current `main`, on mlx 0.32.0 as on 0.31.x — the branch cannot execute at all, so nobody using the keep-floor gets a result.

The fix is one token (`False` → `mx.array(False)`) plus the regression test already in the PR; the test fails on `main` and passes with the change. We have been carrying the same one-line change locally since July with no issues. Happy to rebase if that helps.

zcbenz:
Sure thanks for looking into this, please excuse us there were just too many pull requests and it was easy to mistakenly close valid ones.

zcbenz (approved):
Thanks!

PR #1601, opened 2026-07-22
Add Laguna S 2.1 model support

avlp12:
## Summary

- add the 48-layer Laguna S 2.1 MLX runtime, including heterogeneous attention heads, YARN/default RoPE, per-head softplus gating, correction-bias routing, shared experts, and mixed KV caches
- sanitize 36,096 individual expert tensors into ordered stacked SwitchGLU projections
- exclude routers, norms, correction biases, and other control tensors from quantization

## Validation

- 10 Laguna unit tests pass
- independent tiny PyTorch/MLX parity passes for routing, unbiased weights, attention gating, both RoPE variants, expert ordering, full logits, and cached logits
- maximum observed full/cached logit absolute error: 2.98e-8
- Ruff format/check and git diff check pass

## Downstream release evidence

- end-to-end release tooling and publication validation: **343 tests passed**, plus **161 subtests passed**
- public compact release: [Laguna-S-2.1-Alis-MLX-Dynamic-3.306814bpw](https://huggingface.co/avlp12/Laguna-S-2.1-Alis-MLX-Dynamic-3.306814bpw), measured at **3.306814 bpw**
- public highest-quality release: [Laguna-S-2.1-Alis-MLX-Highest-Quality-4.892631bpw](https://huggingface.co/avlp12/Laguna-S-2.1-Alis-MLX-Highest-Quality-4.892631bpw), measured at **4.892631 bpw**
- the full checkpoint was exercised through **32K context**; **1M context is a calculated support limit only and was not executed**

The PR contains code and weight-free tests only. Full-checkpoint artifacts and quality evidence are published separately in the downstream release repositories above.

### DIFF
--- a/mlx_lm/models/laguna.py
+++ b/mlx_lm/models/laguna.py
@@ -1,0 +1,592 @@
+# Copyright © 2026 Apple Inc.
+
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .activations import swiglu
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .cache import KVCache, RotatingKVCache
+from .rope_utils import initialize_rope
+from .switch_layers import SwitchGLU
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str = "laguna"
+    vocab_size: int = 100352
+    hidden_size: int = 3072
+    intermediate_size: int = 12288
+    num_hidden_layers: int = 48
+    num_attention_heads: int = 48
+    num_key_value_heads: int = 8
+    head_dim: int = 128
+    max_position_embeddings: int = 1048576
+    rms_norm_eps: float = 1e-6
+    hidden_act: str = "silu"
+    attention_bias: bool = False
+    qkv_bias: bool = False
+    attention_dropout: float = 0.0
+    gating: bool | str = "per-head"
+    sliding_window: Optional[int] = 512
+    rope_parameters: Optional[Dict[str, Any]] = None
+    layer_types: Optional[List[str]] = None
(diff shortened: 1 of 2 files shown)


JuanColilla:
Same PR like mine: https://github.com/ml-explore/mlx-lm/pull/1602

Although I see yours was first, which one shall we close?

avlp12:
Thanks for flagging, @JuanColilla — and apologies for the slow reply.

Checking the current state of `main`, I think both of our PRs have been overtaken: **#1334 ("Support for Poolside LagunaXS open source coding model in nvfp4") is already merged**, and it adds `mlx_lm/models/laguna.py` with `model_type = "laguna"`. That is why both #1601 and #1602 now show as CONFLICTING — we are each adding a file that already exists on `main`, for the same `model_type`.

So the question is no longer which of ours to keep, but whether the merged implementation covers **Laguna S 2.1** as well as XS. Both checkpoints declare `model_type: laguna` and differ only in dimensions (`poolside/Laguna-S-2.1`: hidden 3072 / 48 layers / top-10; `poolside/Laguna-XS-2.1`: hidden 2048 / 40 layers / top-8), so a single implementation should serve both if the architecture is identical.

Two things I noticed while comparing my branch against `main` that may be worth a maintainer's eye:

- **Shared-expert attribute name.** The merged file builds `share_expert`, while both checkpoints ship `model.layers.N.mlp.shared_expert.*`. If nothing remaps this, those weights would not load.
- **Router weight path.** The merged file's gate exposes a nested `gate.gate.weight`, while the checkpoints have `model.layers.N.mlp.gate.weight` (plus `mlp.gate.e_score_correction_bias`).

I have not run a full load against a Laguna checkpoint to confirm these are real rather than artifacts of my key comparison, so please treat them as leads rather than bug reports. If the merged implementation does load both models cleanly, I am happy to close this PR as superseded — I would rather not leave a duplicate file sitting in the queue.

PR #1603, opened 2026-07-22
Add native Nanbeige model support

ukint-vs:
## Summary

- Add native `model_type="nanbeige"` support for the released Nanbeige4.2-3B checkpoint.
- Reuse the existing Llama stack for two recurrent passes while keeping 22 physical layers and 44 independent KV-cache entries.
- Split speculative target and draft caches using each model's declared cache layout instead of `len(model.layers)`.
- Reject unsupported vendor architecture modes explicitly rather than loading them with different semantics.

## Why a separate implementation

PRs #1597 and #1599 both implement the core loop and correctly recognize that each pass needs its own cache state. This version differs in a few runtime-level details:

1. It subclasses the existing Llama implementation instead of copying attention, MLP, masking, RoPE, model output, and sharding logic. Nanbeige therefore inherits fixes to those shared paths.
2. It fixes speculative prompt-cache partitioning. Nanbeige exposes 22 physical layers but creates 44 target-cache entries, so splitting at `len(model.layers)` assigns half of the target cache to the draft model.
3. It validates the exact combined cache size and the trimmability of both target and draft caches.
4. It adds the standard model-discovery test plus focused coverage for variable-length batch generation, serialized 44-entry prompt caches, 4-bit and 8-bit weights, 4-bit KV caches, and invalid target/draft cache layouts.
5. It deliberately supports the released two-loop, full-attention checkpoint configuration and names unsupported vendor fields in the load error.

## Validation

The official checkpoint was pinned to revision `5ff54fb7ed86ce8e216d78bff5417ab9981de3d4` and loaded strictly without checkpoint Python.

Float32 vendor parity, evaluated over the full vocabulary at all positions of the fixed prompt:

| Comparison | Max absolute error | Mean absolute error |
| --- | ---: | ---: |
| Vendor vs MLX, full prompt | 0.000141144 | 0.0000143511 |
| Vendor vs MLX, cached | 0.0000257492 | 0.00000285435 |
| Vendor full vs cached | 0.000110626 | 0.0000132345 |
| MLX full vs cached | 0.0000782013 | 0.00000763334 |

All comparisons pass with `rtol=1e-4`, `atol=1e-4`, and vendor and MLX produce the same 32-token greedy continuation.

Fresh BF16, 8-bit group-64, 4-bit group-64, and 4-bit weight plus KV4 generation checks pass. Both weight-only quantized models contain the expected 155 quantized linear modules.

Local focused checks:

```text
69 tests passed:
  tests.test_models.TestModels.test_nanbeige
  tests.test_nanbeige
  tests.test_generate
  tests.test_prompt_cache
  tests.test_utils
  tests.test_evaluate

pre-commit on all changed Python files:
  black 25.1.0: pass
  isort 6.0.0: pass
```

Full `python -m unittest discover tests/`: 219 passed, 1 skipped, and 1 error in `test_datsets.TestDatasets.test_hf`. The same Hugging Face `billsum` URI error reproduces on a clean `origin/main` checkout in this environment.

BF16 does not reproduce the vendor continuation bit for bit. Both PyTorch and MLX show full-versus-cached BF16 drift, while the float32 architecture gate passes. The first BF16 token difference occurs at an exact vendor logit tie; this is recorded as backend numerical behavior rather than hidden by a wider tolerance.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -510,18 +510,31 @@ def speculative_generate_step(
     prev_tokens = None
 
     # Create the KV cache for generation
-    if prompt_cache is None:
-        model_cache = cache.make_prompt_cache(model)
-        draft_cache = cache.make_prompt_cache(draft_model)
-    else:
-        model_cache = prompt_cache[: len(model.layers)]
-        draft_cache = prompt_cache[len(model.layers) :]
+    model_cache = cache.make_prompt_cache(model)
+    draft_cache = cache.make_prompt_cache(draft_model)
+    if prompt_cache is not None:
+        model_cache_size = len(model_cache)
+        expected_cache_size = model_cache_size + len(draft_cache)
+        if len(prompt_cache) != expected_cache_size:
+            raise ValueError(
+                "Speculative decoding requires "
+                f"{expected_cache_size} prompt cache entries, got {len(prompt_cache)}."
+            )
+        model_cache = prompt_cache[:model_cache_size]
+        draft_cache = prompt_cache[model_cache_size:]
 
-    if not cache.can_trim_prompt_cache(model_cache):
-        types = {type(c).__name__ for c in model_cache if not c.is_trimmable()}
-        raise ValueError(
-            f"Speculative decoding requires a trimmable prompt cache " f"(got {types})."
-        )
+    for cache_name, prompt_model_cache in (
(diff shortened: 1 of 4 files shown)


zcbenz:
Closing in favor of https://github.com/ml-explore/mlx-lm/pull/1597.

PR #1608, opened 2026-07-22
Add granitemoe_swa (GraniteMoE + sliding-window attention)

pierre427:
# Add granitemoe_swa (GraniteMoE + sliding-window attention)

Adds `mlx_lm/models/granitemoe_swa.py` for `model_type: "granitemoe_swa"` /
`GraniteMoeSWAForCausalLM`, so mlx-lm can load the new IBM Granite MoE+SWA
checkpoints (e.g. `ibm-granite/granite-swash-3b-a600m`). No existing module or
remapping handles this arch today — loading it currently raises
`Model type granitemoe_swa not supported.`

## Architecture

GraniteMoE with sliding-window attention. Reuses existing mlx-lm infrastructure
throughout (mirrors `granitemoehybrid.py` for the Granite MoE/shared-expert side
and `gpt_oss.py` for the sinks/SWA side):

- **MoE**: `nn.Linear` router + `SwitchGLU` experts, top-k routing computed in
  **fp32** (matches HF Granite and `granitemoe.py`), result cast back to the
  activation dtype.
- **Shared expert**: always-on `shared_mlp` (fused `input_linear` gate/up +
  `output_linear`), added to the routed MoE output.
- **Attention sinks**: one learned logit per head, passed through
  `base.scaled_dot_product_attention(..., sinks=...)`.
- **Sliding-window attention** driven by `layer_types`: full-attention layers use
  `KVCache`, sliding layers use `RotatingKVCache(max_size=sliding_window)`, with
  the correct full/SWA mask selected per layer via
  `create_attention_mask(..., window_size=...)`.
- **Granite scalars**: `embedding_multiplier`, `attention_multiplier` (attention
  scale), `residual_multiplier`, `logits_scaling`; tied embeddings; nested
  `rope_parameters` unwrapped to `rope_theta`.
- `sanitize()` splits the fused `block_sparse_moe.experts.gate_up_proj`
  `[E, 2*inter, hidden]` into `SwitchGLU` gate/up **by halves** (gate = first
  half, up = second half — matching HF `GraniteMoeParallelExperts`, *not*
  gpt-oss interleaving).
- `supports_speculative_rollback = True` (the SWA `RotatingKVCache` records an
  exact in-window rollback, enabling `--draft-model`); `make_cache(max_kv_size=…)`
  caps the full-attention layers for long serving.

## Testing

- `tests/test_models.py::test_granitemoe_swa` (tiny 4-layer full/sliding config)
  passes on a pristine `main` checkout — exercises fp32/fp16 forward, cached
  decode, batch > 1, and deepcopy via `model_test_runner`.
- `black`/`isort` clean.

## Validation on the real checkpoint (`granite-swash-3b-a600m`)

- Loads and generates coherently in **bf16** and after MLX **4-bit** conversion.
- **SWA cache is numerically identical to one-shot prefill** past the 128 window:
  cached vs no-cache next-token logits `max|Δ| = 0.00000`; two-chunk prefill vs
  one-shot `0.00001`; full-attention cache grows with the sequence while sliding
  layers cap at `sliding_window`.
- The fused `gate_up_proj` halves-vs-interleaved split was verified empirically
  (per-token NLL **2.09 halves** vs **10.13 interleaved**) and cross-checked
  against the HF `GraniteMoeParallelExperts` source.

## Performance

Apple M5 Max, mlx `0.32.0-dev`, greedy, 64 generated tokens. `prefill` = prompt
tokens/s, `decode` = generation tokens/s, `peak` = MLX peak memory. Rows above
8192 exceed `max_position_embeddings` (the trained window) — perf-only.

4-bit (`granite-swash-3b-a600m-mlx-4bit`, 1.6 GB weights):

|   ctx | prefill tok/s | decode tok/s | peak GB |
|------:|--------------:|-------------:|--------:|
|   128 |         3,791 |          287 |    1.90 |
|   256 |         7,993 |          299 |    2.05 |
|   512 |        12,664 |          289 |    2.37 |
|  1024 |        16,636 |          282 |    2.63 |
|  2048 |        23,163 |          285 |    2.75 |
|  3072 |        22,959 |          282 |    2.75 |
|  4096 |        24,485 |          281 |    2.83 |
|  6144 |        24,663 |          272 |    2.89 |
|  8192 |        24,502 |          277 |    2.95 |
| 12288 |        23,937 |          282 |    3.04 |
| 16384 |        23,337 |          279 |    3.06 |

bf16:

|   ctx | prefill tok/s | decode tok/s | peak GB |
|------:|--------------:|-------------:|--------:|
|   128 |         4,101 |          242 |    6.12 |
|   256 |         7,590 |          246 |    6.20 |
|   512 |        12,229 |          245 |    6.26 |
|  1024 |        18,771 |          239 |    6.47 |
|  2048 |        23,714 |          241 |    6.70 |
|  3072 |        24,202 |          241 |    6.70 |
|  4096 |        25,158 |          229 |    6.77 |
|  6144 |        25,714 |          238 |    6.83 |
|  8192 |        25,053 |          229 |    6.90 |
| 12288 |        24,857 |          220 |    6.98 |
| 16384 |        24,250 |          176 |    7.02 |

Sliding-window attention keeps decode throughput and memory nearly flat across a
**128× context range** (128 → 16384): 20 of 28 layers cap their KV at
`sliding_window=128`, so only the 8 full-attention layers grow. 4-bit decode
holds ~280 tok/s throughout with peak rising just **1.16 GB** (1.90 → 3.06);
bf16 holds ~230 tok/s to 12k (dipping to 176 at 16k as the 8 global layers' KV
begins to dominate) with peak +0.9 GB. Prefill scales to ~25k tok/s.

Notes: the upstream base checkpoint ships no chat template (it is a base model);
a standard Granite template can be added downstream. Weights are bf16
safetensors; `sanitize()` targets that layout (not pre-quantized `_blocks`/
`_scales` variants).

### DIFF
--- a/mlx_lm/models/granitemoe_swa.py
+++ b/mlx_lm/models/granitemoe_swa.py
@@ -1,0 +1,247 @@
+# Copyright © 2025 Apple Inc.
+
+from dataclasses import dataclass
+from typing import Any, List, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .activations import swiglu
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .cache import KVCache, RotatingKVCache
+from .rope_utils import initialize_rope
+from .switch_layers import SwitchGLU
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str
+    vocab_size: int
+    hidden_size: int
+    intermediate_size: int
+    num_hidden_layers: int
+    num_attention_heads: int
+    num_key_value_heads: int
+    num_local_experts: int
+    num_experts_per_tok: int
+    shared_intermediate_size: int
+    max_position_embeddings: int
+    rms_norm_eps: float
+    embedding_multiplier: float
+    attention_multiplier: float
+    residual_multiplier: float
+    logits_scaling: float
+    sliding_window: int
+    layer_types: List[str]
+    attention_bias: bool = False
+    tie_word_embeddings: bool = True
+    rope_theta: float = 10000.0
+    rope_parameters: Optional[dict] = None
+
+    def __post_init__(self):
+        # transformers 5.x nests rope config under `rope_parameters`
+        if self.rope_parameters:
+            self.rope_theta = float(
+                self.rope_parameters.get("rope_theta", self.rope_theta)
+            )
+
+
+class Attention(nn.Module):
+    def __init__(self, args: ModelArgs):
+        super().__init__()
+        dim = args.hidden_size
+        self.n_heads = args.num_attention_heads
+        self.n_kv_heads = args.num_key_value_heads
+        self.head_dim = head_dim = dim // self.n_heads
+        self.scale = args.attention_multiplier
+        bias = args.attention_bias
+
+        self.q_proj = nn.Linear(dim, self.n_heads * head_dim, bias=bias)
+        self.k_proj = nn.Linear(dim, self.n_kv_heads * head_dim, bias=bias)
+        self.v_proj = nn.Linear(dim, self.n_kv_heads * head_dim, bias=bias)
+        self.o_proj = nn.Linear(self.n_heads * head_dim, dim, bias=bias)
+
+        # gpt-oss-style attention sinks (one learned logit per head)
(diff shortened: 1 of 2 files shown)


zcbenz:

Issue #1587, opened 2026-07-18
QuantizedKVCache shows higher peak memory than fp16 KVCache (worse at longer contexts)


gschaefer3:
## Summary

`QuantizedKVCache` shows *higher* peak MLX memory than plain fp16 `KVCache` in `stream_generate`, at every bit width and context length tested, with the gap widening as context grows — the opposite of the expected result. Filing this as data in case it's useful alongside the compute-stays-float design gap already noted on #941.

## Environment

- mlx-lm 0.31.3, mlx 0.31.2
- Apple M4 Max, 128GB unified memory, macOS 27.0
- Model: `mlx-community/Llama-3.2-3B-Instruct-4bit`

## Reproduction

```python
import mlx.core as mx
from mlx_lm import load, stream_generate
from mlx_lm.models.cache import make_prompt_cache, QuantizedKVCache

model, tokenizer = load("mlx-community/Llama-3.2-3B-Instruct-4bit")

filler = ("The quick brown fox jumps over the lazy dog. " * 2000)
ids = tokenizer.encode(filler)[:32000]
prompt = tokenizer.decode(ids) + "\n\nSummarize the above in one sentence."

# fp16 case (separate process from quantized case, to rule out allocator state bias)
cache = make_prompt_cache(model)
# quantized case
base = make_prompt_cache(model)
cache = [QuantizedKVCache(group_size=64, bits=8) if type(c).__name__ == "KVCache" else c for c in base]

mx.reset_peak_memory()
for _ in stream_generate(model, tokenizer, prompt=prompt, max_tokens=30, prompt_cache=cache):
    pass
print(mx.get_peak_memory() / 1e9, "GB")
```

Each case run in its own fresh process (not just `mx.reset_peak_memory()` between cases in the same process), to rule out allocator-state bias between runs. Results were identical either way.

## Results

| context | case | peak MLX memory | decode speed |
| --- | --- | --- | --- |
| 8,000 tok | fp16 | 3.46 GB | 3.2 tok/s |
| 8,000 tok | `QuantizedKVCache` 8-bit | 4.87 GB (**+1.41 GB**) | 2.6 tok/s |
| 32,000 tok | fp16 | 4.72 GB | 1.0 tok/s |
| 32,000 tok | `QuantizedKVCache` 8-bit | 7.10 GB (**+2.38 GB**) | 0.7 tok/s |
[...]

PR #1611, opened 2026-07-23
docs: note quantized KV-cache prefill memory tradeoff (#1587)

rajanshxrma:
## Summary
Adds a short README note on the `--kv-bits` prefill memory tradeoff surfaced in #1587: quantized KV-cache attention doesn't tile its attention-score matrix the way the fused fp16 kernel does, so on long prompts it can transiently use *more* peak memory than an unquantized cache during prefill (the opposite of what KV-cache quantization is generally expected to do). Reducing `prefill_step_size` is an existing, available mitigation today, with no code changes needed.

Full investigation, raw data, and a correctness-verified prototype: https://github.com/rajanshxrma/mlx-kvcache-1587

## Test plan
- [x] Docs-only change, no code paths affected
- [x] Rendered locally to confirm formatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,18 @@ To use the rotating key-value cache pass the argument `--max-kv-size n` where
 result in worse quality. Larger values like `4096` or higher will use more RAM
 but have better quality.
 
+KV-cache quantization (`--kv-bits`) trades memory for a small quality/latency
+cost, but on long prompts it can transiently use *more* peak memory than an
+unquantized cache during prefill, since the current quantized-attention path
+doesn't tile the attention-score matrix the way the fused fp16 attention
+kernel does. If you hit an out-of-memory error while prefilling a long prompt
+with `--kv-bits` set, reducing `prefill_step_size` (the `--prefill-step-size`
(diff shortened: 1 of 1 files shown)


PhilipJohnBasile (commented):
Accuracy check from the #1587 side, for whoever triages this: every claim in these 12 lines is measured, not asserted. The transient prefill inversion reproduces on three rigs across the thread (M4 Max original report, M1/16GB and M5 Max/128GB in the split-phase cells); the mechanism attribution (untiled attention-score matrix, not cache growth) was isolated by a pre-registered discriminator run independently on two of those rigs, with the predicted scores-tensor size matching measured within 1% ([receipted here](https://github.com/PhilipJohnBasile/scrutineer/blob/main/capsules/quantizedkv-prefill-discriminator.json)); and the `prefill_step_size` mitigation's effect is directly measured in the author's chunk-size sweep (inversion disappears at 512). I also just verified the API names cited: `--prefill-step-size` exists on `mlx_lm.server` and `prefill_step_size` (default 2048) on `generate_step`, current PyPI 0.31.3. Small, correct, and it helps people hitting real OOMs today.

zcbenz:

PR #1613, opened 2026-07-23
Add Kimi K3 (speculative preflight port) — DRAFT, pending 2026-07-27 weights

pierre427:
**Draft / not for merge yet.** This is a *speculative preflight* port of Kimi K3, whose weights and tech report are expected **2026-07-27**. I'm opening it early as a starting point for discussion and to avoid duplicated effort — **no commitment to finish on any timeline**, and it will need revision against the real `config.json` when it lands.

**What this is.** A `kimi_k3.py` model assembled from the public record, reusing existing in-tree components rather than duplicating them:
- **Kimi Delta Attention + hybrid MLA** plumbing from `kimi_linear.py` (Kimi Linear, arXiv:2510.26692)
- **Block attention-residuals** (arXiv:2603.15031)
- **Latent MoE** following the Nemotron-H `moe_latent_size` deployable variant (arXiv:2601.18089; routing stays in model space)
- **Sigmoid attention-output gate** on the MLA layers (arXiv:2505.06708 lineage)

**Every K3-specific assumption is config-gated** (`moe_latent_size`, `mla_gate`, `attnres_block_size`, `linear_attn_config.kda_layers`, …), so once the real config is known the correct path is selected by config rather than a rewrite. Components that don't match will be corrected then.

**Testing.** Adds a `kimi_k3` case to `test_all_models`. Passes the standard runner locally: fp32 + fp16, prefill / prompt-cache / single-token decode, batch > 1, and deepcopy.

**Explicitly unverified until weights drop:** exact layer schedule (which layers are KDA vs full-MLA), the attn-residual block size, whether the MoE latent projection is applied pre- or post-routing, and the gate parametrization. Feedback on any of these from folks with earlier access is very welcome.


zcbenz:

PR #1615, opened 2026-07-24
Add Apertus 1.5 text backbone (apertus1p5_text)

pierre427:
## What

Adds the language backbone for [`swiss-ai/Apertus-v1.5-8B`](https://huggingface.co/swiss-ai/Apertus-v1.5-8B),
the first multimodal Apertus (image + audio + text → text). The backbone is a
plain Apertus decoder (xIELU MLP, qk-norm attention, llama3 RoPE), so this new
model reuses the existing `apertus.py` trunk and overrides only what differs:

1. **Split vocabulary.** The input embedding spans the *extended* vocabulary
   (`vocab_size = 266752`: text + visual + audio code tokens), while the LM head
   is physically pruned to the text-only prefix (`output_vocab_size = 131072`).
   The visual/audio code ids are input-only and never generated, so the pruned
   head is exactly the generatable range.
2. **Checkpoint layout.** Text weights live under the multimodal-wrapper prefix
   (`model.language_model.*`) with a top-level `lm_head.weight`, and rope config
   is a single nested `rope_parameters` dict rather than the legacy
   `rope_theta` + `rope_scaling` pair.

This mirrors the existing `*_text` backbones (`gemma3_text`, `gemma4_text`,
`llama4_text`) — the text tower of a multimodal model, usable directly and by a
future `mlx-vlm` loader. Dispatch is by `model_type` (`apertus1p5_text`); no
registry change is needed.

## Validation

Teacher-forced logits vs the HF reference (`transformers` swiss-ai fork) on
`swiss-ai/Apertus-v1.5-8B`, prompt *"The Swiss Alps are famous for"*:

- **100% next-token argmax parity** across all positions
- min per-position cosine **0.9999**
- residual max-abs-diff ~0.375 is bf16 rounding (both sides bf16)

Greedy generation is coherent (*"…their breathtaking landscapes, enabled by the
unique geological processes that have shaped the region over millions of
years…"*). Verified running on a clean checkout of `main` (this branch is based
on `main` + this one file).

## Scope

Text backbone only. The full multimodal model (EMU3.5 vision tokenizer +
WavTokenizer audio codec + discrete-token fusion) is `mlx-vlm` territory and is
not part of this PR.

### DIFF
--- a/mlx_lm/models/apertus1p5_text.py
+++ b/mlx_lm/models/apertus1p5_text.py
@@ -1,0 +1,113 @@
+# Copyright © 2026 mlx-uag lab.
+#
+# Apertus 1.5 text backbone.
+#
+# Apertus 1.5 (swiss-ai/Apertus-v1.5-8B) is an early-fusion, discrete-token
+# multimodal model (image + audio + text -> text). Its language backbone is a
+# plain Apertus decoder (xIELU MLP, qk-norm attention, llama3 RoPE), with two
+# differences from `apertus.py` that this module handles:
+#
+#   1. Split vocabulary. The input embedding spans the *extended* vocabulary
+#      (`vocab_size` = 266752: text + visual + audio code tokens), while the LM
+#      head is physically pruned to the text-only prefix
+#      (`output_vocab_size` = 131072). Image/audio code ids are input-only and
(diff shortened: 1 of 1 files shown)


zcbenz:

Issue #1604, opened 2026-07-22
qwen3_coder tool parser crashes the server on an unparseable string/object argument (no raw-string fallback)


automate-this:
### Title
qwen3_coder tool parser crashes the server on an unparseable string/object argument (no raw-string fallback)

### Summary
When a model using the Qwen3-Coder XML tool-call format (`qwen3-coder-*`) emits a
tool call whose parameter value can't be parsed as JSON **and** can't be parsed as
a Python literal (e.g. a large free-text / code / shell-command argument, or a
malformed/truncated object), `mlx_lm/tool_parsers/qwen3_coder.py :: _convert_param_value`
lets the second parser's exception propagate. It escapes the HTTP request handler
and crashes the request (unhandled `SyntaxError`/`ValueError`), instead of falling
back to the raw string. A string-valued argument should never be able to crash the
tool parser.

### Environment
- mlx_lm: 0.31.3
- Python: 3.12
- OS: macOS (Apple Silicon)
- Model: an mlx-community Qwen3-Coder-next-8bit build served via `python -m mlx_lm server`

### Repro
1. Serve a Qwen3-Coder model with `mlx_lm server` and register a tool whose schema
   has an `object`/`array` parameter (or any non–string/int/float/bool type).
2. Prompt so the model emits a tool call whose value for that parameter is a large
   blob that is neither valid JSON nor a valid Python literal (in our case ~3,870
   chars of code passed to a "run in terminal" style tool).
3. The server thread throws while formatting the tool call.

### Traceback (key frames)
```
File ".../mlx_lm/tool_parsers/qwen3_coder.py", line 75, in _convert_param_value
return json.loads(param_value)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 3871 (char 3870)

During handling of the above exception, another exception occurred:
...
File ".../mlx_lm/server.py", line 1486, in handle_completion
tool_calls=tool_formatter(tool_calls),
File ".../mlx_lm/tool_parsers/qwen3_coder.py", line 97, in _parse_xml_function_call
param_dict[param_name] = _convert_param_value(...)
File ".../mlx_lm/tool_parsers/qwen3_coder.py", line 77, in _convert_param_value
return ast.literal_eval(param_value)
...
File "<unknown>", line 1
SyntaxError: ...
```

### Root cause
In `_convert_param_value`, the `else:` branch calls `ast.literal_eval` without
guarding it:

```python
else:
[...]

PR #1616, opened 2026-07-24
server: make qwen3_coder tool-call parsing robust (parse unwrapped calls + do not crash on non-literal object/array params)

fryz:
Two closely-related fixes that make `qwen3_coder` tool-call parsing robust in `mlx_lm.server`: **(1)** parse tool calls the model emits without the `<tool_call>` wrapper, and **(2)** don't 500 the whole request when an object/array parameter value isn't a valid literal. Bug 2 is only reliably reachable once Bug 1 is fixed (before the fix, unwrapped calls rarely reached the parser), so they belong together.

## Environment where these reproduce

- **mlx-lm 0.31.3** (Homebrew `0.31.3_2`), **mlx 0.32.0**
- macOS arm64 (Apple M4 Max), Python 3.14
- Model: **`mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit`** (Qwen3-Coder 30B A3B Instruct, 8-bit MLX quant)
- Originally surfaced by the Pi coding agent (`@earendil-works/pi-coding-agent`) driving the local server via the OpenAI-compatible `/v1/chat/completions` endpoint with `tools`, but both reproduce without Pi (see below).

---

## Bug 1 — unwrapped tool calls leak out as content

`mlx_lm.server`'s tool-call text state machine only leaves the `normal` state when it sees the literal `<tool_call>` start marker. Under a non-trivial system prompt, **Qwen3-Coder** frequently emits the inner tool-call block

```
<function=NAME>...<parameter=...>...</parameter>...</function>
```

**WITHOUT** the enclosing `<tool_call>...</tool_call>` wrapper (and sometimes with a stray trailing `</tool_call>`). Because the state machine never leaves `normal`, the entire call leaks out as assistant `content` with `finish_reason: "stop"` and **no `tool_calls`**. The client sees text, not a tool call, and nothing executes. The model's own `qwen3_coder` parser (`parse_tool_call`) parses the inner block fine — the server just never hands it the text.

### Root cause

`make_text_state_machine` (in `mlx_lm/generate.py`) registers only `tokenizer.tool_call_start` (`<tool_call>`) as the `normal -> tool` transition. When the model omits that opener, the machine stays in `normal` and the emitted `<function=...></function>` text is treated as ordinary content. Empirically, with a large system prompt structured `tool_calls` were emitted **0/6** samples; with a minimal prompt, **6/6**. Pure server-side parser brittleness.

### Fix

Add an **opt-in, generic** pair of alternate delimiters, `tool_call_start_alt` / `tool_call_end_alt`, that any tool parser can advertise. When a parser sets them:

- `make_text_state_machine` adds a **separate `tool_unwrapped` state** (`normal`/`reasoning` -> `tool_unwrapped` on `<function=`, back to `normal` on `</function>`).
- The server accumulates the captured text and **reconstructs the wrapped form** (`"<function=" + captured + "</function>"`) before calling the existing `parse_tool_call`. Reconstruction also runs if an EOS ends generation mid-call (no closing `</function>`), keyed on `tool_text`.

Only `qwen3_coder` sets the alt delimiters here. **The well-formed `<tool_call>` path is completely unchanged** — inside the `tool` state there is no `<function=` transition, so wrapped calls behave exactly as before (verified below).

---

## Bug 2 — non-literal object/array params 500 the whole request

`mlx_lm/tool_parsers/qwen3_coder.py` `_convert_param_value` called `ast.literal_eval(param_value)` for object/array/dict/list (and unknown) param types. When a model emits a tool with an **array or object** parameter whose value is multi-line code (unbalanced quotes/brackets) — e.g. an editor tool with an `edits: array` param writing source that contains an unterminated string literal — `ast.literal_eval` raises `SyntaxError: unterminated string literal`. That was **uncaught and 500'd the entire request**: the client sees "Stream ended without finish_reason" and the agent turn dies.

This was latent before the Bug 1 fix because tool calls rarely reached the parser; now that they do, it is hit in practice (reproduced end-to-end: Qwen3-Coder-30B calling an `edit` tool on flask's `config.py`).

### Fix

In `_convert_param_value`, for object/array types try `json.loads` first, then fall back to `ast.literal_eval`, and finally return the **raw string** rather than raising. A best-effort string is far better than a 500. Valid JSON and valid Python literals (e.g. single-quoted dicts) still parse exactly as before.

---

## Files touched

- `mlx_lm/tool_parsers/qwen3_coder.py` — declare `tool_call_start_alt = "<function="`, `tool_call_end_alt = "</function>"` (Bug 1); best-effort fallback in `_convert_param_value` (Bug 2).
- `mlx_lm/tokenizer_utils.py` — plumb the alt delimiters through `TokenizerWrapper` and `load()` (matched on decoded text, so no encoded token tuples needed).
- `mlx_lm/generate.py` — add the `tool_unwrapped` state to `make_text_state_machine`.
- `mlx_lm/server.py` — collect/reconstruct/flush the `tool_unwrapped` state in the response loop.

## Pi-free reproduction

Start the server:

```bash
mlx_lm.server --model mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit
```

**Bug 1** — feeding the following assistant output (as the model emits it under a heavy system prompt) through the server's state machine yields it verbatim as `content` with `finish_reason: "stop"` and no `tool_calls` before this change:

```
Sure, let me do that.
<function=multiply>
<parameter=a>
12234585
</parameter>
<parameter=b>
48838483920
</parameter>
</function></tool_call>
```

End-to-end curl that tends to elicit the unwrapped form (a heavier system prompt makes it far more likely):

```bash
  -d '{
    "model": "mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit",
    "messages": [
      {"role": "system", "content": "You are a coding agent with access to tools. Think carefully, then call the appropriate tool. <long system prompt ...>"},
      {"role": "user", "content": "Multiply 12234585 by 48838483920."}
    ],
    "tools": [
      {"type": "function", "function": {
        "name": "multiply",
        "description": "Multiply two numbers.",
        "parameters": {"type": "object", "required": ["a","b"],
          "properties": {"a": {"type": "number"}, "b": {"type": "number"}}}}}
    ],
    "temperature": 0.0
  }'
```

- **Before:** `choices[0].message.content` contains the raw `<function=...></function>` text, `finish_reason` is `"stop"`, and `tool_calls` is absent.
- **After:** `choices[0].message.tool_calls` contains a structured `multiply` call and `finish_reason` is `"tool_calls"`.

**Bug 2** — an object/array param whose value isn't a valid literal (this is what an `edit`-style tool produces when the model writes real source code):

```python

tools = [{"type": "function", "function": {
    "name": "edit",
    "parameters": {"type": "object", "properties": {"edits": {"type": "array"}}}}}]

model_output = (
    "<function=edit>"
    '<parameter=edits>SECRET_KEY = os.environ.get("SECRET_KEY", "dev\nDEBUG = [True</parameter>'
    "</function>"
)
print(qwen3_coder.parse_tool_call(model_output, tools))
```

- **Before:** raises `SyntaxError: unterminated string literal` (uncaught in the server -> HTTP 500, "Stream ended without finish_reason").
- **After:** returns `{"name": "edit", "arguments": {"edits": 'SECRET_KEY = ...'}}` — the raw string, no crash.

## Verification

- Bug 1, failing prompt on `mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit`: structured `tool_calls` emission **0/6 -> 6/6** (non-streaming and streaming).
- Bug 1 regression, well-formed `<tool_call>...</tool_call>`: **6/6** (unchanged).
- Added `tests/test_server.py::TestServer::test_make_state_machine_unwrapped_tool_call` (Bug 1): drives the state machine token-by-token and asserts a structured tool call for a bare `<function=...></function>` block — bare, with a leading prose sentence, with a trailing stray `</tool_call>`, and truncated (EOS mid-call) — plus a regression case confirming the well-formed wrapped block still yields exactly one call.
- Added `tests/test_tool_parsing.py::TestToolParsing::test_qwen3_coder_non_literal_object_array_params` (Bug 2): asserts an unparseable multi-line-code array param returns the raw string instead of raising, and that valid JSON object/array params still parse.
- Full `tests/test_server.py`, `tests/test_tool_parsing.py`, and `tests/test_tokenizers.py` pass (**37 passed, 28 subtests**). `black` and `isort --profile=black` clean.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1077,6 +1077,21 @@ def make_text_state_machine(tokenizer, stop_words=None):
             [(tokenizer.tool_call_end, "normal")] if tokenizer.tool_call_end else []
         )
 
+        # Optional: recognize a bare "<function=...>...</function>" block emitted
+        # WITHOUT the <tool_call> wrapper (a common failure mode for some models
+        # under a heavy system prompt). Handled as a separate "tool_unwrapped"
+        # state so the well-formed <tool_call> path above is unchanged. The
+        # server reconstructs the wrapped form before parsing.
+        tool_call_start_alt = getattr(tokenizer, "tool_call_start_alt", None)
+        if tool_call_start_alt:
+            transitions["normal"].append((tool_call_start_alt, "tool_unwrapped"))
+            if tokenizer.has_thinking:
+                transitions["reasoning"].append((tool_call_start_alt, "tool_unwrapped"))
+            tool_call_end_alt = getattr(tokenizer, "tool_call_end_alt", None)
+            transitions["tool_unwrapped"] = (
+                [(tool_call_end_alt, "normal")] if tool_call_end_alt else []
+            )
+
     if stop_words:
         for state_name in set(transitions) | {"normal"}:
             for w in stop_words:
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1425,11 +1425,18 @@ class APIHandler(BaseHTTPRequestHandler):
                 # Collect the clean text by state: reasoning, tool, or normal
                 if current_state == "reasoning":
                     reasoning_text += clean_text
-                elif current_state == "tool":
+                elif current_state in ("tool", "tool_unwrapped"):
                     tool_text += clean_text
                 elif current_state == "normal":
                     if prev_state == "tool":
                         tool_calls.append(tool_text)
+                        tool_text = ""
+                        made_tool_call = True
+                    elif prev_state == "tool_unwrapped":
+                        # Reconstruct the wrapped form the parser expects, since
+                        # the "<function=" / "</function>" delimiters were
+                        # stripped on the state transitions.
+                        tool_calls.append("<function=" + tool_text + "</function>")
                         tool_text = ""
                         made_tool_call = True
                     text += clean_text
@@ -1443,7 +1450,7 @@ class APIHandler(BaseHTTPRequestHandler):
 
                 if (
                     self.stream
-                    and current_state != "tool"
+                    and current_state not in ("tool", "tool_unwrapped")
                     and (text or tool_calls or reasoning_text)
                 ):
                     resp = self.generate_response(
@@ -1465,6 +1472,11 @@ class APIHandler(BaseHTTPRequestHandler):
 
             if prev_state == "tool" and tool_text:
                 tool_calls.append(tool_text)
+                made_tool_call = True
+            elif prev_state == "tool_unwrapped" and tool_text:
+                # An EOS ended generation mid-call (no closing "</function>");
+                # flush the reconstructed wrapped form.
+                tool_calls.append("<function=" + tool_text + "</function>")
                 made_tool_call = True
 
             if finish_reason == "stop" and made_tool_call:
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -297,6 +297,8 @@ class TokenizerWrapper:
         tool_call_start=None,
         tool_call_end=None,
         tool_parser=None,
+        tool_call_start_alt=None,
+        tool_call_end_alt=None,
     ):
         self._tokenizer = tokenizer
         self._detokenizer_class = detokenizer_class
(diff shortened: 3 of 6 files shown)


fryz:
Unsure what the etiquette is with submitting AI-assisted PRs, but wanted to confirm that after applying this patch locally on my machine, the Qwen Coder model I'm using went from unusable to actually being able to make edits to files. 

LMK if there's anything I can do to help get this PR merged. Thanks!

fryz:
https://github.com/ml-explore/mlx-lm/pull/1614 seems related

fryz:
Also https://github.com/ml-explore/mlx-lm/issues/1604

zcbenz:

fryz:
godspeed

PR #1617, opened 2026-07-24
Guard numeric param conversions in Qwen3 Coder tool parser

damascoemi:
## Summary

- guard the int/float branches in the same parser against non-numeric model output, mirroring the raw-string fallback #1614 added for the object/array/other-type branch
- also fall back to the raw string for "inf"/"nan" float literals, which pass float() but then crash int() with either OverflowError/ValueError
- cover both cases with a regression test

This is a follow up to PR #1614, which fixed the crash for object/array/custom-typed arguments but not the numeric branches, which was noted in the original issue (#1604). Malformed model output for an int/float-typed tool argument could still crash the request handler.

## Testing

- confirmed the crash reproduces on main for non-numeric and inf/nan numeric arguments, and no longer reproduces on this branch
- `python -m unittest tests.test_tool_parsing`
- `black --check mlx_lm/tool_parsers/qwen3_coder.py tests/test_tool_parsing.py`
- `isort --check-only --profile=black mlx_lm/tool_parsers/qwen3_coder.py tests/test_tool_parsing.py`

Follow up to #1614 / related to #1604

### DIFF
--- a/mlx_lm/tool_parsers/qwen3_coder.py
+++ b/mlx_lm/tool_parsers/qwen3_coder.py
@@ -54,10 +54,16 @@ def _convert_param_value(param_value: str, param_name: str, param_config: dict) 
         or param_type.startswith("short")
         or param_type.startswith("unsigned")
     ):
-        return int(param_value)
+        try:
+            return int(param_value)
+        except ValueError:
+            return param_value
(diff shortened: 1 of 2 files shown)


zcbenz:
Closing in favor of https://github.com/ml-explore/mlx-lm/pull/1417.

PR #1618, opened 2026-07-26
Fail early for unsupported KV-cache quantization

PhilipJohnBasile:
## Stack status

Dependent on #1584, branched at `d98dbe2`. This branch contains #1584 plus **one follow-up commit, `aa5614b8`** — that commit is the whole of this PR's contribution: 3 files, +381/-1.

**Please review `aa5614b8` only.** The diff against `main` reads +1358/-380 because it carries #1584's feature underneath; roughly a thousand of those lines belong to that PR, not this one. This will be rebased onto `main` once #1584 lands, at which point the diff collapses to the follow-up commit alone.

Two housekeeping notes for anyone comparing SHAs: this branch was amended after an earlier push (the previous tip `d981a71` is no longer reachable — same tree, corrected author email), and #1584 has advanced to `ff1e837` since `d98dbe2`. `git merge-tree --write-tree` on the two current heads reports **one conflict**, in `mlx_lm/generate.py` — about 16 lines in `BatchGenerator.insert_segments`. It is mechanical: #1584's `3ebafab` added an explanatory comment and this branch added a `_validate_prompt_cache_quantization(...)` call, both immediately before the same `maybe_quantize_kv_cache(...)`; the resolution is to keep both. It appeared because #1584 advanced two commits (`3ebafab`, `ff1e837`) past the `d98dbe2` I branched from. *(Corrected twice: this line first claimed no conflicts, from a legacy three-argument `git merge-tree` grepped for `^<<<<<<<`. That form does report the conflict, but as a diff, so the markers arrive as `+<<<<<<<` and the pattern was one character short — the pipeline exited non-zero for that reason and I discarded the signal by keeping only the count. It then named the method `insert`; it is `insert_segments`.)*

Related to #1573.

## What changed

- Add explicit cache capabilities: whether KV quantization applies, whether a cache can honor it, and whether it is already quantized.
- Validate `generate_step` before the first model call.
- Validate both target and draft caches in speculative generation.
- Validate and quantize both generated and externally supplied caches in `BatchGenerator`; supplied caches previously bypassed quantization entirely.
- Report every offending layer/cache class together, including the specific `RotatingKVCache(keep>0)` limitation from #1584.
- Reject a pre-quantized cache whose bits or group size do not match the request instead of silently retaining the old configuration.
- Continue to leave non-KV recurrent state (`ArraysCache`) untouched explicitly, while rejecting unsupported KV-cache implementations.

There is deliberately no server flag in this change: `mlx_lm.server` does not currently expose `kv_bits`. The validation lives in the library entry points that actually accept the option, so CLI wrappers inherit it without inventing a server configuration surface.

## Why

The old `hasattr(c, "to_quantized")` check cannot distinguish a working implementation from a stub that raises `NotImplementedError`. It also silently skips cache classes with no method. That defers a capability mismatch until a request crosses the quantization threshold and can leave mixed caches only partially quantized.

After #1584, Gemma 4's `keep=0` rotating/full-attention mix is supported, but the generic `--max-kv-size` fallback still creates `RotatingKVCache(keep=4)`. This guard turns that remaining true-negative into an immediate, layer-specific configuration error before inference.

## Security classification

This is a correctness and configuration-validation bug, not a cybersecurity vulnerability. It does not cross a trust boundary, expose data, execute untrusted input, or alter authentication/authorization; the impact is a late exception or an unhonored performance/memory option.

## Validation

- `MLX_ENABLE_TF32=0 python -m pytest -q tests/test_cache_quantization_capability.py tests/test_prompt_cache.py tests/test_generate.py` — **64 passed** on M5 Max / MLX 0.32.0.
- New unit coverage includes the `keep>0` true-negative, Gemma-style mixed-cache true-positive, non-KV state handling, no silent skip, pre-quantized configuration mismatch, target/draft validation, and generated/external continuous-batching caches.
- Pinned `black` and `isort` pre-commit hooks pass.

Without the TF32 control, the same eight pre-existing M5 batch-equivalence failures already documented on #1584 reproduce; the new capability tests pass independently.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1587,8 +1587,25 @@ class BatchGenerator:
         prefill_batch_size: int = 8,
         prefill_step_size: int = 2048,
         max_kv_size: Optional[int] = None,
+        kv_bits: Optional[int] = None,
+        kv_group_size: int = 64,
+        quantized_kv_start: int = 0,
         stream=None,
     ):
+        if kv_bits is not None and quantized_kv_start != 0:
+            # Validated before any state is set, so a rejected config never
+            # leaves a partially-constructed instance behind (__del__ calls
+            # close(), which needs self._old_wired_limit to already exist).
+            # Per-job caches are created once, empty (offset=0), at insertion
+            # time via _make_new_cache() — maybe_quantize_kv_cache's
+            # offset-threshold gate would never trigger later for a nonzero
+            # quantized_kv_start, since there's no per-step re-check in the
+            # batching path. Only immediate quantization is supported here.
+            raise NotImplementedError(
+                "BatchGenerator only supports quantized_kv_start=0 with kv_bits "
+                "set — delayed/threshold quantization is not implemented for "
+                "the continuous-batching path."
+            )
         self.model = model
         self.max_tokens = max_tokens
         self.sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
@@ -1598,6 +1615,9 @@ class BatchGenerator:
         self.prefill_batch_size = prefill_batch_size
         self.completion_batch_size = max(completion_batch_size, prefill_batch_size)
         self.max_kv_size = max_kv_size
+        self.kv_bits = kv_bits
+        self.kv_group_size = kv_group_size
+        self.quantized_kv_start = quantized_kv_start
 
         self._stream = stream or generation_stream
 
@@ -1631,7 +1651,10 @@ class BatchGenerator:
         return self._stream
 
     def close(self):
-        if self._old_wired_limit is not None:
+        # getattr guards against __del__ firing on an instance whose __init__
+        # raised before this attribute was ever set (e.g. an invalid kv_bits
+        # config) — a bare self._old_wired_limit would AttributeError there.
+        if getattr(self, "_old_wired_limit", None) is not None:
             mx.synchronize(self._stream)
             mx.set_wired_limit(self._old_wired_limit)
             self._old_wired_limit = None
@@ -1732,16 +1755,27 @@ class BatchGenerator:
 
     def _make_new_cache(self):
         if self.max_kv_size is None:
-            return cache.make_prompt_cache(self.model)
+            new_cache = cache.make_prompt_cache(self.model)
+        else:
+            new_cache = [
+                (
+                    RotatingKVCache(max_size=self.max_kv_size)
+                    if isinstance(ci, KVCache)
+                    else ci
+                )
+                for ci in cache.make_prompt_cache(self.model)
+            ]
 
-        return [
-            (
-                RotatingKVCache(max_size=self.max_kv_size)
-                if isinstance(ci, KVCache)
-                else ci
+        if self.kv_bits is not None:
+            # quantized_kv_start is always 0 here (enforced in __init__), so this
+            # quantizes every layer immediately — the cache is empty (offset=0),
+            # so there's no precision to lose. All subsequent update_and_fetch
+            # calls go through the quantized cache classes from this point on;
+            # no further re-quantization is needed anywhere else.
+            maybe_quantize_kv_cache(
+                new_cache, self.quantized_kv_start, self.kv_group_size, self.kv_bits
             )
-            for ci in cache.make_prompt_cache(self.model)
-        ]
+        return new_cache
 
     def _find_uids(self, uids):
         uids = set(uids)
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -287,11 +287,59 @@ class GenerationResponse:
     finish_reason: Optional[str] = None
 
 
+def _validate_prompt_cache_quantization(
+    prompt_cache,
+    *,
+    cache_name,
+    kv_group_size,
+    kv_bits,
+):
+    if kv_bits is None:
+        return
+
+    errors = []
+    if not cache.can_quantize_prompt_cache(prompt_cache):
+        for layer, entry in enumerate(prompt_cache):
+            if entry.is_quantization_applicable() and not entry.is_quantizable():
+                errors.append(
+                    f"{cache_name} layer {layer} ({type(entry).__name__}: "
+                    f"{entry.quantization_unsupported_reason()})"
+                )
+
+    for layer, entry in enumerate(prompt_cache):
+        if entry.is_quantized() and (
+            entry.bits != kv_bits or entry.group_size != kv_group_size
+        ):
+            errors.append(
+                f"{cache_name} layer {layer} ({type(entry).__name__}: already "
+                f"quantized with bits={entry.bits}, group_size={entry.group_size}; "
+                f"requested bits={kv_bits}, group_size={kv_group_size})"
+            )
+
+    if errors:
+        raise ValueError(
+            "KV-cache quantization cannot be applied to every cache entry: "
+            + "; ".join(errors)
+            + "."
+        )
+
+
 def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_bits):
     if kv_bits is None:
         return
+
+    _validate_prompt_cache_quantization(
+        prompt_cache,
+        cache_name="prompt cache",
+        kv_group_size=kv_group_size,
+        kv_bits=kv_bits,
+    )
     for e, c in enumerate(prompt_cache):
-        if hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
+        if (
+            c.is_quantization_applicable()
+            and not c.is_quantized()
+            and c.offset >= quantized_kv_start
+        ):
             prompt_cache[e] = c.to_quantized(group_size=kv_group_size, bits=kv_bits)
 
 
@@ -364,6 +412,13 @@ def generate_step(
             model,
             max_kv_size=max_kv_size,
         )
+
+    _validate_prompt_cache_quantization(
+        prompt_cache,
+        cache_name="model prompt cache",
+        kv_group_size=kv_group_size,
+        kv_bits=kv_bits,
+    )
 
     prompt_progress_callback = prompt_progress_callback or (lambda *_: None)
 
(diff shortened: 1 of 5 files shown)


mabaeyens:
Keep it stacked the way you have it. #1584 first, then rebase this onto main.

Two reasons, and neither is about precedence. You already validated #1584 end to end at `d98dbe2`, including the Gemma 4 hybrid layout and the rotation path, and reordering would throw that away for nothing. And #1584 rewrites large parts of `cache.py`, so whichever of us goes second is rebasing regardless; better that it is the smaller, newer diff.

The one thing I would not accept is your PR being held hostage to mine. #1584 has no maintainer review yet and I have no idea when it will get one. If it sits, do not sit with it. Say so and I will either close mine or rebase under yours, whichever gets the guard in faster, because the guard is useful with or without my cache classes.

On the design, `can_quantize_prompt_cache()` plus a declared `is_quantizable()` is the right call and better than my read of it. I checked the precedent you are copying: `can_trim_prompt_cache()` at `cache.py:88` with `is_trimmable()` on each class, and `generate_step` raising up front with the offending type names. Putting the second capability next to the first, in `generate_step` rather than in one CLI, covers library callers and the server equally, and it does not depend on a `--kv-bits` flag existing anywhere (which, per what I just posted on #1573, it does not in this repo).

For the two classes #1584 adds, `RotatingQuantizedKVCache` and `BatchRotatingQuantizedKVCache`, take them in your rebase rather than having me add `is_quantizable()` here. You own the convention and it should read consistently across all of them. If you would rather I land them with the method already on so your rebase is smaller, tell me and I will add it.

Worth flagging one interaction: mlx-lm#1595 also touches `tests/test_generate.py`, which #1584 touches too. Different regions, imports against test bodies, so it should merge clean, but if all three land close together someone is going to eat a trivial conflict.

PhilipJohnBasile:
Rechecked this draft against the refactored #1353. No code change is required: #1353 no longer touches `generate.py` or `models/cache.py`, while this PR remains the library/runtime capability layer plus `BatchGenerator` support. The newly reported `max_kv_size` + `kv_bits` constraint is already represented here as a real `keep>0` true negative. After #1584 lands, I’ll rebase this draft; server startup probing stays in #1353, and any later batching re-enable should be a small wiring follow-up after both are present.

mabaeyens:
Came here from your note on #1353, since the narrowing lands in this PR rather than that one.

Yours, and the reasoning is right. If the probe has to know which generator will serve the request, that belongs in the capability API and not in a fourth PR of mine sitting on top of two unmerged ones.

One thing to carry into the design, because the comment I wrote at `generate.py:1595` undersells it. `BatchGenerator` refusing `quantized_kv_start != 0` is not only a missing per-step re-check. Per-job caches exist until the batch forms and no longer: `PromptProcessingBatch` merges them at `generate.py:1120` into one cache per layer for the whole batch, so after that there is no per-sequence object left to convert. Rows in a batch also sit at different offsets, so a per-sequence threshold has nothing to attach to on a shared cache. And a cache quantized mid-flight could not rejoin a batch anyway, since `_merge_caches` needs one class across all jobs and `QuantizedKVCache` defines no `merge`. Doing it properly means batch-level threshold semantics plus a `BatchKVCache.to_quantized()`, and no plain `BatchQuantizedKVCache` exists (the rotating side has one, the plain side does not).

So treat it as unimplemented rather than forbidden, but it is a feature and not a plumbing fix. Worth an issue alongside the sink-token one if this PR wants it gone. What I would avoid is the API modelling "cannot honour a nonzero start" as either permanent or cheap, because it is neither.

The `keep` half is genuinely asymmetric and no decision of mine removes it. Sequential goes through `make_prompt_cache` with its hardcoded `keep=4` (`cache.py:37`), batched builds `keep=0` (`generate.py:1777`), so the same two flags mean different things depending on who serves the request. That one your probe has to model either way.

soobrosa:
Filed the sink-token issue as #1631, per your confirmation on #1353.

Relevant to #1618's design rather than just bookkeeping: the issue offers two resolutions and they land differently for you. If sink tokens get implemented, `keep > 0` stays a real capability distinction your API has to express. If instead `cache.py:37` drops to `keep=0` (it is the only `keep > 0` in the repo, and `generate.py:842` already treats nonzero as unsupported), then `keep` stops being a case worth modelling and `quantization_unsupported_reason()` loses its only concrete caller today.

Worth knowing before you finalise the capability surface, since one outcome makes `is_quantizable()` cheaper than what is currently in the draft. Asked @awni on the issue which way he wants it, as he added the default in #1015.

PhilipJohnBasile:
Follow-up from #1631: the historical `keep=4` intent is now resolved, but the current policy decision is not. I am keeping this capability surface unchanged rather than simplifying it from the construction-site count.

The concrete gate on #1631 is now lifecycle-complete: merge equal-policy `keep=4` caches with recognizable sink rows; reject mixed `keep` policies; compare the sink rows immediately after batch wrapping; rotate the batch past `max_size`; extract; then require both `keep == 4` and byte-identical sinks after further plain-cache rotation. The quality A/B must likewise cross the window.

One precision correction to my earlier fork: changing the in-tree generic fallback to `keep=0` would remove today's generic true-negative producer, but it would not make arbitrary `RotatingKVCache` objects unconditionally quantizable. While the public class accepts `keep>0` and quantization cannot preserve it, the Rotating capability check should remain conditional on `keep == 0` unless sink support itself is removed or implemented in the quantized cache.

`ChunkedKVCache`, `ConcatenateKVCache`, and composites independently keep the general capability API useful, so no code simplification to #1618 is justified before that evidence and design decision.

soobrosa:
Agreed, and thank you for the correction. Verified both points against the imported history before conceding them.

`38d78ee` is [mlx-examples#931](https://github.com/ml-explore/mlx-examples/pull/931) and its description says it plainly: the first `keep` tokens are always kept, citing StreamingLLM. `0c8e878` ([mlx-examples#1015](https://github.com/ml-explore/mlx-examples/pull/1015)) only relocated it into `make_prompt_cache`. My "looks unintentional" framing was wrong, and the bare `#1015` in my issue pointed at an unrelated mlx-lm issue rather than the mlx-examples PR, which made it harder to check than it should have been. #1631's body and title are corrected, with the original framing marked rather than quietly rewritten.

On the capability API, you are right and I withdraw the suggestion. I reasoned from in-tree call sites to the public surface, which does not follow: `RotatingKVCache` accepts `keep > 0` from any caller and quantization cannot preserve it, so `is_quantizable()` stays conditional on `keep == 0` whichever way the policy lands. Nothing in #1618 should be simplified on the basis of what I wrote.

Nothing changes in #1353 either way, since it reports whatever reason the cache itself raises rather than encoding a rule about `keep`. One update there that is relevant to your validation split, though: after @Git-on-my-level's 100K-token Laguna data on #1043, #1353 no longer refuses on partial quantizability. It warns and serves when some layers convert, and raises only when none do. The `--max-kv-size` case you call a true negative is still the raising case, since every layer is rotating with sinks, so the boundary between us is unchanged.

PhilipJohnBasile:
Thanks — and specifically for marking the original framing in #1631 rather than rewriting it. A corrected issue that still shows what it used to say is worth more to whoever reads it next than a clean one, since the reasoning that produced the wrong framing is usually the part worth seeing.

Agreed on all three points, and one of them resolves a dependency:

**The capability surface stays as drafted.** `RotatingKVCache.is_quantizable()` is `return self.keep == 0`, with `quantization_unsupported_reason()` reporting `keep={n} sink tokens are not supported`. That holds whichever way #1631 lands — if sinks get implemented, `keep > 0` is a real distinction the API has to express; if `cache.py:37` drops to `keep=0`, the constructor still accepts `keep > 0` from any caller and quantization still cannot preserve it. So the predicate is about what the cache *can* be handed, not about what the repo currently hands it. That means **#1618 no longer waits on @awni's answer** — worth saying explicitly, since your earlier note had it as a blocker.

**#1353's boundary is unchanged**, and I'd only add why the direction is right: warn-and-serve on partial conversion is the correct call given @Git-on-my-level's 100K Laguna data, because refusing a request that would have worked is a worse failure than serving it degraded with a warning. `--max-kv-size` still raises because every layer is rotating with sinks, so nothing converts — that's the "none do" case, not the partial one. The line between us was always that distinction rather than a disagreement about policy.

No changes needed here on my side. If the sink-token work does land in #1631, `is_quantizable()` is the one place that would need revisiting, and I'll watch that issue.

PhilipJohnBasile:
> **Corrected twice.** This comment first told @mabaeyens `git merge-tree` reported no conflicts and the rebase would be uneventful — wrong, there is one. The correction I then wrote explained *why* wrongly, and named the wrong method. Both fixed below; the description carries the same fixes.

Corrected the stack note in the description — it had gone stale in a way that would waste a reviewer's time.

It still said "Draft" (this hasn't been one for a while) and pointed at `d981a71` as the follow-up commit. That commit still resolves on GitHub, but it sits on no branch here: the branch was amended to fix a typo in my own author email, so the current tip is `aa5614b8` with an identical tree.

The substantive part for review: **`aa5614b8` is the entirety of this PR's contribution — 3 files, +381/-1.** The `+1358/-380` shown against `main` is mostly #1584 riding underneath, so the line count overstates this PR by about 3.5x. Reviewing that one commit is reviewing all of it.

@mabaeyens — no action wanted, just keeping the stack honest per your #1584-first ordering. #1584 has moved to `ff1e837` since I branched at `d98dbe2`.

**On the rebase.** `git merge-tree --write-tree` on the two current heads exits 1 with `CONFLICT (content): Merge conflict in mlx_lm/generate.py` — one region of about 16 lines in **`BatchGenerator.insert_segments`** (I previously wrote `insert`; wrong method). Your `3ebafab` added an explanatory comment on the `elif self.kv_bits is not None:` branch; this PR added a `_validate_prompt_cache_quantization(...)` call on the same branch. Both sit immediately before the same `maybe_quantize_kv_cache(...)`, and the resolution is to keep both — the comment describes the in-place quantization, the call guards it. Nothing semantic is in dispute, and it's mine to resolve at rebase time, not yours.

**Why I got it wrong twice**, since the second reason is more interesting than the first. I originally ran the legacy three-argument `git merge-tree` and grepped its output for `^<<<<<<<`. I then wrote that this form "does not emit conflict markers." It does — it reports as a *diff*, so they arrive as `+<<<<<<<`, and my pattern was one character short. `grep -c '^+<<<<<<<'` returns 1. The pipeline had also exited non-zero for exactly that reason, and I threw that signal away by capturing only the count. So the instrument answered correctly twice and I mis-read it twice, the second time while explaining the first.

zcbenz:

PR #1619, opened 2026-07-26
Fix rotated flag round-trip in BatchRotatingKVCache.meta_state

### LINKED ISSUE #1250: BatchRotatingKVCache: rotated=False reloads as True from saved prompt cache

mabaeyens:
`meta_state` stringifies every field, so `rotated` is stored as `"True"` or `"False"`. The setter read it back with `bool()`:

```python
```

Both values are truthy as strings, so a cache that had not rotated came back from a saved prompt cache marked as rotated.

Compares against `"True"` instead. The round-trip test covers both values and fails with `bool()`.

Fixes #1250

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1312,7 +1312,9 @@ class BatchRotatingKVCache(_BaseCache):
             int,
             v[:3],
         )
-        self.rotated = bool(v[3])
+        # meta_state stringifies, so v[3] is "True"/"False" and bool() would
+        # make both truthy, marking an unrotated cache as rotated on reload.
+        self.rotated = v[3] == "True"
 
     def is_trimmable(self):
         return self._offset < self.max_size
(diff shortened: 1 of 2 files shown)


pierre427:
Independently hit this exact class while auditing persisted-cache round-trips: `bool("False") is True`, so an un-rotated batch cache restores as rotated and the ring-order handling silently corrupts from there. Your fix and test look right. We shipped the same `v[3] == "True"`-style parse (plus an int encoding on the write side) in our serving fork and the round-trip test has been green in production since. +1 to merging this.

zcbenz:
I'm closing this in favor of https://github.com/ml-explore/mlx-lm/pull/1778 which fixes the root cause.

PR #1620, opened 2026-07-26
fix(sampler): give each sampler its own PRNG stream (#1245)

### LINKED ISSUE #1245: mlx_lm.server: seed parameter silently ignored

sashko-zakharchuk:
Fixes #1245.

## The bug

`/v1/chat/completions` accepts and validates a `seed`, but it has no effect on the output, and identical requests return byte-identical tokens at temperature > 0.

Measured on `MiniMax-M3-MLX-6bit`, 5 identical requests per cell at `temperature=1.0, top_p=0.95`:

```
noseed: distinct=1/5   max_repeat=[359,359,359,359,359]
seeded: distinct=1/5   (seeds 1..5 all returning the same text)
```

#1245 also reports it on Qwen2.5 and Llama-3.1.

## Cause

Every sampling draw reads the implicit global `mx.random.state`, and the sampling functions are `mx.compile`d with `inputs=mx.random.state`. The issue thread suspected the compile decorator captures the state object, and the reporter's own testing did not confirm that — so I have deliberately not built the fix on that diagnosis. Threading explicit keys removes the dependency on global state entirely, so it holds regardless of the precise mechanism.

Worth noting separately: a global RNG cannot give per-request reproducibility on a server at all. `BatchGenerator` runs multiple generations concurrently, so with one shared stream a request's output depends on what else is in flight.

## The change

`make_sampler` takes an optional `seed` and owns an explicit key chain, splitting a subkey per randomness consumer — each sampling method, plus the final categorical draw. `apply_xtc` and `categorical_sampling` take an optional `key`. `server.py` derives fresh entropy when a request carries no seed, and no longer calls `mx.random.seed` globally, which was resetting the RNG for any concurrently generating request.

A sampler is already constructed per request in `server.py` and per sequence in `BatchGenerator`, so per-sampler state gives per-request isolation with:

- **no changes to `generate.py`** and no arity change at any of the four sampler call sites
- **no public API break** — the documented `Callable[[mx.array], mx.array]` signature still works, so third-party custom samplers are unaffected
- **compilation preserved** — the key is an array input, so nothing is un-compiled

With no explicit seed, one is derived from the global RNG rather than leaving the key as `None`. That keeps each sampler independent, keeps a caller's `mx.random.seed(...)` honoured, and avoids passing `None` into the compiled functions on the hot path.

## Result

Same measurement after the change — 4/4 cells:

```
noseed: distinct=5/5   degen=0/5
seeded: distinct=5/5   degen=0/5
```

## Tests

`tests/test_sampler_rng.py` covers: a seed reproduces the stream; different seeds diverge; interleaved samplers do not perturb each other; the full `top_p`/`min_p`/`top_k` chain stays reproducible; the XTC path is seeded; `temp==0` stays argmax; the legacy one-argument call still works; and the compiled helpers still accept a call with no key.

Verified end to end on Apple Silicon (M3 Ultra, macOS 26.5, mlx 0.32.0) against a live `mlx_lm.server`.

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -16,6 +16,7 @@ def make_sampler(
     xtc_probability: float = 0.0,
     xtc_threshold: float = 0.1,
     xtc_special_tokens: List[int] = [],
+    seed: Optional[int] = None,
 ) -> Callable[[mx.array], mx.array]:
     """
     Make a sampler function for use with ``generate_step``.
@@ -37,6 +38,13 @@ def make_sampler(
             for being sampled.
         xtc_special_tokens (list(int), optional): List of special tokens IDs to
             be excluded from XTC sampling.
+        seed (int, optional): Seed for this sampler's own PRNG stream. The sampler
+            draws from an explicit key chain rather than the implicit global
+            ``mx.random.state``, so the same seed reproduces the same tokens and
+            concurrent samplers cannot perturb one another. When ``None`` a seed is
+            derived from the global RNG, giving each sampler an independent stream
+            while still honouring an ``mx.random.seed(...)`` set beforehand.
+            Default: ``None``.
 
 
     Returns:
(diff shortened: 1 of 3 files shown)


sashko-zakharchuk:
Closing this in favour of a corrected patch.

I re-reviewed it before anyone spent time on it, and found defects I'd rather fix than have reviewed:

- **Distributed serving would diverge.** `_make_sampler` seeded each request with `secrets.randbits(32)` — per-process entropy, so every rank derives a different seed and ranks stop emitting identical tokens. The server synchronises `mx.random.state` across ranks precisely to avoid that.
- **The tests need rebuilding.** I'm not satisfied they fail against unpatched `main`, which would make them worthless as regression tests.
- Smaller: an existing in-repo test breaks, `black` fails on the changed files, and an out-of-range client seed can kill the generation thread.

The underlying bug in #1245 is real and reproducible — identical requests return byte-identical tokens at temperature > 0 (measured: 1 distinct output across 5 identical requests), and the `seed` value has no effect. A revised, much smaller PR will follow.

PR #1622, opened 2026-07-26
FalconH1: apply key_multiplier to k_proj (silently dropped via a dead key_proj branch)

rcfa:
## Bug

`mlx_lm/models/falcon_h1.py` `sanitize()` folds `key_multiplier` only into a weight named `key_proj.weight`:

```python
elif name.endswith("q_proj.weight") or name.endswith("k_proj.weight"):
    param *= args.attention_in_multiplier
elif name.endswith("key_proj.weight"):                 # dead: weight is named k_proj
    param *= args.attention_in_multiplier * args.key_multiplier
```

Falcon-H1 checkpoints name the key projection **`k_proj.weight`**, which matches the *first* branch (only `attention_in_multiplier`, default 1.0). The `key_proj.weight` branch never matches, so **`key_multiplier` is never applied** — and it isn't applied in the attention forward either. HF applies it to the key states (`modeling_falcon_h1`: `key_states = self.k_proj(hidden_states)... * self.key_multiplier`).

For Falcon-H1R-7B `key_multiplier ≈ 0.031`, so the keys end up **~32× too large** → attention scores blow up → softmax collapses toward a hard argmax → the model produces grammatical but degenerate, never-terminating output. This is also baked into every `mlx_lm.convert` output (the fold is saved into the weights), so existing conversions are corrupt and need re-converting after this lands.

## Fix

Split the branch so `k_proj.weight` receives `attention_in_multiplier * key_multiplier`.

## Validation (mlx-lm, bf16 `tiiuae/Falcon-H1R-7B`, greedy, same prompt)

**Before (main):** degenerate, never reaches an answer —
> `<think> We need to parse the problem. ... Wait, the original user message: ... Actually the problem statement seems a bit confusing. Let's rewrite: ... Wait, the original distances: ... Let's read the original:` *(loops until the token cap)*

**After (this PR):** correct —
> `<think> ... 210 miles / 70 mph = 3 hours. ... 240 miles / 80 mph = 3 hours. Total time = 3 + 3 = 6 hours. Thus the trip took 6 hours.`

Same model, prompt, and sampler; the only change is this one branch. (Cross-checked against `transformers`: layer-0 attention output goes from badly mis-scaled to cos 0.999998.)

### DIFF
--- a/mlx_lm/models/falcon_h1.py
+++ b/mlx_lm/models/falcon_h1.py
@@ -471,9 +471,9 @@ class Model(nn.Module):
                 param *= args.embedding_multiplier
             elif name.endswith("lm_head.weight"):
                 param *= args.lm_head_multiplier
-            elif name.endswith("q_proj.weight") or name.endswith("k_proj.weight"):
+            elif name.endswith("q_proj.weight"):
                 param *= args.attention_in_multiplier
-            elif name.endswith("key_proj.weight"):
+            elif name.endswith("k_proj.weight") or name.endswith("key_proj.weight"):
                 param *= args.attention_in_multiplier * args.key_multiplier
             elif name.endswith("o_proj.weight"):
                 param *= args.attention_out_multiplier


rcfa:
One more downstream data point for why this matters beyond runtime inference: because `sanitize()` runs as part of conversion, the dropped `key_multiplier` gets **baked into converted checkpoints**, not just applied wrong at load. Existing Falcon-H1 quants carry keys ~32× too large and can't be repaired at load time (you can't rescale packed-quantized weights), so they need re-converting once this lands.

I hit the same class of problem in a third-party converter that copies this logic — filed as jjang-ai/mlxstudio#129 — which suggests the dead `key_proj` branch has propagated. Cross-linking in case it's useful context for reviewers.

nastya236:
Thanks for your contribution! We recently introduced a new policy regarding the use of AI in pull request descriptions and discussions #1770.

rcfa:
Great! 

We're supposedly at the verge of super intelligence and recursive self improvement...

...but refusing bug fixes, because AI was involved...

...in a project that's all about enabling AI and agentic coding.

Exactly my kind of humor.🤡

I'm doing you guys a favor by reporting a bug along with a fix, and you refuse it without checking on the merits because of some policy 🤦🏻‍♂️

Well, enjoy the bug!🪳

I'm using vMLX-swift where the bug was fixed. 

So: not my circus🎪, not my monkeys🐒

rcfa:
> Thanks for your contribution! We recently introduced a new policy regarding the use of AI in pull request descriptions and discussions #1770.
> 
> Unfortunately, this means that we need to close this pull request. However, you’re very welcome to open a new one that follows the updated policy. Thanks for your understanding!

Great! 

We're supposedly at the verge of super intelligence and recursive self improvement...

...but refusing bug fixes, because AI was involved...

...in a project that's all about enabling AI and agentic coding.

Exactly my kind of humor.🤡

Even more ridiculous: the PR was filed July 26th, the policy is dated August 21st, so being applied retroactively. You guys invented a time machine?💨

And then the policy self contradicts: it both prohibits AI use and requires AI use disclosure. Kind of odd to require the disclosure of something supposedly not allowed 🤣

Never mind that the policy snacks of anti-AI/AL racism/speciesism.🤖

Artificial lives matter!🦾

I'm doing you guys a favor by reporting a bug along with a fix, and you refuse it without checking on the merits because of some policy 🤦🏻‍♂️

Well, enjoy the bug!🪳

I'm using vMLX-swift where the bug was fixed. 

So: not my circus🎪, not my monkeys🐒

PR #1623, opened 2026-07-26
Fix Qwen3.6 converted RMSNorm double shift

PhilipJohnBasile:
## Summary

- use the raw Qwen Conv1D tensor layout as the structural signal that RMSNorm weights still need the MLX `+1` conversion
- stop treating the presence of MTP tensors as evidence that norm weights are zero-centered
- cover dense and MoE Qwen3.5/3.6 conversion, including already-converted checkpoints that still contain MTP tensors

## Root cause

MTP keys are not a reliable checkpoint-format signal. An already-converted VLM/MTP checkpoint can contain MTP tensors while its Conv1D and RMSNorm weights are already in MLX layout. The old predicate shifted every norm by `+1` again, producing incoherent generation.

This implements the maintainer feedback on #1198 without adding a value-distribution heuristic. It also addresses the fresh tensor-level reproduction in #1197.

## Validation

- focused dense and MoE regression: passed
- `black --check mlx_lm/models/qwen3_5.py tests/test_models.py`: passed
- `git diff --check`: passed
- full `tests.test_models`: 75 passed, 1 skipped; the sole failure is the existing, reproducible `test_ssm` numerical comparison in untouched code

Local checkpoint inspection also confirmed the intended raw-conversion case: Qwen3.6 base RMSNorm mean `-0.0238`, with Conv1D shape `(10240, 1, 4)`, so the raw checkpoint still receives the required `+1` conversion.

### DIFF
--- a/mlx_lm/models/qwen3_5.py
+++ b/mlx_lm/models/qwen3_5.py
@@ -346,11 +346,9 @@ class TextModel(nn.Module):
         return [ArraysCache(size=2) if l.is_linear else KVCache() for l in self.layers]
 
     def sanitize(self, weights):
-        has_mtp_weights = any("mtp." in k for k in weights)
         has_unsanitized_conv1d = any(
             "conv1d.weight" in k and v.shape[-1] != 1 for k, v in weights.items()
         )
-        should_shift_norm_weights = has_mtp_weights or has_unsanitized_conv1d
         weights = {k: v for k, v in weights.items() if "mtp." not in k}
 
         if self.args.tie_word_embeddings:
@@ -366,7 +364,7 @@ class TextModel(nn.Module):
         for k, v in weights.items():
             if "conv1d.weight" in k and v.shape[-1] != 1:
                 weights[k] = v.moveaxis(2, 1)
-            if should_shift_norm_weights and any(k.endswith(sfx) for sfx in norm_keys):
+            if has_unsanitized_conv1d and any(k.endswith(sfx) for sfx in norm_keys):
                 if v.ndim == 1:
                     weights[k] = v + 1.0
         return weights
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -647,7 +647,9 @@ class TestModels(unittest.TestCase):
             "max_position_embeddings": 64,
         }
         hf_norm_key = "model.language_model.layers.0.input_layernorm.weight"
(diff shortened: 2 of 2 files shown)


PhilipJohnBasile:
Rebased onto current main (`254d153`). Verified the bug is still live on main and the fix resolves it:

- New regression test `test_qwen3_5_family_convert_then_load_norm_not_shift_twice` **fails on main** (already-converted checkpoint that merely contains `mtp.*` tensors gets its RMSNorm weights +1-shifted a second time — old `should_shift_norm_weights = has_mtp_weights or has_unsanitized_conv1d` gate) and **passes with this patch** (gate on the raw `conv1d` layout alone, which is the reliable converted-GGUF signal).
- `mlx_lm/models/qwen3_5.py` on main still carries the buggy gate; no upstream change has addressed it.

Ready for CI + review.

avlp12:
I verified this patch independently on a different member of the family and it fixes that case too.

My model is `Qwen/Qwen3.8-27B` — dense `qwen3_5`, 64 layers, hybrid GatedDeltaNet/full-attention — converted to 4-bit with the MTP head preserved. I applied only the two predicate lines from this PR to an otherwise stock install and re-ran teacher-forced NLL on a fixed Korean passage with identical weights and prompt:

| mlx-lm | NLL |
| --- | --- |
| stock | 17.460 |
| with this patch | 1.679 |

17.46 is worse than uniform over a 248k vocab, so the logits are actively wrong rather than degraded. The value reproduces to three decimals across runs, and `mlx_lm.generate` is fluent once patched. The patched number matches what I get from my own fork, which had arrived at the same predicate change.

Two things worth adding from my side:

The reports so far are on 35B MoE checkpoints; this one is dense, so the fix is confirmed on both branches.

The bug is invisible unless a checkpoint carries an MTP head — mine without one measured correctly throughout. That made it read as "the MTP-preserving build is broken" rather than "the loader is", and I spent a while suspecting my own quantization before finding it here. Worth keeping in mind when triaging similar reports.

One gap I would suggest closing: the existing regression re-sanitizes weights whose MTP keys were already stripped, so it cannot exercise the case users actually hit — a converted checkpoint that *still contains* MTP tensors at load time. I added that assertion in #1735, which is otherwise the same fix as this PR. Reverting the predicate makes it fail. Feel free to lift it into this PR; I would rather see this one merged than mine.

zcbenz (approved):
Thanks!

PR #853, opened 2026-02-06
Fix Kimi Linear

kernelpool:
This fixes some issues and improves performance of [Kimi Linear](https://huggingface.co/moonshotai/Kimi-Linear-48B-A3B-Instruct), e.g. [mlx-community/Kimi-Linear-48B-A3B-Instruct-4bit](https://huggingface.co/mlx-community/Kimi-Linear-48B-A3B-Instruct-4bit). I've tested the model in OpenCode and tool calling works fine. Note that the original HF repo is missing some tokens in its tokenizer config, but it seems it has been since [fixed](https://huggingface.co/mlx-community/Kimi-Linear-48B-A3B-Instruct-4bit/commit/919175d75427f1c10cfc17e7254b53295edddf72) for the mlx community models.

1. MLA absorption - Similar to #839
2. Fix ShortConv1d variable name - `cache.lengths` -> `lengths`
3. Fix ArraysCache compatibility -  Store conv states as a single concatenated array instead of a tuple (changed in #690)
4. Remove RoPE handling - the model shouldn't be using it as per quote from paper: _"In Kimi Linear, we apply NoPE to all full attention (MLA) layers."_

| Context | Original Prompt TPS | Fixed Prompt TPS | Original Gen TPS | Fixed Gen TPS | Gen TPS Change |                                                                      
|---------|---------------------|------------------|------------------|---------------|----------------|                                                                    
| 1k | 1,485 | 2,185 | 99.2 | 93.8 | -5% |
| 2k | 2,163 | 2,758 | 93.0 | 96.8 | +4% |                                                                                                                                    
| 4k | 2,427 | 2,771 | 89.4 | 95.2 | +6% |
| 8k | 2,494 | 2,698 | 83.3 | 94.0 | +13% |                                                                                                                                   
| 16k | 2,374 | 2,484 | 71.4 | 93.3 | +31% |
| 32k | 2,018 | 2,095 | 57.0 | 89.6 | +57% |

PR #1624, opened 2026-07-27
[kimi_linear] Fix KDA q/k normalization eps to match reference l2norm semantics

avlp12:
The KDA q/k normalization in `kimi_linear` uses `mx.fast.rms_norm(x, None, 1e-6)`. `rms_norm` adds eps to **mean(x²)**, but the reference implementation normalizes with eps added to **sum(x²)** — the official Kimi modeling delegates q/k normalization to FLA's KDA kernels (`use_qk_l2norm_in_kernel=True`), which compute:

```python
# fla/ops/kda/fused_recurrent.py
b_q = b_q / tl.sqrt(tl.sum(b_q * b_q) + 1e-6)
b_k = b_k / tl.sqrt(tl.sum(b_k * b_k) + 1e-6)
```

With `head_dim=128` the effective eps therefore differs by **128×**.

## Fix

```python
eps = 1e-6 / self.head_dim
```

This makes `rms_norm` *exactly* equivalent to the reference l2norm at all magnitudes:

`x / sqrt(mean(x²) + 1e-6/D) == sqrt(D) · x / sqrt(sum(x²) + 1e-6)`

## Impact

Deviation of the full q-preprocessing vs the reference, by activation magnitude (fp32, D=128, max|Δ|/max|ref|):

| amplitude | before (eps=1e-6) | after (eps=1e-6/D) |
|---:|---:|---:|
| 1.0 | 6.2e-07 | 1.0e-07 |
| 0.1 | 7.0e-05 | 1.1e-07 |
| 0.01 | 5.7e-03 | 1.1e-07 |
| 0.001 | 3.3e-01 | 1.1e-07 |
| 0.0001 | **8.7e-01** | 1.5e-07 |

Harmless for typical activation magnitudes, but small-magnitude q/k vectors (possible after the short conv + SiLU) deviate by up to ~87% per component — and because k feeds the recurrent state, the error propagates to *subsequent tokens'* outputs rather than staying local (measured in an end-to-end KDA recurrence harness with mixed near-zero tokens).

<details>
<summary>Repro script</summary>

```python
import mlx.core as mx
import numpy as np

D = 128
scale = D ** -0.5
mx.random.seed(7)

def ref_l2(x):  # reference: eps on sum(x^2)
    return x * mx.rsqrt((x * x).sum(-1, keepdims=True) + 1e-6)

def dev(amp):
    x = mx.random.normal((64, 32, D)) * amp
    ref = scale * ref_l2(x)
    before = (scale ** 2) * mx.fast.rms_norm(x, None, 1e-6)
    after = (scale ** 2) * mx.fast.rms_norm(x, None, 1e-6 / D)
    mx.eval(ref, before, after)
    r = np.array(ref, dtype=np.float64)
    err = lambda y: np.max(np.abs(np.array(y, dtype=np.float64) - r)) / np.max(np.abs(r))
    return err(before), err(after)

for amp in [1.0, 1e-1, 1e-2, 1e-3, 1e-4]:
    print(amp, *dev(amp))
```

</details>

## Testing

- `python -m unittest tests.test_models.TestModels.test_all_models` — OK (includes the `kimi_linear` config)
- Verified against an independently ported KDA reference (FLA/FlashKDA lineage, chunked formulation) across `T ∈ {1, 5, 16, 33, 128, 300}`, with/without initial state, and raw-activation magnitudes `{1.0, 1e-2, 1e-4, mixed near-zero}` — global normalized error ≤ 6e-7 after the fix (was up to 0.87 on the mixed case, with contamination of normal-token outputs through the recurrent state).

Scope: `kimi_linear` only; no other model uses this normalization pattern.

### DIFF
--- a/mlx_lm/models/kimi_linear.py
+++ b/mlx_lm/models/kimi_linear.py
@@ -350,8 +350,11 @@ class KimiDeltaAttention(nn.Module):
         v = v_conv.reshape(B, T, self.num_heads, self.head_dim)
 
         inv_scale = self.scale
-        q = (inv_scale**2) * mx.fast.rms_norm(q, None, 1e-6)
-        k = inv_scale * mx.fast.rms_norm(k, None, 1e-6)
+        # Match the reference l2norm: x / sqrt(sum(x^2) + 1e-6). rms_norm adds
+        # eps to mean(x^2), so the equivalent eps is 1e-6 / head_dim.
+        eps = 1e-6 / self.head_dim
+        q = (inv_scale**2) * mx.fast.rms_norm(q, None, eps)
+        k = inv_scale * mx.fast.rms_norm(k, None, eps)
 
         a_logits = self.f_b_proj(self.f_a_proj(x)).reshape(
             B, T, self.num_heads, self.head_dim


Exorust:
Verified this independently on an M5 today. 

I started by checking the identity first in float64: `x / sqrt(mean(x^2) + 1e-6/D)` against `sqrt(D) * x / sqrt(sum(x^2) + 1e-6)` agrees to ~1e-17 at D=128, it's teh same as the ref l2norm.

When i ran it based on the repro, at amplitude 1e-4 the max deviation against the reference drops from 8.69e-01 to 1.53e-07 with this patch. This adds up to the table. 


But since k feeds the recurrent state, this is worth fixing because tthe error from a near-0k vector lands on later tokens outputs

avlp12:
Thanks for taking the time to reproduce this independently — good to see the 1e-4 case land on the same numbers (8.69e-01 → 1.53e-07), and for confirming the identity holds in float64.

For anyone picking this up: the affected lines are unchanged on `main`, so the fix still applies as-is, and the change is confined to two lines in `kimi_linear.py`.

@awni — would you have a chance to take a look? Happy to rebase or add a regression test if you'd prefer either.

zcbenz (changes requested):
Thanks for noticing this.

> This makes rms_norm exactly equivalent to the reference l2norm at all magnitudes:
> 
> x / sqrt(mean(x²) + 1e-6/D) == sqrt(D) · x / sqrt(sum(x²) + 1e-6)

I don't think stands true, of course it makes error much less though. I would prefer reverting the change https://github.com/ml-explore/mlx-lm/pull/853/changes#diff-75f4a9b1351d9fd0a566964dbf51a532a69c82a46eb055cea599d6f05b3f2113L364

avlp12:
Thanks for the review! Two points, and I'm happy to rework the PR either way.

**On the identity.** You're right that "exactly" was sloppy wording for what a
float kernel does — but the identity itself is plain algebra, not an
approximation:

```
sqrt(mean(x²) + ε/D) = sqrt((sum(x²) + ε)/D) = sqrt(sum(x²) + ε) / sqrt(D)
```

so `rms_norm(x, None, ε/D) = sqrt(D) · x / sqrt(sum(x²) + ε)` holds in real
arithmetic (all three `fast.rms_norm` backends add eps to the mean inside the
rsqrt), with the `sqrt(D)` absorbed by the `inv_scale` powers already in the
code. The ~1e-7 residual in my table is fp32 rounding noise, not a semantic
gap. For completeness I also checked FLA's chunked path — `fla/modules/
l2norm.py` uses the same `1/sqrt(sum(x²) + 1e-6)` form as the fused-recurrent
kernel, so both prefill and decode reference paths agree on the semantics.

**On reverting to the pre-#853 `_l2norm`.** I'd gently push back on that one:
it adds eps *outside* the sqrt — `x / (‖x‖ + 1e-6)` — which is a third
semantics, also different from the FLA kernels (`x / sqrt(‖x‖² + 1e-6)`, eps
inside on the sum). As ‖x‖ → 0 its denominator tends to 1e-6 instead of the
reference's 1e-3, so near-zero vectors get amplified ~1000×. Measured against
the FLA form (fp32, D=128, max|Δ|/max|ref|, order-of-magnitude stable across
seeds):

| amplitude | this PR (ε/D) | pre-#853 `x/(‖x‖+ε)` |
|---:|---:|---:|
| 1.0 | ~1e-07 | ~2e-07 |
| 1e-2 | ~1e-07 | ~4e-05 |
| 1e-4 | ~1e-07 | ~0.4 |
| 1e-6 | ~1e-07 | ~70× |

So the revert would trade the current mismatch for a different (larger)
small-magnitude one.

If the `ε/D` trick reads too clever, the reference can also be implemented
literally — note the scale placement has to change with it, since the
explicit l2norm doesn't carry `rms_norm`'s internal `sqrt(D)` factor:

```python
q = self.scale * q * mx.rsqrt((q * q).sum(-1, keepdims=True) + 1e-6)
k = k * mx.rsqrt((k * k).sum(-1, keepdims=True) + 1e-6)
```

(i.e. the pre-#853 scale structure — `scale` once on q, none on k — with eps
moved inside the sqrt.) That matches the reference the same way this PR does;
the one thing the `fast.rms_norm` form keeps is fp32 accumulation on bf16
inputs. Happy to update the PR to either form — just say the word.

zcbenz (approved):
Thanks for clarification, that makes sense!

PR #1625, opened 2026-07-28
Fix sharing directories with symlinks

cupkk:
# PR Draft

**Title:** Fix sharing directories with symlinks

`DirectoryEntry.from_path()` currently classifies entries with a lookup keyed by `is_dir()` and `is_symlink()`. A symlink to a directory makes both predicates true, so sharing a model directory that contains one fails with `KeyError((True, True))`.

This changes the classification to check symlinks before directories and keeps `dst=None` for entries that are not symlinks. The new test covers a relative directory symlink through `get_files()`.

Local verification:

- reproduced the pre-fix `KeyError` from the exact class definition
- ran the fixed `DirectoryEntry` and `get_files` definitions against a temporary directory
- ran `compileall` and `git diff --check`
- verified the patch applies cleanly

The full test suite was not run locally because this minimal environment does not have MLX installed. The change and test were prepared with AI assistance and manually inspected.

### DIFF
--- a/mlx_lm/share.py
+++ b/mlx_lm/share.py
@@ -44,14 +44,17 @@ class DirectoryEntry:
 
     @classmethod
     def from_path(cls, root, path):
-        entry_type = {
-            (True, False): "directory",
-            (False, True): "symlink",
-            (False, False): "file",
-        }[path.is_dir(), path.is_symlink()]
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1626, opened 2026-07-28
Add Kimi K3

kernelpool:
## Requirements

- AI usage disclosure: This implementation was AI assisted using Claude Fable for researching spec/reference and optimizing gpu kernels. The code has been manually reviewed by the author.

## Model information

- Name: Kimi K3
- Homepage: https://www.kimi.ai/blog/kimi-k3
- Reference implementation:
  - https://huggingface.co/moonshotai/Kimi-K3/blob/main/modeling_kimi_k3.py
  - https://huggingface.co/moonshotai/Kimi-K3/blob/main/modeling_kimi_linear.py

The implementation was cross-checked against sglang and vLLM
- https://github.com/sgl-project/sglang/pull/32541
- https://github.com/vllm-project/vllm/pull/50000

## Supported checkpoints

- https://huggingface.co/moonshotai/Kimi-K3

## Verification

- mlx: 0.32.1
- mlx-lm: this branch (`add-kimi-k3`)
- macOS: 26.5 (25F71)
- Hardware: 2x Mac Studio M3 Ultra
- Python: 3.12.8

Note: I've used a [2-bit mixed quantized model](https://huggingface.co/kernelpool/Kimi-K3-2bit-UVMAX) for this test.

```
mlx.launch --verbose --backend jaccl --hostfile hosts-jaccl.json --env MLX_METAL_FAST_SYNCH=1 --env PYTHONPATH=/Users/optimus/repo/mlx-lm-prs/add-kimi-k3 -- /Users/optimus/repo/mlx-lm/.venv/bin/python /Users/optimus/repo/mlx-lm/mlx_lm/examples/sharded_generate.py --model /Users/optimus/models/catalyst/Kimi-K3-2bit-UVMAX -p "The secret to baking a good cake is" -m 1024
[INFO] Running /Users/optimus/repo/mlx-lm/.venv/bin/python /Users/optimus/repo/mlx-lm/mlx_lm/examples/sharded_generate.py --model /Users/optimus/models/catalyst/Kimi-K3-2bit-UVMAX -p 'The secret to baking a good cake is' -m 1024 
[jaccl] Connection attempt 0 waiting 1000 ms
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
Calling super().encode with {'add_special_tokens': False}
The user has sent a very short prompt: "The secret to baking a good cake is"

This is an open-ended prompt that seems to invite me to complete the sentence or share knowledge about cake baking. It's a casual, friendly question about baking. There's no ambiguity about what they're asking - they want to know what makes a good cake or what the "secret" is to baking one well.

Let me think about what actually makes a good cake from a baking perspective:

1. **Proper ingredient temperatures** - Room temperature eggs, butter, and dairy mix together more smoothly and create better emulsion, leading to better texture.

2. **Accurate measurements** - Baking is chemistry. Weighing ingredients (especially flour) is more accurate than volume measurements.

3. **Proper mixing technique** - Overmixing develops gluten and makes cakes tough; undermixing leaves lumps. The creaming method (butter + sugar) incorporates air for lightness.

4. **Don't overbake** - Cakes continue cooking after removal from oven. A toothpick with a few moist crumbs is better than completely clean.

5. **Quality ingredients** - Real vanilla, good butter, fresh leavening agents (baking powder/soda lose potency).

6. **Proper oven temperature** - An oven thermometer helps; many ovens run hot or cold.

7. **Pan preparation** - Greasing and flouring (or parchment) prevents sticking.

8. **Following the recipe** - Baking isn't the place for improvisation until you understand the science.

9. **Let ingredients come to room temp** - Cold ingredients don't emulsify properly.

10. **The "secret" answer** - Many bakers would say the secret is: room temperature ingredients, proper creaming, not overmixing, and not overbaking.

There's also a playful angle - some might joke the secret is "love" or "a glass of wine while baking" or something whimsical. But I think the user wants genuine baking advice.

Format considerations: This is a casual conversational prompt. The user gave me an incomplete sentence, inviting completion. I could:
- Complete the sentence directly with a punchy answer
- Give a short list of key secrets
- Keep it conversational

I think the best response is conversational but informative. Maybe lead with a direct answer to the "secret" and then support with a few key points. I shouldn't write a massive essay - this is a light, casual prompt. A medium-length response with maybe a few bullet points or a short paragraph would work well.

Let me draft something:

"The secret to baking a good cake is... temperature control and restraint! 

A few things that separate good cakes from mediocre ones:
- Room temperature ingredients (butter, eggs, milk) - they emulsify better
- Don't overmix once flour is added - develops gluten, makes it tough
- Don't overbake - pull it when a toothpick has a few moist crumbs
- Cream butter and sugar properly - this is where the air comes from
- Weigh your flour instead of scooping

Honestly, if I had to pick just one: it's not overbaking. Most home bakers pull cakes too late."

That's a decent structure. Let me refine. I want to be helpful and give genuinely useful baking wisdom. The tone should match the casual prompt.

Actually, I could also lead with the classic answer: "room temperature ingredients" or "proper technique." Let me think about what the actual "secret" most pastry chefs would cite...

Common answers:
- "Room temperature ingredients" - very common
- "Don't overmix" - very common
- "Measure accurately / weigh ingredients"
- "Cream the butter and sugar long enough"
- "Use an oven thermometer"
- "Don't open the oven door early"

I'll give a concise, punchy answer with the top secrets. Keep it friendly and not too long. Maybe 100-150 words with a short list. That feels right for this prompt.<|close|>think<|sep|><|open|>response<|sep|>...not really a secret at all — it's a few small things done right:

- **Room-temperature ingredients.** Cold butter, eggs, and milk don't blend smoothly, and you end up with a dense, uneven crumb.
- **Cream the butter and sugar properly.** This is where the cake gets its air. A minute or two isn't enough — give it 4–5 minutes until it's pale and fluffy.
- **Don't overmix once the flour goes in.** That's when gluten develops and cakes turn tough. Mix just until combined.
- **Don't overbake.** Pull it when a toothpick comes out with a few moist crumbs, not bone-dry clean. Carryover heat finishes the job.
- **Weigh your flour.** Scooping from the bag can add 20–30% more flour than a recipe intends.

If I had to pick just one? **Don't overbake.** More homemade cakes are ruined by an extra five minutes in the oven than by anything else.

What kind of cake
==========
Prompt: 96 tokens, 23.767 tokens-per-sec
Generation: 1024 tokens, 11.737 tokens-per-sec
Peak memory: 420.215 GB
[INFO] Node with rank 1 completed 
[INFO] Node with rank 0 completed 
```

## Reference tests

I don't have access to CUDA hardware, but fwiw I've collected logprobs for a number of generation continuations (using temperature 0) from an external Kimi K3 provider (together.ai) and compared these to the implementation loading the full model (disk streamed).

| prompt | context tokens | scored tokens | mean delta logprob | max delta logprob | top-1 match | provider token in local top-5 |
|---|---:|---:|---:|---:|---:|---:|
| short_fact | 109 | 64 | 0.028 | 0.615 | 63/64 | 64/64 |
| short_code | 118 | 64 | 0.032 | 0.662 | 63/64 | 64/64 |
| short_reasoning | 120 | 64 | 0.005 | 0.075 | 64/64 | 64/64 |
| long_memory | 3364 | 64 | 0.021 | 0.211 | 64/64 | 64/64 |
| long_code_audit | 3847 | 64 | 0.031 | 0.324 | 63/64 | 64/64 |
| **total** | | **320** | | | **317/320** | **320/320** |

On the short_fact miss, the question was "In one sentence: what is the capital of Australia, and in which year did it become the capital?" and it diverged mid-reasoning on a sentence-opener emitting "Many" vs "It".

## Benchmarks

benchmark_chart


guruswami-ai:
**Independent validation on 4× M3 Ultra in TP4, plus one note for anyone applying this to the current release.**

We ran this PR's model file on a 4-node Apple silicon cluster and it works. Some
data in case it is useful, and one compatibility observation.

### What was actually tested — please read before weighing this

**We did not run the complete PR on `main`.** The deployment is mlx-lm
**v0.31.3** with this PR's model, tokenizer and tool-parser changes backported
from `7d505c285b801108a52c23353c7fb6af07204717`. Because v0.31.3 lacks
`make_text_state_machine`, it also carries a **local streaming-aware
response-layer marker filter** in place of your `generate.py` change.

So these fixtures validate our backport plus this PR's model and tool-parser
pieces. They do **not** validate your `generate.py` marker handling directly.

### Setup

- 4× Mac Studio M3 Ultra 512 GB, macOS 26.6, Thunderbolt 5 full mesh, JACCL/RDMA
- Kimi-K3 2.78 T MXFP4, **tensor-parallel across 4 ranks**, 420.8 GB/rank
- `prefill_step_size=512`, prompt cache size 1

### What worked

A four-fixture suite over the HTTP API, all passing:

| fixture | result |
|---|---|
| non-streaming | `content` exactly `391`, no XTML control tokens |
| streaming | same, clean across SSE chunk boundaries |
| reasoning separation | answer in `content`, reasoning in its own channel |
| tool call | structured `tool_calls` → `get_weather(city="Tokyo")` |

Also: load ~5 min at 420.8 GB/rank; prefill throughput flat at ~108–114 tok/s
from 2.5 k to 38.5 k tokens, which is the expected shape when 69 of 93 layers
are KDA linear attention.

**Durability boundary:** this validates model loading, TP4 sharding and
wire-level protocol correctness — **not sustained serving reliability**. Longer
JACCL soaks hit a separate MLX RDMA failure, reported on
[mlx#3933](https://github.com/ml-explore/mlx/pull/3933). No model-output
correctness failure was observed before transport failure.

### `A_log` — independent confirmation that your handling is right

We initially diverged from this PR here and were **wrong**, so it seems worth
recording the evidence.

We reasoned from shapes: released shards store `A_log` as `[128]`, `num_heads`
is 96, `head_dim` is 128 — so we treated it as per-channel and replicated it
across ranks. Then we read the values instead of reasoning about them, across
**all 69 KDA tensors**:

```
shapes: {(128,)}
nonzero-count distribution: [96]
tensors whose [96:] is EXACTLY zero: 69/69
layer 0 A_log[92:100] = [-0.5055, -0.0233, -0.4065, -0.1461, 0.0, 0.0, 0.0, 0.0]
```

96 live values padded to `head_dim`. Your `[:num_heads]` truncation,
`(num_heads, 1)` reshape and per-head sharding are correct; our reading was not.
Worth noting the failure is **silent** either way — 128 divides evenly by a
world size of 4, so every shape still broadcasts and nothing raises.

The one thing we would suggest: **assert the tail is zero before truncating**,
rather than truncating whenever `shape[0] > num_heads`. A future checkpoint with
a non-zero tail would otherwise be silently discarded.

### Compatibility note for the current release

The marker-stripping relies on `structural_markers` in
`make_text_state_machine`. That function exists in `main`, so this PR is
self-consistent — but it is **not in v0.31.3**, which is still the latest
release. **Backporting #1626 to v0.31.3 without its `generate.py` change**
therefore leaves the response markers in `content`:

```
'<|open|>response<|sep|>391<|close|>response<|sep|><|close|>message<|sep|>'
```

Reasoning/response separation still works there; only the response body is
affected. We worked around it with a streaming-aware exact-marker filter at the
response layer — it has to buffer partial marker prefixes, since a marker can
span SSE chunks and a per-chunk `str.replace` misses `"<|op"` + `"en|>"`.

Not a defect in this PR, but it may be worth a line in the description, since
anyone trying K3 on the current release will hit it and the output looks
plausible enough to miss.

### Offer

Happy to share the fixture suite, or to re-run any of this on the 4-node TP4
setup if that would help review.

guruswami-ai:
Thanks again for the Kimi-K3 implementation. We have a small tests-led follow-up
prepared from an independent TP4 validation, if it would be useful. It is based
on `43f5e87` and is three commits, the first of which carries the only proposed
production change: a guard against silently discarding a non-zero `A_log`
padding tail.

The coverage is deliberately independent of the full checkpoint — no downloads,
and only the distributed file needs more than one process:

- **`A_log` padding** — the released checkpoint's zero-padded tail is accepted
  and trimmed to the live prefix; a non-zero tail raises instead of being
  truncated away, naming the tensor key, observed length and expected head
  count. `sanitize()` currently truncates unconditionally.
- **Kimi tensor-parallel equivalence** on a randomized tiny model at TP2 and
  TP4, with direct assertions on the KDA `A_log` and `dt_bias` slices, the
  depthwise convolution groups and weight rows, `projection_dim`, and that
  head-dimension-only parameters (`o_norm`, `f_a_proj`) keep their unsharded
  values. Full rank tensors are compared rather than a checksum, and cached
  prefill plus a decode step are compared against an unsharded reference so the
  recurrent KDA state and MLA KV cache are covered.
- **A separate MoE fixture** for the latent-MoE reduction. This one needs
  explaining: in the main fixture the control for "reduce after normalize"
  cannot fire at all. RMSNorm is `w·y/√(mean(y²)+eps)`, and at a conventional
  ±0.02 init the routed activations sit far below `√eps`, so the norm is nearly
  linear — and a linear map commutes with summation. The separate fixture keeps
  the real epsilon and drives the routed tensor to ~158× `√eps`, where correct
  sharding gives `max_abs 2.98e-07` and moving the norm before the all-reduce
  gives `7.096e-01`. The ratio is asserted so the fixture cannot quietly lose its
  power.
- **XTML streaming state machine** — every structural marker fed split at every
  internal offset across two `step()` calls, plus preservation of legitimate
  partial-prefix text (`"a<|"` then `"b"` → `"a<|b"`). The markers are inferred
  from the vocabulary through `TokenizerWrapper` rather than hardcoded.

We ran negative controls for wrong `A_log` slicing, stale convolution groups and
the MoE reduction ordering, so the tests demonstrably detect the mistakes they
are meant to guard against. The wrong-slice control is the interesting one: it
produces `max_abs 2.3e-10` and passes a numerical comparison, and is caught only
by the structural assertions.

Tolerances are `rtol=1e-5, atol=1e-6`; the whole-model comparison measured `0.0`
at TP2 and `1.164e-10` at TP4, and the isolated MoE comparison `2.980e-07` at
both. TP8 was not run — we only have four nodes. Black 25.1.0 and isort 6.0.0
(`--profile=black`) pass on every touched file.

The three commits are independent and touch disjoint files, so they can be taken
separately if only some are of interest.

The PR already has good parser-level Kimi tool-call cases, so we have not
duplicated those. Would you prefer cherry-pickable commits based on the current
`add-kimi-k3` head, or a follow-up PR after #1626 merges?

zcbenz (commented):

mlx_lm/models/kimi_k3.py
```diff
@@ -0,0 +1,1312 @@
...
+        input_names=["x", "state", "w"],
+        output_names=["y", "new_state"],
+        source=_SHORT_CONV_SOURCE,
+    )
+    if mx.metal.is_available()
+    else None
+)
+
+
+class KimiK3ShortConv(ShortConv1d):
```
zcbenz: Maybe this should just be the default `kimi_linear.ShortConv1d` implementation?

mlx_lm/generate.py
```diff
@@ -1051,6 +1051,9 @@ def make_text_state_machine(tokenizer, stop_words=None):
             [(tokenizer.tool_call_end, "normal")] if tokenizer.tool_call_end else []
         )
 
+    for w in getattr(tokenizer, "structural_markers", None) or ():
+        transitions.setdefault("normal", []).append((w, "normal"))
```
zcbenz: I would prefer explicit code here

```suggestion
    if tokenizer.structural_markers:
        for w in tokenizer.structural_markers:
            transitions.setdefault("normal", []).append((w, "normal"))
```

mlx_lm/tokenizer_utils.py
```diff
@@ -623,6 +675,8 @@ def load(
     tool_parser_type = tokenizer_config.get(
         "tool_parser_type", _infer_tool_parser(tokenizer.chat_template)
     )
+    if tool_parser_type is None and _is_xtml_vocab(tokenizer.get_vocab()):
```
zcbenz: I think this should be part of `_infer_tool_parser`, maybe passing the whole `tokenizer` to it.

mlx_lm/tokenizer_utils.py
```diff
@@ -311,11 +340,28 @@ def __init__(
...
+            or chat_template is not None
+            or custom_renderer
         )
+        self._thinking_kwarg = "enable_thinking"
+        if custom_renderer:
+            try:
+                params = inspect.signature(
+                    type(tokenizer).apply_chat_template
+                ).parameters
+                if "thinking" in params and "enable_thinking" not in params:
```
zcbenz: There are too many details exposed, I think we should isolate it to separate function like:

```python
self._thinking_kwarg, has_custom_renderer = _infer_thinking_kwarg(tokenizer)
```

mlx_lm/models/kimi_k3.py
kernelpool: Ah yes, there was not much to gain from that decode kernel in practice so worthwhile reducing complexity

mlx_lm/tokenizer_utils.py
kernelpool: Agreed, much cleaner

chillosu:
**Bug + tested fix: MLA attention crashes with a quantized KV cache (`--kv-bits`)**

Found while running this PR's `kimi_k3.py` in TP4 across 4× M3 Ultra (MXFP4 weights, `mlx_lm.server`). Any request that takes the non-batch path with `--kv-bits 4` crashes:

```
File ".../mlx_lm/models/kimi_k3.py", line 613, in __call__
```

`QuantizedKVCache.update_and_fetch` returns `(packed, scales, biases)` tuples, and the MLA code calls array ops on them in three places: the rope-score matmul, the `L > 1` latent projections (`embed_q`/`unembed_out`), and implicitly the SDPA dispatch. Since quantized KV is the main lever for long-context K3 (fp16 latent cache is ~0.22 GiB per 1K tokens at TP4 in our measurements), this path matters.

One subtlety worth flagging: the obvious fix — route decode through `quantized_scaled_dot_product_attention` — is wrong for MLA, because that helper mis-broadcasts an additive float mask when `n_kv_heads == 1`: with `n_repeats > 1` it reshapes scores to `(B, 1, H, L, S)` but expands the 4-D mask at `-3` to `(B, H, 1, L, S)`, broadcasting to `(B, H, H, L, S)` and mixing heads. So the patch below does the two quantized matmuls directly for decode, and dequantizes the latent for `L > 1` prefill chunks (same transient footprint as the fp16 path; the storage saving between steps is retained).

```python
# rope scores
if isinstance(k_pe, (tuple, list)):
    pe_scores = mx.quantized_matmul(
        q_pe * self.scale, *k_pe, transpose=True,
        group_size=cache.group_size, bits=cache.bits,
    )
else:

# attention
quantized_kv = isinstance(kv_latent, (tuple, list))
if L == 1:
    q_nope = self.embed_q(q_nope)
    k = v = kv_latent
else:
    if quantized_kv:
        kv_latent = mx.dequantize(
            *kv_latent, group_size=cache.group_size, bits=cache.bits
        )
        quantized_kv = False
    k = self.embed_q(kv_latent, transpose=False)
    v = self.unembed_out(kv_latent)

if quantized_kv:
    # Manual quantized MLA attention (see mask-broadcast note above).
    scores = mx.quantized_matmul(
        q_nope * self.scale, *k, transpose=True,
        group_size=cache.group_size, bits=cache.bits,
    )
    scores = scores + pe_scores
    weights = mx.softmax(scores, axis=-1, precise=True)
    output = mx.quantized_matmul(
        weights, *v, transpose=False,
        group_size=cache.group_size, bits=cache.bits,
    )
else:
    output = mx.fast.scaled_dot_product_attention(
        q_nope, k, v, scale=self.scale, mask=pe_scores
    )

if L == 1:
    output = self.unembed_out(output)
```

Tested on the 4-node TP4 cluster (mlx 0.32.2, this PR's model code): with `--kv-bits 4 --kv-group-size 64`, the previously-crashing path now generates correctly (greedy exact-match canaries pass, `finish_reason: stop`), and the unquantized batch path is byte-identical in behavior. Happy to turn this into a PR against `kernelpool:add-kimi-k3` if that's easier — we're currently running a ~950K-token needle certification on this exact code path and can report long-context results here too.

chillosu:
**Follow-up: the quantized-KV fix above is now certified at ~949K tokens.**

As promised — we ran a needle-retrieval certification through the patched path on the 4× M3 Ultra TP4 cluster (this PR's model code + the fix from my previous comment, mlx 0.32.2, `mlx_lm.server` with `--kv-bits 4 --kv-group-size 64 --prefill-step-size 512 --prompt-concurrency 16`):

- **Prompt: 948,888 tokens** (single request, needle codeword planted in the first sentence)
- **Result: needle retrieved verbatim**, `finish_reason: stop`, with the reasoning quoting the source sentence from the top of the corpus
- Cold prefill wall: 22h36m (dispatch/bandwidth-bound at depth; rate decays ~n^-0.5 from 139 tok/s at 12K to ~6 tok/s at 900K)
- Peak wired ~487GB of a 506GB `iogpu.wired_limit_mb` on 512GB nodes — only viable with the quantized cache; fp16 KV is arithmetically impossible at this depth on this hardware

Two operational notes for anyone running long contexts on this code:

1. **The Metal allocator retains stale size-class buffers across prefill chunks** — wired memory grows ~0.2-0.4 GB per 1K prompt tokens even though transients are freed, which OOMs long prefills. Calling `mx.clear_cache()` once per prefill chunk (we hooked the server's progress callback) flattens it to a healthy sawtooth. Might be worth a server option.
2. **Mixing a batch-engine request with a concurrent non-batch (seed-bearing) request deadlocked all four ranks** — collective busy-wait at 99-100% CPU, GPU idle, no recovery (the distributed backends have no collective timeouts). Easy to reproduce: start a long seeded request, then send a plain chat request mid-flight. Worth a guard serializing the two paths.

Happy to share the full telemetry (prefill rate curve, wired-memory traces, per-attempt failure ledger) if useful.

---

**Correction on finding 2 (deadlock repro), with better data:** we since learned our 22h seeded cert request ran its final ~8 hours with a batch-path completion firing every ~10 minutes (~48 events) and did NOT deadlock — steady-state mixing is evidently tolerated. The deadlock we hit occurred when a batch request landed while a seeded request was in its startup/prompt-cache-restore phase. So the hazard window appears to be the single-path request's initialization, not its whole lifetime — "easy to reproduce" above overstates it. A guard serializing the paths is still the right fix, but repro attempts should target request startup.

chillosu:
**Retraction of my previous correction — the "steady-state mixing is tolerated" claim was based on data I misread.**

In [my earlier comment](https://github.com/ml-explore/mlx-lm/pull/1626#issuecomment-5471673106) I walked back the deadlock repro, saying our 22h seeded certification run had ~48 concurrent batch-path completions interleaved without deadlocking, and concluded the hazard window was single-path *initialization* rather than steady-state coexistence.

That was wrong. An audit of the rank-0 log for the certification run (`server-mac3-20260829-161756.log`, the log covering the full 22h36m request) shows **three** HTTP requests in its entire life: a `GET /v1/models`, a 12K warm-up, and the certification itself. The single gateway request from another host arrives at 15:14:35 — **fifteen minutes after the certification completed at 14:59:46**. I had counted the canary cadence from the *next* server log and projected it backwards onto the certification window.

So there is no evidence for steady-state tolerance. **My original report stands: mixing a batch-engine request with a concurrent non-batch (seed-bearing) request deadlocked all four ranks, and anyone attempting a repro should not restrict it to request startup.** I don't have a clean characterization of the hazard window, and I won't offer one again until I've run a deliberate mixing test rather than inferring from an uncontrolled run.

Two other corrections to the same comment while I'm here:

- **Prefill decay exponent.** I wrote "~n^-0.5". Measured from 1,856 progress lines in the certification log, the correct figure is **~n^-0.73**: 139 tok/s @ 12K, 80 @ 50K, 48 @ 100K, 28 @ 200K, 14 @ 400K, 9.3 @ 600K, 7.0 @ 800K, 6.1 @ 900K (step-512, 4-bit KV). n^-0.5 would have predicted 16 tok/s at 900K — 2.7x the observed value, and it contradicted the two endpoints I quoted in the same sentence.
- **Energy.** I have not cited energy figures here, but for anyone who saw them elsewhere: our numbers were derived from a whole-home meter during a window that included an EV charging session, and are not defensible as a cluster measurement.

Apologies for the noise. The certification result itself is unaffected — 948,888 prompt tokens, needle retrieved verbatim, `finish_reason: stop` — and the quantized-KV fix in my first comment is unchanged.

zcbenz (approved):
Thanks!

zcbenz:
Can you fix the failing test?

PR #1628, opened 2026-07-28
Fix FunctionGemma parser for escaped braces and multiple calls

ask-q-ai:
Port the string-aware balanced-brace approach used by the Gemma 4 parser to FunctionGemma.

This preserves braces, quotes, commas and multiline text inside `<escape>…<escape>` arguments, supports hyphenated function names and multiple calls, and rejects incomplete calls clearly.

Tests cover complete VisMan-like YAML, multiple calls, truncation, and 10,000 fixed-seed source variants.

### DIFF
--- a/tests/test_tool_parsing.py
+++ b/tests/test_tool_parsing.py
@@ -1,4 +1,5 @@
 import unittest
+from random import Random
 from pathlib import Path
 
 from mlx_lm.tool_parsers import (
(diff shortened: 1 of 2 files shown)


tests/test_tool_parsing.py
```diff
@@ -275,6 +276,64 @@ def test_gemma4(self):
...
+                        "language": "visman",
+                        "outputSlot": "scene.red",
+                        "source": source,
+                    },
+                },
+                {"name": "mark-ready", "arguments": {"ready": True}},
+            ],
+        )
+
+    def test_function_gemma_parses_ten_thousand_fixed_source_variants(self):
```
zcbenz, line 307: This makes the test very slow, a single test case would be enough.

zcbenz (changes requested):

zcbenz:

Issue #806, opened 2026-01-24
Add support for GLM-4 MoE architecture (glm4_moe_lite)


Smilefounder:
## Description

MLX-LM currently doesn't support the GLM-4 MoE (Mixture of Experts) architecture, specifically models with `model_type: glm4_moe_lite` in their config.

## Steps to Reproduce

```bash
mlx_lm.generate --model lmstudio-community/GLM-4.7-Flash-MLX-8bit
```

## Error Output

```
ERROR:root:Model type glm4_moe_lite not supported.
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages/mlx_lm/utils.py", line 71, in _get_classes
    arch = importlib.import_module(f"mlx_lm.models.{model_type}")
  File "/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ModuleNotFoundError: No module named 'mlx_lm.models.glm4_moe_lite'

During handling of the above exception, another exception occurred:
ValueError: Model type glm4_moe_lite not supported.
```

## Environment

- Python: 3.14
- - mlx_lm: (latest)
- - - macOS (Apple Silicon)
## Affected Models

Models from the GLM-4 family that use the MoE architecture:
- lmstudio-community/GLM-4.7-Flash-MLX-8bit
- - lmstudio-community/GLM-4.7-Flash-MLX-4bit
- - - Other GLM-4 MoE variants
## Request

Could support for the `glm4_moe_lite` architecture be added to mlx_lm? The GLM-4 family is gaining popularity and having MoE support would be valuable for the MLX community.

## References

- Model repository: https://lmstudio.ai/models/zai-org/glm-4.7-flash
- - Hugging Face: https://huggingface.co/THUDM/glm-4-9b-chat

awni:
It should be supported in the latest mlx-lm. Somehow you are not using the latests mlx-lm. Maybe try:

`python -m pip install -U mlx-lm`

trobriander:
I'm having the same problem as the OP, but with **lmstudio-community/GLM-4.7-Flash-MLX-4bit**

Current Version of mlx-lm:  **0.30.4**

However, I'm on MacOS 14. Is that the problem?

I asked an AI to help me diagnose the problem, used a debug script and it came up with:

```
Model Type in config: glm4_moe_lite
[...]

PR #1630, opened 2026-07-28
Fix logprobs being silently dropped from streamed responses

slipstr34m:
## Problem

`handle_completion` in `mlx_lm/server.py` accumulates logprobs for every token
(`server.py:1438-1442`):

```python
tokens.append(gen.token)
if args.logprobs:
    token_logprobs.append(gen.logprob)
if args.top_logprobs > 0:
    top_tokens.append(gen.top_tokens)
```

The non-streaming response passes them on (`server.py:1496`):

```python
resp = self.generate_response(
    text, finish_reason, len(ctx.prompt), len(tokens), ctx.prompt_cache_count,
    token_logprobs=token_logprobs, top_tokens=top_tokens, tokens=tokens, ...
)
```

The two streaming responses (`server.py:1449` and `1474`) do not:

```python
resp = self.generate_response(
    text, None, tool_calls=..., reasoning_text=...,   # nothing else
)
```

So for an identical request with `logprobs: true, top_logprobs: 3, max_tokens: 8`:

- without `stream`: `choices[0].logprobs.content` has 8 entries, each carrying the
  chosen token, its logprob and 3 alternatives
- with `"stream": true`: no `logprobs` key appears in any chunk

The values are computed per token and then discarded. `logprobs` is accepted and
validated (`server.py:1193`), so the client gets a 200 with no indication the
parameter was ignored: there is no signal at any layer.

If streaming logprobs were deliberately unsupported, the right behaviour would be
a 400. Accepting the parameter and ignoring it is the one option nobody picks on
purpose.

## Why this matters

I hit this routing between a local model and a frontier API. The decision is made
on token-level confidence while the response is still being generated: when the
top two candidates are close: at the token carrying a tool name, or across a
span the small model is unsure about: that request escalates. Without logprobs
on the streaming path you have to wait for the whole completion before you can
act, which defeats the purpose on a backend generating at tens of tokens per
second.

## Fix

Pass the tokens accumulated since the previous chunk:

```python
resp = self.generate_response(
    text, None,
    token_logprobs=token_logprobs[streamed:],
    top_tokens=top_tokens[streamed:],
    tokens=tokens[streamed:],
    ...
)
streamed = len(tokens)
```

The two logprob lists grow in lockstep with `tokens` or stay empty, so a single
offset slices all three. `generate_response` already builds the `logprobs` block
regardless of whether the response is streamed, so nothing changes there.

Non-streaming behaviour is untouched, and a request that does not ask for
logprobs slices empty lists: a no-op.

## Testing

- Added a regression test in `tests/test_server.py` asserting that the logprobs
  reassembled from a stream equal the non-streamed ones for the same request
  (fails on `main`, passes with this change).
- `python -m unittest tests.test_server`: 26 tests, all pass.
- Checked against a running server with the OpenAI Python client on
  Qwen1.5-0.5B, Llama-3.2-1B and LFM2-1.2B, and on the sequential generation path
  via a request with `seed` set. Before: 0 streamed entries in every case. After:
  identical to the non-streamed ones, token strings and `top_logprobs` included.
- `black` / `isort` clean.

Note: the parameters arrived in #806 ("Logprobs info to completion API"), which
added the plumbing and updated the single non-streaming call site; the streaming
ones were never wired up. Separately, `SERVER.md` documents `logprobs` as "an
integer between 1 and 10" while the code takes a bool and puts the integer in
`top_logprobs`: orthogonal to this fix and left untouched here, happy to follow
up.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1402,6 +1402,10 @@ class APIHandler(BaseHTTPRequestHandler):
         tokens = []
         token_logprobs = []
         top_tokens = []
+        # Tokens whose logprobs have already gone out in a stream chunk. The
+        # logprob lists grow in lockstep with `tokens` (or stay empty), so one
+        # offset slices all three.
+        streamed = 0
 
         try:
             for gen in response:
@@ -1449,6 +1453,9 @@ class APIHandler(BaseHTTPRequestHandler):
                     resp = self.generate_response(
                         text,
                         None,
+                        token_logprobs=token_logprobs[streamed:],
+                        top_tokens=top_tokens[streamed:],
+                        tokens=tokens[streamed:],
                         tool_calls=tool_formatter(tool_calls),
                         reasoning_text=reasoning_text,
                     )
@@ -1457,6 +1464,7 @@ class APIHandler(BaseHTTPRequestHandler):
                     reasoning_text = ""
                     text = ""
                     tool_calls = []
+                    streamed = len(tokens)
 
                 if gen.finish_reason is not None:
                     finish_reason = gen.finish_reason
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1632, opened 2026-07-28
Fix unbounded ArraysCache metadata graph during decode

LukaAfonin:
## Summary

- bound the lazy graph retained by `ArraysCache.lengths` and
  `ArraysCache.left_padding`
- evaluate metadata together with an existing cache state
- add a regression test covering repeated `advance()` calls

## Problem

`ArraysCache.advance()` performs lazy subtraction on every decode step. The
metadata arrays retain the preceding graph even when the generation loop
evaluates the model state, so long continuous generations accumulate live
Metal resources and can eventually fail with:

```text
RuntimeError: [metal::malloc] Resource limit (499000) exceeded
```

In a minimal reproduction, 256 calls to `advance(1)` leave 1,280 graph edges
behind each metadata array.

## Fix

Make one populated cache state depend on the current metadata arrays. Existing
generation paths already evaluate cache state, which bounds the metadata graph
without adding a separate synchronization on every decode step.

## Testing

- the regression test fails on unpatched `main` with `1280 not less than or
  equal to 8`
- `python -m unittest tests.test_prompt_cache` (23 tests)
- `pre-commit run --files mlx_lm/models/cache.py tests/test_prompt_cache.py`
- 50,000-step synthetic decode: both metadata graphs remain at 0 edges and
  retain the expected values

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -688,6 +688,15 @@ class ArraysCache(_BaseCache):
         if self.left_padding is not None:
             self.left_padding -= N
 
+        metadata = tuple(
+            value for value in (self.lengths, self.left_padding) if value is not None
+        )
+        if metadata:
+            for index, value in enumerate(self.cache):
+                if value is not None:
+                    self.cache[index] = mx.depends(value, metadata)
+                    break
+
     def make_mask(self, N: int):
         if self.left_padding is not None:
             pos = mx.arange(N)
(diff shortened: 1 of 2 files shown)


pierre427:
Validation data point: we shipped exactly this `mx.depends` mechanism in our serving fork (with a test asserting the `lengths`/`left_padding` graphs stay under a fixed edge count after 256 decode steps — fails before, passes after) and it eliminated the unbounded metadata growth and the downstream Metal object-limit crash (#1642's failure mode) on qwen3_5-family batch serving. The mechanism is sound and minimal; see my note on #1642 for how the two PRs relate.

zcbenz (approved):
I changed the code to take a different approach, please let me know if it is not working.

Issue #1572, opened 2026-07-15
One-shot `mx.eval(model.parameters())` of >300 GB models triggers IOGPU command-buffer watchdog timeout at load (crash; fixed by chunked eval)


IzD680:
## Summary

When `load_model()` / `sharded_load()` in `mlx_lm/utils.py` finish weight loading, they call a single `mx.eval(model.parameters())`. With very large models (~390 GB of parameters per device), this one-shot eval builds a single enormous Metal command buffer and trips the macOS GPU watchdog:

```
(Metal) Execution aborted: Caused GPU Timeout Error (00000002:kIOGPUCommandBufferCallbackErrorTimeout)
void IOGPUCommandQueue::retireCommandBuffer: Deny submissions/ignore app[python3.13] with 2 GPURestarts
```

The resulting `std::runtime_error` is thrown from `mlx::core::metal::EventImpl::check_error()` on an MLX scheduler worker thread and **escapes uncaught**, so the process hard-aborts (SIGABRT) instead of raising a catchable Python exception:

```
libc++abi: terminating due to uncaught exception of type std::runtime_error:
[METAL] Command buffer execution failed: Caused GPU Timeout Error
(00000002:kIOGPUCommandBufferCallbackErrorTimeout).
```

Crash report (python3.13, 2026-07-15, faulting thread 12):

```
abort → __cxa_throw → failed_throw
mlx::core::metal::EventImpl::check_error()
std::__1::__function::__func<mlx::core::Event::wait(mlx::core::Stream)::$_0, void ()>::operator()()
mlx::core::scheduler::StreamThread::thread_fn()
```

(In applications that wait on the event differently — e.g. the exo distributed runner — the same watchdog event is swallowed and manifests as a *silent* wedge: GPU spinning 100% in `Event::wait` with no error ever raised.)

**Evaluating the same parameter tree in slices of a few GB succeeds 100% of the time** (evidence below). We have this deployed as a local patch and it makes loading reliable.

## Environment

| | |
|---|---|
| Hardware | Apple M3 Ultra, 512 GB unified memory (2-node tensor-parallel pair) |
| macOS | 26.2 (25C56) — primary; also reproduced on 26.5.2 (see Related context) |
| MLX | 0.32.0 (`libmlx.dylib`) |
| mlx-lm | 0.31.3 |
| Python | 3.13, arm64 |
[...]

PR #1634, opened 2026-07-29
Materialize shards incrementally in save_model to avoid GPU watchdog timeout

### LINKED ISSUE #1633: save_model: 5 GB shards force one-shot eval of quantized weights, tripping the IOGPU watchdog (leaves a 0-byte shard)
### Environment

- `mlx-lm` 0.31.3, `mlx` 0.32.0
- Python 3.11.14, macOS 26.5.2 (Darwin 25.5.0), Apple M2 Pro, 32 GB
- Model: [`microsoft/MagenticBrain`](https://huggingface.co/microsoft/MagenticBrain) — 14.8B params stored in **float32** (59 GB on disk)

### Actual behavior

`mlx_lm.convert` quantizes successfully and then crashes inside the save step:

```
[INFO] Using dtype: bfloat16
[INFO] Quantizing
[INFO] Quantized model with 4.500 bits per weight.
  File ".../mlx_lm/convert.py", line 259, in main
    convert(**vars(args))
  File ".../mlx_lm/convert.py", line 166, in convert
    save(
  File ".../mlx_lm/utils.py", line 942, in save
    save_model(dst_path, model, donate_model=True)
  File ".../mlx_lm/utils.py", line 756, in save_model
    mx.save_safetensors(str(shard_path), shard, metadata={"format": "mlx"})
RuntimeError: [METAL] Command buffer execution failed: Caused GPU Timeout Error
(00000002:kIOGPUCommandBufferCallbackErrorTimeout).
```

Command:

```bash
python -m mlx_lm convert --hf-path microsoft/MagenticBrain \
  --mlx-path out/MagenticBrain-4bit \
  -q --q-bits 4 --q-group-size 64 --dtype bfloat16
```

**Reproduced twice, failing at the same point both times, with 22 GB free disk.** It is not an out-of-space error — that surfaces separately and distinctly as `RuntimeError: [write] Unable to write N bytes to file` (I hit that too, on a later run, and it is clearly a different failure).

The output directory is left in a **silently corrupt** state:

```
-rw-r--r--  5354381380  model-00001-of-00002.safetensors   # complete, loads fine
-rw-r--r--           0  model-00002-of-00002.safetensors   # zero bytes
```

`model.safetensors.index.json` is never written (it is emitted after the loop), so the directory fails later at load time rather than at save time.

### Expected behavior

Either the save completes, or it fails without leaving a partially-written model directory.

### Analysis

`save_model` builds shards with the default `MAX_FILE_SIZE_GB = 5` and hands each to `mx.save_safetensors`:

https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/utils.py#L757

```python
weights = dict(tree_flatten(model.parameters()))
shards = make_shards(weights)          # 5 GB per shard
...
for i in range(len(shards)):
    shard = shards[i]
    shards[i] = None
    ...
    mx.save_safetensors(str(shard_path), shard, metadata={"format": "mlx"})
```

After `quantize_model`, every quantized weight/scales/biases array is still lazy. `save_safetensors` is the first consumer, so it forces evaluation of **an entire 5 GB shard's worth of quantized tensors as one Metal command buffer**. On this hardware that single buffer exceeds the IOGPU watchdog window and the command buffer is killed.

This is the same class of failure as #1572, which reports a one-shot `mx.eval(model.parameters())` tripping the same watchdog at **load** time and being fixed by chunked evaluation. The difference here is that it happens on the **save** path, and at 14.8B rather than >300 GB — so it is reachable on ordinary hardware.

### Workaround (verified)

Pre-evaluating each shard's tensors in small batches before calling `save_safetensors` fixes it. With `mx.eval` in batches of 16 tensors, the same model saves cleanly:

```
shard 1/8:  92 tensors, 1.03 GB
shard 2/8: 131 tensors, 1.03 GB
...
shard 8/8:  65 tensors, 0.96 GB
done
```

The resulting model loads and generates correctly, and scores 0.731 overall on BFCL v4 AST evaluation (4 categories, n=160), so the weights are sound.

### Minimal repro without the 59 GB download

Builds a synthetic quantized stack big enough to produce one default-sized shard:

```python
import mlx.core as mx, mlx.nn as nn
from mlx.utils import tree_flatten
from mlx_lm.utils import make_shards

DIM, N = 8192, 24                      # ~6 GB once quantized
model = nn.Sequential(*[nn.Linear(DIM, DIM, bias=False) for _ in range(N)])
mx.eval(model.parameters())
nn.quantize(model, group_size=64, bits=4)

shards = make_shards(dict(tree_flatten(model.parameters())))   # default 5 GB
for i, shard in enumerate(shards):
    mx.save_safetensors(f"/tmp/shard-{i}.safetensors", shard, metadata={"format": "mlx"})
```

Increase `N` until a shard reaches the 5 GB boundary if it does not trip on the first try — the trigger is total lazy work inside a single `save_safetensors` call, so the threshold is hardware-dependent.

### Proposed fix

Evaluate incrementally inside `save_model` before writing each shard. This is a few lines, changes no file layout, and does not alter output for models that already save fine. Happy to open a PR — I have the patch and it is what produced the working conversion above.

A secondary point worth considering separately: writing the index only after all shards succeed (or cleaning up on failure) would avoid leaving a half-written directory behind.

axiom-of-choice:
Fixes #1633.

## Problem

Freshly quantized weights are lazy, so `mx.save_safetensors` is their first consumer and forces an entire shard's graph into a single Metal command buffer. With the default `MAX_FILE_SIZE_GB = 5` that buffer can exceed the GPU watchdog window and the save is killed:

```
[INFO] Quantized model with 4.500 bits per weight.
...
  File ".../mlx_lm/utils.py", line 756, in save_model
    mx.save_safetensors(str(shard_path), shard, metadata={"format": "mlx"})
RuntimeError: [METAL] Command buffer execution failed: Caused GPU Timeout Error
(00000002:kIOGPUCommandBufferCallbackErrorTimeout).
```

Quantization reports success and the crash happens afterwards, during the write. The directory is left in a state that looks plausible but is broken:

```
-rw-r--r--  5354381380  model-00001-of-00002.safetensors   # complete, loads fine
-rw-r--r--           0  model-00002-of-00002.safetensors   # zero bytes
```

`model.safetensors.index.json` is emitted after the loop, so it is never written and the failure only surfaces later, at load time.

Reproduced on an M2 Pro / 32 GB converting [`microsoft/MagenticBrain`](https://huggingface.co/microsoft/MagenticBrain) (14.8B stored in fp32) with `--q-bits 4 --q-group-size 64 --dtype bfloat16`. Twice, at the same point, with 22 GB free disk — so it is distinct from running out of space, which surfaces separately and unambiguously as `RuntimeError: [write] Unable to write N bytes to file`.

This is the same class of failure as #1572, which reports a one-shot `mx.eval(model.parameters())` tripping the same watchdog on the **load** path for >300 GB models. This one is on the **save** path and is reachable at 14.8B, i.e. on ordinary hardware.

## Change

Materialize each shard in batches of `EVAL_BATCH_SIZE` (16) arrays before handing it to `save_safetensors`, so no single command buffer is large enough to time out.

```python
shard_values = list(shard.values())
for start in range(0, len(shard_values), EVAL_BATCH_SIZE):
    mx.eval(shard_values[start : start + EVAL_BATCH_SIZE])
del shard_values
```

Fifteen lines including the comment. No change to file layout, shard sizing, or output for models that already save fine — the arrays that `save_safetensors` receives are identical, they are simply already evaluated.

## Verification

With this patch the same conversion completes:

```
shard 1/8:  92 tensors, 1.03 GB
shard 2/8: 131 tensors, 1.03 GB
...
shard 8/8:  65 tensors, 0.96 GB
done
```

The resulting model loads and generates correctly, and scores 0.731 overall on BFCL v4 AST evaluation across 4 categories (n=160), so the saved weights are sound rather than merely present.

`python -m unittest tests.test_utils` — 10/10 pass.

## Test added

`test_save_model_materializes_lazy_quantized_weights` quantizes a small Llama, saves it, then asserts:

- no shard is 0 bytes and every shard is loadable
- `model.safetensors.index.json` exists and its `weight_map` covers exactly the model's parameters
- every tensor round-trips bit-identically

The test does not reproduce the watchdog timeout itself — that needs a multi-GB shard and is hardware-dependent — but it locks in the invariants that the failure violated, so a regression that reintroduces "save writes an empty shard" or breaks the index/round-trip is caught.

## Note on a related weakness

Not addressed here to keep the diff focused, but worth considering separately: because the index is written only after the shard loop, any failure mid-loop leaves a directory with weights and no index. Writing the index first, or cleaning up on failure, would turn a silent late failure into an obvious immediate one.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -57,6 +57,11 @@ MODEL_REMAPPING = {
 
 MAX_FILE_SIZE_GB = 5
 
+# Number of arrays materialized per `mx.eval` call when writing a shard. Keeps
+# each Metal command buffer small enough to stay inside the GPU watchdog window;
+# see save_model.
+EVAL_BATCH_SIZE = 16
+
 
 def _parse_size(x):
     sizes = {"M": 1e6, "G": 1e9, "MB": 1e6, "GB": 1e9, "": 1}
@@ -795,6 +800,16 @@ def save_model(
         shards[i] = None
         shard_name = shard_file_format.format(i + 1, shards_count)
         shard_path = save_path / shard_name
+
+        # Materialize the shard in batches before writing it. Freshly quantized
+        # weights are still lazy, so `save_safetensors` would otherwise be the
+        # first consumer and force a whole shard's graph into a single Metal
+        # command buffer, which can exceed the GPU watchdog window and abort the
+        # save. Evaluating in batches keeps each command buffer small.
+        shard_values = list(shard.values())
+        for start in range(0, len(shard_values), EVAL_BATCH_SIZE):
+            mx.eval(shard_values[start : start + EVAL_BATCH_SIZE])
+        del shard_values
 
         mx.save_safetensors(str(shard_path), shard, metadata={"format": "mlx"})
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -59,6 +59,52 @@ class TestUtils(unittest.TestCase):
         shards = utils.make_shards(dict(weights), 1)
         self.assertTrue(gb <= len(shards) <= gb + 1)
 
+    def test_save_model_materializes_lazy_quantized_weights(self):
+        # Regression test: freshly quantized weights are lazy, and letting
+        # `save_safetensors` be their first consumer forces a whole shard's graph
+        # into one Metal command buffer, which can exceed the GPU watchdog window
+        # and abort the save (leaving a 0-byte shard behind). `save_model` should
+        # materialize each shard in batches before writing it.
+        from mlx_lm.models import llama
+
+        args = llama.ModelArgs(
+            model_type="llama",
+            hidden_size=128,
+            num_hidden_layers=4,
+            intermediate_size=256,
+            num_attention_heads=4,
+            rms_norm_eps=1e-5,
+            vocab_size=1024,
+        )
+        model = llama.Model(args)
+        mx.eval(model.parameters())
+        nn.quantize(model, group_size=64, bits=4)
+
+        save_dir = os.path.join(self.test_dir, "save_model_lazy")
+        utils.save_model(save_dir, model)
+
+        # every shard written must be non-empty and readable
+        written = sorted(Path(save_dir).glob("*.safetensors"))
+        self.assertGreater(len(written), 0)
+        for shard_path in written:
+            self.assertGreater(shard_path.stat().st_size, 0)
+            mx.load(str(shard_path))
+
+        # the index must cover exactly the saved parameters
+        index_path = Path(save_dir) / "model.safetensors.index.json"
+        self.assertTrue(index_path.exists())
+        with open(index_path) as f:
+            weight_map = json.load(f)["weight_map"]
+        expected = {k for k, _ in tree_flatten(model.parameters())}
+        self.assertEqual(set(weight_map), expected)
+
+        # and the round trip must preserve values exactly
+        reloaded = {}
+        for shard_path in written:
+            reloaded.update(mx.load(str(shard_path)))
+        for name, value in tree_flatten(model.parameters()):
+            self.assertTrue(mx.array_equal(reloaded[name], value), msg=name)
+
     def test_quantize(self):
         from mlx_lm.models import llama
 


zcbenz:

PR #1640, opened 2026-07-29
Add Nanbeige (Nanbeige4.2) model

houseofdouglas:
## Add Nanbeige (Nanbeige4.2) model

Adds support for the `nanbeige` architecture (e.g. [Nanbeige/Nanbeige4.2-3B](https://huggingface.co/Nanbeige/Nanbeige4.2-3B)).

### What it is
Under the hood the released checkpoint is a **standard Llama-style GQA decoder** (RMSNorm, SwiGLU MLP, rotate-half RoPE, GQA) with one non-standard runtime behavior: **weight-tied looping**. The full stack of decoder layers is executed `num_loops` times with shared weights. Each loop keeps an **independent KV cache** (`make_cache()` returns `num_loops × num_hidden_layers` entries), and `model.norm` is applied between loops (unless `skip_loop_final_norm`).

All other optional features in the upstream modeling code (n-gram embeddings, hyper-connection / mHC, depth-attention, double-loop split, QK-LayerNorm, attention/MLP bias) are disabled in the released checkpoint and carry no weights. They are not implemented here; `Model.__init__` raises a clear `NotImplementedError` if a config enables any of them, so we never silently produce wrong output.

### Validation
The implementation was validated numerically against the original Hugging Face modeling code in fp32: **identical argmax on every position** and **max |Δ| ≈ 1e-4** on the output logits. The per-loop KV cache was checked for prompt-forward vs. token-by-token consistency, and end-to-end generation (bf16 / 8-bit / 4-bit) produces coherent output.

Converted models for testing: https://huggingface.co/vote-for-pedro/Nanbeige4.2-3B-mlx-4bit (also `-8bit`, `-bf16`).

### Tests
Adds `test_nanbeige` to `tests/test_models.py` (uses `num_loops=2` to exercise the loop and per-loop KV cache; asserts the cache has `num_loops × num_hidden_layers` entries).

```
$ python -m pytest tests/test_models.py -k nanbeige
1 passed
```

Formatted with `black` and `isort --profile=black`.

> Opened as a **draft** for maintainer feedback on placement/style before final review.

### DIFF
--- a/mlx_lm/models/nanbeige.py
+++ b/mlx_lm/models/nanbeige.py
@@ -1,0 +1,278 @@
+# Copyright © 2024 Apple Inc.
+#
+# MLX implementation of the Nanbeige (Nanbeige4.2) architecture.
+#
+# Nanbeige is a standard Llama-style GQA decoder (RMSNorm, SwiGLU MLP,
+# rotate-half RoPE) with ONE non-standard runtime behavior that is active in
+# the released Nanbeige4.2-3B checkpoint: weight-tied looping. The full stack of
+# `num_hidden_layers` decoder layers is executed `num_loops` times with shared
+# weights. Each loop keeps an INDEPENDENT KV cache (cache slot for a layer is
+# `layer_idx + loop_idx * num_hidden_layers`), and `model.norm` is applied at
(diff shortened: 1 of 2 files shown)


zcbenz:

Issue #1332, opened 2026-05-31
DeepSeek-V4 (Flash/Pro) on Apple Silicon: unbounded Metal residency growth during decode → [metal::malloc] Resource limit (499000) exceeded after ~11K tokens


snagnever:
### Summary
On Apple Silicon, the DeepSeek‑V4 port (#1192) aborts generation with
`RuntimeError: [metal::malloc] Resource limit (499000) exceeded` after **~11,300 generated tokens, independent of prompt length**. After the first failure the Metal command queue is wedged until a full process restart. This is the "looping / long‑context" failure several testers have hit in the #1192 thread. Root cause is an unbounded per‑decode‑step leak of **live** Metal buffers in the attention caches. A fix is in https://github.com/Blaizzy/mlx-lm/pull/25.

### Environment
- mlx‑lm @ #1192 head (`Blaizzy:pc/add-deepseekv4flash-model`), **mlx 0.31.2**
- Apple **M4 Max, 128 GB**, macOS 26.3 (also reproduced on M3 by others in the thread, ~4000 tokens)
- `mlx-community/DeepSeek-V4-Flash-2bit-DQ`

### Reproduce
Server (`mlx_lm.server --max-tokens 65536`), then stream a long generation — it breaks at ~11,314 tokens:

```python
import json, urllib.request
body = {"model": MODEL, "stream": True, "max_tokens": 20000, "temperature": 0.0,
        "messages": [{"role":"user","content":"Count from 1 to 20000, one number word per line."}]}
# ... stream and count delta tokens; the connection drops at ~11,314 with the server-side OOM
```

In‑process (no server, converges in <1 min) — measures the leak rate directly:

```python
import mlx.core as mx
from mlx_lm import load
from mlx_lm.models.cache import make_prompt_cache
model, tok = load(MODEL); cache = make_prompt_cache(model); MB = 1024**2
y = mx.argmax(model(mx.array(tok.encode("Count upward:\n"))[None], cache=cache)[:, -1], -1); mx.eval(y)
base = mx.get_active_memory() / MB
for s in range(500):
    y = mx.argmax(model(y[:, None], cache=cache)[:, -1], -1); mx.eval(y)
    if s % 100 == 0: print(s, f"{mx.get_active_memory()/MB - base:+.1f} MB")
# active memory grows ~linearly, no plateau (~0.2 MB/step). Adding
# `mx.eval([c.state for c in cache])` after the per-step eval flattens it to ~0.
```

[...]

Issue #1185, opened 2026-04-23
LoRA training: `[metal::malloc] Resource limit (499000) exceeded` on qwen3_5 — descriptor-count leak (not OOM)


Ogilthorp3:
## Summary

`python -m mlx_lm lora` consistently crashes with `RuntimeError: [metal::malloc] Resource limit (499000) exceeded` partway through LoRA training on a Qwen3.5-architecture model (specifically `mlx-community/Qwen3.5-27B-4bit` and a 35B A3B variant that also declares `model_type: qwen3_5`). The crash location is deterministic per config but shifts with `clear_cache_threshold`: lower threshold → later crash, but never eliminated.

This is the same error signature reported for the *serving* path in #831 (fixed by the work around that timeframe), but reproduces in the *training* path on the latest `mlx_lm==0.31.2` / `mlx==0.31.1`. Peak memory is stable across the run (~60 GB), so this is a Metal descriptor / buffer handle accumulation, not an allocation-size problem.

## Environment

| Item | Value |
|---|---|
| Hardware | MacBook Pro M4 Max, 128 GB unified memory |
| macOS | 26.4.1 (Build 25E253) |
| Kernel | Darwin 25.4.0 |
| `iogpu.wired_limit_mb` | 122000 (sysctl) |
| Python | 3.12 (venv) |
| `mlx` | 0.31.1 |
| `mlx-lm` | 0.31.2 |

## Repro

```bash
python -m mlx_lm lora --config lora-config.yaml
```

`lora-config.yaml`:

```yaml
model: /path/to/Qwen3.5-27B-4bit   # or any qwen3_5 checkpoint
train: true
data: /path/to/chat-jsonl-dir
fine_tune_type: lora
optimizer: adamw
num_layers: 32
lora_parameters:
  rank: 16
  alpha: 32
  dropout: 0.05
  scale: 2.0
batch_size: 1
grad_accumulation_steps: 4
iters: 800
learning_rate: 3.0e-6
clear_cache_threshold: 10     # tried 25, 10; see Observations
lr_schedule:
  name: cosine_decay
  arguments: [2.0e-6, 600]
  warmup: 80
  warmup_init: 1.0e-7
val_batches: 10
steps_per_eval: 200
steps_per_report: 10
save_every: 100
mask_prompt: false
max_seq_length: 1024          # also tried 2048; same crash
grad_checkpoint: true
seed: 42
```

[...]

PR #1642, opened 2026-07-29
Fix per-token Metal buffer-object leak in `ArraysCache.advance()` (qwen3_5/qwen3_next batch decode crashes at the resource limit on long completions)

### LINKED ISSUE #1641: `ArraysCache.advance()` leaks Metal buffer objects in qwen3_5/qwen3_next batch decode, crashing long completions at the resource limit
## 84-second repro (stock pip install, no downloads)

```bash
pip install mlx mlx-lm
python repro_leak.py                    # RuntimeError: [metal::malloc] Resource limit
                                        # (499000) exceeded — at ~10.3k decode tokens,
                                        # 84 s wall on an M5 Max
python repro_leak.py --attention-only   # identical depth, zero SSM layers: flat, no crash
```

`repro_leak.py` (inlined at the bottom) builds a random-weight 50-layer tiny qwen3_5_moe (hidden size 64, a few MB of weights, built in seconds) and decodes through `BatchGenerator` — the exact `mlx_lm.server` decode path. The leak rate scales with the number of GatedDeltaNet (SSM) layers, not model size, so a deep-but-tiny model hits the real production crash in minutes. No instrumentation needed to see it on stock mlx: process RSS climbs monotonically 113 MB → 1.44 GB (~130 KB/token of graph-chain descriptors) while `mx.get_active_memory()` stays ~20 MB.

## The measurement

Decoding any qwen3_5 / qwen3_next (hybrid GatedDeltaNet) model through the batch pipeline accumulates live Metal buffer **objects** — not bytes — at exactly **(#SSM layers − 1) per decoded token**. Measured on `mlx-community/Qwen3.5-122B-A10B-6bit` (36 SSM layers), M5 Max 128 GB, macOS 26.5.2, mlx 0.32.0 + mlx-lm 0.31.3, stock code:

| measurement | value |
|---|---|
| live-object slope inside a request | **+35.001 / decode token** (predicted: 36 SSM layers − 1 = 35) |
| single 16.5k-token completion, fresh process | **crash at 14,179 tokens** (predicted (499000 − 5053 baseline)/35 = 14,113; Δ0.5%) |
| same workload with the fix ([#1642](https://github.com/ml-explore/mlx-lm/pull/1642)) | survives — **2,725 objects total, flat**; greedy output token-identical |
| gpt-oss-120b control on the identical stack | flat for hours |

Byte memory stays flat (the leaked objects are 4-byte scalars ≈ 140 B/token), so `mx.get_active_memory()` / `mx.get_cache_memory()` see nothing, and `mx.clear_cache()` cannot help — these are live handles, not cache. The process dies at the per-process object cap (499,000 on Apple silicon) as soon as a **single completion** exceeds ~499000/(#SSM−1) tokens — ~14.1k for the 122B. For isolated or fully drained batches, requests below that threshold are unaffected (the chains die with the request), which is why this presents as flaky long-generation crashes. (A second, much slower variant of the same pattern accumulates across batch *churn* — lazy gather/concat on the same unread metadata in `filter`/`extend` — so a long-lived continuously batched server is not strictly safe below the threshold either; the fix covers both.)

## Root cause

`ArraysCache.advance()` ([cache.py#L685](https://github.com/ml-explore/mlx-lm/blob/v0.31.3/mlx_lm/models/cache.py#L685) @ v0.31.3; identical on main):

```python
def advance(self, N):
    if self.lengths is not None:
        self.lengths -= N
    if self.left_padding is not None:
        self.left_padding -= N
```

`left_padding`/`lengths` are mx.arrays in the batch/server path, so `-=` is **lazy graph arithmetic**: each call appends an unevaluated `Subtract` node whose inputs are (previous node, fresh scalar array holding a live 4-byte `MTL::Buffer` from construction). Every GatedDeltaNet layer calls `cache.advance(S)` per forward (qwen3_next.py:302), but the model rebuilds the SSM mask from only **one** of those caches — `ssm_idx = 0` (qwen3_next.py:401,415). For every other SSM layer the chain is a dead branch: never evaluated, never detached, one link per token, each link pinning one scalar buffer. Rate = (#SSM layers − 1) per decode token. The mlx allocator counts objects, not bytes, against the resource limit (mlx `backend/metal/allocator.cpp:142`).

Three confirming experiments (tiny random model, count-probe build of mlx):

- `mx.eval(cache.left_padding)` each step → count flattens completely (mechanism pinned);
- attention-only twin of the same depth → flat;
- leak rate scales as (#SSM−1) across configs: 9 SSM → +7.99/tok, 11 SSM → +10.65/tok, 49 SSM → +48/tok.

**Why the server path and not plain `generate()`:** `advance()` no-ops when both fields are `None` — the plain `make_prompt_cache()` path. The batch pipeline arms the leak even at batch size 1: `_merge_caches` (generate.py:1036) → `ArraysCache.merge` sets `left_padding = mx.array([0] * B)` for all-empty caches (cache.py:709), and `finalize()` (which nulls the fields) only runs `if max_padding > 0` (generate.py:1166) — never for a single request.

**Why vanilla transformers are immune:** `BatchKVCache.offset` is the same shape of per-step mx.array state (cache.py:961), but attention **consumes it every step** via `self.rope(q, offset=cache.offset)`, forcing and freeing the chain. A dead chain requires per-step array state the forward pass never reads; only the hybrid-SSM `ArraysCache` bookkeeping qualifies.

## Fix ([#1642](https://github.com/ml-explore/mlx-lm/pull/1642))

`ArraysCache` tracks the cumulative `advance()` decrement as a Python int per metadata field, folded in place into the stored array when that field is read (property access or `make_mask`) — metadata-side subtraction wrapped to the metadata dtype's modular range, so integer mask arithmetic is identical to the eager decrement (floating metadata rounds at coalesced, int64-gate-sized chunks rather than per step — not a dtype this pipeline produces); `left_padding`/`lengths` stay mx.array-valued properties. `advance()` creates zero graph nodes. Validation (measured with the initial build of the `advance()` fix; the PR adds review-driven hardening for slower leak variants in the same pattern class, covered by unit tests): the 122B crash config survives with 2,725 objects flat; 205k-token tiny-model endurance flat (+0.007/tok, KV-block noise); greedy decode token-identical to stock (64/64).

## How it was measured ([ml-explore/mlx#3942](https://github.com/ml-explore/mlx/pull/3942))

The numbers above come from three read-only counters added to the mlx allocator — `mx.get_active_buffer_count()` / `get_cache_buffer_count()` / `get_buffer_histogram()` — the probe #1185 asked for. The histogram pinned this leak to the 4-byte size class in a single run (3,268 → 480,470 objects over 10k tokens, every other class flat).

## Likely duplicates / related issues

- #1332 — DeepSeek-V4 decode, "unbounded Metal residency growth", crash after ~11k tokens: the signature matches exactly, but that thread implicates the `PoolingCache`/`RotatingKVCache` path rather than `ArraysCache` — if the mechanism is the same, it is a **sibling instance of the pattern at a different site** (per-step lazy arithmetic on state the forward pass never reads), which this patch does not touch. The probe API makes checking it a one-run exercise.
- #1185 — same crash signature in LoRA **training** on qwen3_5. Training doesn't go through `BatchGenerator`, so the trigger there may likewise be a different site with the same pattern; the probe PR is the tool for confirming.
- #831 (closed as resolved by #839) — `mlx_lm.server` distributed, same error string, different mechanism (MLA).
- ml-explore/mlx#3564, ml-explore/mlx#3539 (closed) — same error family; their reports implicate residency-set commit and training-time allocation respectively, so they may be entirely distinct mechanisms. The triage notes below are for telling these apart.

Triage notes for that family:

- There is exactly **one** throw site of `Resource limit (N) exceeded` (mlx `allocator.cpp:142`): the process's own live-object counter against a limit read once at startup (`iogpu.rsrc_limit` sysctl, hardcoded 499000 fallback). A constant N across crash reports carries no device-state information. Driver-side refusals surface differently (`Unable to allocate`, command-buffer errors, SIGABRT in residency-set commit).
- The dominant crash horizon is **per-request**: at request end the chains release into the buffer cache and are recycled (total plateaus at ~35 × longest completion), so sub-threshold workloads on drained batches run for a very long time, which makes the bug look nondeterministic. (The slower churn variant above accumulates across requests on a never-drained batch.)
- Byte-side knobs (cache limit 0/1G/8G, forced `clear_cache`, wired limit) measurably do nothing for this leak — it is an object-count mechanism.

## Environment

M5 Max 128 GB, macOS 26.5.2 (Darwin 25.5.0), mlx 0.32.0 (pip), mlx-lm 0.31.3 (pip), Python 3.12.

<details>
<summary><code>repro_leak.py</code> (self-contained, ~380 lines)</summary>

```python
#!/usr/bin/env python3
"""Smoking-gun repro: qwen3_5/qwen3_next decode leaks live Metal buffer OBJECTS.

    pip install mlx mlx-lm
    python repro_leak.py                    # hybrid: dies with
                                            #   RuntimeError: [metal::malloc]
                                            #   Resource limit (499000) exceeded
                                            # after ~10.4k decode tokens, minutes
    python repro_leak.py --attention-only   # same depth, no SSM: flat, no crash

One script, one knob. The model is a random-weight qwen3_5_moe built in
seconds (hidden 64, vocab 128, ~MBs of weights — NO download): the leak rate
scales with the number of GatedDeltaNet (SSM) layers, not model size, so a
DEEP-but-tiny model (default: 50 layers = 49 SSM + 1 attention) hits the real
crash quickly on a stock pip install.

What leaks
----------
`ArraysCache.advance()` decrements `left_padding`/`lengths` with lazy
mx.array arithmetic once per SSM layer per decoded token
(mlx_lm/models/qwen3_next.py:302 -> mlx_lm/models/cache.py:685). The model
only rebuilds the SSM mask from cache[ssm_idx=0] (qwen3_next.py:401,415), so
for every OTHER SSM layer the graph chain is dead: never evaluated, one new
node per token, each node pinning a fresh scalar array whose live 4-byte
MTL::Buffer counts against the per-process resource limit
(mlx/backend/metal/allocator.cpp:142, 499000 on Apple silicon). Bytes stay
flat; the object count climbs at ~(#SSM layers - 1)/token; `mx.clear_cache()`
cannot help (live handles, not cache). The server/batch path arms this even
at batch size 1 (`ArraysCache.merge` sets left_padding, generate.py:1036 /
cache.py:709); attention layers have equivalent per-step array state
(`BatchKVCache.offset`) but RoPE consumes it every step, which is why vanilla
transformers (gpt-oss etc.) do not leak.

Observables
-----------
- stock mlx: deterministic crash at ~499000/(#SSM-1) decode tokens; monotonic
  process-RSS growth (chain nodes) and a slow monotonic
  `mx.get_active_memory()` creep (~4 B/object) while cache memory is flat.
- probe build (mlx + buffer-count probes: get_active_buffer_count et al.):
  exact live-object slope per token, and a per-size-class histogram pinning
  the growth to the smallest size class.
- with the ArraysCache fix (mlx-lm patch): slope 0, no crash, same tokens.

Exit codes: 0 flat/survived, 1 leak detected (probe build), 2 resource-limit
crash reproduced.
"""

import argparse
import json
import resource
import sys
import time

import mlx.core as mx


def _mx_fn(*names):
    """First available mx function: upstream buffer-count probe naming first,
    then the name used by the local v0.32.0 probe wheel."""
    for n in names:
        if hasattr(mx, n):
            return getattr(mx, n)
    return None


_active_count = _mx_fn("get_active_buffer_count", "get_active_resource_count")
_cache_count = _mx_fn("get_cache_buffer_count", "get_cache_count")
_histogram = _mx_fn("get_buffer_histogram", "get_resource_histogram")
HAVE_COUNT_PROBE = _active_count is not None


def build_model(num_layers: int, interval: int, experts: int, hidden: int = 64):
    from mlx_lm.models import qwen3_5_moe

    cfg = dict(
        model_type="qwen3_5_moe",
        hidden_size=hidden,
        intermediate_size=hidden * 2,
        num_hidden_layers=num_layers,
        num_attention_heads=4,
        num_key_value_heads=2,
        head_dim=16,
        vocab_size=128,
        full_attention_interval=interval,
        linear_num_value_heads=4,
        linear_num_key_heads=2,
        linear_key_head_dim=16,
        linear_value_head_dim=16,
        linear_conv_kernel_dim=4,
        num_experts=experts,
        num_experts_per_tok=2,
        moe_intermediate_size=32,
        shared_expert_intermediate_size=32,
    )
    args = qwen3_5_moe.ModelArgs.from_dict(
        {"model_type": "qwen3_5_moe", "text_config": cfg}
    )
    model = qwen3_5_moe.Model(args)
    mx.eval(model.parameters())
    n_ssm = sum(1 for l in model.layers if getattr(l, "is_linear", False))
    print(
        f"[model] layers={num_layers} (ssm={n_ssm} attn={num_layers - n_ssm}) "
        f"interval={interval} experts={experts} hidden={hidden}",
        flush=True,
    )
    return model, n_ssm


def rss_mb() -> float:
    return resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1e6  # macOS: bytes


def probe(sync: bool = True) -> dict:
    """Sample allocator state. Synchronizes first so in-flight command buffers
    (whose completion handlers release temporaries) have drained."""
    if sync:
        mx.synchronize()
    out = {
        "active_mb": mx.get_active_memory() / 1e6,
        "cache_mb": mx.get_cache_memory() / 1e6,
        "rss_mb": rss_mb(),
    }
    if HAVE_COUNT_PROBE:
        out["count"] = _active_count()
        out["cache_count"] = _cache_count()
        # Live handles are what crash the process; cached objects count
        # against the limit too but are reclaimable (the allocator clears the
        # cache under count pressure, and byte limits bound its growth).
        out["live"] = out["count"] - out["cache_count"]
        out["hist"] = dict(_histogram())
    return out


def fmt(p: dict) -> str:
    base = (
        f"active={p['active_mb']:9.2f}MB cache={p['cache_mb']:7.1f}MB "
        f"rss={p['rss_mb']:7.0f}MB"
    )
    if HAVE_COUNT_PROBE:
        return (
            f"live={p['live']:>7d} (+cached={p['cache_count']} "
            f"= {p['count']}) " + base
        )
    return base + "  (stock mlx: no count probe)"


def hist_delta(a: dict, b: dict, top: int = 8) -> str:
    keys = sorted(set(a.get("hist", {})) | set(b.get("hist", {})))
    deltas = [
        (k, b.get("hist", {}).get(k, 0) - a.get("hist", {}).get(k, 0)) for k in keys
    ]
    deltas = [d for d in deltas if d[1] != 0]
    deltas.sort(key=lambda kv: -abs(kv[1]))
    if not deltas:
        return "  (no size-class drift)"
    return "\n".join(f"  size<= {k:>10d} B : {d:+d}" for k, d in deltas[:top])


def _cache_arrays(c):
    st = getattr(c, "state", None)
    if st is None:
        st = []
    arrs = [a for a in (st if isinstance(st, (list, tuple)) else [st]) if a is not None]
    # ArraysCache.state exposes only the state slots; the leaking graph chains
    # hang off left_padding/lengths, so include those explicitly.
    for attr in ("left_padding", "lengths", "offset"):
        v = getattr(c, attr, None)
        if isinstance(v, mx.array):
            arrs.append(v)
    return arrs


def run_plain(model, n_tokens, report_every, samples, eval_cache_every):

    y = mx.array([[1, 2, 3, 4, 5, 6, 7, 8]])
    logits = model(y, cache=cache)
    y = mx.argmax(logits[:, -1], axis=-1)
    mx.eval(y)
    samples.append((0, probe()))
    print(f"tok {0:>6d}  {fmt(samples[-1][1])}", flush=True)
    for t in range(1, n_tokens + 1):
        logits = model(y[:, None], cache=cache)
        y = mx.argmax(logits[:, -1], axis=-1)
        mx.async_eval(y)
        if eval_cache_every and t % eval_cache_every == 0:
            mx.eval([a for c in cache for a in _cache_arrays(c)])
        if t % report_every == 0:
            samples.append((t, probe()))
            print(f"tok {t:>6d}  {fmt(samples[-1][1])}", flush=True)
    return samples


def run_batch(model, n_tokens, report_every, samples, eval_cache_every):
    """Drive mlx_lm's BatchGenerator directly -- the exact mlx_lm.server decode
    path (left-padded batch caches, masked SSM kernel, async_eval streaming)."""

    bg = BatchGenerator(model, max_tokens=n_tokens + 8)
    bg.insert([[1, 2, 3, 4, 5, 6, 7, 8]], max_tokens=[n_tokens])
    t = 0
    samples.append((0, probe()))
    print(f"tok {0:>6d}  {fmt(samples[-1][1])}", flush=True)
    next_report = report_every
    done = False
    idle = 0
    while not done:
        prompt_responses, gen_responses = bg.next()
        idle = idle + 1 if not prompt_responses and not gen_responses else 0
        if idle > 32:
            break
        for r in gen_responses:
            t += 1
            if getattr(r, "finish_reason", None) is not None:
                done = True
        if eval_cache_every and t >= 1 and t % eval_cache_every == 0:
            for entry in bg._generation_batch.prompt_cache:
                arrs = _cache_arrays(entry)
                if arrs:
                    mx.eval(arrs)
        if t >= next_report:
            samples.append((t, probe()))
            print(f"tok {t:>6d}  {fmt(samples[-1][1])}", flush=True)
            next_report += report_every
    bg.close()
    return samples


def main():
    ap = argparse.ArgumentParser(
        description="qwen3_5 hybrid-SSM Metal buffer-object leak repro"
    )
    ap.add_argument("--mode", choices=["plain", "batch"], default="batch")
    ap.add_argument(
        "--attention-only",
        action="store_true",
        help="THE control knob: same depth, zero SSM layers -- no leak, no crash",
    )
    ap.add_argument("--layers", type=int, default=50)
    ap.add_argument(
        "--interval",
        type=int,
        default=0,
        help="full_attention_interval (0 = auto: == layers for the hybrid "
        "smoking gun, 1 with --attention-only). Production qwen3_5 uses 4.",
    )
    ap.add_argument("--experts", type=int, default=8)
    ap.add_argument("--hidden", type=int, default=64)
    ap.add_argument(
        "--tokens",
        type=int,
        default=12000,
        help="decode budget; the default hybrid config is expected to crash at "
        "~499000/(#SSM-1) ~= 10.4k tokens before reaching it",
    )
    ap.add_argument("--report-every", type=int, default=250)
    ap.add_argument("--disable-compile", action="store_true")
    ap.add_argument(
        "--ops",
        action="store_true",
        help="force the pure-ops gated delta path instead of the custom metal kernel",
    )
    ap.add_argument(
        "--eval-cache-every",
        type=int,
        default=0,
        help="mx.eval all cache state arrays every N tokens (0 = never). "
        "Flattening the count here proves the leak is unevaluated cache-state "
        "graph chains.",
    )
    ap.add_argument("--fail-slope", type=float, default=0.5)
    ap.add_argument("--json-out", type=str, default="")
    args = ap.parse_args()

    interval = args.interval
    if interval == 0:
        interval = 1 if args.attention_only else args.layers
    elif args.attention_only:
        interval = 1

    if args.disable_compile:
        mx.disable_compile()
        print("[env] mx.disable_compile()")

    if args.ops:
        from mlx_lm.models import gated_delta

        orig = gated_delta.gated_delta_update

        def ops_update(*a, **kw):
            kw["use_kernel"] = False
            return orig(*a, **kw)

        import mlx_lm.models.qwen3_5 as _q35
        import mlx_lm.models.qwen3_next as _q3n

        for m in (gated_delta, _q3n, _q35):
            if hasattr(m, "gated_delta_update"):
                m.gated_delta_update = ops_update
        print("[env] gated_delta_update forced to ops path (no custom metal kernel)")

    mx.random.seed(0)
    model, n_ssm = build_model(args.layers, interval, args.experts, args.hidden)
    expected = None
    if n_ssm >= 2 and args.mode == "batch":
        expected = 499_000 // (n_ssm - 1)
        print(
            f"[predict] leak ~{n_ssm - 1} objects/token -> resource-limit crash "
            f"near token ~{expected:,} (if unpatched)",
            flush=True,
        )

    samples: list = []
    t0 = time.time()
    crashed = None
    runner = run_batch if args.mode == "batch" else run_plain
    try:
        runner(model, args.tokens, args.report_every, samples, args.eval_cache_every)
    except RuntimeError as e:
        if "Resource limit" not in str(e):
            raise
        crashed = str(e)
    dt = time.time() - t0
    last_t = samples[-1][0] if samples else 0

    verdict = None
    slope = None
    leaking = False
    if crashed:
        leaking = True
        verdict = (
            f"CRASH REPRODUCED: {crashed!r} after ~{last_t}+ decode tokens "
            f"(predicted ~{expected:,}; {dt:.0f}s wall)"
        )
    elif HAVE_COUNT_PROBE and len(samples) >= 4:
        # Slope over the mid-run steady state: skip warmup AND the final
        # sample (after the request completes, the batch drops its caches and
        # the chains release into the buffer cache -- by design).
        half = samples[len(samples) // 4 : -1]
        if len(half) < 2:
            half = samples[-2:]
        (t_a, p_a), (t_b, p_b) = half[0], half[-1]
        if t_b > t_a:
            slope = (p_b["live"] - p_a["live"]) / (t_b - t_a)
            leaking = slope > args.fail_slope
            verdict = (
                f"LEAKING: +{slope:.2f} LIVE buffer objects/token "
                f"(tokens {t_a}..{t_b}: live {p_a['live']} -> {p_b['live']}; "
                f"predicted ~{max(n_ssm - 1, 0)})"
                if leaking
                else f"FLAT: {slope:+.3f} live objects/token (tokens {t_a}..{t_b})"
            )
        print("\n[size-class drift over second half]")
        print(hist_delta(p_a, p_b))
    else:
        verdict = (
            "SURVIVED (stock build, no count probe -- see rss/active byte proxies)"
        )

    print(f"\n[verdict] {verdict}")
    print(f"[run] {last_t} tokens in {dt:.1f}s ({last_t/max(dt,1e-9):.0f} tok/s)")

    if args.json_out:
        payload = {
            "argv": sys.argv[1:],
            "n_ssm": n_ssm,
            "interval": interval,
            "have_count_probe": HAVE_COUNT_PROBE,
            "crashed": crashed,
            "slope_per_token": slope,
            "leaking": leaking,
            "expected_crash_token": expected,
            "samples": [
                (t, {k: v for k, v in p.items() if k != "hist"}) for t, p in samples
            ],
            "hist_first": samples[0][1].get("hist") if samples else None,
            "hist_last": samples[-1][1].get("hist") if samples else None,
            "seconds": dt,
        }
        with open(args.json_out, "w") as f:
            json.dump(payload, f, indent=2)
        print(f"[json] {args.json_out}")

    sys.exit(2 if crashed else (1 if leaking else 0))


if __name__ == "__main__":
    main()
```

</details>

thinkroth:
Closes #1641. Related crash reports share the error signature but likely not this exact site: #1332's DeepSeek-V4 path implicates `PoolingCache`/`RotatingKVCache` rather than `ArraysCache` (a sibling instance of the same lazy-metadata pattern, if so — not fixed here), and #1185 / #831 / ml-explore/mlx#3564 / ml-explore/mlx#3539 are triaged in the issue.

## What was happening

`advance()` decremented `left_padding`/`lengths` with lazy mx.array arithmetic once per GatedDeltaNet layer per decoded token. Only `cache[ssm_idx]` ever has its mask rebuilt, so for every other SSM layer the resulting graph chain is dead — never evaluated, never detached — and each node pins a live 4-byte Metal buffer object. That's (#SSM layers − 1) objects per decode token; the Metal allocator counts objects (not bytes) against the per-process resource limit, so any single completion over ~499000/(#SSM−1) tokens (~14.1k on Qwen3.5-122B, 36 SSM layers) kills the process with `[metal::malloc] Resource limit (499000) exceeded`. Byte telemetry stays flat and `mx.clear_cache()` can't help. Full mechanism, file:line walkthrough, and an 84-second no-download repro are in #1641.

## The fix (commit 1)

Track the cumulative `advance()` decrement as a Python int **per metadata field** (`_lp_advance`/`_len_advance`), folded into the stored array when that field is read (property access or `make_mask`), replaced, or finalized — so a workload where both fields coexist (uneven batched prefill) but only one is consumed cannot chain nodes onto the unread one. Folds are applied **in place** (`-=`, preserving the array object) **on the metadata side** and scheduled with `mx.async_eval`, so no public operation sequence — discarded reads, discarded masks, repeated item assignment, repeated lazy reassignment, batch churn — accumulates unevaluated graph on metadata nothing consumes, and nothing blocks on the generation stream. `advance()` itself creates zero graph nodes and evaluates nothing; it validates its argument **unconditionally** (regardless of whether metadata is set): `mx.array` offsets raise `TypeError` (silently coercing one would force an eval and accept size-1 non-scalars) and `operator.index` rejects floats (stock silently float-drifted the metadata). Offsets outside the scalar range MLX accepts for the metadata dtype raise `ValueError` at `advance()`, per armed field in stock's field order — exactly where stock's eager subtraction rejected them. The check mirrors MLX's conversion precisely (python int scalars convert through int64, per `python/src/utils.cpp`): the int64 gate applies to *every* dtype including floating, integer dtypes narrower than 64 bits are range-checked besides, and uint64 accepts negatives, which wrap modulo 2^64 (stock's gate rejections surfaced as an opaque `std::bad_cast`; here it is uniformly `ValueError`).

**Arithmetic scope:** deferred totals fold as the dtype-congruent scalar. Sequential wrapping subtractions equal one subtraction of the sum modulo 2^bits, so integer metadata folds to exactly stock's per-step value — including totals that overflow the dtype, where stock's sequential in-range subtractions wrap; uint64 uses the congruent *signed* representative, since that is what the int64 conversion accepts. Floating totals that exceed the int64 gate (individually valid offsets can sum past it) fold in gate-sized chunks rather than failing the scalar conversion, each chunk scheduled as it is applied so a many-chunk fold cannot itself build an unbounded chain; per-step float *rounding* is not reproduced — a dtype nothing in the pipeline produces; documented in the class note. Fold counters commit chunk-by-chunk, before each subtraction: a decrement is never applied twice, and a failure mid-fold leaves the unapplied remainder pending except for at most the single in-flight chunk (an asynchronous interrupt between counter commit and subtraction drops it).

`filter()`/`extend()` additionally schedule materialization of the metadata after mutating it (`mx.async_eval`, so batch reorganization never waits behind a queued forward pass). Each successful `extend()` field concatenation commits its replacement array and fresh bookkeeping together: if a later field fails, the earlier stock-compatible partial mutation stays coherent instead of retaining a shallow alias's old counter. These operations run once per batch composition change, and their lazy gather/concat nodes are the same dead-chain pattern at per-request rate (present upstream too): on a long-lived server under continuous batching, request churn would otherwise still grow an unevaluated chain for every layer whose metadata the mask path never reads.

**Documented deviations from the eager decrement** (deferred visibility; nothing in this repo relies on the old timing): a decrement becomes visible to aliases of a metadata array at the *next fold of that field* — property access, `make_mask`, replacement, or `finalize()` (the latter two fold the outgoing array first, so a decrement is never silently discarded) — rather than at `advance()` time. An alias inspected before any fold shows the pre-advance value, alias reads/writes before the fold interleave accordingly, and one array object stored in both fields applies each field's decrement at its own fold. The stale-until-fold sequence, fold-through-`make_mask`, fold-on-replacement, fold-on-finalize, and the shared-object case are each pinned by tests as documented behavior.

Further documented deviations and limitations, all pinned by tests:

- A read that encounters a **tracer** — metadata swapped into a trace by `mx.compile(..., inputs=vars(cache))` — is pure: no fold, no scheduling (`async_eval` is disallowed on tracers), matching stock's plain attribute read; the pending decrement folds at the next *eager* access, after which the captured array shows it. Without this, a captured read would raise mid-trace and could lose a pending decrement on the error path. A **closure-captured** cache holds concrete arrays during tracing (`mx.compile` without `inputs=`, `mx.vmap`), so those reads fold eagerly — producing exactly the values stock's eager decrement produced.
- Compiling a function that **calls `advance()`** is not supported: the decrement is host-side bookkeeping and runs at trace time only, so the compiled graph contains no subtraction (the trace-time call is not lost — it stays pending and folds at the next eager read). Stock incidentally supported this because its decrement was lazy graph arithmetic — the same arithmetic that leaks. Call `advance()` outside compiled functions.
- **Mutating** captured metadata (assignment, `finalize`, `filter`, `extend`) while a pending decrement cannot fold — the field is a tracer — raises `RuntimeError` rather than silently discarding the decrement; the never-silently-discarded guarantee holds. This guard cannot fire on eager paths, where folds always land. One mutation cannot be intercepted: **in-place item assignment** through a captured tracer read (`__setitem__` on the returned array bypasses the property machinery) — it falls into the documented alias-write window, the write landing before the deferred decrement where stock's eager decrement landed first; pinned by a test.
- `copy.copy` and `copy.deepcopy` fold before copying. A shallow copy shares the backing arrays, reentrant cache lock, pending counters, per-array fold lock, and bool-promotion record; this is one alias group, so either object folds a shared field's total exactly once. A standalone deep copy gets independent arrays, bookkeeping, and locks. When an entire shallow-alias graph is deep-copied together, Python's shared memo preserves the copied array aliases and their copied bookkeeping/lock topology. Reconstruction follows Python's reduction protocol rather than copying only `__dict__`, so state in subclass `__slots__` is preserved too. If the fold cannot land (captured tracer, decrement pending), both copiers raise like the other guarded mutations — a copy taken mid-trace would otherwise escape holding the transient tracer and fail on any later use.
- **Reads serialize per cache alias group and per backing-array identity**: stock's pure attribute reads were trivially thread-safe, while deferred folding makes reads mutate. A pickleable reentrant cache lock guards advance bookkeeping and a weak identity registry associates a second reentrant lock with each live metadata array, so two independently constructed caches whose public setters receive the same array cannot fold it concurrently on different thread-local streams. `meta_state`/`save_prompt_cache` acquire the relevant locks for a complete linearizable two-field snapshot. Generic pickle preserves array/state/lock alias topology only for a **quiescent source graph**; concurrent generic pickle is explicitly unsupported because the outer Pickler traversal continues after any owning-object reducer hook returns. Cross-thread *mutation* (`filter`/`extend`/`finalize` racing anything) remains unsupported, as in stock. The registry holds arrays weakly; a 2,000-array retention probe returned exactly to its zero baseline.
- bool metadata arithmetic is reproduced at **int32 precision** (stock's promotion): the dtype stays bool through zero-net advance histories, but validation tracks stock's promotion faithfully — after the *first* `advance()` on a bool field (even a zero or cancelling one, through a bool array shared by both fields, through a shallow cache alias, or through independently constructed caches assigned the same array), later offsets validate against int32 exactly as stock's already-promoted stored array would; assigning a genuinely new array resets the tracking (re-assigning the same backing object does not — the flag is the only record of stock's physical promotion). `extend()` materializes this logical promotion before concatenation, so a promoted bool field combined with a narrow integer remains int32 like stock rather than being narrowed by dtype promotion. Offsets are normalized by `operator.index` (python bool and numpy integer scalars become python ints, value preserved) — stock's array-promotion side effects for exotic offset types (a bool offset keeping bool metadata bool; numpy scalar offsets promoting the metadata dtype) are not reproduced.

## Serialization fix (commit 2, pre-existing bug)

`ArraysCache.state` serializes only the state slots, so `save_prompt_cache` silently dropped `left_padding`/`lengths` — a loaded cache had `None` metadata and `make_mask()` returned `None`. (`BatchKVCache`/`BatchRotatingKVCache` already round-trip `left_padding` via `state`.) Metadata now round-trips through `meta_state` as `<dtype>:<comma-joined values>` with `""` for absent fields, folding any pending decrement at save time. Decoding is strict: the delimiter is required and the dtype must come from an explicit numeric whitelist with per-dtype value parsing, so truncated or malformed file metadata raises instead of silently loading as empty metadata. Unsupported physical metadata dtypes (bool, complex) fail loudly at save — including a logically promoted bool whose zero-net history intentionally leaves the physical dtype bool — as does non-1-D metadata (a 0-d `filter` index is the public route to a scalar field), rather than emitting an entry the decoder cannot parse. The encoding keeps an empty metadata array distinct from an absent field, though a zero-row cache cannot pass through `save_prompt_cache` at all — safetensors rejects empty state arrays — so that distinction is pinned at the codec level. Files written before this change carry an empty `meta_state` entry and load unchanged. The existing batch-cache serialization test asserted on the *source* cache rather than the loaded one, which is why this never surfaced — the test now checks both.

## Measurements (advance-only build; see validation status below)

Measured on Qwen3.5-122B-A10B-6bit, M5 Max 128 GB, macOS 26.5.2, mlx 0.32.0:

| | stock 0.31.3 | measured fix build |
|---|---|---|
| live-object slope inside a request | +35.001 / decode token | **+0.000** (zero size-class drift) |
| single 16.5k-token completion | crash at 14,179 tokens | **survives, 2,725 objects total, flat** |
| greedy decode vs stock | — | **token-identical (64/64)** |
| tiny-model endurance, 205k decode tokens | (crashes at ~10.3k) | **flat** (+0.007/tok, KV-block noise) |

**Validation status, stated precisely:** the 122B Metal measurements above were taken with the initial integer-bookkeeping implementation of the `advance()` fix — the mechanism this PR is built on and the dominant leak — and remain historical measurements of that build. The complete PR tip `985af30` has now been executed on Metal after the final alias/copy/partial-extend hardening: `test_prompt_cache.py` **37/37** on Metal and with `DEVICE=cpu`; the companion MLX probe tip `cb83965c` has `test_memory.py` **4/4**, `test_zero_copy.py` **7/7**, and full Python discovery at **790 passed / 45 skipped**. The independently assigned shared-array race passed **200/200** Metal trials; another **200/200** stress with two caches swapping the same arrays between fields produced only linearizable `3/3` and `4/4` snapshots. Bool promotion propagated across independent caches; multi-level subclass slot copy/deepcopy, failed-extend detachment, quiescent pickle protocols 2–5, pre-array-lock pickle state, and independent-array-alias pickle topology passed. (MLX array pickling itself rejects protocols 0–1.) A 2,000-array registry-retention probe returned from 2,000 entries exactly to zero. Full mlx-lm discovery ran 200 tests: 9 persistent failures previously reproduced on exact base, 3 missing-optional-dependency errors (`datasets`, `lm_eval`, `requests`), and 1 skip. A fresh 1,000-token/49-SSM `BatchGenerator` run was flat at **+0.000 live buffers/token** and 106 tok/s. The earlier 10,000 discarded reads, 10,000 discarded masks, 2,000 filter-churn iterations, 2,048-chunk float fold, and 12,800 randomized integer comparisons remain applicable. Slopes were measured with the allocator count probes proposed in ml-explore/mlx#3942 (the API requested in #1185); the repro script in #1641 reproduces the crash on stock pip installs in 84 s.

## Tests (commits 3–5)

- `test_arrays_cache_advance` (no Metal required; runs on whatever device CI selects): semantics-equivalence guards — property values, `make_mask`, all-empty `merge` (the server path that arms the leak), `filter`/`extend` with pending advances on both sides, repeated filter/extend churn with no intermediate reads (the continuous-batching pattern), field assignment, in-place item assignment with a pending advance, `finalize` — plus the pinned deferred-visibility semantics (stale-until-fold, fold via read/`make_mask`/replacement/`finalize`, shared-object folding) and white-box guards: per-field counters stay Python ints, `advance()` raises `TypeError` for mx.array/float offsets on armed, unarmed, and finalized caches alike, `advance()` never rebinds the stored arrays, `make_mask()` folds only the field it uses, consuming one metadata field never touches the other, and 2048-iteration loops of discarded reads, discarded masks, item assignment, and lazy reassignment each stay within a 4 KB active-memory bound instead of chaining. Dtype boundary cases pin the congruent fold against stock's values (int32 `2**31-1` + `1` → `INT32_MIN`, int64 `2**62` twice → `INT64_MIN`, int8/uint8 wraparound, uint64 totals crossing `2**63` and negative offsets wrapping modulo 2^64, floating totals past the int64 gate folding chunked to deterministic values, a 2048-chunk fold staying within the 4 KB memory bound) and the per-field range validation (out-of-range offsets raise `ValueError` for every dtype class incl. uint64, float, and bool-after-first-advance validating as int32 — through zero and cancelling histories, reset by fresh assignment, propagated across a shared bool array and shallow alias, and materialized before mixed-dtype extension; an offset valid for one field but not the other leaves the first field's decrement applied, in stock's order). Shallow-copy semantics (fold-before-copy, shared backing array/bookkeeping, post-copy advances visible to both objects) and the bool-metadata deviations (zero-net dtype preservation, `advance(True)` normalization, int32-precision folds) are pinned. A 1024-iteration `filter` churn loop with the state consumed every pass but the metadata never read stays within the same 4 KB bound — it fails if the materialization in `filter`/`extend` is removed, which the value-equivalence churn test alone would not catch.
- `test_arrays_cache_compile_captured_reads`: `mx.compile` with `inputs=vars(cache)` — a property read and a `make_mask` during tracing stay pure (no raise, counters intact, stale value used), and the captured array observes the decrement after the next eager fold.
- `test_arrays_cache_closure_captured_reads`: closure-captured reads inside `mx.compile` (no `inputs=`) and `mx.vmap` fold eagerly with stock-identical values.
- `test_arrays_cache_compiled_advance_not_recorded`: pins the documented limitation — a compiled function calling `advance()` records no decrement in the traced graph; the trace-time call stays pending and folds at the next eager read.
- `test_arrays_cache_captured_mutation_guard`: replacing — or `copy.copy`-ing — captured metadata during tracing with a decrement pending raises `RuntimeError` and preserves the counter (mid-trace exceptions leave compile's captured containers mid-swap — an MLX-level property — so only the raise and counter are asserted).
- `test_arrays_cache_captured_item_assignment`: pins the documented alias-write window — an item write through a captured tracer read lands before the deferred decrement.
- `test_arrays_cache_fold_failure_keeps_remainder`: injects a scheduling failure mid-fold and asserts the counter holds exactly the unapplied remainder, which the next fold completes.
- `test_arrays_cache_concurrent_reads`: two threads racing one pending decrement both complete (bounded joins) and observe the folded value exactly once — the unserialized fold deadlocked here.
- `test_arrays_cache_alias_copy_state`: shallow aliases share the backing array, lock, counter, and promotion state; concurrent property reads fold their combined advances once. Deep-copying a shallow-alias graph preserves the copied array/lock/state topology while remaining independent from the original graph.
- `test_arrays_cache_external_array_alias_state`: independently constructed caches assigned the same metadata array retain separate pending counters/cache locks but share the backing-array fold lock and bool-promotion record; concurrent reads complete with one fold per counter. The same topology is checked after quiescent pickle.
- `test_arrays_cache_subclass_slot_copy`: shallow and deep copies retain subclass slot state while preserving their expected shared/independent array topology.
- `test_arrays_cache_failed_extend_detaches_state`: when the `left_padding` concatenation succeeds and the `lengths` concatenation fails, the committed first field has fresh state; a shallow alias cannot consume the target's later decrement.
- `test_arrays_cache_pickle`: populated cache state and pending counters round-trip through a pickleable lock; an entire shallow-alias graph retains one copied array/lock/state group across protocols 2–5. The test explicitly records that concurrent generic pickle is outside the supported contract.
- `test_arrays_cache_metadata_serialization_is_atomic`: an event-controlled interleaving pauses after folding `left_padding`, races `advance()`, and proves the serialized pair is a linearizable `9/19`, never the torn `9/18`.
- Eager standalone `copy.deepcopy` (folds first, independent counters/arrays/lock), self-assignment preserving the bool-promotion record, and numpy integer-scalar offset normalization are pinned in `test_arrays_cache_advance`.
- `test_arrays_cache_metadata_serialization`: save/load round-trip with pending folds; absent fields and dtype preservation; codec-level round-trip keeping empty (zero-row) metadata distinct from absent; strict decoding (malformed or non-whitelisted entries raise; non-numeric dtypes rejected at save; scalar metadata from a 0-d `filter` index and 2-D metadata rejected at save); mask reconstruction on the loaded cache; the existing batch serialization test now asserts on the loaded cache.
- `pre-commit` (black 25.1.0, isort) clean.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -610,6 +610,32 @@ def _try_schedule(*arrays):
         return False
 
 
+class _ArraysCacheFoldLock(type(threading.RLock())):
+    """An RLock that recreates itself across pickle/deepcopy.
+
+    A shallow cache copy keeps the same lock by normal object sharing.
+    Pickle and deepcopy create a new native lock; their memo preserves
+    sharing when an alias graph contains the same lock more than once.
+    """
+
+    def __reduce__(self):
+        return (type(self), ())
+
+
+class _ArraysCacheFieldState:
+    """Mutable bookkeeping shared by shallow aliases of one field.
+
+    ``promotion`` is separate from the pending counter because the two
+    metadata fields can reference the same bool array: stock promoted
+    both logical views together, while each field still applied its own
+    decrement.
+    """
+
+    def __init__(self, advance=0, promoted=False, promotion=None):
+        self.advance = advance
+        self.promotion = [promoted] if promotion is None else promotion
+
+
 class ArraysCache(_BaseCache):
     # ``left_padding`` and ``lengths`` are logically decremented by
     # ``advance()`` on every layer call during decode. Doing that decrement
@@ -683,37 +709,68 @@ class ArraysCache(_BaseCache):
     # machinery): the write lands before the deferred decrement -- the
     # alias-write window above -- where stock's eager decrement landed
     # first. Reads are serialized per cache: a reentrant lock guards the
-    # advance bookkeeping, folds, scheduling, and copies, because
-    # stock's pure attribute reads were trivially thread-safe while
-    # unserialized concurrent folds of one array deadlock on their
-    # thread-local streams. Cross-thread *mutation* (filter/extend/
-    # finalize racing anything) remains unsupported, as in stock.
+    # advance bookkeeping, folds, scheduling, copies, and metadata
+    # serialization, because stock's pure attribute reads were trivially
+    # thread-safe while unserialized concurrent folds of one array deadlock
+    # on their thread-local streams. Shallow aliases share that lock and
+    # their per-field bookkeeping for as long as they share the backing
+    # array; deepcopy preserves that topology within an alias graph.
+    # Cross-thread *mutation* (filter/extend/finalize racing anything)
+    # remains unsupported, as in stock.
 
     def __new__(cls, *args, **kwargs):
         instance = super().__new__(cls)
         instance._left_padding = None
         instance._lengths = None
-        instance._lp_advance = 0
-        instance._len_advance = 0
+        instance._lp_state = _ArraysCacheFieldState()
+        instance._len_state = _ArraysCacheFieldState()
         # bool-metadata promotion trackers: stock's eager subtraction
         # promoted a bool field to int32 on the FIRST advance() -- even a
         # zero or cancelling one -- so validation must remember that
         # independently of the pending total (see _logical_dtype)
-        instance._lp_promoted = False
-        instance._len_promoted = False
         # Serializes the deferred-decrement machinery (advance
-        # bookkeeping, folds, scheduling, copies). Stock's pure
-        # attribute reads were trivially thread-safe; deferred folding
-        # makes reads mutate, and two unserialized folds of the same
-        # array from different threads deadlock on their thread-local
-        # streams. Reentrant: mutators call _sync/_fold under it.
-        instance._fold_lock = threading.RLock()
+        # bookkeeping, folds, scheduling, copies, serialization). The
+        # native lock subclass is pickleable and deepcopy-aware without
+        # adding a wrapper on the per-token advance() hot path.
+        instance._fold_lock = _ArraysCacheFoldLock()
         return instance
 
     def __init__(self, size, left_padding: Optional[List[int]] = None):
         self.cache = [None] * size
         if left_padding:
             self.left_padding = mx.array(left_padding)
+
+    @property
+    def _lp_advance(self):
+        return self._lp_state.advance
+
+    @_lp_advance.setter
+    def _lp_advance(self, value):
+        self._lp_state.advance = value
+
+    @property
+    def _len_advance(self):
+        return self._len_state.advance
+
+    @_len_advance.setter
+    def _len_advance(self, value):
+        self._len_state.advance = value
+
+    @property
+    def _lp_promoted(self):
+        return self._lp_state.promotion[0]
+
+    @_lp_promoted.setter
+    def _lp_promoted(self, value):
+        self._lp_state.promotion[0] = value
+
+    @property
+    def _len_promoted(self):
+        return self._len_state.promotion[0]
+
+    @_len_promoted.setter
+    def _len_promoted(self, value):
+        self._len_state.promotion[0] = value
 
     # Integer metadata dtypes as (bits, signed): used to mirror stock's
     # eager per-call scalar range check in advance() and to wrap the
@@ -858,11 +915,17 @@ class ArraysCache(_BaseCache):
                 # Bound chains from repeatedly assigning lazy expressions
                 _try_schedule(v)
             self._left_padding = v
-            self._lp_advance = 0
             if not same:
                 # Only a genuinely new array resets the bool-promotion
                 # record; re-assigning the backing array is not a reset
-                self._lp_promoted = False
+                promotion = (
+                    self._len_state.promotion
+                    if v is not None and v is self._lengths
+                    else None
+                )
+                self._lp_state = _ArraysCacheFieldState(promotion=promotion)
+            else:
+                self._lp_advance = 0
 
     @property
     def lengths(self):
@@ -890,9 +953,15 @@ class ArraysCache(_BaseCache):
             if v is not None:
                 _try_schedule(v)
             self._lengths = v
-            self._len_advance = 0
             if not same:
-                self._len_promoted = False
+                promotion = (
+                    self._lp_state.promotion
+                    if v is not None and v is self._left_padding
+                    else None
+                )
+                self._len_state = _ArraysCacheFieldState(promotion=promotion)
+            else:
+                self._len_advance = 0
 
     def _sync(self):
         """Fold both pending integer decrements into the stored arrays."""
@@ -915,12 +984,12 @@ class ArraysCache(_BaseCache):
 
     def __copy__(self):
         # A shallow copy shares the backing arrays (as stock's did) --
-        # and the fold lock, so folds on the shared arrays stay
-        # serialized across both objects. Fold first: otherwise the
-        # copied pending counters would apply this cache's decrement
-        # once per object to the shared array. If the fold cannot land
-        # (captured tracer), a copy would escape the trace holding the
-        # transient tracer -- raise like the other guarded mutations.
+        # plus the fold lock and mutable per-field bookkeeping, so a
+        # pending total is folded only once across the alias graph.
+        # Fold first to preserve stock's copy-time visibility. If the
+        # fold cannot land (captured tracer), a copy would escape the
+        # trace holding the transient tracer -- raise like the other
+        # guarded mutations.
         with self._fold_lock:
             self._sync()
             self._require_folded("__copy__")
@@ -932,15 +1001,16 @@ class ArraysCache(_BaseCache):
         # Same guard as __copy__: python's default deepcopy would copy
         # __dict__ directly, letting a deep copy taken mid-trace escape
         # with the transient tracer and the pending counter. The copy
-        # gets its own fold lock (its arrays are independent).
+        # gets an independent lock for an independent array, while a
+        # shared deepcopy memo preserves lock/state sharing within a
+        # copied shallow-alias graph.
         with self._fold_lock:
             self._sync()
             self._require_folded("__deepcopy__")
             new = self.__class__.__new__(self.__class__)
             memo[id(self)] = new
             for k, v in self.__dict__.items():
-                if k != "_fold_lock":
-                    new.__dict__[k] = copy.deepcopy(v, memo)
+                new.__dict__[k] = copy.deepcopy(v, memo)
             return new
 
     def _materialize(self):
@@ -1005,10 +1075,6 @@ class ArraysCache(_BaseCache):
         # "" for an absent field, "<dtype>:<comma-joined values>"
         # otherwise (an empty array stays distinct from None and dtype
         # survives the round trip).
-        if self._left_padding is None and self._lengths is None:
-            return ""
-        self._sync()
-
         def encode(a):
             if a is None:
                 return ""
@@ -1027,7 +1093,11 @@ class ArraysCache(_BaseCache):
                 )
             return dtype + ":" + ",".join(str(x) for x in a.tolist())
 
-        return (encode(self._left_padding), encode(self._lengths))
+        with self._fold_lock:
+            if self._left_padding is None and self._lengths is None:
+                return ""
+            self._sync()
+            return (encode(self._left_padding), encode(self._lengths))
 
     @meta_state.setter
     def meta_state(self, v):
@@ -1059,9 +1129,13 @@ class ArraysCache(_BaseCache):
         self._sync()
         self._require_folded("filter")
         if self._left_padding is not None:
+            promoted = self._lp_promoted
             self._left_padding = self._left_padding[batch_indices]
+            self._lp_state = _ArraysCacheFieldState(promoted=promoted)
         if self._lengths is not None:
+            promoted = self._len_promoted
             self._lengths = self._lengths[batch_indices]
+            self._len_state = _ArraysCacheFieldState(promoted=promoted)
         self._materialize()
 
     def extend(self, other):
@@ -1091,15 +1165,32 @@ class ArraysCache(_BaseCache):
 
             return mx.concatenate([a, b])
 
+        def materialize_promoted_bool(a, promoted):
+            # Stock's first subtraction physically promoted bool metadata
+            # to int32. Preserve that logical dtype before mixed-dtype
+            # concatenation (bool + int8 would otherwise become int8).
+            if a is not None and promoted and a.dtype == mx.bool_:
+                return a.astype(mx.int32)
+            return a
+
         self._sync()
         other._sync()
         self._require_folded("extend")
         other._require_folded("extend")
         self.cache = [cat(c, o) for c, o in zip(self.cache, other.cache)]
-        self._left_padding = cat(self._left_padding, other._left_padding)
-        self._lengths = cat(self._lengths, other._lengths)
-        self._lp_promoted = self._lp_promoted or other._lp_promoted
-        self._len_promoted = self._len_promoted or other._len_promoted
+        self._left_padding = cat(
+            materialize_promoted_bool(self._left_padding, self._lp_promoted),
+            materialize_promoted_bool(other._left_padding, other._lp_promoted),
+        )
+        self._lengths = cat(
+            materialize_promoted_bool(self._lengths, self._len_promoted),
+            materialize_promoted_bool(other._lengths, other._len_promoted),
+        )
+        # Concatenation rebinds both fields. Any logical bool promotion
+        # was materialized above, so the new arrays start with fresh
+        # per-field state and no alias bookkeeping from either input.
+        self._lp_state = _ArraysCacheFieldState()
+        self._len_state = _ArraysCacheFieldState()
         self._materialize()
 
     def extract(self, idx):
@@ -1117,8 +1208,8 @@ class ArraysCache(_BaseCache):
         self._require_folded("finalize")
         self._lengths = None
         self._left_padding = None
-        self._lp_promoted = False
-        self._len_promoted = False
+        self._lp_state = _ArraysCacheFieldState()
+        self._len_state = _ArraysCacheFieldState()
 
     def advance(self, N):
         # Integer bookkeeping only: building this with mx.array arithmetic
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -3,7 +3,9 @@
 import copy
 import operator
 import threading
+import weakref
 from collections import deque
+from contextlib import ExitStack, contextmanager
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional
 
@@ -611,15 +613,65 @@ def _try_schedule(*arrays):
 
 
 class _ArraysCacheFoldLock(type(threading.RLock())):
-    """An RLock that recreates itself across pickle/deepcopy.
+    """An RLock used by a cache alias group or a metadata-array alias group.
 
-    A shallow cache copy keeps the same lock by normal object sharing.
-    Pickle and deepcopy create a new native lock; their memo preserves
-    sharing when an alias graph contains the same lock more than once.
+    Normal object sharing keeps one lock across aliases. Pickle and deepcopy
+    create a new native lock; their memo preserves sharing when an alias graph
+    contains the same lock more than once.
     """
 
     def __reduce__(self):
         return (type(self), ())
+
+
+class _ArraysCacheArraySync:
+    """Synchronization and physical-promotion state for one backing array."""
+
+    def __init__(self, promotion=None, lock=None):
+        self.promotion = [False] if promotion is None else promotion
+        self.lock = _ArraysCacheFoldLock() if lock is None else lock
+
+
+# Public setters can attach one mx.array to independently constructed caches.
+# Keep synchronization by backing-array identity so their deferred reads cannot
+# concurrently mutate that array on different thread-local MLX streams. The
+# weak registry does not retain arrays after their last real owner is gone.
+_ARRAYS_CACHE_ARRAY_SYNCS = {}
+_ARRAYS_CACHE_ARRAY_SYNCS_LOCK = threading.Lock()
+
+
+def _register_arrays_cache_array(arr, state):
+    if arr is None:
+        return state
+    if not hasattr(state, "array_lock"):
+        # Compatibility with cache pickles produced before array-level locks.
+        state.array_lock = _ArraysCacheFoldLock()
+    key = id(arr)
+    with _ARRAYS_CACHE_ARRAY_SYNCS_LOCK:
+        item = _ARRAYS_CACHE_ARRAY_SYNCS.get(key)
+        if item is not None and item[0]() is arr:
+            sync = item[1]
+            state.promotion = sync.promotion
+            state.array_lock = sync.lock
+            return state
+
+        sync = _ArraysCacheArraySync(state.promotion, state.array_lock)
+
+        def remove(ref, key=key):
+            with _ARRAYS_CACHE_ARRAY_SYNCS_LOCK:
+                current = _ARRAYS_CACHE_ARRAY_SYNCS.get(key)
+                if current is not None and current[0] is ref:
+                    del _ARRAYS_CACHE_ARRAY_SYNCS[key]
+
+        ref = weakref.ref(arr, remove)
+        _ARRAYS_CACHE_ARRAY_SYNCS[key] = (ref, sync)
+        return state
+
+
+def _arrays_cache_field_state(arr=None, advance=0, promoted=False):
+    return _register_arrays_cache_array(
+        arr, _ArraysCacheFieldState(advance=advance, promoted=promoted)
+    )
 
 
 class _ArraysCacheFieldState:
@@ -631,9 +683,10 @@ class _ArraysCacheFieldState:
     decrement.
     """
 
-    def __init__(self, advance=0, promoted=False, promotion=None):
+    def __init__(self, advance=0, promoted=False, promotion=None, array_lock=None):
         self.advance = advance
         self.promotion = [promoted] if promotion is None else promotion
+        self.array_lock = _ArraysCacheFoldLock() if array_lock is None else array_lock
 
 
 class ArraysCache(_BaseCache):
@@ -684,10 +737,13 @@ class ArraysCache(_BaseCache):
     # bool metadata bool, numpy scalar offsets promoting the metadata
     # dtype) are not reproduced. bool metadata arithmetic is reproduced
     # at int32 precision (stock's promotion): the dtype stays bool
-    # through zero-net advance histories, and after the first advance()
-    # on the field -- even a zero or cancelling one, and through a bool
-    # array shared by both fields -- later offsets validate against
-    # int32 exactly as stock's already-promoted array would. A read that encounters a *tracer*
+    # through zero-net advance histories. Physical promotion and fold
+    # synchronization follow the backing array even when public setters
+    # attach it to independently constructed caches; shallow aliases
+    # additionally share their per-field pending counter. After the first
+    # advance() on a bool field -- even a zero or cancelling one -- later
+    # offsets through any of those aliases validate against int32 exactly
+    # as stock's already-promoted array would. A read that encounters a *tracer*
     # -- metadata swapped into a trace by
     # ``mx.compile(..., inputs=vars(cache))`` -- is pure: no fold, no
     # scheduling, like stock's attribute access, and the pending
@@ -711,10 +767,16 @@ class ArraysCache(_BaseCache):
     # first. Reads are serialized per cache: a reentrant lock guards the
     # advance bookkeeping, folds, scheduling, copies, and metadata
     # serialization, because stock's pure attribute reads were trivially
-    # thread-safe while unserialized concurrent folds of one array deadlock
-    # on their thread-local streams. Shallow aliases share that lock and
-    # their per-field bookkeeping for as long as they share the backing
-    # array; deepcopy preserves that topology within an alias graph.
+    # thread-safe while unserialized concurrent folds of one array
+    # deadlock on their thread-local streams. A second, array-identity
+    # lock serializes independently constructed caches that share a
+    # backing array. Shallow aliases share their cache lock and per-field
+    # bookkeeping; deepcopy preserves that topology within an alias
+    # graph. Quiescent generic pickle preserves the same topology, but
+    # pickle does not expose a lifecycle hook that can hold the source
+    # lock for the complete outer Pickler traversal: concurrent generic
+    # pickle is unsupported. ``meta_state`` and ``save_prompt_cache``
+    # remain linearizable complete metadata snapshots.
     # Cross-thread *mutation* (filter/extend/finalize racing anything)
     # remains unsupported, as in stock.
 
@@ -838,7 +900,7 @@ class ArraysCache(_BaseCache):
                 f"for {dtype} metadata"
             )
 
-    def _fold(self, arr_attr, adv_attr):
+    def _fold(self, arr_attr, state_attr):
         """Fold the pending decrement into the backing array -- in place,
         so aliases observe it -- and schedule the evaluation. A no-op on
         tracers (see _try_schedule): the fold stays pending, the counter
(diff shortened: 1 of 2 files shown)


patrick-sunderland:
> so any single completion over ~499000/(#SSM−1) tokens (~14.1k on Qwen3.5-122B, 36 SSM layers) kills the process with [metal::malloc] Resource limit (499000) exceeded.

FYI: Testing this PR on a M3 Ultra 256GB and the Qwen3.5-122B-A10B (mxfp8) model worked well with a 43k prompt. The above issue was resolved in my case, no more `[metal::malloc] Resource limit (499000) exceeded` errors. 

Caching/cache hits functioned as expected.

pierre427:
Confirming this bug class from an independent reproduction: on a 36-SSM-layer qwen3_5-family model served through the batch path, long completions died at the Metal per-process object limit right around the predicted token count, with flat byte telemetry — exactly the dead-lazy-node-per-layer-per-token mechanism described here.

Data point for review: in our serving fork we shipped the minimal `mx.depends` variant from #1632 (tie `lengths`/`left_padding` into the state graph in `advance()`), with a regression test bounding the metadata graph's edge count after 256 decode steps; it fails before the fix and passes after, and the crash class is gone in production. That mechanism alone was sufficient for the leak; the deferred-integer bookkeeping here additionally removes the per-step array arithmetic and adds serialization thread-safety, which the smaller fix does not address.

zcbenz:
Closing in favor of #1632.

PR #1645, opened 2026-08-01
Fix divide-by-zero in BatchGenerator.stats()

th4ruka:
## Summary
The `BatchGenerator.stats()` computes `prompt_tps`/`generation_tps` in its `finally` block without guarding a zero denominator:

```python
stats.prompt_tps = stats.prompt_tokens / stats.prompt_time
...
stats.generation_tps = stats.generation_tokens / stats.generation_time
```
When no prompt/generation time has accrued, prompt_time (or generation_time) is 0, so this raises ZeroDivisionError. Because it runs in a finally, it also masks the original exception whenever the with gen.stats() body raises before prompt processing completes. The real traceback is replaced by a confusing "division by zero".

## Reproduction
```python
from mlx_lm.utils import load

model, tok = load("mlx-community/SmolLM-135M-Instruct-4bit")
gen = BatchGenerator(model, stop_tokens=[[t] for t in tok.eos_token_ids])

# 1) A stats context that accrues no time
try:
    with gen.stats():
        pass
    print("case 1: no error")
except Exception as e:
    print("case 1:", type(e).__name__)

# 2) A real error raised inside the body
try:
    with gen.stats():
        raise ValueError("boom")
except Exception as e:
    print("case 2:", type(e).__name__)
```
On `main`:
```
case 1: ZeroDivisionError
case 2: ZeroDivisionError   # the real ValueError is masked
```
With this change:
```
case 1: no error
case 2: ValueError
```
## Fix
Fall back to 0.0 tps when the elapsed time is zero (matching the BatchStats default of tps: float = 0). This keeps the finally block from raising, so a zero-work stats context no longer crashes and a body exception now propagates unmasked.

## Testing
Added two regression tests in `tests/test_generate.py`:
- `test_batch_stats_zero_time_no_divide_by_zero`: a zero-work stats context yields `0.0` tps instead of raising.
- `test_batch_stats_does_not_mask_exceptions`:  an exception in the body propagates instead of being masked.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1654,10 +1654,16 @@ class BatchGenerator:
             gen_time = total_time - self._prompt_time_counter
             stats.prompt_tokens += self._prompt_tokens_counter
             stats.prompt_time += self._prompt_time_counter
-            stats.prompt_tps = stats.prompt_tokens / stats.prompt_time
+            stats.prompt_tps = (
+                stats.prompt_tokens / stats.prompt_time if stats.prompt_time else 0.0
+            )
             stats.generation_tokens += self._gen_tokens_counter
             stats.generation_time += gen_time
-            stats.generation_tps = stats.generation_tokens / stats.generation_time
+            stats.generation_tps = (
+                stats.generation_tokens / stats.generation_time
+                if stats.generation_time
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1646, opened 2026-08-02
Add opt-in q4_0 lattice calibration to quantization

GoodOlClint:
## Proposed changes

Adds an opt-in **calibration** to quantization that reproduces a checkpoint's existing `Q4_0` grid instead of deriving a new one.

### The problem

Some checkpoints ship weights that were already trained against a specific quantization grid — Google's Gemma 4 QAT releases are the motivating case. `nn.quantize` derives a fresh grid from those weights by min/max, discarding the alignment they were trained for. The converted model then sits further from the original than it needs to, at the same bit width and the same file size.

The grid is recoverable: 32-element groups scaled by the **signed** element of largest magnitude, `d = extremum / -8`, expressible as ordinary affine `scales = d`, `biases = -8 * d`.

### What this changes

```bash
mlx_lm.convert --hf-path google/gemma-4-12B-it-qat-q4_0-unquantized -q --q-calibration q4_0
```

- `quantize_model(..., calibration="standard" | "q4_0")`
- `convert(..., q_calibration=...)` — appended after existing parameters, so positional callers are unaffected
- `--q-calibration` CLI flag
- `utils.q4_0_quantize(w)` exposed for direct use

`q4_0` resolves to 4 bits / group size 32 / affine and **rejects** conflicting values rather than silently overriding them, failing before the model is loaded. Per-layer predicate parameters that conflict are rejected by path, and each calibrated module's geometry is verified after conversion rather than assumed.

**This is a derivation, not a storage format.** Output is ordinary affine 4-bit; converted models load and run through the existing path unchanged, with no decode-side change.

### Evidence

Each converted model compared against its own bfloat16 reference over 200 chat-formatted prompts, teacher-forced. Mean KL divergence per token position (lower is better) and exact top-1 agreement, stock group-32 conversion vs this calibration **at identical file size**:

| model | KL, stock | KL, calibrated | top-1, stock | top-1, calibrated |
|---|---|---|---|---|
| Gemma 4 E2B | 0.0220 | **0.0175** | 95.7% | **96.2%** |
| Gemma 4 E4B | 0.0160 | **0.0112** | 96.6% | **97.2%** |
| Gemma 4 12B | 0.1219 | **0.0631** | 92.8% | **94.3%** |
| Gemma 4 31B | 0.0321 | **0.0205** | 96.5% | **96.9%** |

Correctness is established independently of those measurements: the derivation reproduces Google's own shipped `Q4_0` GGUF **exactly** — 100% of codes and scales across 721M weights on 12B — and `q4_0_quantize` was then checked tensor-by-tensor against that verified artifact on E2B: **275 tensors, 0 mismatches**.

Worth noting for anyone reproducing this: simply passing `--q-group-size 32` does **not** get you there. It costs ~11% more disk and, on the dense models measured, buys nothing — 12B actually got slightly worse (0.1048 → 0.1219). The gain is in how the grid is derived, not in the group size.

### Scope and limits, stated plainly

- **Only `nn.Linear` is calibrated.** Other quantizable modules still land on the same 4-bit/group-32 affine grid but derive it the standard way. Affects accuracy, not loadability; documented in the docstring.
- **No benefit on mixture-of-experts.** The same measurement on Gemma 4 26B-A4B showed no significant change (0.0707 → 0.0680, overlapping confidence intervals). No claim is made there.
- **Opt-in only.** No model-name or fingerprint auto-detection.
- Single-run measurements on one prompt corpus with a fixed seed.

### Notes for review

- A mirrored change for `mlx-swift-lm` is prepared so the two conversion paths stay in step.
- `tests/test_models.py::test_ssm` and 8 cases in `tests/test_generate.py` fail on this branch, but they fail identically on a clean `main` — pre-existing and unrelated.

## Checklist

- [x] `black` clean via pre-commit
- [x] Tests added
- [x] Documentation updated (docstrings + CLI help)

Tests: 14 added to `tests/test_utils.py` covering hand-computable codes/scales/biases, positive and negative extrema, clipping at both ends, degenerate all-zero groups, per-group scaling across rows, dequantization onto the intended lattice, rejection of incompatible global and per-layer settings, shape validation, that the predicate is evaluated exactly once per module, and one pinning the signed-extremum rule apart from `-abs(w).max()/8` — the error that silently produces a different grid on roughly half of all groups.

---

**Published weights using this path** — 4-bit MLX conversions of the Gemma 4 QAT checkpoints, which load with released `mlx-lm` (they are ordinary affine 4-bit; none of this PR is needed to *use* them):

- [gemma-4-E2B-it-qat-q4_0-mlx](https://huggingface.co/goodolclint/gemma-4-E2B-it-qat-q4_0-mlx)
- [gemma-4-E4B-it-qat-q4_0-mlx](https://huggingface.co/goodolclint/gemma-4-E4B-it-qat-q4_0-mlx)
- [gemma-4-12B-it-qat-q4_0-mlx](https://huggingface.co/goodolclint/gemma-4-12B-it-qat-q4_0-mlx)
- [gemma-4-31B-it-qat-q4_0-mlx](https://huggingface.co/goodolclint/gemma-4-31B-it-qat-q4_0-mlx)

Each was produced by this branch and verified bit-identical to Google's shipped `Q4_0` grid before upload.

### DIFF
--- a/mlx_lm/convert.py
+++ b/mlx_lm/convert.py
@@ -14,6 +14,7 @@ from .utils import (
     quantize_model,
     save,
     upload_to_hub,
+    validate_calibration,
 )
 
 
@@ -97,6 +98,7 @@ def convert(
         Union[Callable[[str, nn.Module, dict], Union[bool, dict]], str]
     ] = None,
     trust_remote_code: bool = False,
+    q_calibration: str = "standard",
 ):
     # Check the save path is empty
     if isinstance(mlx_path, str):
@@ -107,6 +109,9 @@ def convert(
             f"Cannot save to the path {mlx_path} as it already exists."
             " Please delete the file/directory or specify a new path to save to."
         )
+
+    if quantize:
+        validate_calibration(q_calibration, q_group_size, q_bits, q_mode)
 
     print("[INFO] Loading")
     model, tokenizer, config = load(
@@ -156,6 +161,7 @@ def convert(
             q_bits,
             mode=q_mode,
             quant_predicate=quant_predicate,
+            calibration=q_calibration,
         )
 
     if dequantize:
@@ -204,6 +210,16 @@ def configure_parser() -> argparse.ArgumentParser:
         help="Group size for quantization.",
         type=int,
         default=None,
+    )
+    parser.add_argument(
+        "--q-calibration",
+        choices=["standard", "q4_0"],
+        default="standard",
+        help=(
+            "How the quantization grid is derived. 'q4_0' reproduces the Q4_0 grid "
+            "for checkpoints already trained against it, instead of deriving a new "
+            "one. Requires --q-bits 4, --q-group-size 32 and affine mode."
+        ),
     )
     parser.add_argument(
         "--q-bits",
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -715,8 +715,7 @@ def upload_to_hub(path: str, upload_repo: str):
     else:
         provenance = ""
 
-    card.text = dedent(
-        f"""
+    card.text = dedent(f"""
         # {upload_repo}
         {provenance}
         ## Use with mlx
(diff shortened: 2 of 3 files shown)


zcbenz:

PR #1647, opened 2026-08-03
Fix MiniMax M2 nullable tool arguments

Mr-Neutr0n:
## Summary

- preserve non-null MiniMax M2 tool arguments when their schema also allows `null`
- continue mapping explicit `null`, `none`, and `nil` values to Python `None`
- add regression coverage for nullable string and integer arguments

## Why

The converter checked whether `null` appeared anywhere in a parameter's allowed types and returned `None` before inspecting the actual value. As a result, values such as `"weather"` and `"10"` were discarded for nullable parameters.

## Tests

- `uv run python -m unittest tests.test_tool_parsing` (6 passed)
- `uvx pre-commit run --files mlx_lm/tool_parsers/minimax_m2.py tests/test_tool_parsing.py` (Black and isort passed)

### DIFF
--- a/mlx_lm/tool_parsers/minimax_m2.py
+++ b/mlx_lm/tool_parsers/minimax_m2.py
@@ -86,15 +86,11 @@ def _extract_types_from_schema(schema: Any) -> list[str]:
 
 
 def _convert_param_value_with_types(value: str, param_types: list[str]) -> Any:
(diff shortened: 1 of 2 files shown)


zcbenz:
This looks like an AI fabricated issue so I'm closing, please provide a way to reproduce the problem if it is true bug.

PR #1656, opened 2026-08-04
Fix Falcon-Mamba long prompt resource exhaustion

### LINKED ISSUE #1637: Mamba2 models (mlx_lm.generate and mlx_lm.server) crash/hang silently on prompts beyond ~1-2k tokens
Both mlx_lm.generate (direct CLI call) and mlx_lm.server fail on Mamba2-architecture models once the input prompt exceeds roughly 300–2000 tokens, depending on the run. Short prompts (~264 tokens) work correctly and produce coherent output. This is reproducible independent of the HTTP server layer, and independent of prompt content (tested both repetitive filler text and varied/random text).

## To Reproduce

pip install mlx-lm==0.31.3

### Works fine:
mlx_lm.generate --model mlx-community/Falcon3-Mamba-7B-Instruct-8bits \
  --prompt "<~260 token prompt>" --max-tokens 50 --temp 0

### Crashes:
mlx_lm.generate --model mlx-community/Falcon3-Mamba-7B-Instruct-8bits \
  --prompt "<~1900 token prompt>" --max-tokens 50 --temp 0

## Actual behavior

- Via mlx_lm.generate: process aborts with SIGABRT (exit code 134). No Python traceback, no stderr output beyond the model-download progress bar. No macOS crash report is generated in ~/Library/Logs/DiagnosticReports/.
- Via mlx_lm.server: request returns HTTP 200, server log shows Prompt processing progress: N/N completing normally, but the HTTP response body is empty (no content, no error) and no traceback appears in server logs.
- Reproduced identically on two different checkpoints/quantizations: mlx-community/Mamba-Codestral-7B-v0.1-4bit and mlx-community/Falcon3-Mamba-7B-Instruct-8bits, ruling out a single-checkpoint quantization defect.

## Expected behavior

Model should process prompts up to its stated context window (both checkpoints advertise 32k+ tokens) without crashing or silently dropping the response.

## Comparison / evidence this is mlx-lm-specific

The same Hudson/falcon-mamba-instruct:7b-q4_0 model (GGUF, llama.cpp backend) served via Ollama handles a ~7.2k token prompt cleanly (prompt_eval_count: 7231, done_reason: "stop", coherent output). This suggests the bug is specific to mlx-lm's Mamba2 implementation (likely the SSM-scan kernel or ArraysCache handling at longer sequence lengths), not the Mamba architecture itself.

## Attempted debugging

Tried capturing a backtrace via lldb, but the traced process hung indefinitely during Metal GPU operations (suspected mach-exception-port conflict between lldb and the Metal driver) rather than reaching the crash point, so no native backtrace was obtained.

## Environment

- mlx: 0.32.0
- mlx-lm: 0.31.3
- macOS: 15.7.7 (24G720)
- Hardware: Apple M2 Max, 32GB RAM
- Python: (venv-based install)

mrjoema:
## Reproduction

- Apple M4 Pro (48 GB), macOS 27.0, Python 3.13.5
- `mlx==0.32.0`, `mlx-lm==0.31.3`
- `mlx-community/Falcon3-Mamba-7B-Instruct-8bits`
- A 1,298-token effective prompt fails with `[metal::malloc] Resource limit (499000) exceeded`.

## Root cause and fix

Falcon-Mamba normalized `delta`, `B`, and `C` twice, and each call allocated an `mx.ones` scale. The lazy graph therefore accumulated six temporary scales per token per layer and exceeded Metal's resource-count limit.

Normalize each value once with `weight=None`. The new code also replaces the nested `map`/`lambda` expression with an explicit split and three normalization calls for readability.

## Verification

- Full model tests: 78 passed, 1 skipped
- Real-model prompts of 1,909 and 4,105 effective tokens complete successfully
- A regular 50-token greedy generation exactly matches the pre-fix output

Fixes #1637

### DIFF
--- a/mlx_lm/models/mamba.py
+++ b/mlx_lm/models/mamba.py
@@ -63,10 +63,6 @@ class MambaBlock(nn.Module):
         self.time_step_rank = int(args.time_step_rank)
         self.use_conv_bias = args.use_conv_bias
         self.use_bcdt_rms = args.use_bcdt_rms
-        if self.use_bcdt_rms:
-            self.mixer_norm = lambda x: mx.fast.rms_norm(
-                x, mx.ones(x.shape[-1], x.dtype), eps=args.mixer_rms_eps
-            )
 
         self.in_proj = nn.Linear(
             self.hidden_size, self.intermediate_size * 2, bias=args.use_bias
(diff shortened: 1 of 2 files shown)


michalk8 (approved):
I tested this locally with a prompt `The quick brown fox jumps over the lazy dog. " * 2000`, this PR fixes it.
LGTM, thanks a lot!

Issue #361, opened 2025-08-08
NaN values doing fine tuning with gpt-oss:20b


nfsecurity:
First of all, many thanks to all the people who collaborate with the mlx project.

I am trying to fine tune gpt-oss:20b but I am receiving "nan" values in the Train Loss and Value Loss in every iteration. 

I am using the latest mlx-lm in an Mac Studio M1 Ultra 64GB:

```
% pip3.11 show mlx-lm

Name: mlx-lm
Version: 0.26.3
Summary: LLMs with MLX and the Hugging Face Hub
Home-page: https://github.com/ml-explore/mlx-lm
Author: MLX Contributors
Author-email: mlx@group.apple.com
License: MIT
Location: /opt/homebrew/lib/python3.11/site-packages
Requires: jinja2, mlx, numpy, protobuf, pyyaml, transformers
Required-by: 

% pip3.11 show mlx   

Name: mlx
Version: 0.27.1
Summary: A framework for machine learning on Apple silicon.
Home-page: https://github.com/ml-explore/mlx
Author: MLX Contributors
Author-email: mlx@group.apple.com
License: MIT
Location: /opt/homebrew/lib/python3.11/site-packages
Requires: mlx-metal
Required-by: mlx-lm
```

Here are the steps I am following:

**Download the model**

```
% mlx_lm.convert --hf-path openai/gpt-oss-20b
```
I also tried with "--dtype bfloat16" and "-q --q-bits 8", but the result is the same.

**Create config.yaml file**
```
model: "./mlx_model"
train: true
fine_tune_type: lora
data: "dataset"
seed: 0
num_layers: 16
batch_size: 1

iters: 50
val_batches: 25
learning_rate: 1e-5
steps_per_report: 10
steps_per_eval: 100
resume_adapter_file: null
adapter_path: "adapters"
save_every: 100
test: true
test_batches: -1
max_seq_length: 8192
grad_checkpoint: true
lora_parameters:
  keys: [
    "self_attn.q_proj",
    "self_attn.v_proj",
    "self_attn.k_proj",
    "self_attn.o_proj",
    "mlp.gate_proj",
    "mlp.down_proj",
    "mlp.up_proj"
  ]
  rank: 8
  scale: 2.0
  dropout: 0.0

lr_schedule: false
```
I also tried deleting "keys" parameters, but the result is the same.

**Run training**
```
% mlx_lm.lora --train -c config.yaml
Loading configuration file config.yaml
Loading pretrained model
[...]

Issue #49, opened 2025-03-24
Val and Training Loss NaN on M4 Max starting on mlx-lm 0.22.2


ivanfioravanti:
Hi,
we are experiencing a training loss going to NaN with 0.22.2 version.

Here an easy command to replicate:
```bash
mlx_lm.lora \                                                                                                                                                    
    --model mlx-community/Mistral-7B-Instruct-v0.3 \
    --train \
    --batch-size 4 \
    --num-layers 16 \
    --data mlx-community/wikisql
```

here a screenshot:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/1e7e0041-14c5-404a-a13a-6408d1c3c0a3" />

zhengxixuan:
+1

awni:
This is fixed. Upgrade mlx to 0.24.1

PR #1659, opened 2026-08-05
Add optional gradient clipping and non-finite guard to the LoRA trainer

### LINKED ISSUE #1658: LoRA trainer has no gradient clipping or non-finite guard: one bad batch loses the whole run
### Summary

The LoRA trainer has no gradient clipping and no guard against non-finite gradients. A single batch whose gradient overflows puts the LoRA weights into NaN permanently, and every subsequent iteration reports `nan` with no warning and no way to recover. The only way to notice is to watch the loss.

`mx.optimizers.clip_grad_norm(grads, max_norm)` already exists in MLX and returns `(clipped_grads, total_norm)`. `mlx_lm/tuner/trainer.py` never calls it. A grep for `clip`, `isnan`, `isfinite` or `norm` in that file returns nothing, and there is no `--grad-clip` / `--max-grad-norm` flag on `mlx_lm.lora`.

### What I hit

Fine-tuning `Qwen/Qwen3-1.7B` (LoRA rank 8, 16 layers, bf16), three runs died the same way:

| Run | lr | batch | scale | optimizer | schedule | Result |
|---|---|---|---|---|---|---|
| 1 | 1e-5 | 2 | 20.0 | adam | none | NaN at iter 380, after 370 stable iters (loss 0.95→1.17) |
| 2 | 5e-6 | 4 | 20.0 | adam | none | NaN at iter 10, resuming from run 1's iter-200 checkpoint |
| 3 | 5e-6 | 4 | 8.0 | adamw | cosine + warmup 100 | **NaN at iter 10, at lr 4.5e-7** |

Run 3 is the one I think matters: NaN at a learning rate of **4.5e-7**, still inside warmup, where `lora_b` is barely off its zero init and the model is essentially the base model. This is not divergence from too high an lr, and lowering the lr did not fix it.

```
Trainable parameters: 0.289% (4.981M/1720.575M)
Iter 1: Val loss 1.472, Val took 49.392s
Iter 10: Train loss nan, Learning Rate 4.500e-07, It/sec 0.071, ...
```

I bisected the layers separately before filing, so this isn't a bad-data report:

- **Data**: 8786 rows, token ids within `[0, 151668]` for vocab 151936, no empty rows, shortest 420 tokens.
- **Forward**: 20 steps, loss 1.09-1.87, `|logit|max` 58-66, no NaN or inf.
- **Backward** (grad-checkpoint on): 25 steps, all gradients finite, `|grad|max` 0.39-1.48.
- **`mx.compile` + optimizer update**: 15 steps with run 3's exact config, no NaN.

Individual batches and the step function are both fine. The NaN only shows up in the real training sequence, on a specific batch.

**Correction to an earlier version of this issue:** I originally wrote that the batch order is not reproducible, because `train()` calls `iterate_batches` without passing `seed`. That was wrong, and I'd rather fix it than leave it standing. `run()` in `lora.py` calls `np.random.seed(args.seed)` before training, and nothing in the CLI training path draws from numpy in between, so batch order for a normal `mlx_lm.lora` run *is* reproducible. There is a real but much smaller bug there, `iterate_batches` gates on `if seed:` so an explicit `seed=0` is ignored, which I've filed separately as #1660. It is not a factor in this issue.

### Two things I measured that shape the fix

**1. Clipping alone does not fix this.** `clip_grad_norm` propagates non-finite values rather than removing them:

```python
>>> g = {"a": mx.array([float("nan"), 1.0]), "b": mx.array([2.0, 3.0])}
>>> out, norm = optim.clip_grad_norm(g, 1.0)
>>> float(norm), tree_flatten(out)
(nan, [('a', array([nan, nan])), ('b', array([nan, nan]))])
```

The finite entries get contaminated too, because everything is scaled by `max_norm / norm` and `norm` is NaN. The `inf` case is worse: the normalizer becomes `0`, and `inf * 0` is NaN, so infinite gradients also poison every parameter.

So clipping covers large-but-finite gradients, which is a different problem from the one above. Fixing the NaN case needs both halves: clipping, plus a guard that drops the update when the norm is not finite.

**2. The guard has to stay in the graph.** `step()` is wrapped in `@partial(mx.compile, inputs=state, outputs=state)`. Branching on `float(norm)` in Python would force a device sync every iteration. This composition compiles and works:

```python
grad, norm = clip_grad_norm(grad, max_norm)
keep = mx.isfinite(norm)
grad = tree_map(lambda g: mx.where(keep, g, 0), grad)
optimizer.update(model, grad)
```

I verified `clip_grad_norm` works inside `mx.compile`, including across a step with a 1000x gradient spike, so the data-dependent value is not a tracing problem.

One caveat worth documenting rather than fixing: zeroing the gradient is not a complete no-op. A momentum optimizer still moves the weights from its existing state, and AdamW applies weight decay independently of the gradient (`w -> w*(1 - lr*wd)`), measured at 1.07e-07 relative at lr=5e-6. With Adam, or AdamW at `weight_decay=0.0`, the weights are exactly unchanged. Either way the bad batch contributes no learning signal, which is the point.

### Prior NaN reports

- **#49** "Val and Training Loss NaN on M4 Max" — closed, resolved by upgrading mlx to 0.24.1. That was an mlx bug.
- **#361** "NaN values doing fine tuning with gpt-oss:20b" — closed, 8 comments, several people reporting the same symptom, went toward data format and MXFP4 quantization.

Neither was caused by missing clipping, and I'm not claiming otherwise. I mention them only as evidence that people do run into NaN during fine-tuning, and that today there's nothing in the trainer that limits the damage.

### Proposed change

Deliberately minimal, and off by default so no existing run changes behavior:

1. `grad_clip: Optional[float] = None` on `TrainingArgs`.
2. `--grad-clip` on `mlx_lm.lora`, plus its `CONFIG_DEFAULTS` entry.
3. Apply it in the compiled `step()` between `value_and_grad` and `optimizer.update`, on the already-accumulated and averaged gradient rather than per micro-batch, so it composes correctly with `grad_accumulation_steps` and `average_gradients`.
4. The non-finite guard via `mx.where`, as above.

I have this working with tests, and would be happy to open a PR. Two questions before I do:

- Is the non-finite guard something you want in the same change? It's less standard than clipping, and I'd rather ask than assume. Clipping on its own would not have saved any of the three runs above.
- Do you prefer one `--grad-clip` flag controlling both, or the guard separated behind its own flag?

Environment: mlx-lm 0.31.3, mlx 0.32.0, M2 Pro 32 GB, macOS 26.5.2. Verified against `main` at e5baded.

axiom-of-choice:
Closes #1658 (opened for discussion; opening the PR since the implementation is done and tested, happy to change either design decision).

### Summary

The LoRA trainer has no bound on gradient size and no guard against non-finite gradients. One batch whose gradient overflows puts the LoRA weights into NaN permanently, and every later iteration reports `nan` with no warning and no recovery path.

This adds `grad_clip` to `TrainingArgs` and `--grad-clip` to `mlx_lm.lora`. Off by default, so no existing run changes behavior.

### High-level changes

- `mlx_lm/tuner/trainer.py`: `grad_clip: Optional[float] = None` on `TrainingArgs`; clipping plus a non-finite guard inside the compiled `step()`; `max_grad_norm` and `non_finite_grad_steps` surfaced to training callbacks; a warning when a step is dropped.
- `mlx_lm/lora.py`: `--grad-clip` flag and its `CONFIG_DEFAULTS` entry.
- `mlx_lm/LORA.md`: a "NaN Loss" section.
- `mlx_lm/examples/lora_config.yaml`: `grad_clip: null`.
- `tests/test_tuner_grad_clip.py`: 8 tests driving the real `train()`.

Clipping is applied to the already-accumulated and averaged gradient, immediately before `optimizer.update`, so it composes with `grad_accumulation_steps` and `average_gradients` rather than clipping each micro-batch separately.

### Why clipping alone is not enough

`clip_grad_norm` propagates non-finite values rather than removing them. A NaN in one entry makes the norm NaN, which scales every parameter to NaN. An `inf` norm makes the normalizer `0`, and `inf * 0` is NaN, so infinite gradients also poison every parameter:

| grads | total_norm | clipped result |
|---|---|---|
| `[nan, 1.0]`, `[2.0, 3.0]` | nan | all four entries nan |
| `[inf, 1.0]`, `[2.0, 3.0]` | inf | `[nan, 0.0]`, `[0.0, 0.0]` |

So clipping covers large-but-finite gradients, a different failure from the one that motivated the issue. The guard zeroes the gradient when the norm is not finite:

```python
grad, grad_norm = clip_grad_norm(grad, args.grad_clip)
keep = mx.isfinite(grad_norm)
grad = tree_map(lambda g: mx.where(keep, g, 0), grad)
```

`mx.where` keeps the check inside the compiled graph. A Python branch on `grad_norm` would force a device sync every iteration.

This is not a full no-op: a momentum optimizer still moves the weights from its existing state, and AdamW applies weight decay independently of the gradient (measured 1.07e-07 relative at lr=5e-6; exactly 0 with Adam or `weight_decay=0.0`). The non-finite values never reach the weights, which is the guarantee. Documented in `LORA.md` and in a code comment.

### Evaluation

**The default is inert.** 30 iterations, same seed, `main` at e5baded vs this branch with no `--grad-clip`:

```
main   losses: 4.169967 4.170533 4.171922 4.167962 4.160542 4.160848
branch losses: 4.169967 4.170533 4.171922 4.167962 4.160542 4.160848
```

Losses bit-identical. Weights differ by 2.98e-08, but `main` differs from *itself* across two runs by exactly the same 2.98e-08, so that is run-to-run nondeterminism, not this change.

**It catches a real overflow and the run recovers.** Qwen3-1.7B, LoRA rank 8 over 16 layers, batch 4, max_seq 2048, AdamW, cosine + warmup 100. This config previously died at iter 10 with `Train loss nan` and never recovered. With `--grad-clip 1.0`:

```
Iter 10: Train loss 1.3797, LR 4.500e-07, grad_norm 4.232,         skipped 0
Iter 20: Train loss 1.4634, LR 9.500e-07, grad_norm 4.200,         skipped 0
Iter 30: Train loss 1.4008, LR 1.450e-06, grad_norm 5044.074,      skipped 0
Iter 40: SKIPPED non-finite batch (loss 1.2308621406555176, norm nan) -- weights untouched
Iter 40: Train loss 1.2680, LR 1.950e-06, grad_norm 177650662.281, skipped 1
Iter 50: Train loss 1.1788, LR 2.450e-06, grad_norm 2.361,         skipped 1
Iter 60: Train loss 1.0549, LR 2.950e-06, grad_norm 1.511,         skipped 1
```

Two things worth noting. The loss was **finite (1.2309) while the norm was NaN**, so the overflow is in the backward pass and is invisible to loss-watching; by the time the loss prints `nan` the weights are already lost. And the norm escalated `4.2 → 5044 → 1.78e8 → nan` with clipping active the whole time, so clipping alone would not have saved this run. One skip in 60 iterations, and the lowest loss of the run came after it.

**Per-iteration cost is below measurement noise.** 5 interleaved pairs of 40 iterations on a synthetic model: median 32.09 ms without, 33.78 ms with, stdev 15.33 and 10.19. I am not claiming a number from that; the difference is inside the noise floor. The added work is one `square().sum()` tree-reduce plus a `where` per parameter, all inside the existing compiled graph.

Environment: mlx 0.32.0, M2 Pro 32 GB, macOS 26.5.2.

### Risks

- The main technical risk was whether `clip_grad_norm` survives `mx.compile`, since it introduces a data-dependent value. Verified it does, including across a step with a 1000x gradient spike, and the real run above is 60+ compiled iterations with no tracing error.
- Behavior change when the flag is unset: none, per the equivalence check above.
- The zeroed step is not a complete no-op with a momentum optimizer, as described above. This is the one behavior a user could find surprising, hence documenting it rather than hiding it.
- `--grad-clip <= 0` raises `ValueError` rather than silently doing nothing.

### Tests

```
python -m unittest tests.test_tuner_grad_clip tests.test_tuner_trainer tests.test_finetune
Ran 24 tests in 0.285s -- OK
```

The new tests drive the real `train()` rather than a reimplementation, and include a control asserting that **without** the flag the weights do go non-finite, so the guard is tested against the actual current behavior. Note that a bad batch has to be marked in the batch data rather than by a Python-side counter, since `mx.compile` traces the loss once and would otherwise bake the overflow into every later iteration.

### Open questions

1. Do you want the non-finite guard in this change, or clipping only? Clipping alone would not have saved the run above, which is why both are here, but the guard is the less standard half and I would rather ask.
2. One flag for both, or a separate flag for the guard?

Happy to split, rename, or drop either half.

### Pre-merge

Nothing required beyond review. No new dependencies, no migration, no config changes for existing users.

### Post-merge

If you want, `--grad-clip 1.0` could be mentioned in the LoRA docs as a first thing to try when a fine-tune reports `nan`, since #49 and #361 both drew several reports of that symptom.

### Not in this PR

Two adjacent things I found while diagnosing this and deliberately left out:

- `iterate_batches` gates its seeding on `if seed:`, so an explicit `seed=0` is silently ignored, and `0` is the default in `CONFIG_DEFAULTS`. Filed as #1660. Note batch order for a normal CLI run is still reproducible, since `run()` seeds global numpy before training; this only bites when calling the API directly with `seed=0`, or when something else draws from numpy in between.
- `CacheDataset.itemlen` does `len(self._data[idx])`, and all three dataset classes return the raw record from `__getitem__`, so for `{"text": ...}` it returns 1 for every row and the length sort in `iterate_batches` never sorts. That is a throughput and memory issue, not a correctness one. I measured 31.6% extra padded compute on one 8786-row dataset and will open it separately.

### DIFF
--- a/mlx_lm/LORA.md
+++ b/mlx_lm/LORA.md
@@ -362,6 +362,29 @@ If you are unsure of the format to use, the `chat` or `completions` are good to
 start with. For custom requirements on the format of the dataset, use the
 `text` format to assemble the content yourself.
 
+## NaN Loss
+
+By default there is no guard on the size of the gradient, so a single batch
+that overflows can put the weights in a state the run never recovers from and
+every later iteration reports `nan`.
+
+Use `--grad-clip <N>` to bound the global gradient norm to `N` (`1.0` is a
+common choice). This does two things:
+
+1. Gradients whose norm exceeds `N` are scaled down to `N`.
+2. If the norm comes out `inf` or `nan`, the gradient for that step is zeroed
+   so the non-finite values never reach the weights. Note that a momentum
+   optimizer such as Adam still moves the weights on that step using its
+   existing state; this bounds the damage of a bad batch rather than making the
+   step a complete no-op.
+
+Both are off unless you pass the flag, so the default behavior is unchanged.
+Steps with a non-finite norm are reported as a warning alongside the loss.
+
+If a run still diverges with clipping on, the usual next levers are a lower
+learning rate, adding warmup via `lr_schedule`, and lowering the LoRA `scale`
+in `lora_parameters`.
+
 ## Memory Issues
 
 Fine-tuning a large model with LoRA requires a machine with a decent amount
--- a/mlx_lm/examples/lora_config.yaml
+++ b/mlx_lm/examples/lora_config.yaml
@@ -50,6 +50,10 @@ steps_per_eval: 200
 # Number of micro-steps to accumulate before each optimizer update.
 grad_accumulation_steps: 1
 
+# Clip the global gradient norm to this value and skip the update when the norm
+# is not finite. Null disables both.
+grad_clip: null
+
 # Load path to resume training with the given adapter weights.
 resume_adapter_file: null
 
--- a/mlx_lm/lora.py
+++ b/mlx_lm/lora.py
@@ -71,6 +71,7 @@ CONFIG_DEFAULTS = {
     "config": None,
     "grad_checkpoint": False,
     "grad_accumulation_steps": 1,
+    "grad_clip": None,
     "clear_cache_threshold": 0,
     "lr_schedule": None,
     "lora_parameters": {"rank": 8, "dropout": 0.0, "scale": 20.0},
@@ -150,6 +151,14 @@ def build_parser():
         "--grad-accumulation-steps",
         type=int,
         help="Number of steps to accumulate before each optimizer update.",
+    )
+    parser.add_argument(
+        "--grad-clip",
+        type=float,
+        help=(
+            "Clip the global gradient norm to this value and skip the update "
+            "when the norm is not finite. Off by default."
+        ),
     )
     parser.add_argument(
         "--resume-adapter-file",
@@ -279,6 +288,7 @@ def train_model(
         max_seq_length=args.max_seq_length,
         grad_checkpoint=args.grad_checkpoint,
         grad_accumulation_steps=args.grad_accumulation_steps,
+        grad_clip=args.grad_clip,
     )
 
     # Initialize the selected optimizer
--- a/mlx_lm/tuner/trainer.py
+++ b/mlx_lm/tuner/trainer.py
@@ -5,11 +5,13 @@ import time
 from dataclasses import dataclass, field
 from functools import partial
 from pathlib import Path
+from typing import Optional
 
 import mlx.core as mx
 import mlx.nn as nn
 import numpy as np
 from mlx.nn.utils import average_gradients
+from mlx.optimizers import clip_grad_norm
 from mlx.utils import tree_flatten, tree_map
 
 from ..cli_ui import TrainUI, rprint
@@ -73,6 +75,15 @@ class TrainingArgs:
         default=1,
         metadata={
             "help": "Number of steps to accumulate gradients before applying an optimizer update."
+        },
+    )
+    grad_clip: Optional[float] = field(
+        default=None,
+        metadata={
+            "help": (
+                "Clip the global gradient norm to this value. Also skips the "
+                "optimizer update when the norm is not finite. Disabled by default."
+            )
         },
     )
     clear_cache_threshold: int = field(
@@ -241,6 +252,9 @@ def train(
     if grad_accum_steps < 1:
         raise ValueError("grad_accumulation_steps must be at least 1")
 
+    if args.grad_clip is not None and args.grad_clip <= 0:
+        raise ValueError("grad_clip must be positive")
+
     state = [model.state, optimizer.state, mx.random.state]
 
     @partial(mx.compile, inputs=state, outputs=state)
@@ -250,14 +264,29 @@ def train(
         if prev_grad is not None:
             grad = tree_map(lambda x, y: x + y, grad, prev_grad)
 
+        grad_norm = None
         if do_update:
             grad = average_gradients(grad)
             if grad_accum_steps > 1:
                 grad = tree_map(lambda x: x / grad_accum_steps, grad)
+            if args.grad_clip is not None:
+                # Clip the accumulated and averaged gradient, not each
+                # micro-batch, so accumulation stays equivalent to a large batch.
+                grad, grad_norm = clip_grad_norm(grad, args.grad_clip)
+                # A batch that overflows makes the norm inf or nan, and
+                # clip_grad_norm propagates that to every entry, so clipping
+                # alone would still poison the weights. Zero the gradient for
+                # that step instead. Note this is not a full no-op: a momentum
+                # optimizer still moves the weights by its existing state. It
+                # does keep the non-finite values out of them.
+                # mx.where keeps the check in the compiled graph; a Python
+                # branch on grad_norm would sync the device every step.
+                keep = mx.isfinite(grad_norm)
+                grad = tree_map(lambda g: mx.where(keep, g, 0), grad)
             optimizer.update(model, grad)
             grad = None
 
-        return lvalue, toks, grad
+        return lvalue, toks, grad, grad_norm
 
     model.train()
     losses = 0
@@ -266,6 +295,8 @@ def train(
     trained_tokens = 0
     train_time = 0
     grad_accum = None
+    max_grad_norm = mx.array(0.0)
+    n_nonfinite = mx.array(0, mx.int32)
 
     ui = TrainUI(args.iters, rank=rank)
     with ui:
@@ -318,7 +349,7 @@ def train(
 
                 tic = time.perf_counter()
 
-            lvalue, toks, grad_accum = step(
+            lvalue, toks, grad_accum, grad_norm = step(
                 batch,
                 grad_accum,
                 it % grad_accum_steps == 0,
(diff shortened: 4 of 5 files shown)


axiom-of-choice:
Additional diagnostics, because I wanted to rule out the obvious competing explanation before leaning on "a safety net is the right fix".

The batch that overflowed was 4x2048 with all four sequences at ~2030 tokens, i.e. about as long as this dataset gets. So "long sequences overflow" was the natural alternative. It does not hold up. Re-running forward and backward on those exact rows with weights at init and no optimizer updates, batch size 1:

```
=== the exact rows that exploded, one at a time ===
row 647  (len 2030)  loss=1.2015  norm=4.0593  |g|max=0.9331  ok
row 1493 (len 2030)  loss=2.1618  norm=6.4187  |g|max=1.6782  ok
row 1554 (len 2031)  loss=1.4870  norm=5.3354  |g|max=1.5249  ok
row 6634 (len 2031)  loss=0.8357  norm=2.8471  |g|max=0.7074  ok

=== the 8 longest rows in the dataset (2049 tokens, longer than the culprits) ===
norms 2.7646 - 6.0173, all finite

=== control: the 4 shortest rows (420-480 tokens) ===
norms 6.4126 - 9.8167, all finite
```

Two conclusions. The culprit rows are perfectly healthy in isolation: the row that produced a NaN mid-run gives norm 4.06 at init. And length does not correlate in the direction the hypothesis needs, in fact it inverts, the **shortest** rows produce the *highest* gradient norms here.

Not a context-window effect either: `Qwen3-1.7B` has `max_position_embeddings: 40960` with `sliding_window: null` and `rope_scaling: null`, so a 2048-token sequence uses 5% of the supported window.

So the overflow requires the accumulated weight state of a real training run. It's an interaction between the adapted LoRA weights and a batch, not a property of any batch or of any sequence length. No single row reproduces it in isolation.

That is the case for a guard rather than for finding and removing the offending data: there is nothing wrong with the data to remove. The run that skipped the batch kept training and the loss kept falling.

I mention this mainly because #361 was closed after moving toward data format, and #49 after an mlx upgrade. This failure mode is neither, and without something in the trainer there is nothing a user can do about it beyond lowering the learning rate and hoping, which as noted in #1658 did not work here either (NaN at lr 4.5e-7).

axiom-of-choice:
Note: this and #1665 both touch `TrainingArgs` and the `train()` loop in `mlx_lm/tuner/trainer.py`. They are independent and branch off `main` separately, so whichever lands second will need a rebase. Happy to do that as soon as either merges, in whichever order suits you.

zcbenz:

PR #1661, opened 2026-08-05
Honor seed=0 in iterate_batches

### LINKED ISSUE #1660: iterate_batches ignores seed=0 due to `if seed:`, and 0 is the default seed
### Summary

`iterate_batches` gates its seeding on `if seed:`, so `seed=0` is silently ignored. `0` is mlx-lm's own default (`CONFIG_DEFAULTS["seed"] = 0`), which makes the one value most users run with the one value that does not work.

`mlx_lm/tuner/trainer.py`:

```python
if seed:
    np.random.seed(seed)
```

Measured, passing `seed=0` explicitly and polluting global numpy state in between:

```
seed=42 isolated from global state: True
seed=0  honored                   : False   <-- `if seed:` treats 0 as unset
```

With `seed=42` the batch sequence is identical regardless of surrounding numpy use. With `seed=0` it is not, because the branch never runs and batch order falls back to whatever global state numpy happens to be in.

`if seed is not None:` would fix it.

### Scope, and what is *not* broken

I want to be precise here, because I initially overstated this to myself and it's worth not repeating.

For a normal `mlx_lm.lora` CLI run, batch order **is** reproducible today. `run()` in `lora.py` calls `np.random.seed(args.seed)` at line 348, before training, and `train()` then calls `iterate_batches` without a `seed` argument, so the permutations draw from that seeded global state. I checked every `np.random` consumer in the package and nothing in the CLI training path draws from numpy between the seed call and the training loop. So the common case works, by accident of global state rather than by the parameter that exists for it.

The falsy-zero bug bites when:

- calling `train()` / `iterate_batches` directly from Python and passing `seed=0`, which reasonably looks like it should work, and is the documented parameter for exactly this;
- anything else consumes numpy randomness between `np.random.seed` and training, at which point relying on global state is fragile. Verified: injecting three `np.random.rand()` calls in between changes the batch sequence.

So this is a small, contained bug rather than a claim that training is non-reproducible. Two suggestions, and I'm happy to send either:

1. Change `if seed:` to `if seed is not None:`. Strictly a bug fix, no behavior change for anyone not passing `0`.
2. Optionally, have `train()` pass `args.seed` through to `iterate_batches` so batch order depends on an explicit seed rather than on ambient global numpy state. That is a behavior change (batch order would differ from today for the same seed), so it should be its own decision, not bundled with the first.

### Why I noticed

While diagnosing #1658, a gradient overflow on a specific batch. Being unable to rely on replaying an exact batch sequence makes that class of failure harder to isolate, which is what sent me looking at the seeding path. Not offered as an argument for anything, just context for how it surfaced.

Environment: mlx-lm at e5baded, mlx 0.32.0, M2 Pro, macOS 26.5.2.

axiom-of-choice:
Closes #1660.

### Summary

`iterate_batches` gated its seeding on `if seed:`, so an explicit `seed=0` was silently ignored and batch order fell back to whatever state global numpy happened to be in. `0` is the default in `mlx_lm.lora`'s `CONFIG_DEFAULTS`, which made the most common value the one that did not work.

```diff
-    if seed:
+    if seed is not None:
         np.random.seed(seed)
```

### Evaluation

Measured before the change, passing the seed explicitly and drawing from numpy in between to see whether the seed actually isolates the sequence:

```
seed=42 isolated from global state: True
seed=0  honored                   : False   <-- `if seed:` treats 0 as unset
```

After the change both are `True`, and different seeds still produce different orders.

The new test asserts exactly that, for `seed=0` and `seed=42`. It fails on `main`:

```
-  [25, 25, 25, 25, 25, 25, 25, 25]]
+  [33, 33, 33, 33, 33, 33, 33, 33]]
FAILED (failures=1)
```

and passes with the fix. Full suite:

```
python -m unittest tests.test_tuner_trainer tests.test_finetune
Ran 17 tests in 0.204s -- OK
```

### Scope

Deliberately just the bug fix. Worth being clear about what this does *not* change, since #1660 could be read as a larger claim:

Batch order for a normal `mlx_lm.lora` run is already reproducible today. `run()` in `lora.py` calls `np.random.seed(args.seed)` before training, `train()` calls `iterate_batches` without a `seed` argument, and nothing in the CLI training path draws from numpy in between, so the permutations come from that seeded global state. This fix does not alter that path at all.

What it fixes is the case where `seed` is passed explicitly and happens to be `0`, which is what you hit calling `train()` or `iterate_batches` directly from Python, and which is the documented parameter for this.

A second, larger question is whether `train()` should pass `args.seed` through to `iterate_batches` so batch order depends on an explicit seed rather than ambient global state. That would change batch order for a given seed relative to today, so I left it out. Happy to send it separately if you want it.

### Risks

Low. The only behavior change is that `seed=0` now seeds instead of being ignored. Anyone previously passing `0` was getting unseeded behavior by accident; they now get deterministic ordering, which is what the argument promises. No change for `seed=None` or any non-zero seed, and no change to the CLI path.

### Pre-merge

Nothing. One-line change, no new dependencies, no config or docs updates needed.

### Post-merge

Nothing required. If you'd like the `train()` passthrough discussed above, I can open it as a follow-up.

Environment: mlx-lm at e5baded, mlx 0.32.0, M2 Pro, macOS 26.5.2.

### DIFF
--- a/mlx_lm/tuner/trainer.py
+++ b/mlx_lm/tuner/trainer.py
@@ -135,7 +135,7 @@ def iterate_batches(
         idx[i + offset : i + offset + batch_size : step]
         for i in range(0, len(idx) - batch_size + 1, batch_size)
     ]
-    if seed:
+    if seed is not None:
         np.random.seed(seed)
     while True:
         indices = np.random.permutation(len(batch_idx))
--- a/tests/test_tuner_trainer.py
+++ b/tests/test_tuner_trainer.py
@@ -3,6 +3,7 @@
 import unittest
 
 import mlx.core as mx
+import numpy as np
 
 from mlx_lm.tuner.trainer import iterate_batches
 
@@ -48,6 +49,26 @@ class TestTunerTrainer(unittest.TestCase):
         run(1, 4, 8)
         run(2, 4, 8)
         run(3, 4, 8)
+
+    def test_iterate_batches_seed(self):
+        # One distinct token id per row so the batch order is observable.
+        data = [([i + 1] * 8, 0) for i in range(64)]
+
+        def order(seed, consume=0):
+            np.random.seed(1)
+            if consume:
+                # Stand-in for anything else drawing from numpy in between.
+                np.random.rand(consume)
+            batches = iterate_batches(data, 4, 8, loop=True, seed=seed)
+            return [b[0].tolist()[0] for b, _ in zip(batches, range(5))]
+
+        # seed=0 must be honored like any other seed. It is also the default
+        # in mlx_lm.lora's CONFIG_DEFAULTS, so `if seed:` silently dropped it.
+        for seed in (0, 42):
+            with self.subTest(seed=seed):
+                self.assertEqual(order(seed), order(seed, consume=3))
+
+        self.assertNotEqual(order(0), order(42))
 
 
 if __name__ == "__main__":


michalk8 (approved):
LGTM, thanks!

PR #1664, opened 2026-08-05
Fix evaluate crash on tasks with no stop sequences

choipilkyu:
### Problem

`_rstrip_until` in `mlx_lm/evaluate.py` calls `min()` on the list of stop-sequence positions. When a task specifies no stop sequences the list is empty and it raises:

```
ValueError: min() iterable argument is empty
```

lm-eval's `ifeval` does exactly that — `until: []` in `ifeval.yaml`, because the completion is meant to run to `max_gen_toks` (1280) rather than stop at a marker. Every multilingual variant (`ifeval_ca`, `ifeval_es`, …) sets it too. So:

```
mlx_lm.evaluate --model <any> --task ifeval
```

fails outright instead of producing a score.

### Fix

With no stop sequences there is nothing to truncate at, so the completion is returned whole. `min(f, default=l)` does that; the non-empty case is unchanged.

### Tests

Added `TestRstripUntil` to `tests/test_evaluate.py` covering truncation at the first stop sequence, the earliest of several, no match, the empty-`untils` case, and empty-`untils` on an empty string.

```
$ python -m pytest tests/test_evaluate.py -q
......                                                                   [100%]
6 passed
```

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -35,7 +35,10 @@ def _rstrip_until(s, untils):
     l = len(s)
     f = [s.find(u) for u in untils]
     f = [l if x < 0 else x for x in f]
-    return s[: min(f)]
+    # A task may legitimately specify no stop sequences -- ifeval sets
+    # `until: []` because the completion is meant to run to max_gen_toks --
+    # and there is then nothing to truncate at.
+    return s[: min(f, default=l)]
 
 
 def _lstrip(s, pattern):
--- a/tests/test_evaluate.py
+++ b/tests/test_evaluate.py
@@ -5,7 +5,7 @@ from unittest.mock import MagicMock, patch
 
 import mlx.core as mx
 
-from mlx_lm.evaluate import MLXLM
+from mlx_lm.evaluate import MLXLM, _rstrip_until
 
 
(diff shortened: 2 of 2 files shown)


choipilkyu:
Reproduction against `main` (e5baded), using the request lm-eval actually builds for `ifeval` — no model needed, since the failure is in post-processing:

```python
from lm_eval.tasks import TaskManager, get_task_dict
from mlx_lm.evaluate import _rstrip_until

task = get_task_dict(["ifeval"], TaskManager())["ifeval"]
task.build_all_requests(limit=1, rank=0, world_size=1)
ctx, opt = task.instances[0].args
print(opt)
_rstrip_until("some completion", opt["until"])
```

```
{'until': [], 'do_sample': False, 'temperature': 0.0, 'max_gen_toks': 1280}
  File "mlx_lm/evaluate.py", line 38, in _rstrip_until
    return s[: min(f)]
               ~~~^^^
ValueError: min() iterable argument is empty
```

`generate_until` passes `opt["until"]` straight through at `evaluate.py:360` with no normalisation, so the empty list from the task config reaches `_rstrip_until` unchanged.

lm-eval 0.4.12, `lm_eval/tasks/ifeval/ifeval.yaml`.

zcbenz:

PR #1665, opened 2026-08-05
Save the best-validation-loss adapter, and add optional early stopping

axiom-of-choice:
@awni this picks up #331, which you reviewed in July 2025 with one inline suggestion and "please address then we can merge". That comment was never addressed and the PR has been idle since; #509 later tried to finish it and was closed without review. Credit for the original approach goes to @hansvdam, and I'm happy to close this if they'd rather come back and finish it themselves.

Opening a fresh PR rather than pushing to that branch because it's a year stale, and because the diagnosis below led to a different implementation than either earlier attempt.

### The underlying problem is broader than early stopping

Validation loss is computed, reported, and then discarded. Every save is `it % steps_per_save`, so validation loss never influences which checkpoint survives. `adapters.safetensors` therefore holds the weights training *ended* on, not the best ones. Even a user who wants all their iterations wants the best checkpoint kept.

So this PR does two things, and the first stands on its own:

1. Track the lowest validation loss and write those weights to `best_adapters.safetensors`. `adapters.safetensors` still always holds the final weights, so nothing changes for existing workflows. `--no-save-best` disables it.
2. `--patience N` stops after N validations without improvement; `--min-delta` sets how much the loss must drop to count. Early stopping is off unless `--patience` is passed.

### Why not the approach in #331

#331 inferred the best weights from save timing, skipping the final save so "the last checkpoint is the best model". Two problems, both of which are why this PR captures the best weights in memory at the moment validation improves instead:

1. **The last periodic save isn't necessarily the best-validated iteration.** `steps_per_eval` defaults to 200 and `steps_per_save` to 100, and the two are independent. Any `steps_per_eval` that isn't a multiple of `steps_per_save` can leave `adapters.safetensors` holding weights from an iteration that was never validated.
2. **`NameError` on short runs.** `adapter_weights` is only bound inside `if it % args.steps_per_save == 0`. With `--iters 50` and the default `--save-every 100` no periodic save happens, so the final save references an unbound name.

Capturing on improvement also means the final save can stay unconditional, which addresses your inline point on #331 about making sure the last model is still saved when early stopping never triggers. Here that's automatic: the two files are independent, so no configuration loses the final weights.

There are regression tests for both issues above (`test_short_run_without_periodic_save`, `test_best_tracked_when_eval_and_save_cadences_differ`).

### Evaluation

Validation loss from a real LoRA run (Qwen3-1.7B, 8786 rows, batch 4, lr 5e-6 cosine + warmup):

| Iter | Val loss | Δ |
|---|---|---|
| 0 | 1.4537 | (base) |
| 200 | 0.7359 | −0.7178 |
| 400 | 0.7085 | −0.0274 |
| 600 | 0.6958 | −0.0127 |

96% of the improvement lands in the first 200 of 3000 iterations; iterations 600-3000 are ~6.5 hours on this machine for maybe 0.02-0.03 more.

To be straight about what this does and doesn't show: that run flattened, but I stopped it before validation loss actually rose, so it demonstrates wasted compute rather than a measurably worse final artifact. The degradation case is covered in tests (`test_best_differs_from_final_when_val_loss_rises`, driving validation loss down and then up and asserting the two files differ) but I don't yet have it from a long real run. I'd rather say so than overclaim.

```
python -m unittest tests.test_tuner_early_stopping tests.test_tuner_trainer tests.test_finetune
Ran 28 tests in 0.937s -- OK
```

### Design decisions worth your call

- **`save_best` defaults to on.** There's no reason to prefer a worse checkpoint, and it's additive: a new file appears, nothing existing changes. Happy to default it off if you'd rather every new file be opt-in.
- **Separate file rather than overwriting `adapters.safetensors`.** Safer, no surprises, but users must know to pick it up. The alternative is friendlier and a behavior change; your call.
- **Patience counts validations, not iterations**, so its meaning depends on `--steps-per-eval`. Documented in LORA.md.
- **Distributed:** only rank 0 writes, as before. The stop decision uses the validation loss, which `evaluate` already averages across ranks via `all_sum`, so every rank breaks on the same iteration. I've reasoned this through but have not tested on multiple machines.

### Risks

- `--patience` without a validation set raises `ValueError` rather than silently never stopping. `save_best` without one is a harmless no-op.
- One extra ~20 MB write per improvement, which in practice is a handful of writes per run.
- No behavior change unless `--patience` is passed, apart from the additional `best_adapters.safetensors` file.

### Pre-merge

Nothing beyond review. No new dependencies. If you take this, #331 and #509 can be closed as superseded.

### Post-merge

Worth mentioning `best_adapters.safetensors` in release notes, since users currently assume `adapters.safetensors` is the model to ship.

### DIFF
--- a/mlx_lm/LORA.md
+++ b/mlx_lm/LORA.md
@@ -22,6 +22,7 @@ LoRA (QLoRA).[^qlora] LoRA fine-tuning works with the following model families:
   - [Generate](#Generate)
 - [Fuse](#Fuse)
 - [Data](#Data)
+- [Choosing the Best Checkpoint](#Choosing-the-Best-Checkpoint)
 - [Memory Issues](#Memory-Issues)
 
 ## Run
@@ -361,6 +362,40 @@ How can I assistant you today.<|im_end|>
 If you are unsure of the format to use, the `chat` or `completions` are good to
 start with. For custom requirements on the format of the dataset, use the
 `text` format to assemble the content yourself.
+
+## Choosing the Best Checkpoint
+
+Training reports validation loss periodically but the weights saved in
+`adapters.safetensors` are the ones training *ended* on, which are not
+necessarily the best. If validation loss bottoms out partway through a long run,
+the final adapter is worse than one you already had.
+
+By default the adapter with the lowest validation loss is also written to
+`best_adapters.safetensors` alongside it, so you can pick whichever you want:
+
+```shell
+mlx_lm.generate --adapter-path adapters/best_adapters.safetensors ...
+```
+
+`adapters.safetensors` still always holds the final weights, so nothing changes
+for existing workflows. Pass `--no-save-best` to skip the extra file.
+
+To stop a run once it stops improving, use `--patience N`: training ends after
+`N` consecutive validations with no improvement. Since patience counts
+validations rather than iterations, its meaning depends on `--steps-per-eval`.
+`--min-delta` sets how much validation loss must drop to count as an
+improvement, which is useful when the loss is noisy or nearly flat:
+
+```shell
+mlx_lm.lora \
+    --model <path_to_model> \
+    --train \
+    --data <path_to_data> \
+    --patience 3 \
+    --min-delta 0.01
+```
+
+Early stopping is off unless `--patience` is given.
 
 ## Memory Issues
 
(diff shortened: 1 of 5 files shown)


axiom-of-choice:
Note: this and #1659 both touch `TrainingArgs` and the `train()` loop in `mlx_lm/tuner/trainer.py`. They are independent and branch off `main` separately, so whichever lands second will need a rebase. Happy to do that as soon as either merges, in whichever order suits you.

zcbenz:

PR #1668, opened 2026-08-07
Validate per-prompt max_tokens in batch_generate

kupilikula:
## Summary

- reject `max_tokens` lists whose length differs from the prompt count
- perform validation before constructing `BatchGenerator`
- document and test the invalid-input behavior

## Motivation

`batch_generate` accepts one token limit per prompt, but if the supplied list
has a different length, `BatchGenerator.insert_segments` uses `zip`, which can
silently drop trailing prompts or ignore extra limits. Returning fewer
generations than requested without an error is difficult for callers to
diagnose.

Scalar limits and correctly sized lists retain their existing behavior.

## Test plan

- `python -m unittest tests.test_generate.TestBatchGenerateValidation -v`
- `black --check mlx_lm/generate.py tests/test_generate.py`
- `git diff --check`

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1992,7 +1992,7 @@ def batch_generate(
           won't be updated in-place.
        verbose (bool): If ``True``, print tokens and timing information.
           Default: ``False``.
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1676, opened 2026-08-08
Fuse MoE gate/up projections into one gather_qmm at load time

dudududukim:
## Proposed changes

`SwitchGLU` runs `gate_proj` and `up_proj` as two `gather_qmm` calls on the
same input. This adds `SwitchGLU.fuse_gate_up()`, which concatenates the two
weights along the output axis into a single projection, and enables it by
default in `load()` (skipped when `adapter_path` is given, opt-out with
`fuse=False`).

Decode on Apple silicon is kernel-launch bound (~10 µs per dispatch measured
on M5 Pro), so removing one dependent kernel per layer per step pays directly:

- Qwen3-30B-A3B-4bit, M5 Pro 24 GB, ctx 512 decode: **84.8 → 89.2 tok/s (+5.2%)**
- prefill: +3.0% at 512 tokens, neutral at 2048 (compute-bound there)
- prefill logits bit-exact (maxdiff 0.0), identical memory (17.25 GB)

Applies to every MoE model built on `SwitchGLU`. Some HF checkpoints
(qwen3.5-moe, qwen3-vl-moe) already ship fused `gate_up_proj` weights that
`sanitize()` splits apart — with this change the fused execution is recovered
at load time without touching any checkpoint format.

The fuse pass in `load()` collects the `SwitchGLU` modules before fusing so
the replaced gate/up buffers are freed as it goes; fusing while holding the
full `named_modules()` list transiently retains every original `up_proj`
(~5 GB on a 48-layer model).

Verified: new `test_switch_glu_fuse_gate_up` covers fp/quantized, bias, and
both the sorted and unsorted dispatch paths; `tests/test_models.py` passes
(`test_ssm` fails identically on clean main on this machine, unrelated).

## Checklist

- [x] I have run `pre-commit run --all-files` to format my code

### DIFF
--- a/mlx_lm/models/switch_layers.py
+++ b/mlx_lm/models/switch_layers.py
@@ -178,7 +178,7 @@ class SwitchGLU(nn.Module):
             return
         gate, up = self.gate_proj, self.up_proj
         for k in ("weight", "scales", "biases", "bias"):
-            if getattr(gate, k, None) is not None:
+            if gate.get(k) is not None:
                 setattr(gate, k, mx.concatenate([gate[k], up[k]], axis=1))
         del self.gate_proj, self.up_proj
         self.gate_up_proj = gate
--- a/mlx_lm/models/switch_layers.py
+++ b/mlx_lm/models/switch_layers.py
@@ -173,6 +173,17 @@ class SwitchGLU(nn.Module):
         self.down_proj = SwitchLinear(hidden_dims, input_dims, num_experts, bias=bias)
         self.activation = activation
 
+    def fuse_gate_up(self):
+        if "gate_up_proj" in self:
+            return
+        gate, up = self.gate_proj, self.up_proj
+        for k in ("weight", "scales", "biases", "bias"):
+            if getattr(gate, k, None) is not None:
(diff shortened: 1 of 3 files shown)


dudududukim:
I need to walk back the speedup claim in the description.

Re-running the same A/B more carefully across sessions (M5 Pro, AC power,
single process, fixed context via cache-offset rewind), the Qwen3-30B-A3B
delta flips sign between +4.7% and -2.1% from run to run, and other MoE
models move within the same ±2-4% band. The fused path stays bit-exact;
the speedup just doesn't reproduce.

The reason is that the premise was wrong. Measuring directly: a chain of 8
dependent 2048x2048 4-bit qmv runs at the same speed as 8 independent ones
(dispatch boundaries cost ~0 at real sizes), and the gate/up matmuls are
independent so they already overlap on the GPU. The expert path streams at
~207 GB/s with or without the merge — there is no launch overhead for this
fusion to remove.

So in practice this change is performance-neutral, and I don't think it
carries its weight on speed grounds. Unless there's interest in the smaller
dispatch count or in matching checkpoints that ship a fused `gate_up_proj`,
I'll close it. Sorry for the noise.

PR #1687, opened 2026-08-09
Require authentication for remote server binding

### LINKED ISSUE #1695: Remote server binding exposes model selection without authentication
### Problem

When an operator explicitly binds the server to a non-loopback interface, model, draft-model, and adapter selection endpoints remain unauthenticated while wildcard CORS is enabled.

The default loopback binding is not a public exposure; this concerns the explicit remote-binding configuration.

### Expected behavior

Remote binding should require an authentication token before startup, protect the relevant endpoints, and avoid wildcard CORS for authenticated deployments.

### Proposed fix

Draft PR #1687 adds remote-binding authentication and focused authorization tests while preserving the loopback default.

Thump604:
## What changed

- Require an API key when binding to a non-loopback address.
- Authenticate GET and POST endpoints with a constant-time bearer-token comparison.
- Support `--api-key` and `MLX_LM_API_KEY`.

## Root cause

Explicit remote binding exposed dynamic model, draft, and adapter selection without authentication while wildcard CORS was enabled by default.

## Validation

- Regressions verify unauthenticated requests return 401 and remote binding without a key is rejected.

Fixes #1695

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -1,8 +1,11 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import argparse
+import hmac
+import ipaddress
 import json
 import logging
+import os
 import pickle
 import platform
 import socket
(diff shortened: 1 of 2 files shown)


zcbenz:
At the moment we don't consider adding authentication, put the server behind a reverse proxy when necessary.

PR #1696, opened 2026-08-09
fix(server): fail closed on unregistered model IDs

SashaMIT:
## Summary
- `ModelProvider.load` treated unknown request `model` / `draft_model` / `adapters` values as Hub or local paths (`dict.get(key, key)`), so a client could force arbitrary `snapshot_download` / path loads against a running `mlx_lm.server`.
- Fail closed: only accept registered aliases from the CLI maps, then resolve them.
- Resolve the CLI `--adapter-path` against the request model id (fixes the silent drop that open #1249 also targets; this PR additionally rejects free-form adapter paths).

## Test plan
- [x] `pytest tests/test_server.py::TestModelProviderFailClosed` (4 passed)
- Tip parent: `254d153`

Commit is SSH-signed.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -367,9 +367,34 @@ class ModelProvider:
             self.load("default_model", None, "default_model")
 
     def load(self, model_path, adapter_path=None, draft_model_path=None):
-        model_path = self._model_map.get(model_path, model_path)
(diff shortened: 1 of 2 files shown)


zcbenz:

Issue #1378, opened 2026-06-08
Missing support for `laguna` model architecture ("Model type laguna not supported")


dougdaulton:
**mlx-lm version:** main at commit `39c4019f` (also reproducible at earlier pinned commit `e476a222`)
**Platform:** macOS 15.x, Apple Silicon (M4)
**Python:** 3.11

## What I observed

Loading a Poolside Laguna model (e.g. `mlx-community/Laguna-XS.2-4bit`) fails with:

```
Model type laguna not supported.
```

The model's `config.json` has:
- `model_type: "laguna"`
- `architectures: ["LagunaForCausalLM"]`

## Expected behavior

Laguna should load and serve generations like any other supported decoder-only causal LM. Poolside has published several Laguna checkpoints on the Hub (including `mlx-community` MLX conversions), so users hitting this error today have no path forward without a downstream patch.

## Reproducer

```python
from mlx_lm import load
model, tokenizer = load("mlx-community/Laguna-XS.2-4bit")
# ValueError: Model type laguna not supported.
```

## Verification (source inspection)

`mlx_lm/utils.py` at main (`39c4019f`):

```python
# Lines 45-57
MODEL_REMAPPING = {
    "mistral": "llama",
    "llava": "mistral3",
    "phi-msft": "phixtral",
    "falcon_mamba": "mamba",
    ...
    # no "laguna" entry
}

# Lines 185-192 (_get_classes)
model_type = config["model_type"]                              # "laguna"
model_type = MODEL_REMAPPING.get(model_type, model_type)       # still "laguna"
try:
    arch = importlib.import_module(f"mlx_lm.models.{model_type}")
except ImportError:
    msg = f"Model type {model_type} not supported."
    raise ValueError(msg)
```

There is no `mlx_lm/models/laguna.py` on main, and no `MODEL_REMAPPING` entry that aliases `laguna` to a compatible existing implementation.

## Related open PRs

Two PRs are already proposing laguna support — flagging here in case this issue helps either one land:

- #1223 — Add Laguna XS.2 (Blaizzy) — full `mlx_lm/models/laguna.py` + tool parser + tests; open since April, awaiting review
- #1334 — Support for Poolside LagunaXS open source coding model in nvfp4 (tzachicohen) — nvfp4 quantization variant; changes requested

Filing this as a tracking issue since neither PR is linked from a Hub-discoverable issue yet, and users hitting the error string have no obvious landing page. Happy to help test once either PR is ready to merge.

[...]

PR #1704, opened 2026-08-10
laguna: load the public checkpoint layouts (original poolside + repacks)

pierre427:
The `laguna` module from #1334 loads checkpoints whose keys already match its module tree, but the public Laguna checkpoints on the Hub use two other layouts, and all of them currently fail `load_weights(strict=True)`:

**Original poolside layout** (`poolside/Laguna-S-2.1-bf16`): experts are stored individually rather than stacked for `SwitchGLU`, and the router correction bias lives under `experts.`:

```
ValueError: Received 36143 parameters not in model:
model.layers.1.mlp.experts.0.down_proj.weight, ...
```

**Public repacks** (`AtomicChat/Laguna-XS-2.1-MLX-8bit`, `mlx-community/Laguna-XS-2.1-bf16`): every tensor is wrapped in a VLM-style `language_model.` prefix, and the router is stored as a Linear (`mlp.gate.proj.weight`) — quantized to `weight`/`scales`/`biases` in the 8-bit repack:

```
ValueError: Received 1634 parameters not in model:
language_model.lm_head.biases, language_model.lm_head.scales, ...
```

### Changes

- `Model.sanitize()`: strips the `language_model.` prefix, stacks individually-stored experts into the `SwitchGLU` layout, and remaps `experts.e_score_correction_bias` → `gate.e_score_correction_bias`.
- `MoEGate` stores the router as an `nn.Linear` (`gate.proj`) instead of a bare matrix, so quantized routers (the 8-bit repack ships one) load directly — a bare matrix can't hold a `weight`/`scales`/`biases` triple, and `nn.quantize` can't target it either, so `mlx_lm.convert -q` on a Laguna checkpoint currently leaves the router unquantizable. The bare `gate.weight` layout (the original checkpoint, and any conversion made with the current module) is mapped onto `gate.proj.weight` by `sanitize()`, so existing conversions keep loading. Router math is unchanged.

### Verification

- Strict lazy load succeeds on all three public checkpoints above (previously: the errors quoted).
- Full CPU materialization + greedy continuation on `AtomicChat/Laguna-XS-2.1-MLX-8bit` (33B, quantized router): `def quicksort(arr):` → `"\n    if len(arr) <="`.
- New `test_laguna_sanitize` exercises the three layouts (native / prefixed / original per-expert) on a tiny config with strict loading and a forward pass; `test_all_models` still passes.

### DIFF
--- a/mlx_lm/models/laguna.py
+++ b/mlx_lm/models/laguna.py
@@ -168,11 +168,15 @@ class MoEGate(nn.Module):
         self.top_k = args.num_experts_per_tok
         self.num_experts = args.num_experts
         self.softcap = args.moe_router_logit_softcapping
-        self.weight = mx.zeros((args.num_experts, args.hidden_size))
+        # A Linear rather than a bare matrix so quantized checkpoints (which
+        # ship the router as a quantized layer, e.g. the public 8-bit repacks)
+        # can load it directly; sanitize() maps the bare `gate.weight` layout
+        # onto `gate.proj.weight`.
+        self.proj = nn.Linear(args.hidden_size, args.num_experts, bias=False)
         self.e_score_correction_bias = mx.zeros((args.num_experts,))
 
     def __call__(self, x):
-        logits = (x @ self.weight.T).astype(mx.float32)
+        logits = self.proj(x).astype(mx.float32)
         if self.softcap > 0.0:
             logits = mx.tanh(logits / self.softcap) * self.softcap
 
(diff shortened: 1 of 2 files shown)


zcbenz:

michalk8 (approved):
Thanks a lot, I've tested this and it works.

michalk8:
closes #1378 
supersedes #1223

PR #1707, opened 2026-08-11
pythonic tool parser: parse every call in a multi-call block

pierre427:
The pythonic format's block is a list of calls, e.g. `[f1(a=1), f2(b="x")]`, but `parse_tool_call` used a single non-greedy search over the whole block. For multi-call payloads it returned one call named `f1` whose argument string was `a=1), f2(b="x` — the arguments of all calls merged together and calls 2..n silently dropped. Every other family parser in `tool_parsers/` drains multi-call payloads; pythonic was the only one that could not.

This matches each call individually with a quote-aware pattern (a `)` inside a quoted argument does not terminate the call). Single-call blocks still return a bare dict, so existing behavior and tests are unchanged; multi-call blocks return a list, which the server's tool-call formatter already flattens.

Found while replaying multi-token commits from speculative decoding through the streaming parsers: a single delta carrying several complete calls is exactly the shape that triggered this.

Tests: multi-call, quoted-paren argument, and no-call regression cases added to `tests/test_tool_parsing.py`; existing parser cases pass unchanged (6 passed, 22 subtests).

### DIFF
--- a/mlx_lm/tool_parsers/pythonic.py
+++ b/mlx_lm/tool_parsers/pythonic.py
@@ -13,36 +13,48 @@ Parses assistant responses containing tool calls in formats like:
 """
 
 
-_tool_call_regex = re.compile(r"\[(\w+)\((.*?)\)\]", re.DOTALL)
+# The block is a LIST of calls: [f1(a=1), f2(b="x")]. Match each call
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1708, opened 2026-08-11
stream_generate: return cleanly when no tokens are generated

pierre427:
With `max_tokens=0` (or any token generator that yields nothing) the final `GenerationResponse` in `stream_generate` reads `token` / `logprobs` / `n` before the loop has ever bound them, raising `UnboundLocalError`.

Return after finalizing the detokenizer instead, so callers see an empty stream.

Test: `max_tokens=0` yields no responses (added next to the existing `max_tokens` test).


zcbenz:

PR #1709, opened 2026-08-11
Add opt-in acceptance rules for speculative decoding (accept_rule)

pierre427:
## Motivation

`speculative_generate_step` accepts a draft token only when it is bit-identical
to the token the target model sampled at that position. This is correct, but at
temperature > 0 it rejects many drafts that classic speculative sampling would
accept: two independent samples from nearly identical distributions rarely
collide (for `p == q` the exact-match acceptance rate is `sum(p^2)`, while
rejection sampling accepts with probability 1). The result is that speculative
decoding loses most of its speedup as soon as sampling is turned on.

This PR adds an opt-in `accept_rule` parameter with three settings:

- **`"exact"`** (default): today's behavior, unchanged — the default path is
  structurally identical to the current code, including RNG consumption.
- **`"residual"`**: standard speculative rejection sampling (Leviathan et al.,
  [arXiv 2211.17192](https://arxiv.org/abs/2211.17192); Chen et al.,
  [arXiv 2302.01318](https://arxiv.org/abs/2302.01318)) — accept a draft
  `x ~ q` with probability `min(1, p(x)/q(x))`; on rejection, resample from
  the normalized residual `relu(p - q)`.
- **`"block"`**: block verification (Sun et al.,
  [arXiv 2403.10444](https://arxiv.org/abs/2403.10444)) — verify the drafted
  block through cumulative likelihood ratios of its prefixes; a later position
  can rescue an earlier failure, so it accepts at least as many tokens in
  expectation as `"residual"` (their Theorem 1) with the same output
  distribution.

All three rules commit tokens distributed exactly as the target model's
sampling distribution; `"residual"` and `"block"` only raise the acceptance
rate. The acceptance ratio is computed in log space
(`exp(min(log p - log q, 0))`) rather than with a linear-space floor on `q`,
so tokens with representable but tiny probabilities are accepted at the
correct rate.

The parameter threads through `stream_generate`/`generate` (existing kwargs
plumbing) and the CLI (`--accept-rule`).

## Supported sampler configurations

The new rules need the distribution the tokens were *actually sampled from*.
`make_sampler` now records the temperature on the sampler it returns when no
filtering is active, and `speculative_generate_step` uses it to recover the
sampled distribution (`softmax(logprobs / temp)`) for both models:

- Supported: plain temperature samplers from `make_sampler` (any `temp`,
  including 0), with or without `logits_processors` (processors apply
  identically to both models' distributions, so the math is unaffected).
- Not supported: samplers with top-p / min-p / top-k / xtc filtering, or
  custom sampler callables — the sampled distribution cannot be recovered
  from an opaque sampler, so these raise a `ValueError` rather than silently
  biasing the output.
- Greedy (`temp == 0`, or no sampler): every rule coincides with exact
  matching, so the code takes the unchanged exact path.

## Validation

- **Default no-regression**: with `accept_rule` unset the code path is
  unchanged. Verified bit-for-bit against `main` on seeded tiny-model streams
  (temperatures 0 / 0.7 / 1.0, multiple seeds, with and without
  `logits_processors`): identical tokens, draft/target labels, and logprobs.
  `tests/test_accept_rules.py` additionally pins default ≡ `"exact"` and
  greedy-identical-across-rules.
- **Kernel exactness** (constructed p/q, no model): acceptance probabilities
  match `min(1, p/q)` exactly, including sub-floor probabilities (`p = 1e-34`,
  `q = 1e-35` accepts with probability 1; the reverse accepts at 0.1); for
  `p == q` the residual rule accepts everything while exact-match sits at
  `sum(p^2)`.
- **Committed-token distribution, verified by simulation**: seeded
  simulations check that the residual rule's committed marginal matches the
  target pmf (total variation < 0.05 against `p`, > 0.4 against `q`) with
  acceptance rate `sum(min(p, q))`; block verification at k=2 preserves the
  target marginal while accepting measurably more tokens than the per-token
  rule on the same drafts, and at k=1 reproduces the per-token rule's
  acceptance rate and marginal.
- **End-to-end distribution sanity**: with tiny random models at temp 0.8,
  per-position committed marginals of all three rules match plain
  non-speculative sampling under a seeded binned chi-square; on this fixture
  the residual rule accepts far more drafts per block than exact match, and
  block accepts at least as many as residual. (Both tiny-model distributions
  are near uniform, so this test is a consistency check; the kernel tests
  carry the exactness burden.)

All trials are explicitly seeded, so the statistical bounds are deterministic
guards against math regressions, not flaky sampling assertions.

`tests/test_generate.py` + `tests/test_accept_rules.py`: 38 tests, all pass
(26 baseline + 12 new; baseline unchanged).

## Caveats

- The default is unchanged; the new rules are opt-in.
- `"residual"` and `"block"` consume RNG differently from `"exact"` (uniform
  draws for acceptance tests and categorical draws for residual corrections),
  so seeded streams differ across rules at temperature > 0 — as expected,
  since the rules only guarantee the same distribution, not the same sample
  path.
- `"block"` does k extra scalar syncs per verification block for the
  cumulative thresholds; with typical `num_draft_tokens` (2–8) this is noise
  next to the model forward passes.
- The per-block verification cost of the new rules is otherwise the same as
  exact matching (no extra model evaluations; the draft's logprobs, already
  computed, are kept instead of discarded).

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1,10 +1,11 @@
-# Copyright © 2023-2024 Apple Inc.
+# Copyright © 2023-2026 Apple Inc.
 
 import argparse
 import contextlib
 import copy
 import functools
 import json
+import math
 import sys
 import time
 from collections import deque
@@ -209,6 +210,16 @@ def setup_arg_parser():
         type=int,
         help="Number of tokens to draft when using speculative decoding.",
         default=3,
+    )
+    parser.add_argument(
+        "--accept-rule",
+        type=str,
+        choices=["exact", "residual", "block"],
+        help="Draft-token acceptance rule for speculative decoding. "
+        "'residual' and 'block' raise the acceptance rate at temperature > 0 "
+        "without changing the sampled distribution; they require plain "
+        "temperature sampling.",
+        default="exact",
     )
     return parser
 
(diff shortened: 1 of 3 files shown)


zcbenz:

PR #1710, opened 2026-08-11
Add Muse Glimmer (Meta) text model support

nicedreamzapp:
## Requirements

- AI usage disclosure: I'm the idea and the brain on this one. I see the bottleneck and direct Claude to write the code that fixes it, then I oversee it and test everything once it's done, the things the AI can't see. Claude wrote the code under that direction, and it also helped me fill in this template from my answers. The testing on my M5 is mine.

## Model information

- Name: Muse Glimmer 30B (Meta), `muse_glimmer` architecture, released 2026-08-10 under Apache-2.0. This PR adds the text tower; the vision encoder is dropped in `sanitize()`.
- Homepage: https://huggingface.co/meta-models/Muse-Glimmer-30B
- Reference implementation: https://github.com/huggingface/transformers/tree/main/src/transformers/models/muse_glimmer

## Supported checkpoints

- https://huggingface.co/mlx-community/Muse-Glimmer-30B-bf16
- https://huggingface.co/mlx-community/Muse-Glimmer-30B-8bit
- https://huggingface.co/mlx-community/Muse-Glimmer-30B-4bit
- https://huggingface.co/divinetribe/Muse-Glimmer-30B-Abliterated-MM-bf16 (multimodal checkpoint, 815 vision tensors present and removed by `sanitize()`)

Both Meta's original layout (nested `text_config`) and the flat mlx-community config load, with `strict=True`.

## Verification

Run mlx-lm command:

```console
mlx_lm.generate --model mlx-community/Muse-Glimmer-30B-bf16 -p "The secret to baking a good cake is" -m 1024
```

Output (the model is a harmony-channel reasoning model; its analysis channel comes first, the final answer is under `assistant to=user`):

<details>
<summary>analysis channel</summary>

```
to=self<|message|>The secret to baking a good cake is

The secret to baking a good cake is

Probably user wants completion? Might be a prompt. Could be they want secret. Could be incomplete sentence. Might be asking for secret to baking a good cake is...

We can give tips. Possibly they want a witty answer. Could be "The secret to baking a good cake is..." Provide answer.

Probably give list: measuring ingredients accurately, room temperature ingredients, don't overmix, preheat oven, etc.

Maybe they want a specific secret? Could be "love". Or "not opening oven door".

Provide comprehensive answer.

Possibly user wants continuation. Could respond with common secrets.

We can give several secrets.

Maybe ask clarification? Probably just give tips.

We should respond friendly.

Possibly in context of AI? Could be a trick? The secret to baking a good cake is... ?

Maybe they want "The secret to baking a good cake is..." and they want a completion.

Provide answer: measuring correctly, using quality ingredients, proper temperature, etc.

Could give bullet points.

Maybe also mention: don't overmix, sift flour, etc.

Provide concise but thorough.

Possibly user expects a humorous answer: "The secret to baking a good cake is... not eating the batter before it bakes."

We can give both practical and fun.

Probably best to give practical tips.

Use friendly tone.

No disallowed content. Safe.

We can respond in English.

Maybe ask follow up.

Provide final answer.

Let's output.

Potential format:

The secret to baking a good cake is...

1. ...
2. ...

Add final tip: patience.

Ok.

Maybe also mention: accurate measurements, room temp ingredients, preheat oven, etc.

Provide final.

Let's answer.

Possibly user wants a single sentence completion. Could be "The secret to baking a good cake is..." Could be "The secret to baking a good cake is..." Might be a meme.

I'll give comprehensive.

Use bullet points.

Ok.<|eom|>
```

</details>

```
The secret to baking a good cake is really 5 secrets in one:

**1. Measure like a chemist, not like a cook**
Baking is chemistry. Use a kitchen scale for flour and sugar if you can, and spoon flour into the cup then level it off. Too much flour = dry, dense cake.

**2. Start with room temperature ingredients**
Eggs, butter, and milk should be at room temp before you mix. Cold ingredients don't emulsify well, which means a tough crumb and uneven baking.

**3. Don't overmix**
Mix just until combined, especially once you add flour. Overmixing develops gluten and gives you a tough, rubbery cake instead of a tender one.

**4. Respect the oven**
Preheat fully before the cake goes in, and never open the door in the first 2/3 of baking time. An oven thermometer is worth it - most home ovens are off by 25°F.

**5. Time > color**
Cakes are done when a toothpick comes out clean or with a few moist crumbs, not when the top is brown. Every oven is different.

Bonus secret: Use good ingredients and don't rush the cool down. Let it cool in the pan for 10-15 min, then on a rack completely before frosting, or you'll get a soggy mess.

Want a specific cake recipe to try with these rules?

==========
Prompt: 64 tokens, 14.150 tokens-per-sec
Generation: 700 tokens, 10.098 tokens-per-sec
Peak memory: 55.777 GB
```

Run on an M5 Max 128 GB, branch at a8468ab, mlx 0.32.0.

## Reference output (if you have access to CUDA hardware)

No CUDA hardware here. The reference comparison was done on the Mac instead, against Meta's `transformers` implementation (`MuseGlimmerForConditionalGeneration`, bf16, `device_map="mps"`) on held-out prompts:

| Metric | Result |
|---|---|
| Top-1 next-token match | 5/5 |
| Mean top-10 overlap | 90% |
| Mean logit cosine (4-bit vs bf16 reference) | 0.9965 |

Full methodology is in my first comment below. As @neuromechanist pointed out, those metrics are blind to `final_logit_softcapping` and `output_multiplier`; both are now applied after `lm_head` (a8468ab) and the tiny-config unit test confirms logits are bounded at ±20.

## Architecture notes

- Gated attention output: `attn * sigmoid(gate_proj(x))` before `o_proj`
- Scaleless RMSNorm on Q/K, then Q scaled by `qk_scale_factor`
- `[local, local, local, global]` layer pattern: RoPE on the sliding-window layers only, global layers run NoPE
- Scaleless RMSNorm on the token embeddings (not Gemma's sqrt scaling)
- Centered RMSNorm (`norm(x) * (1 + weight)`) on the four per-block norms; standard RMSNorm on the final norm
- `output_multiplier` and `final_logit_softcapping` applied after `lm_head`

Unit coverage in `tests/test_models.py`: a shrunken config exercising construction and a forward pass over the `[local, local, local, global]` pattern, so CI hits the NoPE path on the global layer.

### DIFF
--- a/mlx_lm/models/muse_glimmer.py
+++ b/mlx_lm/models/muse_glimmer.py
@@ -1,0 +1,275 @@
+# Copyright © 2026 Apple Inc.
+# Muse Glimmer (Meta, 2026-08-10) — text-tower support for mlx-lm.
+#
+# Glimmer is multimodal; this implements the language model only, which is what
+# text/agentic inference needs. Architecture verified against Meta's reference
+# transformers implementation (transformers 5.15 modeling_muse_glimmer.py).
+#
+# Glimmer-specific details (vs a standard Gemma/Llama block):
+#   * gated attention output: attn * sigmoid(gate_proj(x)) before o_proj
+#   * scaleless RMSNorm on Q and K, then Q scaled by qk_scale_factor
+#   * [local, local, local, global] layer pattern; RoPE on local layers only,
+#     global layers run NoPE (no positional embedding)
+#   * scaleless RMSNorm applied to the token embeddings
+#   * "centered" RMSNorm (output = norm(x) * (1 + weight)) on the four
+#     per-block norms; standard RMSNorm on the final norm
+from dataclasses import dataclass, field
+from typing import Any, List, Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .cache import KVCache, RotatingKVCache
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str = "muse_glimmer"
+    hidden_size: int = 6656
+    num_hidden_layers: int = 52
+    intermediate_size: int = 19968
+    num_attention_heads: int = 32
+    num_key_value_heads: int = 2
+    head_dim: int = 128
+    vocab_size: int = 202048
+    rms_norm_eps: float = 1e-5
+    post_norm_eps: float = 1e-8
+    sliding_window: int = 2048
+    qk_scale_factor: float = 3.87
+    hidden_activation: str = "silu"
+    max_position_embeddings: int = 131072
+    rope_theta: float = 500000.0
+    rope_parameters: Optional[dict] = None
+    layer_types: Optional[List[str]] = None
+    layer_rope_theta: Optional[List[float]] = None
+    tie_word_embeddings: bool = False
+
+    @classmethod
+    def from_dict(cls, params):
+        # Meta's original checkpoint nests the language-model fields under
+        # "text_config"; the mlx-community conversions flatten them. Accept both.
+        if "text_config" in params:
+            merged = dict(params)
+            merged.update(params["text_config"])
+            params = merged
+        fields = {f.name for f in cls.__dataclass_fields__.values()}
+        return cls(**{k: v for k, v in params.items() if k in fields})
+
+    def __post_init__(self):
+        # rope_theta may live under rope_parameters (as in Meta's config)
+        if self.rope_parameters and "rope_theta" in self.rope_parameters:
+            self.rope_theta = self.rope_parameters["rope_theta"]
+        if self.layer_types is None:
+            # default [local, local, local, global] repeating
+            pat = ["sliding_attention"] * 3 + ["full_attention"]
+            self.layer_types = (pat * (self.num_hidden_layers // 4 + 1))[
+                : self.num_hidden_layers
+            ]
+        if self.layer_rope_theta is None:
+            self.layer_rope_theta = [
+                0 if t == "full_attention" else self.rope_theta
+                for t in self.layer_types
+            ]
+
+
+class CenteredRMSNorm(nn.Module):
+    """RMSNorm with a zero-centered scale: out = norm(x) * (1 + weight)."""
+
+    def __init__(self, dim: int, eps: float):
+        super().__init__()
+        self.weight = mx.zeros((dim,))
+        self.eps = eps
+
+    def __call__(self, x: mx.array) -> mx.array:
+        return mx.fast.rms_norm(x, 1.0 + self.weight, self.eps)
+
+
+class Attention(nn.Module):
+    def __init__(self, args: ModelArgs, layer_idx: int):
+        super().__init__()
+        dim = args.hidden_size
+        self.n_heads = args.num_attention_heads
+        self.n_kv_heads = args.num_key_value_heads
+        self.head_dim = args.head_dim
+        self.scale = self.head_dim**-0.5
+        self.qk_scale_factor = args.qk_scale_factor
+
+        self.layer_type = args.layer_types[layer_idx]
+        self.is_sliding = self.layer_type == "sliding_attention"
+        self.use_rope = bool(args.layer_rope_theta[layer_idx])
+
+        self.q_proj = nn.Linear(dim, self.n_heads * self.head_dim, bias=False)
+        self.k_proj = nn.Linear(dim, self.n_kv_heads * self.head_dim, bias=False)
+        self.v_proj = nn.Linear(dim, self.n_kv_heads * self.head_dim, bias=False)
+        self.o_proj = nn.Linear(self.n_heads * self.head_dim, dim, bias=False)
+        self.gate_proj = nn.Linear(dim, self.n_heads * self.head_dim, bias=False)
+
+        self.eps = args.rms_norm_eps
+        if self.use_rope:
+            self.rope = nn.RoPE(self.head_dim, traditional=False, base=args.rope_theta)
+
+    def __call__(self, x: mx.array, mask=None, cache=None) -> mx.array:
+        B, L, _ = x.shape
+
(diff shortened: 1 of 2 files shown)


nicedreamzapp:
Adding the verification methodology, since the cosine figure in the description could read as lower than expected without it.

**Reference side.** Meta's `transformers` implementation (`MuseGlimmerForConditionalGeneration`, `dtype=torch.bfloat16`, `device_map="mps"`), taking last-token logits for five short prompts. For each prompt I stored the argmax token id, the top-10 ids and values, and a sparse fingerprint of the full logit vector (every 2000th entry).

**MLX side.** `mlx-community/Muse-Glimmer-30B-4bit` loaded through this port, same prompts, same last-token logits.

The two sides run as separate processes — bf16 at 30B doesn't fit alongside the MLX load without swapping.

**Scoring.** Three independent checks per prompt:

- top-1 next-token match against the reference argmax
- top-10 set overlap
- cosine similarity over the sparse logit fingerprint

**On the 0.9965.** That number compares 4-bit MLX against a bf16 reference, so it's bounded by quantization error, not by architecture fidelity — I don't think it should read as unexplained drift. The relevant signal is that top-1 matched 5/5: a wrong architecture here (missing the attention output gate, RoPE on the global layers, or non-centered block norms) garbles predictions outright rather than shifting them slightly, and each of those produced obvious token-level divergence while I was working through the port.

Happy to run a bf16-vs-bf16 comparison if that would be more convincing for review, or to share the two verification scripts.

nicedreamzapp:
Friendly ping on this one.

The port is complete and verified against Meta's `transformers` implementation — methodology in the comment above. CI hasn't run yet; the workflow is still awaiting maintainer approval, since this is my first PR to the repo. If someone can approve the run, I'll fix anything it turns up the same day.

Happy to adjust the approach or split it into smaller pieces if that would make it easier to review.

jbejar:
any update?

nicedreamzapp:
nothing new yet. the port is finished and verified against meta's transformers implementation, but CI still hasn't run because the workflow needs a maintainer to approve it on a first time contributor PR, and nobody has reviewed it yet.

if you want it before it lands you can pull the branch and run it against mlx-community/Muse-Glimmer-30B-4bit, that's what i tested with. i'll post here as soon as anything moves.

neuromechanist:
Nice port — the block structure (gated attention output, scaleless QK-norm with `qk_scale_factor`, NoPE globals, centered per-block norms, rotating cache on the sliding layers) matches what we found bringing this arch up on llama.cpp. Two output-path transforms are missing though, and the verification in the description is structurally unable to catch either of them.

Meta's `text_config` carries two fields this port currently drops in `ModelArgs.from_dict`'s field filter:

```
final_logit_softcapping = 20.0
output_multiplier = 0.19611613513818404   # 1/sqrt(26); llama.cpp maps this to logit_scale
```

The reference applies both after `lm_head` (transformers `modeling_muse_glimmer.py`, `MuseGlimmerForConditionalGeneration.forward`):

```python
logits = self.lm_head(hidden_states[:, slice_indices, :])
logits = logits * self.config.text_config.output_multiplier
logits = logits / self.config.text_config.final_logit_softcapping
logits = torch.tanh(logits)
logits = logits * self.config.text_config.final_logit_softcapping
```

The GGUF metadata on Meta's own conversion agrees (`muse-glimmer.final_logit_softcapping = 20.0`, `muse-glimmer.logit_scale = 0.1961161`), and llama.cpp applies both.

Why the described checks cannot see this: both transforms are strictly monotonic, so argmax and top-k sets are invariant — top-1 5/5 and the top-10 overlap pass with or without them by construction. Cosine similarity is scale-invariant, so the missing 0.196 multiplier is invisible to the 0.9965 figure. And a fingerprint sampled every 2000th vocab entry is almost entirely near-zero logits, where `tanh(x/20)*20 ≈ x`, so the softcap barely moves it either. A check that does catch it: compare `max(abs(logits))` or softmax entropy against the reference — unscaled/uncapped logits run roughly 5x hotter and are unbounded, the reference's are capped at ±20.

Practical impact: greedy decoding is unaffected, but temperature sampling runs ~5x colder than Meta calibrated it, and logprobs, perplexity, and speculative-decoding acceptance are all computed on the wrong scale.

Minimal patch:

```python
@dataclass
class ModelArgs(BaseModelArgs):
    ...
    final_logit_softcapping: float = 20.0
    output_multiplier: float = 0.19611613513818404
```

and in `Model.__call__`:

```python
    def __call__(self, inputs, cache=None):
        out = self.model(inputs, cache=cache)
        if self.tie_word_embeddings:
            out = self.model.embed_tokens.as_linear(out)
        else:
            out = self.lm_head(out)
        out = out * self.args.output_multiplier
        cap = self.args.final_logit_softcapping
        return mx.tanh(out / cap) * cap
```

Happy to help land this: we run a long-context eval harness (NIAH/RULER/MRCR at 8K to 131K) against this model on llama.cpp (M4 Pro, 64 GB) and can validate the patched port across the full context range and post numbers here.

nicedreamzapp:
good catch, thank you — you're right on both counts. the port dropped `final_logit_softcapping` and `output_multiplier`, and my verification was structurally blind to them for exactly the reasons you laid out: argmax and top-k are invariant under monotonic transforms, and cosine is scale-invariant.

fixed in a8468ab. both fields are now in `ModelArgs` (defaults matching Meta's config, config values win when present) and both transforms are applied after `lm_head`, per your patch. verified so far: the tiny-config unit test passes with output logits now bounded at ±20, and config overrides flow through correctly. to be transparent, i have not rerun the full 30B comparison against the transformers reference — my stored fingerprints can't see this change anyway, so a meaningful rerun needs the max|logit| / entropy checks you described.

i'd gladly take you up on the NIAH/RULER/MRCR offer — validation from your harness across the full context range would be much stronger than anything i can produce locally. and thanks for the detail on why each of my checks couldn't catch this; that writeup made the failure mode obvious in a way i won't forget.

zcbenz (commented):

zcbenz (approved):
Thanks!

Issue #1627, opened 2026-07-28
qwen3_coder tool parser rejects float-formatted integer args ("140.0"), silently dropping the call and causing client infinite loops


wmorgan85:
# qwen3_coder tool parser rejects float-formatted integer args ("140.0"), and server silently drops the call — causing infinite client loops

## Summary

Two compounding problems, observed with `mlx_lm.server` 0.31.3 serving a Qwen3-Coder variant to an agentic client (opencode):

1. **`tool_parsers/qwen3_coder.py`**: `_convert_param_value` handles schema-typed integers with a bare `int(param_value)`. Qwen3-Coder sometimes emits float-formatted integers (e.g. `140.0` for a `limit` parameter typed `integer`), and `int("140.0")` raises `ValueError`. The float branch immediately below already coerces tolerantly in both directions; the int branch does not.

2. **`server.py` (`ToolCallFormatter.__call__`)**: when the parser raises, the tool call is dropped with only a server-side log warning, and the response is returned as HTTP 200 with `finish_reason: "tool_calls"` but no tool call. The client gets no error signal at all.

The combination is worse than either bug alone: an agentic client sees a tool-call turn with no tool call, retries, and — with greedy/deterministic sampling over an identical (fully prompt-cached) context — the model regenerates the exact same rejected output. The result is a tight infinite loop. In our case the client looped **~390 times at ~1.3s per iteration** (55 tokens generated, 56k tokens cache-read each time) until manually killed:

```
2026-07-28 10:15:28,688 - WARNING - Failed to parse tool call (ValueError: invalid literal for int() with base 10: '140.0') — tool text was likely truncated mid-generation.
127.0.0.1 - - [28/Jul/2026 10:15:28] "POST /v1/chat/completions HTTP/1.1" 200 -
2026-07-28 10:15:30,026 - WARNING - Failed to parse tool call (ValueError: invalid literal for int() with base 10: '140.0') — tool text was likely truncated mid-generation.
127.0.0.1 - - [28/Jul/2026 10:15:30] "POST /v1/chat/completions HTTP/1.1" 200 -
... (repeats indefinitely)
```

Note the warning text ("tool text was likely truncated mid-generation") is misleading here — the tool text was complete and well-formed XML; only the type coercion failed.

## Reproduction

```python
from mlx_lm.tool_parsers.qwen3_coder import parse_tool_call

tools = [{"type": "function", "function": {"name": "read", "parameters": {
[...]

PR #1712, opened 2026-08-11
Make Qwen3-Coder tool parsing tolerant of multiline code

bnsd55:
## Summary

Qwen3-Coder can emit source-code tool arguments in two non-standard forms:

- literal newlines inside JSON string values
- backtick-delimited multiline strings inside JSON-like arrays

The Qwen parser now normalizes backtick-delimited values before parsing and uses `json.loads(..., strict=False)` for model-generated JSON. Valid JSON behavior is unchanged.

The server formatter also catches `SyntaxError` from malformed tool arguments and reports that it skipped the malformed call, instead of allowing the request handler to crash with an uncaught exception.

## Testing

- `python -m unittest tests.test_tool_parsing` (7 passed)
- exact React-style backtick payload regression check
- direct `ToolCallFormatter` `SyntaxError` check
- Black check on changed Python files
- `git diff --check`

The full local suite has 3 unrelated errors because optional dependencies are not installed locally: `datasets`, `lm_eval`, and `requests`.

Related: #1627

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -78,10 +78,10 @@ class ToolCallFormatter:
         for tool_text in tool_calls:
             try:
                 parsed = self._tool_parser(tool_text, self._tools)
-            except (ValueError, json.JSONDecodeError) as e:
+            except (SyntaxError, ValueError, json.JSONDecodeError) as e:
                 logging.warning(
-                    f"Failed to parse tool call ({type(e).__name__}: {e}) — "
-                    f"tool text was likely truncated mid-generation."
+                    f"Failed to parse tool call ({type(e).__name__}: {e}); "
+                    f"skipping malformed tool call."
                 )
                 continue
             if not isinstance(parsed, list):
--- a/mlx_lm/tool_parsers/qwen3_coder.py
+++ b/mlx_lm/tool_parsers/qwen3_coder.py
@@ -33,6 +33,47 @@ def _get_arguments_config(func_name: str, tools: Optional[Any]) -> dict:
     return {}
 
 
+def _normalize_backtick_strings(param_value: str) -> str:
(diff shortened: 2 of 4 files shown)


FU-max-boop (changes requested):
The new normalization path silently changes a previously supported tool argument when a literal backtick appears inside a single-quoted string.

For example, with an array-typed `tags` parameter:

```python
<function=edit><parameter=tags>['use `code` here']</parameter></function>
```

`main` returns `['use `code` here']`, while this branch returns `['use "code" here']`. I reproduced this at `7037ba1`; the existing seven `tests.test_tool_parsing` tests pass, so they do not catch the regression.

The scanner at `mlx_lm/tool_parsers/qwen3_coder.py:60-69` tracks double-quoted context only, so both backticks inside a single-quoted Python-style string are interpreted as delimiters and replaced with a JSON string. This can silently alter code or prose passed to a tool. Please preserve single-quoted string context (or try the existing `ast.literal_eval` path before applying backtick normalization) and add this case as a regression test.

bnsd55:
@FU-max-boop Fixed in 3ce2ef1.

The parser now tries the existing JSON and `ast.literal_eval` paths before backtick normalization, so literal backticks inside single-quoted strings are preserved. I also added the exact reported case as a regression test.

Verified with `tests.test_tool_parsing`, `TestToolCallFormatter`, Black, and `git diff --check`. Could you please re-review?

FU-max-boop (approved):
Re-reviewed at `3ce2ef1`. The reported single-quoted backtick regression is fixed, the exact regression test is present, and the focused parser plus additional edge-case matrix pass. Thanks for addressing it.

zcbenz:
I'm going to close this in favor of #1239, please open new PRs for cases uncovered.

PR #1717, opened 2026-08-12
feat(generate): expose --repetition-penalty in the CLI (#1669)

### LINKED ISSUE #1669: mlx_lm.generate: repetition_penalty has no default (unlike other sampling params), model can loop into repeated text with no warning
**Description**
Running `mlx_lm.generate` with default sampling parameters, a model looped
into repeated text with no warning. Looking at `generate.py` on `main`,
other sampling parameters have explicit `DEFAULT_*` constants defined at
module level (`DEFAULT_TEMP`, `DEFAULT_TOP_P`, `DEFAULT_MIN_P`,
`DEFAULT_TOP_K`, etc.), but there is no `DEFAULT_REPETITION_PENALTY`.
`SERVER.md` also documents defaults for `repetition_context_size` (20),
`presence_penalty` (0.0), and `frequency_penalty` (0.0), but doesn't state
a default for `repetition_penalty` itself.

**To reproduce**
1. Run `mlx_lm.generate --model <model> --prompt "<prompt>"` with no
   `--repetition-penalty` flag set.
2. With some prompts/models, generation degenerates into repeated text
   or loops.

**Expected behavior**
Either a sensible default repetition penalty applied out of the box, or
at minimum a clear note in the docs/CLI help that repetition_penalty is
off by default and users should set it explicitly for longer generations.

**Question**
Is the lack of a default intentional (e.g. to match upstream sampling
behavior exactly)? If so, a one-line doc/help-text note would still help
first-time users avoid this.

**Environment**
mlx-lm version: [will confirm]
Python version: [will confirm]
OS: macOS [version], chip [M1/M2/M3/...]

Happy to submit a small docs PR to add this note if useful.

Anai-Guo:
### What

`mlx_lm.generate`'s CLI builds a sampler and passes **no** logits processors, so the
repetition penalty — which lives in `make_logits_processors` — is unreachable from the
command line. Meanwhile `mlx_lm.server` already exposes it (`_make_logits_processors`
→ `make_logits_processors(..., repetition_penalty, repetition_context_size, ...)`).
So a `generate` user who hits a repetition loop has no CLI knob to fix it, and `--help`
gives no hint that the capability exists.

This adds `--repetition-penalty` and `--repetition-context-size` to `mlx_lm.generate`
and threads them into `generate()` via `make_logits_processors`, mirroring the server.

### Behavior is unchanged by default

Both flags default to *off* (`repetition_penalty=None`). `make_logits_processors`
already skips any penalty that is `None` or `0`, so with no flags the processor list is
empty and generation is byte-for-byte identical to today. The knob only does something
when a user opts in (e.g. `--repetition-penalty 1.2`).

I intentionally did **not** change the *default* penalty value — the linked issue also
asks whether a non-zero default would be desirable, but that's a behavior change and a
separate decision. This PR is just the missing plumbing so the existing feature is
reachable from `generate`, matching `server`.

### Verification

mlx isn't available off Apple silicon, so I exercised the argument parser directly
(stubbing the mlx imports) against `main` and this branch:

- **`main`**: `--repetition-penalty` / `--repetition-context-size` — not present.
- **this branch**: both present; `--repetition-penalty 1.3 --repetition-context-size 40`
  parses to `1.3 / 40`; with no flags the defaults are `None / 20` → empty processor list.

Scope: `mlx_lm/generate.py` only, +19 −1.

Fixes #1669. Thanks @arjunneema09 for the report and for pinpointing that the gap is the
missing CLI surface rather than a docs nit.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -29,7 +29,7 @@ from .models.cache import (
     TokenBuffer,
     load_prompt_cache,
 )
-from .sample_utils import make_sampler
+from .sample_utils import make_logits_processors, make_sampler
 from .tokenizer_utils import TokenizerWrapper
 from .utils import does_model_support_input_embeddings, load
 
@@ -42,6 +42,8 @@ DEFAULT_TOP_K = 0
 DEFAULT_XTC_PROBABILITY = 0.0
 DEFAULT_XTC_THRESHOLD = 0.0
 DEFAULT_MIN_TOKENS_TO_KEEP = 1
+DEFAULT_REPETITION_PENALTY = None
+DEFAULT_REPETITION_CONTEXT_SIZE = 20
 DEFAULT_SEED = None
 DEFAULT_MODEL = "mlx-community/Llama-3.2-3B-Instruct-4bit"
 DEFAULT_QUANTIZED_KV_START = 5000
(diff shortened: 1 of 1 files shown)


zcbenz:

PR #1718, opened 2026-08-12
Give a helpful error when a model has only .bin weights

xquantize:
`load_model` raised a bare `FileNotFoundError: No safetensors found` when a repo ships only PyTorch .bin/.pt weights, with no hint about the cause or fix.

I hit this converting a real model whose repo only had .bin weights. This detects that case and raises a message naming the .bin/.pt files found and pointing to the safetensors converter, the same conversion the CONTRIBUTING "Adding New Models" section already directs people to.

Kept it message-only rather than auto-loading .bin, to avoid pulling torch into the load path. The genuine "no weights at all" case is unchanged.

Added `test_load_model_bin_weights_error`; `tests/test_utils.py` passes.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -335,6 +335,17 @@ def load_model(
     weight_files = glob.glob(str(model_path / "model*.safetensors"))
 
     if not weight_files and strict:
+        bin_files = glob.glob(str(model_path / "*.bin")) + glob.glob(
+            str(model_path / "*.pt")
+        )
+        if bin_files:
+            raise FileNotFoundError(
+                f"No safetensors found in {model_path}, but PyTorch weight "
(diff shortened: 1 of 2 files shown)


FU-max-boop (approved):
Reviewed and tested at `01460c4`.

- `tests.test_utils.TestUtils.test_load_model_bin_weights_error` passes.
- I separately exercised both `.bin` and `.pt` inputs and confirmed that the diagnostic names the discovered file and explains the safetensors requirement.
- I verified that the genuine no-weight path keeps the previous `No safetensors found in ...` error.
- Black, isort, and `git diff --check` pass for the changed files.

Keeping this message-only instead of adding a PyTorch load path is a good scope boundary. The implementation and regression test look good to me.

zcbenz:

PR #1719, opened 2026-08-13
Fix threadgroup memory race in fused KL and JS Metal kernels (fixes #1700)

### LINKED ISSUE #1700: Fused Metal KL-divergence kernel overflows (−3.39e38) in backward pass during DWQ for models >~1B
## Description

While running DWQ (`mlx_lm.quant.dwq`) on models larger than ~1B parameters on an M-series Mac (M5 Max, 137 GB), the fused Metal KL-divergence kernel's **backward pass** overflows to `-3.39e38` (≈ `-FLT_MAX`) on certain training batches. The initial validation loss computes fine; the **first training step corrupts the parameters**, and training diverges immediately.

## Environment

- mlx-lm 0.31.3, macOS (Apple Silicon M5 Max, 137 GB unified memory)
- Observed during DWQ distillation (8-bit teacher → 4-bit student), e.g. Qwen2.5-1.5B-Instruct; smaller (<~1B) models unaffected on the same setup

## Where

`mlx_lm/tuner/losses.py`:

- `_make_kl_backward_kernel()` — the fused backward kernel
- `_kl_div_loss.vjp` — dispatches it as `grid=(1024, cotangent.size, 1)`, `threadgroup=(1024, 1, 1)`, templated on `("V", logits_q.shape[-1])`
- `can_run_metal()` — returns `mx.default_device() == mx.gpu and mx.metal.is_available()`; `kl_div_loss()` falls back to the `nn.losses.kl_div_loss` reference path when it is false

## Reproduction signals

- Triggered by the default fused path; also triggered independently by `--grad-checkpoint` and by `--batch-size > 1`.
- Forcing the numerically stable reference KL path avoids it entirely:

```python
import mlx_lm.tuner.losses as _losses
_losses.can_run_metal = lambda: False   # use reference KL path (stable)
from mlx_lm.quant.dwq import main
main()
```

With that one-line monkeypatch, DWQ completes normally and the resulting weights pass our perplexity gate (teacher 8.34 < DWQ 9.45 < plain-affine 9.86 on our pilot).

## Possible mechanism (hypothesis — not isolated)

Offered only as a starting point; we have **not** confirmed this and would not want it taken as a diagnosis.

Both KL kernels pad the tail block with a `-1e30` sentinel:

```c
vals_q[j] = (offset + j < V) ? logits_q[offset + j] : -1e30;
```

and later rescale partial sums with `metal::fast::exp(prev_max_q - max_q)`. For a thread whose whole slice falls in the padding region, that argument is ≈ `-1e30`. Metal's `metal::fast::` math functions are only defined over a limited input domain, so if `fast::exp` does not return exactly 0 there, `sum_exp_q` — and therefore `lse_q = max_q + fast::log(sum_exp_q)` — is garbage, and the per-element write

```c
out[offset + j] = static_cast<T>(c * (fast::exp(vals_q[j] - lse_q) - fast::exp(vals_p[j] - lse_p)));
```

saturates. That would also explain why the forward pass looks fine: it reduces to one scalar per row, whereas the backward writes a full `V`-wide gradient, so a corrupted `lse` is far more consequential.

The same padding-and-rescale pattern appears in the JS kernels in this file, if the theory holds up.

## Expected

The fused Metal KL kernel's backward should match the reference implementation numerically (or clamp/guard against the overflow), for models above ~1B params.

## Notes

Happy to provide the exact DWQ invocation and gate script — we've documented the workaround in our internal recipe and can share timings/configs. We can also test candidate patches on the M5 Max setup where this reproduces.

AxelNoun:
Recreates #1714, which was closed automatically on 12 Aug when the fork
hosting its branch was detached during an account cleanup on my side —
my mistake, apologies for the noise.

The head commit is byte-identical to the original PR (`ee4fae4ec8`).
All prior context and discussion: https://github.com/ml-explore/mlx-lm/pull/1714

---

## Problem

#1700 reports that during DWQ on models above roughly 1B parameters, the
backward pass of the fused Metal KL divergence kernel intermittently
saturates to -3.39e38 (about -FLT_MAX) and corrupts training on the first
step. The failure is nondeterministic and load dependent; forcing the
reference KL path avoids it entirely.

## Root cause

The four fused kernels in mlx_lm/tuner/losses.py reduce across simdgroups
through a single threadgroup buffer, first for the row max, then for the
sum of exponentials. After the barrier, every thread reads the maxes back
from `shared`, and lane 0 of each simdgroup then writes its partial sum
into the same slots. There is no barrier between those reads and writes.
Simdgroups make independent progress, so a fast group can overwrite a slot
before a slow group has read it. The affected thread folds a partial sum
into what it believes is a max, the row logsumexp collapses, and
exp(vals - lse) saturates in the gradient write. The forward pass shares
the hazard but reduces to a single scalar per row, which is why the
corruption surfaces in the backward first.

Two related issues in the same kernels are fixed alongside the race.
Several barriers separating accesses to `shared` used
mem_flags::mem_none, which orders execution but does not fence
threadgroup memory. And the tail block feeds a -1e30 sentinel into
metal::fast::exp, outside the documented input range of the fast math
functions; that did not appear to be the trigger here (the forward pass
performs the same out of range call on every row and computes correctly),
but those sites are now guarded as hardening in a separate commit that
can be dropped if preferred.

## Additional finding: deterministic JS softplus overflow

While validating the race fix on the diagnostic suite, a second, fully
deterministic bug showed up in the JS kernels only. They evaluated
log(1 + exp(logq - logp)) directly. At logit scale 20 the log-ratio
reaches several hundred, exp overflows to inf, and the reduction yields
-inf or NaN (including via 0 * inf when a probability underflows). That
matched the residual scale-20 saturation that remained after the barrier
fix alone: same on every retry of identical inputs, independent of
scheduling. The JS softplus sites now use the standard stable form
(mathematically identical for in-range inputs, finite where the naive
form overflowed). This is a separate commit from the race fix and from
the fast-math hardening; unlike the hardening, it intentionally changes
outputs exactly on the overflowing inputs.

## Changes

1. Add a threadgroup_barrier(mem_flags::mem_threadgroup) between the max
   read back and the sum exchange, in all four kernels.
2. Promote the mem_none barriers guarding `shared` to mem_threadgroup.
3. Hardening (separate commit): bounds guard on tail block accumulation,
   zero guard on the rescale factor below the f32 underflow range, and
   clamping of the exp arguments in the gradient writes to their
   mathematical range. Exact no ops for in range inputs.
4. Stable softplus in the JS kernels (separate commit): replaces the
   naive log(1+exp(x)) that overflowed at large log-ratios.
5. Regression tests: large non-block-aligned vocabulary determinism for
   KL and JS vjp, plus a JS scale-20 softplus case that asserts finite
   outputs and parity with the CPU reference path.

## Validation

Measured on a GitHub hosted macos-15 arm64 runner (virtualized Apple
M series, Metal available), mlx 0.32.0, mlx-lm 0.31.3.

Determinism, repeated backward on identical inputs, worst KL
configuration (float32, scale 6, shape (512, 151936)):
  before: 50/50 nondeterministic, 0 saturated
  after:  0/50 nondeterministic, 0 saturated

Pipelined stress (bf16, 8 concurrent backwards of shape (256, 151936)):
  before: 3/400 nondeterministic
  after:  0/400 nondeterministic

Parity with the reference path after the fix (max abs difference on the
gradient): float32 7.4e-6, bfloat16 2.0e-3 across vocabularies of 17,
4095, 4097, 100003 and 151936. Finite cells are small but nonzero, so
the fused Metal path is the one under test rather than a silent fallback
to the reference implementation.

No measurable performance regression on the same runner class for the
(512, 151936) bf16 backward; run to run variance on virtualized hosts
dominates any barrier cost on the once-per-row reduction path.

I do not have access to Apple hardware beyond CI runners, so I could
not run the DWQ recipe end to end. The reporter of #1700 offered to
test candidate patches on the M5 Max where the failure reproduces;
exact commands are posted in the issue. The new regression tests in
tests/test_losses.py will also exercise the fix on the project CI.

### DIFF
--- a/mlx_lm/tuner/losses.py
+++ b/mlx_lm/tuner/losses.py
@@ -98,6 +98,8 @@ def _make_kl_forward_kernel():
         max_q = simd_max(max_q);
         max_p = simd_max(max_p);
 
+        threadgroup_barrier(mem_flags::mem_threadgroup);
+
         // Share the sum_exp across the threadgroup
         sum_exp_q *= metal::fast::exp(prev_max_q - max_q);
         sum_exp_p *= metal::fast::exp(prev_max_p - max_p);
@@ -117,7 +119,7 @@ def _make_kl_forward_kernel():
         lse_q_minus_p = max_q + metal::fast::log(sum_exp_q) - lse_p;
     }
 
-    threadgroup_barrier(mem_flags::mem_none);
+    threadgroup_barrier(mem_flags::mem_threadgroup);
 
     {
         float kl = 0;
@@ -157,7 +159,7 @@ def _make_kl_forward_kernel():
         if (simd_lane_id == 0) {
             shared[simd_group_id] = kl;
         }
-        threadgroup_barrier(mem_flags::mem_none);
+        threadgroup_barrier(mem_flags::mem_threadgroup);
         kl = shared[simd_lane_id];
         kl = simd_sum(kl);
 
@@ -267,6 +269,8 @@ def _make_kl_backward_kernel():
         max_q = simd_max(max_q);
         max_p = simd_max(max_p);
 
+        threadgroup_barrier(mem_flags::mem_threadgroup);
+
         // Share the sum_exp across the threadgroup
         sum_exp_q *= metal::fast::exp(prev_max_q - max_q);
         sum_exp_p *= metal::fast::exp(prev_max_p - max_p);
@@ -286,7 +290,7 @@ def _make_kl_backward_kernel():
         lse_q = max_q + metal::fast::log(sum_exp_q);
     }
 
-    threadgroup_barrier(mem_flags::mem_none);
+    threadgroup_barrier(mem_flags::mem_threadgroup);
 
     {
         float kl = 0;
@@ -477,6 +481,8 @@ def _make_js_forward_kernel():
         max_q = simd_max(max_q);
         max_p = simd_max(max_p);
 
+        threadgroup_barrier(mem_flags::mem_threadgroup);
+
         // Share the sum_exp across the threadgroup
         sum_exp_q *= metal::fast::exp(prev_max_q - max_q);
         sum_exp_p *= metal::fast::exp(prev_max_p - max_p);
@@ -496,7 +502,7 @@ def _make_js_forward_kernel():
         lse_q = max_q + metal::fast::log(sum_exp_q);
     }
 
-    threadgroup_barrier(mem_flags::mem_none);
+    threadgroup_barrier(mem_flags::mem_threadgroup);
 
     {
         float kl_p = 0;
@@ -664,6 +670,8 @@ def _make_js_backward_kernel():
         max_q = simd_max(max_q);
         max_p = simd_max(max_p);
 
+        threadgroup_barrier(mem_flags::mem_threadgroup);
+
         // Share the sum_exp across the threadgroup
         sum_exp_q *= metal::fast::exp(prev_max_q - max_q);
         sum_exp_p *= metal::fast::exp(prev_max_p - max_p);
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1721, opened 2026-08-13
Fix empty shard for oversized first weight

### LINKED ISSUE #1720: make_shards prepends an empty shard when the first tensor exceeds the size limit
## Description

`make_shards()` flushes the current shard whenever the next tensor would exceed `max_file_size_gb`, even when the current shard is still empty. If the first model tensor is larger than the 5 GiB default, the result starts with `{}`:

```python
from types import SimpleNamespace
from mlx_lm.utils import make_shards

large_weight = SimpleNamespace(nbytes=(5 << 30) + 1)
shards = make_shards({"model.embed_tokens.weight": large_weight})

assert shards == [{"model.embed_tokens.weight": large_weight}]
# Current result: [{}, {"model.embed_tokens.weight": large_weight}]
```

## Impact

A single tensor cannot be split by this helper, so an oversized tensor correctly has to live in an oversized shard. The extra empty shard has no useful representation. `save_model()` nevertheless counts and writes it, producing an unreferenced `model-00001-of-00002.safetensors` and placing the only real tensor in part 2. `mx.save_safetensors()` accepts the empty mapping and writes a 41-byte file, so the error is silent: the saved model has a misleading, noncanonical shard set and consumers that enumerate shard files process a shard containing no weights.

This is reachable for large-vocabulary embedding or output matrices that individually exceed the configured shard limit.

## Expected behavior

Only flush a shard that already contains at least one weight. For an oversized first tensor, return one oversized, non-empty shard. Normal multi-tensor shard boundaries should remain unchanged.

## Proposed fix

Guard the flush with `if shard and shard_size + v.nbytes > max_file_size_bytes`, plus a regression for an oversized first tensor.

I searched open and closed issues and pull requests for `make_shards`, empty/oversized shards, and single large tensors. I found no existing report or implementation; #1633 concerns eager evaluation during saving, not shard construction. I am preparing a focused patch.

FU-max-boop:
## Summary

Fixes #1720.

`make_shards()` flushed the current shard whenever the next weight would cross the size limit, even when no weight had been added yet. If the first tensor itself exceeded the limit, the function returned an empty shard followed by the real oversized shard.

`save_model()` then silently wrote that empty mapping as part 1 and placed the only real weight in part 2, producing a misleading, unreferenced shard in the model artifact.

The flush now requires the current shard to be non-empty. Oversized tensors still remain intact in an oversized shard, since this helper cannot split a tensor, and normal shard boundaries are unchanged.

## Validation

- red-to-green regression using an oversized first weight without allocating a multi-GiB tensor
- existing and new `make_shards` tests: 2 passed
- Black and isort repository hooks on both changed files
- Python compile check and `git diff --check`
- verified `mx.save_safetensors()` silently writes an empty mapping (41-byte file), confirming the artifact path
- exact open/closed issue and PR duplicate search; branch synchronized with `main`

AI assistance was used to identify the boundary condition, design the allocation-free regression, and review the diff; every changed line and test result was manually verified.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -653,7 +653,7 @@ def make_shards(weights: dict, max_file_size_gb: int = MAX_FILE_SIZE_GB) -> list
     shards = []
     shard, shard_size = {}, 0
     for k, v in weights.items():
-        if shard_size + v.nbytes > max_file_size_bytes:
+        if shard and shard_size + v.nbytes > max_file_size_bytes:
             shards.append(shard)
             shard, shard_size = {}, 0
         shard[k] = v
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -5,6 +5,7 @@ import os
 import tempfile
 import unittest
 from pathlib import Path
+from types import SimpleNamespace
 
 import mlx.core as mx
 import mlx.nn as nn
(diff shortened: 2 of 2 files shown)


zcbenz:

michalk8 (approved):
LGTM, thanks!

Issue #1689, opened 2026-08-09
Batch generation can silently omit prompts and mishandles max_tokens=0


Thump604:
### Problem

Two generation boundary cases fail unclearly:

1. A per-prompt list such as `max_tokens` that is shorter than the prompt list is consumed with unchecked `zip`, silently omitting trailing prompts.
2. `stream_generate(max_tokens=0)` skips the generation loop and later reads locals that were never initialized, raising `UnboundLocalError`.

### Expected behavior

Per-prompt option lengths should be validated before generation, and a zero-token request should terminate normally without entering token generation.

### Proposed fix

Draft PR #1681 validates per-prompt list lengths and handles the zero-token path explicitly, with focused regressions for both cases.

Issue #1631, opened 2026-07-28
make_prompt_cache(max_kv_size=N) sink tokens: policy is enforced in one place and dropped in another


soobrosa:
Splitting this out of #1353 and #1584 at @PhilipJohnBasile's suggestion.

*Edited 2026-08-01: the original text called `keep=4` unintentional and dated it
to the wrong PR. Corrected below, see
[the correction comment](https://github.com/ml-explore/mlx-lm/issues/1631#issuecomment-5150910092).*

### Summary

`make_prompt_cache(model, max_kv_size=N)` builds sink tokens unconditionally:

```python
# mlx_lm/models/cache.py:37
RotatingKVCache(max_size=max_kv_size, keep=4) for _ in range(num_layers)
```

#1584 implements `RotatingKVCache.to_quantized()` but explicitly refuses
`keep > 0`, since `RotatingQuantizedKVCache` would have to preserve the sink
region across rotation in the packed `(packed, scales, biases)` layout. The
consequence is that `--max-kv-size` and `--kv-bits` together cannot work on the
sequential path, before or after #1584, for any model that does not provide its
own `make_cache`. #1353 refuses to start on exactly that combination, since no
layer would be quantizable.

### `keep=4` was deliberate, and is now the only producer of an unsupported shape

It was introduced in [ml-explore/mlx-examples#931](https://github.com/ml-explore/mlx-examples/pull/931)
(commit `38d78ee`, Aug 2024), which states the intent explicitly: "The first
`keep` tokens (default 4) of the prompt are always kept, everything else gets
overwritten in the circular buffer," citing
[StreamingLLM](https://arxiv.org/abs/2309.17453).
[ml-explore/mlx-examples#1015](https://github.com/ml-explore/mlx-examples/pull/1015)
(commit `0c8e878`, Oct 2024) then moved that `keep=4` from the `generate_step`
call site into `make_prompt_cache`. Provenance traced by @PhilipJohnBasile,
correcting my original claim that it was undocumented and dated to #1015.

What has changed since is the surrounding code, not the intent. `cache.py:37` is
now the only `keep > 0` in the repository. Every model that constructs rotating
caches itself passes `keep=0` explicitly:

- `cohere2.py:211`, `exaone4.py:211`, `exaone_moe.py:379`, `gemma3n.py:561`,
  `gemma4_text.py:685`

and the remaining sites take the class default of `keep=0`. @mabaeyens counted
[...]

PR #1722, opened 2026-08-13
Feature/refactor generate

michalk8:
This PR refactors some choices in the `generate.py` file and fixes (in some part) the issues below:

- #1689: can be closed.
insert_segments validates option lengths up front and commits only after every sequence is built; stream_generate rejects max_tokens=0 rather than crashing. Raising is the intended contract, not the empty stream the issue suggests.
- #1681: batch half addressed, stream half rejected.
Ours names the offending option and is atomic. Its max_tokens < 0 guard would close off the working unbounded range (verified at -1 and -7).
- #1708: overridden by design.
Same crash fixed, opposite contract: max_tokens=0 raises, because it already means prefill-only in generate_step and an empty stream would silently skip the prefill.
- #1668: addressed.
On more surface too — BatchGenerator.insert, not just batch_generate.
- #1657 / issue #1472: Closed.
_normalize_logits_processors turns None into () at insert, so it can't reach the per-token loop.
- #1645: addressed.
Guarded tps properties in generate_utils.py; stats()'s finally only accumulates, so it can't mask the body's exception.
- #1631: partially addressed.
keep > 0 guard moved into _merge_caches; the unconditional keep=4 blocking --kv-bits stands.

Also fixes some issues in the KV cache's left padding (syncing too often); now does ` self.left_padding.tolist()[idx]` instead of computing this using `mx`).


mlx_lm/models/cache.py
```diff
@@ -1023,8 +1023,8 @@ def filter(self, batch_indices):
         self.offset = self.offset[batch_indices]
         self.left_padding = self.left_padding[batch_indices]
 
-        # Shift left to reduce padding
-        min_left_pad = self.left_padding.min().item()
+        mx.eval(self.left_padding)
+        min_left_pad = min(self.left_padding.tolist())
```
zcbenz, line 1027: What this change does is moving computation of min from gpu to cpu?

zcbenz (commented):
Is it possible to split the PR into smaller ones?

michalk8:
> Is it possible to split the PR into smaller ones?

Yes, splitting this!

mlx_lm/models/cache.py
michalk8: I explain it here #1824

michalk8:
Closing this in favor of the linked PRs.

PR #1727, opened 2026-08-14
Validate adapter tensors in load_adapters

### LINKED ISSUE #1725: load_adapters silently accepts unknown or shape-incompatible adapter tensors
## Reproduction

On current `main` (`254d153`, mlx-lm 0.31.3), both invalid adapters below load without an error:

```python
import json
import tempfile
from pathlib import Path

import mlx.core as mx
import mlx.nn as nn

from mlx_lm.tuner.utils import load_adapters


class TinyLayer(nn.Module):
    def __init__(self):
        super().__init__()
        self.proj = nn.Linear(4, 3, bias=False)


class TinyModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.layers = [TinyLayer()]


def make_adapter(weights):
    path = Path(tempfile.mkdtemp())
    config = {
        "fine_tune_type": "lora",
        "num_layers": 1,
        "lora_parameters": {
            "keys": ["proj"],
            "rank": 2,
            "scale": 1.0,
            "dropout": 0.0,
        },
    }
    (path / "adapter_config.json").write_text(json.dumps(config))
    mx.save_safetensors(str(path / "adapters.safetensors"), weights)
    return path


cases = {
    "unknown tensor name": {
        "layers.0.proj.lora_a": mx.zeros((4, 2)),
        "layers.0.proj.lora_b": mx.zeros((2, 3)),
        "layers.0.proj.lora_typo": mx.zeros((1,)),
    },
    "wrong tensor shape": {
        "layers.0.proj.lora_a": mx.zeros((5, 2)),
        "layers.0.proj.lora_b": mx.zeros((2, 3)),
    },
}

for label, weights in cases.items():
    load_adapters(TinyModel(), make_adapter(weights))
    print(f"{label}: loaded")
```

Observed output:

```text
unknown tensor name: loaded
wrong tensor shape: loaded
```

`load_adapters()` currently calls `load_weights(..., strict=False)`, which ignores unknown names and does not check tensor shapes.

## Impact

A typo or an adapter built for an incompatible target can appear to load successfully. The adapter may then be partially applied, produce incorrect results, or fail only later during inference or fusion.

## Expected behavior

After adapter modules are instantiated, every supplied adapter tensor should:

- name a parameter present in the model; and
- have the same shape as that parameter.

Otherwise, `load_adapters()` should raise a clear `ValueError`.

## Compatibility

Using `strict=True` directly is not suitable because an adapter intentionally contains only a subset of model parameters. Missing base-model parameters should remain allowed; validation only needs to cover the tensors supplied by the adapter before retaining the non-strict subset load.

I have a focused implementation and regression tests ready, but wanted to confirm that this fail-fast behavior is desired before opening a PR.

xquantize:
Fixes #1725.

load_adapters loaded adapters with load_weights(strict=False), which silently ignores adapter tensors whose names match no model parameter and tensors whose shapes are incompatible. A typo'd key or an adapter built for a different target then loads without error and produces silently wrong results, failing only later at inference or fusion.

This validates every supplied adapter tensor against the model's parameters after the LoRA layers are built: each tensor must name a real parameter and match its shape, otherwise a clear ValueError is raised listing all mismatches. An adapter is still allowed to provide only a subset of parameters, so the non-strict subset load is preserved.

Adds regression tests in tests/test_tuner_utils.py covering the unknown-name and wrong-shape cases from the issue, plus a valid adapter that still loads.

Thanks to @FU-max-boop for the precise diagnosis and reproduction.

### DIFF
--- a/mlx_lm/tuner/utils.py
+++ b/mlx_lm/tuner/utils.py
@@ -4,6 +4,7 @@ import types
 from pathlib import Path
 from typing import Dict
 
+import mlx.core as mx
 import mlx.nn as nn
 import mlx.optimizers as opt
 from mlx.utils import tree_flatten, tree_unflatten
@@ -134,7 +135,23 @@ def load_adapters(model: nn.Module, adapter_path: str) -> nn.Module:
             config.lora_parameters,
             use_dora=(fine_tune_type == "dora"),
         )
-    model.load_weights(str(adapter_path / "adapters.safetensors"), strict=False)
+    weights = mx.load(str(adapter_path / "adapters.safetensors"))
+    params = dict(tree_flatten(model.parameters()))
+    errors = []
+    for name, w in weights.items():
+        if name not in params:
+            errors.append(f"  {name}: not a parameter in the model")
+        elif w.shape != params[name].shape:
+            errors.append(
+                f"  {name}: adapter shape {tuple(w.shape)} does not match "
+                f"model parameter shape {tuple(params[name].shape)}"
+            )
+    if errors:
+        raise ValueError(
+            f"Adapter at {adapter_path} contains incompatible tensors:\n"
+            + "\n".join(errors)
+        )
+    model.load_weights(list(weights.items()), strict=False)
     return model
 
 
--- a/mlx_lm/tuner/utils.py
+++ b/mlx_lm/tuner/utils.py
@@ -137,14 +137,23 @@ def load_adapters(model: nn.Module, adapter_path: str) -> nn.Module:
         )
     weights = mx.load(str(adapter_path / "adapters.safetensors"))
     params = dict(tree_flatten(model.parameters()))
+    if fine_tune_type == "full":
+        allowed = params
+    else:
+        adapter_leaves = {"lora_a", "lora_b", "m"}
+        allowed = {
+            name: shape
+            for name, shape in params.items()
+            if name.rsplit(".", 1)[-1] in adapter_leaves
+        }
     errors = []
     for name, w in weights.items():
-        if name not in params:
-            errors.append(f"  {name}: not a parameter in the model")
-        elif w.shape != params[name].shape:
+        if name not in allowed:
+            errors.append(f"  {name}: not an adapter parameter in the model")
+        elif w.shape != allowed[name].shape:
             errors.append(
                 f"  {name}: adapter shape {tuple(w.shape)} does not match "
-                f"model parameter shape {tuple(params[name].shape)}"
+                f"model parameter shape {tuple(allowed[name].shape)}"
             )
     if errors:
         raise ValueError(
--- a/tests/test_tuner_utils.py
+++ b/tests/test_tuner_utils.py
@@ -1,14 +1,18 @@
 # Copyright © 2024 Apple Inc.
 
+import json
 import sys
+import tempfile
 import unittest
 from io import StringIO
+from pathlib import Path
 from unittest.mock import MagicMock
 
+import mlx.core as mx
 import mlx.nn as nn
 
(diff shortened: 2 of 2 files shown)


FU-max-boop:
Thanks for picking this up and for the credit. This matches the intended fail-fast behavior while preserving valid subset adapter loading. Happy to help test additional edge cases if useful.

xquantize:
Thanks! 

Since you offered, I think a couple of edge cases worth covering if you have cycles:

- DoRA adapters (fine_tune_type: "dora"), which add a magnitude vector alongside lora_a/lora_b, worth confirming the validation accepts the full DoRA tensor set and doesn't false-positive on the extra params.
- Quantized base models, where the LoRA layers wrap QuantizedLinear, just to check the parameter names/shapes still line up as expected.
- A valid adapter that's a genuine subset (fewer layers than the model), to confirm we didn't accidentally tighten that into an error.

If any of those surface something, happy to fold it in. Otherwise this is ready for review whenever a maintainer has a moment.

FU-max-boop:
Thanks — I tested these against the current PR head (`4f3649b2`) on macOS/Metal with Python 3.12.13 and MLX 0.32.0. I generated the files through the training path's actual save shape (`tree_flatten(model.trainable_parameters())` -> `mx.save_safetensors` -> `load_adapters`).

Requested cases:

- dense DoRA: a complete `lora_a` / `lora_b` / `m` set loaded and round-tripped exactly;
- quantized base: LoRA and DoRA both loaded against a real 4-bit `nn.QuantizedLinear` with the logical adapter shapes, including `m` for DoRA;
- genuine subset: a one-of-three-layer adapter loaded successfully; I also confirmed that converting all three layers while supplying only one layer's complete A/B pair remains accepted.

The focused tuner suites also pass: `22 passed, 2 subtests passed`. I found no false positive in those three requested cases.

I did find one adjacent, pre-existing name-scope hole that this validation does not close. The allowed-name map currently comes from all `model.parameters()`, so a LoRA file can include a shape-compatible base parameter such as `layers.0.proj.linear.weight`; it passes validation and overwrites the base weight. I reproduced this by loading an all-`42.0` `(3, 4)` tensor and verifying that the wrapped base weight became exactly `42.0`.

There is also a completeness boundary: because validation only walks supplied tensors, a DoRA target missing `m` (or a LoRA target missing `lora_b`) still passes and retains the runtime-initialized value. That is compatible with allowing module-level subsets, but completeness could be enforced per included target without requiring every model layer.

If you want to cover these here or in a follow-up, my suggested contract would be:

- for `lora` / `dora`, accept only adapter-owned tensor names and require a complete tensor group for each included target;
- continue allowing any subset of targets/layers;
- keep full model parameters valid only for `fine_tune_type: full`.

Happy to turn these reproductions into focused regression tests if that would help.

xquantize:
This is great, thank you!

Eespecially catching the base-weight overwrite. Both look right to me at a glance; let me take a proper look at the name-scoping and get back to you. Yes please on the reproductions as regression tests, the 42.0 overwrite one especially.

xquantize:
Pushed (85eca44). 

Scoped the allowed names to adapter-owned tensors (lora_a/lora_b/m) for lora/dora, with the full parameter set valid only for fine_tune_type: full, so the layers.0.proj.linear.weight overwrite you found is now rejected. 

Added your 42.0 repro as a regression test, plus a valid-DoRA-with-m load. Subset-of-targets loading still works. Left per-target completeness (a target missing m/lora_b) for a follow-up as discussed. 

Credited you as co-author on the commit, thanks again!

zcbenz (approved):
Thanks!

Issue #1662, opened 2026-08-05
Models that discard `update_and_fetch`'s values return leak one Metal buffer per layer per generated token — deterministic crash at Metal's buffer-count limit (deepseek_v4 dies at ~11.5k generated tokens)


aidiffuser:
## Summary

`GenerationBatch` never evaluates cache state during decode (only during
chunked prefill). Any cache field that is not consumed by the logits graph
therefore accumulates an **unevaluated lazy `slice_update` chain** — one node
per step — and each node pins that step's freshly materialized input buffer.
The bytes are tiny; the **Metal buffer-object count** is not. On a 43-layer
model leaking one buffer per layer per token, the process hits Metal's
`resource_limit` (499,000 buffer objects on M3 Ultra) at ~11,470 generated
tokens and dies with:

```
RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
```

This message is easy to misread as a byte-size failure (the same hardware's
max recommended working set is ~499,000 MB — an unfortunate coincidence); it
is a buffer-count failure. All byte metrics (`get_active_memory`,
`get_cache_memory`, `get_peak_memory`) look healthy right up to the crash.

`deepseek_v4` is a shipping first-party instance: its attention uses the
K==V trick (SDPA is fed the same rows for keys and values), so both call
sites discard the values half:

```python
win_keys, _ = win_cache.update_and_fetch(k4, k4)   # values return unused
```

Nothing in the graph ever reads `cache.values`; at S == 1 the rotating
cache's in-place update rebinds it to `slice_update(old_values, new_row)`
each step, growing the dead chain.

## Observed behavior (production, 2× M3 Ultra 512GB, TP-2)

- Deterministic crash at generated ≈ 11,456–11,488, bracketed to ±8 tokens
  with a per-step probe; independent of prompt length (reproduced with 1k,
  10k, and 23k-token prompts — totals 12.5k/21.7k/34.3k, same generated
  count at death).
- 43 layers × 1 buffer/token: 43 × 11,470 + ~5.8k baseline ≈ 499,000.
- Prompt tokens do not leak because chunked prefill evaluates
  `[c.state for c in prompt_cache]` every chunk; only decode steps do.
- Total leaked bytes at death: ~0.5 GB (1 KB rows) — invisible.
- Decode throughput stays flat (the dead chain is never executed).
- Severity beyond the process: dying with ~499k live Metal buffers has
[...]

PR #1731, opened 2026-08-14
Guard batch cache metadata against lazy-graph accumulation

rohnach29:
## Guard batch cache metadata against lazy-graph accumulation (extends #960's guard to BatchKVCache and ArraysCache)

  `BatchKVCache.offset` and `ArraysCache.left_padding`/`lengths` are rewritten on every step and every `filter()`/`extend()`, but nothing in the decode forward graph is guaranteed to read them. Left lazy, each
  rewrite extends an unevaluated graph chain that transitively pins every Metal buffer that ever fed it. #960 added exactly this guard to `BatchRotatingKVCache` (`mx.depends(self.keys, (self.left_padding,
  self.offset))`); this PR extends the same idiom to `BatchKVCache.update_and_fetch` and to `ArraysCache.filter()/extend()` (stapled onto its state arrays — the pattern `qwen3_5.py:308` already uses).

  **Scope** mlx-lm's own `BatchGenerator` loop does *not* exhibit unbounded growth today: its prompt path evaluates full cache state (which includes the metadata), so chains keep getting truncated — we measured a flat, bounded edge count over sustained never-draining load. The hazard is real for **external drivers of the batch-cache API** that evaluate only their own forward outputs: a churn probe (500 `filter`/`extend`/`update_and_fetch` cycles evaluating only forward outputs, counting Metal buffer-handle headroom before/after `mx.eval(metadata)`) pins **12,000 handles on main vs 0 with this change** (~24/cycle), and in production behind vllm-mlx this exact mechanism killed a continuously saturated Qwen3.6-35B server at the 499,000 buffer-handle limit after 19h45m (`[metal::malloc] Resource limit (499000) exceeded.`; downstream mitigation: waybarrios/vllm-mlx#708). The class-level guard removes the hazard for every consumer without each having to know which internal arrays are unsafe to leave lazy — and makes the native path's currently-incidental safety structural.

  **Relationship to existing threads:** #1232 attempted this direction and was closed awaiting a clean repro — the probe above is that repro. #1632 (open) applies the same staple to `ArraysCache.advance()`, the per-step path; this PR deliberately does not touch `advance()` to avoid overlapping it — the two compose, and I'm happy to rebase over it. #1662/#1672 report the same resource-limit signature; this PR does not claim to fix them (the native path self-bounds in our measurements).

  **Cost:** none — slightly faster if anything (synthetic decode loop, B=8: ~284 vs ~312 µs/step; keeping the metadata graph small beats carrying it). The metadata is evaluated as a rider on arrays the forward pass
  already computes; no eager eval is introduced. The `ArraysCache` staple covers every non-None state entry, so a layer that overwrites one entry without reading it cannot silently drop the guard.

  **Tests** (`tests/test_prompt_cache.py`): behavioral graph-bound churn tests for both classes using `mx.export_to_dot` edge counting — a correct guard leaves 0 pending edges, unpatched measures in the hundreds,
  and the test fails if the staple targets anything the forward pass doesn't read — plus an `mx.depends` spy test for the `BatchKVCache` update path (following `test_models.py`'s mock pattern from #1656),
  value-integrity checks, and the metadata-absent path. Three fail on current main with the defect. Full `test_prompt_cache`: 27/27; black + isort clean.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -629,6 +629,23 @@ class ArraysCache(_BaseCache):
     def state(self, v):
         self.cache = v
 
+    def _sync_metadata(self):
+        # Tie left_padding/lengths into the state graph so the next forward
+        # pass evaluates them (mirroring BatchRotatingKVCache's guard on
+        # left_padding/offset). They are rewritten on every filter/extend but
+        # nothing in the decode graph reads them, so left lazy their graphs
+        # accumulate across membership changes and pin buffers for the
+        # lifetime of the batch.
+        deps = tuple(a for a in (self.left_padding, self.lengths) if a is not None)
+        if not deps:
+            return
+        # Staple every entry so the guard holds as long as the model reads
+        # any part of its state before overwriting it; a single-entry staple
+        # would be silently dropped by a blind write to that entry.
+        self.cache = [
+            mx.depends(c, deps) if c is not None else None for c in self.cache
+        ]
+
     def filter(self, batch_indices):
         """
         In-place filter to keep just the given indices in the cache.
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1736, opened 2026-08-15
Fix batched generation for hybrid-cache models (silent corruption + broadcast crash)

benrey:
## Problem

Batched generation is broken for hybrid-architecture models (Gated DeltaNet / SSM layers using `ArraysCache`, e.g. Qwen3.5) in two ways:

1. **Silent output corruption** with mixed-length prompts at batch >= 2: `ArraysCache.make_mask` returns only the left-padding mask whenever `left_padding` is set — and `ArraysCache.merge` always sets `left_padding = [0] * B` — so the `lengths` (right-padding) mask installed by `prepare()` during chunked prefill is silently dropped. Pad tokens then update the recurrent state of any sequence whose final partial chunk is co-scheduled with longer chunks. Attention layers happen to survive (causality excludes trailing pads); a recurrent state does not.

2. **Broadcast crash**: `BatchKVCache.filter` on an *empty* cache with nonzero left padding decrements `_idx` below zero; the next `make_mask` builds a zero-length causal mask (`arange(offset + N)` with a negative offset) which crashes attention with `[broadcast_shapes] Shapes (1,1,1,0) and (1,16,1,N) cannot be broadcast`.

Both were found serving `mlx-community/Qwen3.5-27B-6bit` behind a continuous-batching server: the engine loop died under concurrency, and — worse — sequences that avoided the crash could return corrupted tokens. Related: #980, #1403 (this appears to be the concrete mechanism behind the batch >= 2 corruption reports for GDN models).

## Repro

Staggered-arrival harness driving `BatchGenerator` on a small random-weight `qwen3_5` `TextModel` (greedy sampling), comparing each sequence's batched output against solo generation:

- Before: 3 of 5 staggered mixed-length sequences produce different tokens than solo (both the Metal kernel and the ops path — the kernel is not at fault); specific arrival timings crash with the broadcast error above.
- After: batched == solo for every sequence in every tested configuration (same-time and staggered arrivals, mixed lengths, chunked prefill through `prefill_step_size`).

## Changes

- `ArraysCache.make_mask` combines the left-padding and lengths constraints instead of prioritizing left padding.
- `BatchKVCache.filter` only performs the left-shift compaction when keys exist.
- Regression tests for both in `tests/test_prompt_cache.py` (`TestHybridBatchCache`).

`tests/test_prompt_cache.py` and `tests/test_generate.py` pass.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -689,14 +689,21 @@ class ArraysCache(_BaseCache):
             self.left_padding -= N
 
     def make_mask(self, N: int):
+        # Combine the constraints instead of prioritizing left padding: after ``merge`` the
+        # cache carries ``left_padding = [0] * B`` (an array, not ``None``), and returning the
+        # all-true left-padding mask here silently discarded the ``lengths`` (right-padding)
+        # mask during chunked prefill — pad tokens then updated the recurrent state of every
+        # sequence whose final partial chunk was co-scheduled with longer ones, corrupting
+        # batched generation for hybrid (SSM/gated-delta) models.
+        mask = None
         if self.left_padding is not None:
             pos = mx.arange(N)
-            return pos >= self.left_padding[:, None]
-        elif self.lengths is not None:
+            mask = pos >= self.left_padding[:, None]
+        if self.lengths is not None:
             pos = mx.arange(N)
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1740, opened 2026-08-16
Native Qwen MTP speculative decoding with transactional prompt-cache reuse

PhilipJohnBasile:
## Summary

This builds on @AirRunner's native Qwen MTP reference implementation in #990 and carries it forward with correctness, compatibility, and transactional multi-turn prompt-cache hardening.

Qwen3.5/3.6/3.8-family checkpoints ship with a trained MTP head, but MLX conversion currently does not provide a complete path for preserving and safely using that head with the models' hybrid attention + GatedDeltaNet state. This PR provides that lower-layer MLX-LM implementation.

### What this adds

- preserves and loads vendor-trained Qwen MTP weights during conversion
- supports dense and MoE Qwen3.5/3.6/3.8-family checkpoints
- performs exact GatedDeltaNet recurrent/conv-state rollback when a draft is rejected
- provides native MTP draft/verify generation with probabilistic acceptance
- safely reuses completed native-MTP conversation prefixes across later turns
- fails closed when cache state or an unsupported feature cannot preserve correctness

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -3,10 +3,12 @@
 import argparse
 import contextlib
 import copy
-import functools
+import inspect
 import json
+import math
 import sys
 import time
+import warnings
 from collections import deque
 from dataclasses import dataclass
 from functools import partial
@@ -29,7 +31,7 @@ from .models.cache import (
     TokenBuffer,
     load_prompt_cache,
 )
-from .sample_utils import make_sampler
+from .sample_utils import categorical_sampling, make_sampler, make_sampler_chain
 from .tokenizer_utils import TokenizerWrapper
 from .utils import does_model_support_input_embeddings, load
 
@@ -45,6 +47,7 @@ DEFAULT_MIN_TOKENS_TO_KEEP = 1
 DEFAULT_SEED = None
 DEFAULT_MODEL = "mlx-community/Llama-3.2-3B-Instruct-4bit"
 DEFAULT_QUANTIZED_KV_START = 5000
+_CACHE_CLEAR_INTERVAL = 256
 
 
 def str2bool(string):
@@ -210,6 +213,12 @@ def setup_arg_parser():
         help="Number of tokens to draft when using speculative decoding.",
         default=3,
     )
+    parser.add_argument(
+        "--mtp",
+        action="store_true",
+        help="Use native Multi-Token Prediction for speculative decoding "
+        "(requires a model with an MTP head, e.g. Qwen3.5).",
+    )
     return parser
 
 
@@ -287,12 +296,50 @@ class GenerationResponse:
     finish_reason: Optional[str] = None
 
 
+@dataclass
+class _MTPPromptFinalizePlan:
+    mtp_hidden: Optional[mx.array]
+    mtp_tokens: Optional[mx.array]
+    target_token: Optional[mx.array]
+    last_hidden: Optional[mx.array]
+    num_tokens: int
+
+
 def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_bits):
     if kv_bits is None:
         return
     for e, c in enumerate(prompt_cache):
         if hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
             prompt_cache[e] = c.to_quantized(group_size=kv_group_size, bits=kv_bits)
+
+
+def validate_prompt_and_embeddings(model, prompt, input_embeddings):
+    if input_embeddings is not None:
+        if not does_model_support_input_embeddings(model):
+            raise ValueError("Model does not support input embeddings.")
+        elif len(prompt) > 0 and len(prompt) != len(input_embeddings):
+            raise ValueError(
+                f"When providing input_embeddings, their sequence length ({len(input_embeddings)}) "
+                f"must match the sequence length of the prompt ({len(prompt)}), or the "
+                "prompt must be empty."
+            )
+    elif len(prompt) == 0:
+        raise ValueError(
+            "Either input_embeddings or prompt (or both) must be provided."
+        )
+
+
+def _model_supports_mtp(model):
+    # Detect a loaded native MTP head without relying on method presence alone.
+    supported = getattr(model, "supports_mtp", None)
+    if supported is not None:
+        return bool(supported)
+    if not hasattr(model, "mtp_forward") or not hasattr(model, "make_mtp_cache"):
+        return False
+    try:
+        return len(model.make_mtp_cache()) > 0
+    except (AttributeError, TypeError):
+        return False
 
 
 def generate_step(
@@ -342,19 +389,7 @@ def generate_step(
     Yields:
         Tuple[mx.array, mx.array]: One token and a vector of log probabilities.
     """
-    if input_embeddings is not None:
-        if not does_model_support_input_embeddings(model):
-            raise ValueError("Model does not support input embeddings.")
-        elif len(prompt) > 0 and len(prompt) != len(input_embeddings):
-            raise ValueError(
-                f"When providing input_embeddings, their sequence length ({len(input_embeddings)}) "
-                f"must match the sequence length of the prompt ({len(prompt)}), or the "
-                "prompt must be empty."
-            )
-    elif len(prompt) == 0:
-        raise ValueError(
-            "Either input_embeddings or prompt (or both) must be provided."
-        )
+    validate_prompt_and_embeddings(model, prompt, input_embeddings)
 
     tokens = None
 
@@ -367,7 +402,7 @@ def generate_step(
 
     prompt_progress_callback = prompt_progress_callback or (lambda *_: None)
 
-    quantize_cache_fn = functools.partial(
+    quantize_cache_fn = partial(
         maybe_quantize_kv_cache,
         quantized_kv_start=quantized_kv_start,
         kv_group_size=kv_group_size,
@@ -455,7 +490,7 @@ def generate_step(
         if n == max_tokens:
             break
         yield y.item(), logprobs
-        if n % 256 == 0:
+        if n % _CACHE_CLEAR_INTERVAL == 0:
             mx.clear_cache()
         y, logprobs = next_y, next_logprobs
         n += 1
@@ -525,7 +560,7 @@ def speculative_generate_step(
 
     sampler = sampler or (lambda x: mx.argmax(x, axis=-1))
 
-    quantize_cache_fn = functools.partial(
+    quantize_cache_fn = partial(
         maybe_quantize_kv_cache,
         quantized_kv_start=quantized_kv_start,
         kv_group_size=kv_group_size,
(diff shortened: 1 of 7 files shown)


PhilipJohnBasile:
Maintainer-ready validation receipt for the current one-commit head
`a9fd7ef1032419a584ead9a38bdb66635f2d85c3`:

- Exact-tree macOS arm64 validation:
  https://github.com/PhilipJohnBasile/mlx-lm/actions/runs/32036066929
  - Python compilation, Black, isort, diff, and clean-tree checks passed
  - 21/21 focused native-MTP tests passed
  - 232/232 discovered unit tests passed, 1 skipped, zero failures
  - two-rank model-parallel smoke passed on both ranks

- Downstream exact-head vLLM Metal validation:
  https://github.com/PhilipJohnBasile/vllm-metal/actions/runs/32037507438
  - native Metal extension and shader builds passed
  - raw MPS and MLX-to-PyTorch bridge preflights passed
  - Ruff, formatting, mypy, and diff checks passed
  - 1,811 combined non-slow tests passed, zero failures

The official Build and Test workflow is currently `action_required` because
this is a first-time fork PR. Could a maintainer please approve that workflow
and review the patch when convenient?

The PR is non-draft, mergeable, one signed-off commit, and
`maintainer_can_modify` is enabled. Its downstream consumer is
vllm-project/vllm-metal#618, which is intentionally remaining draft until this
PR merges so it can pin the actual merged commit rather than a synthetic PR
merge SHA.

PhilipJohnBasile:
@michalk8 @nastya236 — could one of you please review or redirect this PR and approve its first-time-fork workflow?

#1740 is now a non-draft, mergeable, one-commit implementation of native Qwen MTP for the Qwen3.5/3.6/3.8 family.

Validation already completed:

- 21/21 focused native-MTP tests
- 232/232 discovered MLX-LM unit tests passed, 1 skipped
- two-rank model-parallel smoke passed on both ranks
- downstream vLLM Metal validation passed 1,811 tests with zero failures
- physical M5 Max qualification achieved exact four-request parity through 128 generated tokens
- native MTP remained active with 242/267 drafts accepted, or 90.64%

The downstream consumer is vllm-project/vllm-metal#618 and is intentionally blocked until this PR lands so it can pin the actual commit merged to mlx-lm/main.

I will address any requested changes immediately. Thank you.

PhilipJohnBasile:
Maintainer update after the latest rebase:

#1740 is now the proposed merge path for native Qwen MTP support, while preserving full attribution to @AirRunner's reference implementation in #990.

Current lineage / scope:

#990 established the native Qwen MTP reference implementation and core draft/verify approach.
#1623 has since merged the Qwen3.5/3.6 converted RMSNorm sanitization fix into main.
#1740 carries the #990 implementation forward on current main and adds the correctness/compatibility work needed for production downstream use: dense + MoE handling, exact hybrid GatedDeltaNet rollback, transactional multi-turn prompt-cache reuse, and fail-closed cache/feature handling.
The downstream consumer is vllm-project/vllm-metal#618, which is intentionally waiting for the actual commit that lands on mlx-lm/main before pinning the dependency.

Current #1740 head: 5e5ceab2732082b616238d0b1a5ab1254a3cb08a.

The previously linked upstream Actions run 32035742340 is stale: it was created for the older a9fd7ef... head and ended action_required before any jobs ran. Could a maintainer please approve CI for the current head and review #1740 (or redirect it to the right reviewer) when convenient?

I will address any requested changes immediately. Thank you.

gltanaka:
Independent user-side data point on why preserving the MTP heads matters. To be clear up front: I have not run this branch. I benchmarked native MTP as shipped via oMLX with fcmeyer's MTP-preserving conversions, so this speaks to the value of the capability, not to the correctness of this implementation.

Setup: Qwen3.8-27B, 128GB M4 Max, oMLX 0.6.1, 40-task coding harness, deterministic generation so outputs could be diffed byte-for-byte rather than eyeballed.

The question I cared about was whether speculative decoding changes what the model produces. Same weights, MTP off vs on: 39/40 outputs were byte-for-byte identical, 37/40 passed in both modes with the same three failures, and there were zero pass/fail changes.

Decode speedup over standard decoding was 1.84x at 1K context, 2.18x at 4K, and 2.19x at 16K.

So on this sample the draft/verify path buys roughly 2x decode without changing generations, which is the property I would most want evidence for before merging something that touches decode semantics.

Caveat: single session, 3 reps per cell, machine was not fully idle, so treat the speed ratios as indicative rather than precise. The output-identity result is not affected by that, since generation was deterministic.

Harness, raw outputs, and hashes: https://github.com/promptdriven/pdd/tree/main/research/omlx-qwen38-quantization (disclosure: my company's repo).

Happy to rerun against this branch on the same harness if that would be useful for review.

waybarrios:
Thanks for carrying native Qwen MTP forward. I reviewed the current head from the downstream `vllm-mlx` perspective and found one checkpoint-compatibility detail that may be worth making explicit.

In [vllm-mlx #660](https://github.com/waybarrios/vllm-mlx/pull/660#issuecomment-5128076125), we ran a deterministic same-artifact A/B on a Qwen3.6-27B MTP conversion:

| MTP input | Acceptance | Mean tok/s |
|---|---:|---:|
| post-norm | 80.30% | 36.2 |
| pre-norm | 75.24% | 34.3 |
| post-norm repeat | 80.30% | 36.2 |

The repeat reproduced the original counters exactly. This does not establish that pre-norm is wrong for vendor checkpoints, and the real-checkpoint acceptance reported on this PR is good evidence for the native path. It does show that already-converted MTP artifacts can carry a different hidden-state contract.

We handled that downstream with an explicit checkpoint setting:

```python
_QWEN_MTP_HIDDEN_STATE_MODES = {"post_norm", "pre_norm"}

mode = text_config.get(
    "mtp_hidden_state_mode",
    config.get("mtp_hidden_state_mode", "post_norm"),
)

hidden_for_mtp = hidden_states if mode == "pre_norm" else normed
```

The current implementation is [here](https://github.com/waybarrios/vllm-mlx/blob/515dd1427e233d535ec622c0f71f2768b891e4b4/vllm_mlx/patches/qwen3_5_mtp.py#L49-L70), and the selection is applied at [`return_hidden`](https://github.com/waybarrios/vllm-mlx/blob/515dd1427e233d535ec622c0f71f2768b891e4b4/vllm_mlx/patches/qwen3_5_mtp.py#L395-L407).

Would you consider keeping pre-norm as the default for conversions produced by this PR while accepting an explicit post-norm compatibility override? Having the converter persist the contract would preserve the native path without silently reducing acceptance for existing converted artifacts.

I also found a few concrete fail-closed and transactional-cache edge cases on this exact head. I am leaving those as focused inline comments so they can be discussed independently.

mlx_lm/generate.py
```diff
@@ -645,12 +680,525 @@ def _draft_generate(y, num_draft):
...
+        mtp_nonempty = any(not c.empty() for c in mtp_cache)
+        if not mtp_cache:
+            if model_nonempty:
+                raise ValueError(
+                    "A pre-populated native-MTP prompt_cache must include its "
+                    "MTP cache entries and boundary state."
+                )
+            mtp_cache = fresh_mtp_cache
+            prompt_cache.extend(mtp_cache)
+        elif prompt_state is None and (model_nonempty or mtp_nonempty):
```
waybarrios: This rejects populated caches only when `prompt_state` is absent. A present but empty `MTPPromptCacheState` passes even when the target or MTP cache is populated, and the alignment step below is skipped because `prompt_state.empty()` is true.

A reachable case is a cache exposed before generator finalization or left after an exception. Reusing it continues without the hidden state needed to align the one-token-lagged MTP cache.

Please reject when either cache group is populated and the boundary is missing **or empty**. A regression test should retain the valid all-empty case while asserting that any populated cache plus an empty boundary fails closed.

mlx_lm/generate.py
```diff
@@ -645,12 +680,525 @@ def _draft_generate(y, num_draft):
...
+        if prompt_state is not None:
+            if len(prompt_cache) == n_main + n_mtp:
+                prompt_cache.append(prompt_state)
+            else:
+                prompt_cache[n_main + n_mtp] = prompt_state
+                del prompt_cache[n_main + n_mtp + 1 :]
+
+    def _finalize_prompt_cache():
+        nonlocal finalize_plan
+        if prompt_state is None or finalize_plan is None:
```
waybarrios: An exception can leave the caller-owned in-place cache speculative while `finalize_plan` is `None`, so this returns without rollback or invalidation.

For example, `_step_backbone()` mutates the target cache before invoking an allowed plain-function logits processor. If that processor raises during verification, the supplied cache keeps unconfirmed target state while the MTP cache and boundary remain behind. Retrying with the same cache can then continue from incompatible states.

Please snapshot and restore the cache transaction on unsuccessful exit, or at minimum mark the boundary invalid so the next reuse is rejected. A focused test can inject a processor failure after prefill and assert exact restoration or deterministic rejection on retry.

mlx_lm/generate.py
```diff
@@ -689,33 +1239,72 @@ def stream_generate(
...
+                    prompt_tokens=prompt.size,
+                    prompt_tps=prompt_tps,
+                    generation_tokens=n + 1,
+                    generation_tps=(n + 1) / (time.perf_counter() - tic),
+                    peak_memory=mx.get_peak_memory() / 1e9,
+                    finish_reason=None,
+                )
 
+            detokenizer.finalize()
             yield GenerationResponse(
```
waybarrios: The terminal response is yielded before the `finally` below closes `token_generator`. With native MTP, the inner generator can still be suspended at a yield with a pending finalize plan.

A direct reproduction is `g = stream_generate(..., mtp=True, max_tokens=1, prompt_cache=cache); response = next(g)`. `response.finish_reason` is already `length`, but the emitted token is not committed and the MTP boundary can still be empty until `g` is resumed or closed.

Please close the inner token generator before yielding the terminal response, keeping the outer `finally` as an idempotent safety net. The regression test should inspect target offset, MTP offset, and boundary metadata immediately when the terminal response is received.

mlx_lm/generate.py
```diff
@@ -689,33 +1239,72 @@ def stream_generate(
...
+        kwargs.pop("max_kv_size", None)
+        kwargs.pop("prompt_progress_callback", None)
+        token_generator = speculative_generate_step(
+            prompt, model, draft_model, **kwargs
+        )
+    elif mtp and _model_supports_mtp(model):
+        kwargs.pop("max_kv_size", None)
+        kwargs.pop("prompt_progress_callback", None)
+        kwargs.pop("num_draft_tokens", None)
+        kwargs.pop("sampler", None)  # mtp_generate_step does not accept sampler=
```
waybarrios: Native MTP silently drops a caller-provided `sampler`, while normal and external-draft generation honor the same public argument.

For example, `stream_generate(..., mtp=True, sampler=forcing_sampler)` with default `temp=0` follows argmax and never calls `forcing_sampler`. MTP needs the proposal distribution for correct acceptance, so arbitrary samplers may indeed be unsupported, but silently changing the requested policy is not fail-closed behavior.

Please reject `sampler` explicitly when MTP is active, or introduce a richer sampler interface that exposes the distribution. A dispatch test should assert that a custom sampler is honored or rejected, never ignored.

mlx_lm/generate.py
```diff
@@ -689,33 +1239,72 @@ def stream_generate(
...
-    if draft_model is None:
+    if draft_model is not None:
+        kwargs.pop("max_kv_size", None)
+        kwargs.pop("prompt_progress_callback", None)
+        token_generator = speculative_generate_step(
+            prompt, model, draft_model, **kwargs
+        )
+    elif mtp and _model_supports_mtp(model):
+        kwargs.pop("max_kv_size", None)
+        kwargs.pop("prompt_progress_callback", None)
```
waybarrios: Dropping `prompt_progress_callback` removes all prefill progress reporting from the native-MTP server path. `_serve_single()` still passes the callback, but `mtp_generate_step()` has no equivalent reporting around its chunked prefill.

With `--mtp` and a long uncached prompt, streaming clients receive no progress or keepalive events until the first generated token, unlike normal generation. Please pass the callback into `mtp_generate_step()` and mirror the initial, incremental, and completed progress calls from `generate_step()`. A multi-chunk prompt test can record and compare those events.

mlx_lm/generate.py
```diff
@@ -645,12 +680,525 @@ def _draft_generate(y, num_draft):
...
+            "Native MTP generation does not yet support input_embeddings. "
+            "Use standard generation for embedding or multimodal inputs."
+        )
+    validate_prompt_and_embeddings(model, prompt, input_embeddings)
+
+    if logits_processors:
+        unsafe_processors = [
+            type(processor).__name__
+            for processor in logits_processors
+            if not inspect.isfunction(processor)
```
waybarrios: `inspect.isfunction()` proves callable form, not statelessness. A plain closure can mutate captured or global state, and `partial` can wrap a stateful callable. Both are currently accepted and are explicitly recommended by the error message.

MTP invokes processors for proposed and discarded positions, so a mutable call counter advances differently from serial generation and can change the processed target distribution. Please require an explicit stateless marker or protocol for user processors, and mark the processors created by `make_logits_processors()` as safe. Tests should reject a stateful closure and a partial wrapping a stateful callable.

tests/test_mtp.py
```diff
@@ -0,0 +1,504 @@
...
+            5,
+            f"{label}Too few rejection events observed; increase n_runs",
+        )
+        self.assertGreater(
+            len(set(rejection_tokens)),
+            1,
+            f"{label}Rejection tokens are always identical, argmax bug likely present",
+        )
+
+    def test_mtp_rejection_residual_sampling(self):
```
waybarrios: This test never reaches MTP draft generation or residual rejection sampling. `_collect_rejection_tokens()` uses `max_tokens=1`, so `mtp_generate_step()` yields the initial backbone token and returns before calling `_step_mtp()`. Every collected `from_draft=False` token is therefore an ordinary first token.

The test can still pass if the residual branch is deleted or replaced with argmax. Please generate beyond the first token and force a deterministic draft rejection with controlled logits. The test should assert that verification and rejection actually occurred before checking the replacement distribution.

zcbenz:
Closing per https://github.com/ml-explore/mlx-lm/pull/990#pullrequestreview-4989083778.

PR #1741, opened 2026-08-16
models/qwen3_5: fused conv-state append + depthwise conv + silu for decode steps

jonathan308:
At decode (S==1) each GatedDeltaNet linear-attention layer runs 4 dependent kernels on a (B,4,C) window — `concatenate(state, x)` → `contiguous` (state write) → depthwise conv1d → `silu` — plus 2 allocations. On a 48-linear-layer hybrid (Qwen3.6-27B class) this chain costs ~0.9 ms/token. This PR fuses the whole step into one `mx.fast.metal_kernel` (one thread per channel: state shift+append, 4-tap dot in fp32, single round to working dtype, stabilized-sigmoid silu).

Bit-exactness required replicating two MLX internals (documented in code): conv1d accumulates in fp32 and rounds once; `nn.silu`'s sigmoid is MLX's stabilized `1/(1+exp(|x|))` form evaluated in working-dtype arithmetic. With both replicated the kernel is **bit-identical** to the composed path (isolated maxdiff 0.0; 64/64 greedy tokens identical e2e, including across the ctx-1024 SDPA 1-pass/2-pass switch).

Stock path retained for prefill (S>1), masks, `cache.lengths`, and non-Metal situations; the fused branch triggers only for S==1 + active cache + no mask. Toggle: `_USE_FUSED_CONV` module flag (default on).

### Benchmarks (M3 Ultra, Qwen3.6-27B-4bit, interleaved A/B)
| ctx | stock | fused | Δ |
|---|---|---|---|
| 1024 | 25.31 ms/tok | 24.88 | **−0.43 ms (+1.7%)** |
| short (32-tok prompt) | 25.30 | 24.74 | −0.56 ms (+2.2%) |

Micro: 4-kernel chain 28.3 µs → fused 8.8 µs.

### Scope / follow-ups
- Same pattern shared by `qwen3_next.py` / `qwen3_5_moe.py` via `gated_delta.py` users; this PR covers `qwen3_5.py` (verified with the real model); extending is mechanical.
- Verified with the full mlx-lm suite (145 passed + 52 subtests).

(draft-note for maintainers: happy to move the kernel into a shared helper module if you prefer that layout.)


zcbenz:

PR #1742, opened 2026-08-16
generate: optional compute_logprobs=False (sample from raw logits); top_p shift-invariant softmax

jonathan308:
`generate_step` normalizes logits to logprobs every token (`logits - mx.logsumexp(logits)` — a 2-kernel reduce+elementwise chain over the full vocab, on the serial decode path), and every built-in sampler is shift-invariant, so the normalization has no effect on sampling at all. Its only purpose is the *yielded* logprob vector.

This PR adds `compute_logprobs: bool = True` to `generate_step`:

- `True` (default): current behavior, byte-for-byte.
- `False`: sample from raw logits and yield `None` for logprobs. `generate()` (text-only entry) now passes `False` by default.

Sampler shift-invariance audit (`sample_utils.py`): argmax, top-k (ranking), min-p (threshold cancels), xtc (softmax), categorical (softmax) are all invariant under a per-row constant shift. **One fix was needed**: `apply_top_p` used `mx.exp(logprobs)`, which only equals probabilities when the input is normalized — it now uses `mx.softmax`, which is identical on true logprobs and also correct on raw logits (verified: identical masks and 0.0 prob diff on both input kinds).

Greedy decode (temp=0, the CLI default) produces **bit-identical token streams** with the flag off (verified at 8B and 27B; 64-token streams equal).

### Benchmarks (M3 Ultra, greedy)
- Qwen3-8B-4bit: 7.96 → 7.92 ms/tok
- Qwen3.6-27B-4bit (vocab 248k): 25.15 → 25.09 ms/tok

Small but free; also removes 2 launches/token from the serial chain (compounds with other fusion work).

### Compatibility
- `stream_generate` yields `GenerationResponse.logprobs=None` when disabled; `server.py` keeps the default (`True`) — no API change for serving users.
- `speculative_generate_step` unchanged (its acceptance math needs true logprobs).

### Tests
33/33 existing generate/sample tests pass; added: greedy stream identity with flag on/off; top_p on raw logits vs normalized (mask + distribution equality). Full suite: 145 passed + 52 subtests.


zcbenz (changes requested):
Lazy graph is capable of removing unused computations.

PR #1743, opened 2026-08-16
fix(utils): honor donate_model in save() so --export-gguf can read weights (#1723)

### LINKED ISSUE #1723: mlx_lm.fuse --export-gguf fails with row-major error — root cause: save() ignores donate_model=False
mlx_lm.fuse --export-gguf fails with "[save_gguf] can only serialize row-major arrays" — root cause: save() ignores donate_model=False, empties the model before GGUF export reads it

Environment

mlx-lm 0.31.3
mlx 0.32.0
mlx_lm_lora 3.0.0 (not implicated, just noting the adapter was trained with it)
macOS, Apple Silicon
Base model: mlx-community/Llama-3.2-3B-Instruct-4bit

Command

python -m mlx_lm.fuse \
  --model mlx-community/Llama-3.2-3B-Instruct-4bit \
  --adapter-path <lora adapter dir> \
  --save-path <out>/fused_model \
  --export-gguf \
  --gguf-path <out>/ggml-model-f16.gguf \
  --dequantize

Observed error

  ...
  File ".../mlx_lm/fuse.py", line 100, in main
    convert_to_gguf(save_path, weights, config, str(save_path / args.gguf_path))
  File ".../mlx_lm/gguf.py", line 313, in convert_to_gguf
    mx.save_gguf(output_file_path, weights, metadata)
ValueError: [save_gguf] can only serialize row-major arrays

This looked at first like the array-layout issue described in #353, but it isn't — see root cause below. The row-major message is mx.save_gguf reacting to a batch of empty arrays (mx.array([]), shape (0,)), not a stride/permutation problem. Confirmed by patching mx.save_gguf to dump per-tensor shapes right before the real call: every single tensor (token_embd.weight, every blk.N.*, output_norm.weight — ~250 tensors) reported shape=(0,).

Root cause

fuse.py's main():

python
save(
    save_path,
    args.model,
    model,
    tokenizer,
    config,
    donate_model=False,   # <- intentional: model.parameters() is still needed below
)

if args.export_gguf:
    ...
    weights = dict(tree_flatten(model.parameters()))   # <- reads EMPTY arrays
    convert_to_gguf(save_path, weights, config, str(save_path / args.gguf_path))

utils.py's save() accepts donate_model but its body hardcodes True when calling save_model(), ignoring the argument entirely:

python
def save(
    dst_path: Union[str, Path],
    src_path_or_repo: Union[str, Path],
    model: nn.Module,
    tokenizer: TokenizerWrapper,
    config: Dict[str, Any],
    donate_model: bool = True,
):
    ...
    save_model(dst_path, model, donate_model=True)   # <- always True, ignores the param above

utils.py's save_model(), when donate_model=True, frees memory by replacing every parameter with an empty array right after writing it to disk:

python
if donate_model:
    model.update(tree_map(lambda _: mx.array([]), model.parameters()))

So fuse.py deliberately passes donate_model=False to save() because it still needs model.parameters() for the --export-gguf step immediately afterward — but save() silently ignores that and always donates. By the time --export-gguf's tree_flatten(model.parameters()) runs, every parameter has already been swapped out for mx.array([]).

Reproduces on a stock fuse, not anything specific to this adapter/base model — any mlx_lm.fuse --export-gguf call goes through this same save() → donate → re-read-parameters sequence.

Suggested fix

Either:

save() should actually honor its own donate_model parameter:
python
   save_model(dst_path, model, donate_model=donate_model)
Or fuse.py's main() should capture the weights dict before calling save(), not after:
python
   if args.export_gguf:
       weights = dict(tree_flatten(model.parameters()))

   save(save_path, args.model, model, tokenizer, config, donate_model=False)

   if args.export_gguf:
       convert_to_gguf(save_path, weights, config, str(save_path / args.gguf_path))

(GGUF conversion still needs to run after save(), since convert_to_gguf reads tokenizer.json from save_path, which is written by save() itself — so only the extraction moves earlier, not the conversion call.)

Fix #1 is more general (fixes save() for any other caller relying on donate_model=False), but #2 is what I verified as a workaround locally and confirmed resolves the issue end-to-end (fuse → dequantize → GGUF export → quantize with llama-quantize all completed successfully afterward).

Workaround used in the meantime

A local wrapper script that mirrors fuse.py's main() with fix #2 applied, invoked in place of python -m mlx_lm.fuse. Happy to share it if useful for a regression test.

Anai-Guo:
## Summary

Fixes #1723.

`save()` in `mlx_lm/utils.py` accepts a `donate_model` parameter, but its body hardcoded `donate_model=True` when calling `save_model()`, ignoring the argument entirely:

```python
def save(..., donate_model: bool = True):
    ...
    save_model(dst_path, model, donate_model=True)   # <- always True
```

When `save_model()` donates, it frees memory by replacing every parameter with an empty array right after writing it to disk:

```python
if donate_model:
    model.update(tree_map(lambda _: mx.array([]), model.parameters()))
```

`mlx_lm.fuse.main()` deliberately passes `donate_model=False` to `save()` precisely because it still needs `model.parameters()` for the `--export-gguf` step immediately afterward:

```python
save(save_path, args.model, model, tokenizer, config, donate_model=False)
if args.export_gguf:
    ...
    weights = dict(tree_flatten(model.parameters()))   # reads EMPTY arrays
    convert_to_gguf(save_path, weights, config, ...)
```

Because `save()` ignored the flag and donated anyway, every tensor handed to `mx.save_gguf` had shape `(0,)`. That surfaces as the misleading error:

```
ValueError: [save_gguf] can only serialize row-major arrays
```

(the row-major message is `mx.save_gguf` reacting to a batch of empty arrays, not a stride/permutation problem).

## Fix

Pass the parameter through so `save()` respects its own contract:

```python
save_model(dst_path, model, donate_model=donate_model)
```

This is the general fix (fix #1 in the issue): any caller relying on `donate_model=False` now keeps its weights. Existing callers that use the default `donate_model=True` are unaffected. The issue reporter confirmed this resolves the end-to-end `fuse → dequantize → GGUF export → quantize` flow.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -982,7 +982,7 @@ def save(
         hf_repo = None
 
     dst_path = Path(dst_path)
-    save_model(dst_path, model, donate_model=True)
+    save_model(dst_path, model, donate_model=donate_model)
     save_config(config, config_path=dst_path / "config.json")
     tokenizer.save_pretrained(dst_path)
 


michalk8 (approved):
LGTM, thanks!

PR #1744, opened 2026-08-17
Add load() overloads so return_config narrows the return type

### LINKED ISSUE #1732: Code in the first example for Python SDK do not pass type check on load()'s returned values
Hi,  the example run fine, but when I want to use same pattern in my own programs which I always type check, errors are thrown.

```python
    model, tokenizer = load("mlx-community/Mistral-7B-Instruct-v0.3-4bit") 
    prompt = "Write a short story about ..."

    messages = [{"role": "user", "content": prompt}]
        messages,
        add_generation_prompt=True,
    )

    text = generate(model, tokenizer, prompt, verbose=True)
    print(text)
```

With `pyrefly`, the following error is thrown:
```
$ hatch check types
 INFO Checking project configured at `/Users/xxxxx/Library/Application Support/hatch/env/virtual/.config/asA_QThn/pyrefly.toml`
ERROR Cannot unpack tuple[Module, TokenizerWrapper] | tuple[Module, TokenizerWrapper, dict[str, Any]] (of size 3) into 2 values [bad-unpacking]
 --> src/myprog/myprog:6:5
  |
6 |     model, tokenizer = load("mlx-community/Mistral-7B-Instruct-v0.3-4bit")
  |     ^^^^^^^^^^^^^^^^
```

With `mypy --strict`, the following error is thrown:
```
$ hatch check types
src/myprog/myprog:6: error: Too many values to unpack (2 expected, 3 provided)  [misc]
```

My current workaround is to disable type check for that line (`# type: ignore`), but it's not ideal, as LLM inference is critical path.

My understanding, is  that which one of the two components of the union is returned depends on whether `return_config` parameter is `True` or `False`.

On Stack Overflow[^1], the accepted solution mentions using overload to fix the problem.

Something like that I guess:
```python
@overload
def load(return_config: Literal[True]) -> tuple[Module, TokenizerWrapper, dict[str, Any]]: 
    ...
@overload
def load(return_config: Literal[False] = ...) -> tuple[Module, TokenizerWrapper]:
    ...
    
def load(return_config: bool = False) -> tuple[Module, TokenizerWrapper] | tuple[Module, TokenizerWrapper, dict[str, Any]]:
    ...
       if return_config:
        return model, tokenizer, config
    else:
        return model, tokenizer
```

[^1]: https://stackoverflow.com/a/72415876/6518111

Awshesh12:
## What

Add `@overload` declarations for `load()` so its return type narrows on `return_config`

## Why

Closes #1732. `load()` is annotated as `Union[Tuple[nn.Module, TokenizerWrapper], Tuple[nn.Module, TokenizerWrapper, Dict[str, Any]]]`, so a checker sees a union of a 2-tuple and a 3-tuple and cannot tell which one a call returns. The two-value unpack in the README fails:

```
error: Too many values to unpack (2 expected, 3 provided)  [misc]
```

pyrefly reports the same as `bad-unpacking`. Today the documented entry point needs a `# type: ignore` to pass a typed build.

## How

Two overloads keyed on `return_config`, `Literal[False]` returning the 2-tuple and `Literal[True]` returning the 3-tuple. The implementation signature and body are unchanged, and the docstring now describes both forms. Overload declarations are erased at import, so runtime behavior is identical.

`sharded_load` has the same `return_config` branch but carries no return annotation, so it does not error today. I left it alone to keep this to one change, happy to follow up if you want it typed as well.

## Test plan

Ran on Apple Silicon with mlx 0.32.0 and Python 3.12. The README example type checks after the change, and every call form resolves, with `return_config=True` narrowing the third element to `dict[str, Any]`. `pre-commit run --files mlx_lm/utils.py` passes black and isort with the file unchanged.

`python -m unittest discover tests/` ran 211 tests with 1 skip and 1 error. That error, `tests/test_datsets.py` loading `billsum`, reproduces on an unmodified tree and is already addressed by #1434.

Since this only changes annotations I did not add a runtime test, as there is no type-checking gate in the suite today. Happy to add one if you would like the overloads pinned.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -15,10 +15,12 @@ from typing import (
     Callable,
     Dict,
     List,
+    Literal,
     Optional,
     Tuple,
     Type,
     Union,
+    overload,
 )
 
 import mlx.core as mx
@@ -479,6 +481,32 @@ def load_tokenizer(model_path, tokenizer_config_extra=None, eos_token_ids=None):
     )
 
 
+@overload
+def load(
+    path_or_hf_repo: str,
+    tokenizer_config: Optional[Dict[str, Any]] = ...,
+    model_config: Optional[Dict[str, Any]] = ...,
+    adapter_path: Optional[str] = ...,
+    lazy: bool = ...,
+    return_config: Literal[False] = ...,
+    revision: Optional[str] = ...,
+    trust_remote_code: bool = ...,
+) -> Tuple[nn.Module, TokenizerWrapper]: ...
+
+
+@overload
+def load(
+    path_or_hf_repo: str,
+    tokenizer_config: Optional[Dict[str, Any]] = ...,
+    model_config: Optional[Dict[str, Any]] = ...,
+    adapter_path: Optional[str] = ...,
+    lazy: bool = ...,
+    return_config: Literal[True] = ...,
+    revision: Optional[str] = ...,
+    trust_remote_code: bool = ...,
+) -> Tuple[nn.Module, TokenizerWrapper, Dict[str, Any]]: ...
+
+
 def load(
     path_or_hf_repo: str,
     tokenizer_config: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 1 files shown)


zcbenz:

PR #1746, opened 2026-08-17
Fix batch cache extract() leaking pending right padding

### LINKED ISSUE #1384: Gemma 3n: mlx_lm.server returns garbage output (shared-KV layers incompatible with batched cache)

### LINKED ISSUE #1493: mlx_lm.server: generation hangs at 0% CPU right after prompt processing on ~22-26k-token streaming requests (gemma-4-26b-a4b, 0.31.3)

jstray:
## Summary

When the server batches several requests together, shorter rows get junk filler cells at the end to square off the rectangle (right padding), and a cleanup pass (`finalize()`) later shuffles each row so the junk sits where
the bookkeeping expects and each row's real-token count is correct again. Between processing and cleanup there is a window where the batch contains junk cells and the counts are temporarily off — by design, because `finalize()` is about to fix it.

The bug: `BatchKVCache.extract()` and `BatchRotatingKVCache.extract()` — which copy one request's row out of the batch to save as a prompt-cache entry — can run during that window. The saved copy then includes the junk filler cells and records the row length *counting* them ("this cache holds 332 tokens" when it's really 322 real ones plus 10 junk). This happens whenever `mlx_lm.server` stores a segment-boundary cache during a mixed-length batched prefill.

When a later request reuses that saved entry, every piece of bookkeeping that derives from the cache length — where new tokens get appended, how wide the attention mask should be — disagrees with the actual contents by
exactly the junk amount. That surfaces as `[broadcast_shapes]` crashes in `scaled_dot_product_attention` (mask width vs. key length differ by the pad amount), as livelocked generation, or quietest of all, as junk cells treated as real context. 

Possibly fixes #1384 where we are seeing Gemma garbage and crashes. Possibly fixes #1493, where the hang is on the same model family under exactly this workload. 

## Reproduction

Serve any model whose requests share long prompt prefixes with differing tails (we hit this with gemma-4-26b-a4b under 4 concurrent single-token-classification requests of unequal lengths; the server wedges
or crashes within a few hundred requests once prompt-cache restores begin). Unit-level, before this fix:

```python
batch = BatchKVCache.merge(four_caches_with_52_token_prefix)
batch.prepare(lengths=[270, 280, 272, 271], right_padding=[10, 0, 8, 9])
batch.update_and_fetch(k, v)          # padded chunk, S=280
batch.extract(0).offset               # 332 -- should be 322 (52 + 270)
```

## Fix

When snapshotting, cut off the junk cells and record the true count — i.e., `extract()` now produces exactly what it would have produced after `finalize()`: it subtracts the sequence's pending right padding from the slice end and (for the rotating cache) from the reported offset. Post-`finalize()` behavior is unchanged.

## Tests

`TestBatchCacheExtract.test_extract_with_pending_right_padding` covers both batch cache classes, pre- and post-finalize; it fails without the fix and passes with it. The full `tests/test_prompt_cache.py` suite passes (23 tests).

## Validation at scale

With this fix applied, our previously-crashing workload (three consecutive 4-way concurrent runs, ~650 prompt-cache stores against a gemma-4-26b-a4b server) completes with zero internally-inconsistent stored entries,
verified by per-layer instrumentation; unpatched, corrupted entries appear within the first dozen stores and the server crashes or hangs shortly after.

We also prototyped a follow-up that avoids `fetch_nearest_cache`'s per-fetch `copy.deepcopy` of the full entry; happy to open that separately if there's interest.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -1080,8 +1080,15 @@ class BatchKVCache(_BaseCache):
     def extract(self, idx):
         cache = KVCache()
         padding = self.left_padding[idx].item()
-        cache.keys = mx.contiguous(self.keys[idx : idx + 1, :, padding : self._idx])
-        cache.values = mx.contiguous(self.values[idx : idx + 1, :, padding : self._idx])
+        # Pending (un-finalized) right padding must not leak into the
+        # extracted cache: those trailing positions are pad garbage, and
+        # counting them corrupts both content and the offset bookkeeping of
+        # any prompt-cache entry stored from a segment boundary.
+        end = self._idx
+        if self._right_padding is not None:
+            end -= int(self._right_padding[idx].item())
+        cache.keys = mx.contiguous(self.keys[idx : idx + 1, :, padding:end])
+        cache.values = mx.contiguous(self.values[idx : idx + 1, :, padding:end])
         cache.offset = cache.keys.shape[2]
         return cache
 
@@ -1426,8 +1433,17 @@ class BatchRotatingKVCache(_BaseCache):
             cache.keys = mx.roll(cache.keys, -self._idx, axis=2)
             cache.values = mx.roll(cache.values, -self._idx, axis=2)
             cache._idx = self.max_size
-        cache.keys = mx.contiguous(cache.keys[:, :, padding : cache._idx])
-        cache.values = mx.contiguous(cache.values[:, :, padding : cache._idx])
+        # Pending (un-finalized) right padding must not leak into the
+        # extracted cache; see BatchKVCache.extract. _lengths holds the
+        # per-sequence true end offsets while a right-padded chunk is in
+        # flight, so the pad amount is offset - _lengths clamped at 0.
+        end = cache._idx
+        if self._lengths is not None:
+            pad = max(0, int((self.offset - self._lengths).tolist()[idx]))
+            end -= pad
+            offset -= pad
+        cache.keys = mx.contiguous(cache.keys[:, :, padding:end])
+        cache.values = mx.contiguous(cache.values[:, :, padding:end])
         cache.offset = offset
         cache._idx = cache.keys.shape[2]
         return cache
--- a/tests/test_prompt_cache.py
+++ b/tests/test_prompt_cache.py
@@ -769,5 +769,77 @@ class TestPromptCache(unittest.TestCase):
         self.assertTrue(mx.array_equal(mask, expected))
 
 
+class TestBatchCacheExtract(unittest.TestCase):
+    """Regression test: extract() from a batch that has pending
+    (un-finalized) right padding must not leak the padding into the
+    extracted cache.
+
+    mlx_lm.server hits this path when it stores a prompt-cache entry at a
+    segment boundary while a mixed-length batched prefill is in flight:
+    between update_and_fetch() on a right-padded chunk and finalize(), the
+    batch buffer holds pad cells and per-sequence bookkeeping that
+    finalize() has not yet reconciled. An extract() in that window used to
+    return a cache containing the pad cells, with an offset that counted
+    them. Restoring such an entry later desynchronizes cache length vs.
+    attention-mask width by exactly the pad amount (broadcast errors in
+    attention, hung generation).
+
+    The constants mirror the production incident this was root-caused
+    from: four concurrent requests sharing a 52-token cached prompt
+    prefix, with per-request remainders of 270/280/272/271 tokens, so
+    sequence 0 carries 10 tokens of right padding.
+    """
+
+    def _restored_prefix_cache(self, cls, **kwargs):
+        # A single-sequence cache holding a 52-token prefix, built the way
+        # the server produces one: fill with more tokens (92), then trim
+        # back (by 40), as fetch_nearest_cache() does when it restores a
+        # stored entry to the shared-prefix point. Key contents are
+        # arange() so positions are identifiable when debugging.
+        c = cls(**kwargs)
+        k = mx.broadcast_to(
+            mx.arange(92, dtype=mx.float32).reshape(1, 1, 92, 1), (1, 2, 92, 4)
+        )
+        c.update_and_fetch(k, k)
+        c.trim(40)
+        return c
+
+    def test_extract_with_pending_right_padding(self):
+        # Both patched classes: the plain batch cache and the rotating
+        # (sliding-window) one. max_size=1024 keeps the rotating cache
+        # un-rotated, matching the production state (total length < window).
+        for batch_cls, single_cls, kwargs in (
+            (BatchKVCache, KVCache, {}),
+            (BatchRotatingKVCache, RotatingKVCache, {"max_size": 1024}),
+        ):
+            singles = [
+                self._restored_prefix_cache(single_cls, **kwargs) for _ in range(4)
+            ]
+            batch = batch_cls.merge(singles)
+
+            # One batched prefill chunk of the four remainders, padded to
+            # the longest (280): sequence 0 has 270 real tokens + 10 pad.
+            lengths = [270, 280, 272, 271]
+            S = max(lengths)
+            batch.prepare(lengths=lengths, right_padding=[S - l for l in lengths])
+            k = mx.random.normal((4, 2, S, 4))
+            batch.update_and_fetch(k, k)
+
+            # Extract in the pre-finalize window. Sequence 0 holds
+            # 52 (prefix) + 270 (real remainder) = 322 tokens; without the
+            # fix both assertions see 332 (322 + 10 pad) for both classes.
+            extracted = batch.extract(0)
+            self.assertEqual(extracted.keys.shape[2], 52 + 270)
+            self.assertEqual(extracted.offset, 52 + 270)
+
+            # After finalize() the padding is reconciled; extract() must
+            # give the same answer (this path was already correct, and
+            # pins that the fix did not change it).
+            batch.finalize()
+            extracted = batch.extract(0)
+            self.assertEqual(extracted.keys.shape[2], 52 + 270)
+            self.assertEqual(extracted.offset, 52 + 270)
+
+
 if __name__ == "__main__":
     unittest.main()


zcbenz (commented):
> The bug: `BatchKVCache.extract()` and `BatchRotatingKVCache.extract()` — which copy one request's row out of the batch to save as a prompt-cache entry — can run during that window

I don't understand why this is happening, are you running a custom server implementation?

zcbenz:

jstray:
Sorry, missed your question.
Yes, I am running a custom server. Here's the code:

```
// Batched CoPE scorer: policy-prefix KV broadcast + packed-lane decoding.
//
// Replaces the llama-server round-trip path for annotate.py's workload
// (one grammar-locked "0"/"1" token per (policy, post); the score is
// P("1") from the logits). The policy prefix is decoded once per policy
// and copied to every lane with llama_memory_seq_cp, so each post costs
// only its own ~40 suffix tokens, packed up to ROUND_BUDGET per decode.
// Validated against the server path on 1000 real corpus posts x 6
// policies (2026-08-18): mean |dp1| 0.0040, p99 0.041, and 3/6000 label
// flips at the 0.5 threshold (0.05%, vs ~1% between GGUF quants).
//
// Line protocol on stdin/stdout (payload bytes follow the header line,
// no trailing newline; responses are flushed):
//   POLICY <nbytes>\n<bytes>   -> POLICYOK <n_prefix_tokens>
//   POST <id> <nbytes>\n<bytes>   (queued; no response)
//   FLUSH                      -> SCORE <id> <p1> per queued post, then
//                                 DONE <n_posts> <n_tokens> <ms>
// Shutdown is SIGTERM from the parent (annotate.py's atexit hook).
// All llama.cpp logging goes to stderr; stdout carries only the protocol.

#include "llama.h"
#include <math.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>

#define MAX_TOK     4096
#define MAX_QUEUE   512
#define LANES       96      // max posts per decode round; sized so that
                            // ROUND_BUDGET, not lane count, is the binding
                            // constraint on typical ~40-token posts
#define PREFIX_SEQ  127     // n_seq_max - 1; lanes use 0..LANES-1
#define SUFFIX_CAP  1664    // tokens; deterministic stand-in for the server
                            // path's 400-and-retry on pathological posts
#define ROUND_BUDGET 4096   // max suffix tokens per llama_decode = n_batch.
                            // Posts can span the internal 512-token ubatch
                            // splits, so big rounds amortize the one partial
                            // trailing ubatch over ~8 full ones; with the
                            // ~360-token prefix this stays under n_ctx 8192

// The CoPE policy-prompt wrapper. This file is now its only home; the cache
// rows it produced are keyed by annotate.py's PROMPT_VERSION ("p1"), so any
// edit here must bump that constant or 280k cached annotations go stale
// silently.
static const char *WRAPPER_HEAD =
    "Examine the given POLICY and determine if the given CONTENT meets "
    "the criteria for ANY of the LABELS. Answer \"1\" if yes, and \"0\" if "
    "no.\n\n\nPOLICY\n======\n\n";
static const char *WRAPPER_MID = "\n\n\nCONTENT\n=======\n\n";

// Gemma-4 turn format, byte-identical to llama-server's Jinja rendering
// (verified against /apply-template and usage.prompt_tokens; the GGUF's
// template is full Jinja, which llama_chat_apply_template cannot parse).
static const char *TPL_PRE  = "<|turn>system\n<|think|>\n<turn|>\n<|turn>user\n";
static const char *TPL_TAIL = "<turn|>\n<|turn>model\n";

static double now_ms(void) {
    struct timespec ts; clock_gettime(CLOCK_MONOTONIC, &ts);
    return ts.tv_sec * 1000.0 + ts.tv_nsec / 1e6;
}

static char *read_payload(size_t n) {
    char *buf = malloc(n + 1);
    if (!buf || fread(buf, 1, n, stdin) != n) {
        fprintf(stderr, "cope_scorer: short read on %zu-byte payload\n", n);
        exit(1);
    }
    buf[n] = 0;
    return buf;
}

typedef struct {
    char        id[64];
    llama_token *tok;
    int         n;
} queued_post;

int main(int argc, char **argv) {
    if (argc != 2) { fprintf(stderr, "usage: %s <model.gguf>\n", argv[0]); return 1; }
    llama_backend_init();

    struct llama_model_params mp = llama_model_default_params();
    mp.n_gpu_layers = 99;
    struct llama_model *model = llama_model_load_from_file(argv[1], mp);
    if (!model) { fprintf(stderr, "model load failed\n"); return 1; }

    struct llama_context_params cp = llama_context_default_params();
    cp.n_ctx = 8192; cp.n_batch = 4096; cp.n_ubatch = 512;
    cp.n_seq_max = 128;
    cp.n_threads = 10; cp.n_threads_batch = 10;
    cp.flash_attn_type = LLAMA_FLASH_ATTN_TYPE_ENABLED;
    cp.type_k = GGML_TYPE_Q8_0; cp.type_v = GGML_TYPE_Q8_0;
    cp.swa_full = true;
    // seq_cp requires the unified KV buffer (split per-sequence streams
    // assert "only supported for full KV buffers").
    cp.kv_unified = true;
    struct llama_context *ctx = llama_init_from_model(model, cp);
    if (!ctx) { fprintf(stderr, "context init failed\n"); return 1; }

    const struct llama_vocab *vocab = llama_model_get_vocab(model);
    llama_memory_t mem = llama_get_memory(ctx);

    llama_token t0, t1;
    if (llama_tokenize(vocab, "0", 1, &t0, 1, false, false) != 1 ||
        llama_tokenize(vocab, "1", 1, &t1, 1, false, false) != 1) {
        fprintf(stderr, "'0'/'1' are not single tokens\n"); return 1;
    }

    int n_prefix = 0;
    queued_post queue[MAX_QUEUE];
    int n_queue = 0;

    printf("READY\n"); fflush(stdout);

    char line[256];
    while (fgets(line, sizeof line, stdin)) {
        if (strncmp(line, "POLICY ", 7) == 0) {
            size_t nb = strtoul(line + 7, NULL, 10);
            char *policy = read_payload(nb);
            size_t cap = strlen(TPL_PRE) + strlen(WRAPPER_HEAD) + nb
                       + strlen(WRAPPER_MID) + 1;
            char *head = malloc(cap);
            snprintf(head, cap, "%s%s%s%s", TPL_PRE, WRAPPER_HEAD, policy,
                     WRAPPER_MID);
            static llama_token tokP[MAX_TOK];
            n_prefix = llama_tokenize(vocab, head, (int)strlen(head), tokP,
                                      MAX_TOK, true, true);
            free(head); free(policy);
            if (n_prefix <= 0) { fprintf(stderr, "prefix tokenize failed\n"); return 1; }
            llama_memory_clear(mem, true);
            struct llama_batch b = llama_batch_init(n_prefix, 0, 1);
            b.n_tokens = n_prefix;
            for (int j = 0; j < n_prefix; j++) {
                b.token[j] = tokP[j]; b.pos[j] = j;
                b.n_seq_id[j] = 1; b.seq_id[j][0] = PREFIX_SEQ;
                b.logits[j] = false;
            }
            if (llama_decode(ctx, b)) { fprintf(stderr, "prefix decode failed\n"); return 1; }
            llama_batch_free(b);
            // Broadcast the prefix to every lane once per policy; rounds
            // then only trim suffixes (seq_cp per round measured ~10x the
            // decode cost).
            for (int k = 0; k < LANES; k++) {
                llama_memory_seq_rm(mem, k, -1, -1);
                llama_memory_seq_cp(mem, PREFIX_SEQ, k, 0, n_prefix);
            }
            printf("POLICYOK %d\n", n_prefix); fflush(stdout);

        } else if (strncmp(line, "POST ", 5) == 0) {
            char id[64]; size_t nb;
            if (sscanf(line + 5, "%63s %zu", id, &nb) != 2 || n_queue >= MAX_QUEUE) {
                fprintf(stderr, "bad POST or queue full\n"); return 1;
            }
            char *post = read_payload(nb);
            queued_post *q = &queue[n_queue];
            strncpy(q->id, id, sizeof q->id - 1);
            q->tok = malloc(MAX_TOK * sizeof(llama_token));
            // Joint post+tail tokenization matches the server byte-for-byte;
            // only over-cap posts fall back to a split at the truncation.
            size_t cap2 = nb + strlen(TPL_TAIL) + 1;
            char *suf = malloc(cap2);
            snprintf(suf, cap2, "%s%s", post, TPL_TAIL);
            q->n = llama_tokenize(vocab, suf, (int)strlen(suf), q->tok,
                                  MAX_TOK, false, true);
            if (q->n <= 0 || q->n > SUFFIX_CAP) {
                int np = llama_tokenize(vocab, post, (int)strlen(post), q->tok,
                                        MAX_TOK, false, true);
                if (np < 0) np = MAX_TOK;             // still too long: hard cap
                if (np > SUFFIX_CAP - 16) np = SUFFIX_CAP - 16;
                llama_token tt[32];
                int nt = llama_tokenize(vocab, TPL_TAIL, (int)strlen(TPL_TAIL),
                                        tt, 32, false, true);
                memcpy(q->tok + np, tt, nt * sizeof(llama_token));
                q->n = np + nt;
            }
            free(suf); free(post);
            n_queue++;

        } else if (strncmp(line, "FLUSH", 5) == 0) {
            double ms = 0; long total_tokens = 0;
            int base = 0;
            while (base < n_queue) {
                int nb = 0, round_tokens = 0;
                while (base + nb < n_queue && nb < LANES
                       && round_tokens + queue[base + nb].n <= ROUND_BUDGET) {
                    round_tokens += queue[base + nb].n;
                    nb++;
                }
                if (nb == 0) { fprintf(stderr, "post exceeds round budget\n"); return 1; }
                // Clear EVERY lane's suffix, not just the nb lanes this
                // round reuses: a lane skipped by a smaller round keeps its
                // old suffix cells, and with long posts (residual pass:
                // ~120 tokens/lane) 96 stale lanes overflow the 8192-cell
                // cache ("failed to find a memory slot"). Clearing all
                // bounds live cells at prefix + ROUND_BUDGET.
                for (int k = 0; k < LANES; k++) {
                    llama_memory_seq_rm(mem, k, n_prefix, -1);
                }
                struct llama_batch b = llama_batch_init(round_tokens, 0, 1);
                b.n_tokens = round_tokens;
                int idx = 0, last_idx[LANES];
                for (int k = 0; k < nb; k++) {
                    queued_post *q = &queue[base + k];
                    for (int j = 0; j < q->n; j++) {
                        b.token[idx] = q->tok[j]; b.pos[idx] = n_prefix + j;
                        b.n_seq_id[idx] = 1; b.seq_id[idx][0] = k;
                        b.logits[idx] = (j == q->n - 1);
                        if (j == q->n - 1) last_idx[k] = idx;
                        idx++;
                    }
                }
                // Metal is async: llama_decode returns at submit and the
                // GPU wait lands in the first logits read, so the logits
                // reads stay inside the timed region.
                double t = now_ms();
                if (llama_decode(ctx, b)) { fprintf(stderr, "decode failed\n"); return 1; }
                float p1[LANES];
                for (int k = 0; k < nb; k++) {
                    const float *lg = llama_get_logits_ith(ctx, last_idx[k]);
                    p1[k] = 1.0f / (1.0f + expf(lg[t0] - lg[t1]));
                }
                ms += now_ms() - t;
                total_tokens += round_tokens;
                for (int k = 0; k < nb; k++) {
                    printf("SCORE %s %.6f\n", queue[base + k].id, p1[k]);
                }
                llama_batch_free(b);
                base += nb;
            }
            printf("DONE %d %ld %.1f\n", n_queue, total_tokens, ms);
            fflush(stdout);
            for (int i = 0; i < n_queue; i++) free(queue[i].tok);
            n_queue = 0;

        } else if (line[0] != '\n') {
            fprintf(stderr, "unknown command: %s", line);
            return 1;
        }
    }

    llama_free(ctx);
    llama_model_free(model);
    llama_backend_free();
    return 0;
}
```

zcbenz:
But that is llama.cpp code?

jstray:
It is not llama.cpp code, but it's calling into the llama.cpp API. Or are you saying this use case is not supported?

zcbenz:
Can you elaborate how the C++ code using llama.cpp API is related to mlx-lm?

PR #1747, opened 2026-08-17
Fix adapter_path lookup using aliased model_path in ModelProvider.load

### LINKED ISSUE #1745: server: --adapter-path is silently ignored (adapter map looked up after model name resolution)
## Bug

`mlx_lm server --model <path> --adapter-path <adapters>` never applies the adapter to any request, silently.

In `ModelProvider.load()` (current `main`, also reproduced on 0.31.3):

```python
    model_path = self._model_map.get(model_path, model_path)          # resolves "default_model" -> real path
    adapter_path = self._adapter_map.get(model_path, adapter_path)    # looks up the *resolved path*...
```

`self._adapter_map` is keyed by the literal string `"default_model"`:

```python
```

Since `model_path` has already been resolved to the real path on the previous line, the adapter lookup always misses, `adapter_path` stays `None`, and the model is served without the adapter — no warning, no error.

## Repro

1. Train any LoRA (we used a Qwen3-4B with a small factual dataset; train loss 0.05).
2. `python -m mlx_lm generate --model <base> --adapter-path <adapters> --prompt "<train question>"` → recites the learned fact ✅
3. In-process `load(<base>, adapter_path=<adapters>)` + `apply_chat_template` + `generate` → recites ✅
4. `python -m mlx_lm server --model <base> --adapter-path <adapters>` then POST `/v1/chat/completions` with `"model": "default_model"` (or the explicit base path) → base-model answer, adapter ignored ❌

## Suggested fix

Look up the adapter map **before** resolving the model name (or key the map by the resolved path at startup):

```python
    adapter_path = self._adapter_map.get(model_path, adapter_path)   # while it's still "default_model"
```

Happy to open a PR if useful.

Environment: mlx-lm 0.31.3, mlx 0.32.0, macOS 26.x, Apple Silicon (M3 Ultra).

mrjoema:
## Summary
- `ModelProvider.load` resolved `model_path` through `_model_map` before looking up `adapter_path` in `_adapter_map`, so adapters registered against a model alias (e.g. `default_model`) were never applied.
- Reorder so the adapter and draft-model lookups happen against the original `model_path` before it gets remapped.

Fixes #1745

## Test plan
- [x] Added `TestModelProvider.test_load_resolves_adapter_for_default_model` in `tests/test_server.py` covering the regression.
- [x] `python -m pytest tests/test_server.py`

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -367,9 +367,9 @@ class ModelProvider:
             self.load("default_model", None, "default_model")
 
     def load(self, model_path, adapter_path=None, draft_model_path=None):
-        model_path = self._model_map.get(model_path, model_path)
         adapter_path = self._adapter_map.get(model_path, adapter_path)
         draft_model_path = self._draft_model_map.get(draft_model_path, draft_model_path)
+        model_path = self._model_map.get(model_path, model_path)
 
         model_key = (model_path, adapter_path, draft_model_path)
         if self.model_key != model_key:
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,10 +1,12 @@
 # Copyright © 2024 Apple Inc.
 
+import argparse
 import http
 import io
 import json
 import threading
 import unittest
+from unittest.mock import patch
 
 import mlx.core as mx
 import requests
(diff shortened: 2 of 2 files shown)


zcbenz:
Fixed by https://github.com/ml-explore/mlx-lm/pull/1249.

Issue #1737, opened 2026-08-15
Distributed JACCL server crashes when two requests arrive concurrently


nachogala:
### Summary

With a distributed `mlx_lm.server` deployment using the JACCL backend, a
single request completes successfully, but two HTTP requests submitted at the
same time cause both distributed ranks to terminate. The clients receive an
empty reply (`curl` 52/HTTP 000).

### Environment

- macOS on two Apple Silicon hosts with unified memory
- `mlx==0.32.0`
- `mlx-metal==0.32.0`
- `mlx-lm==0.31.3`
- Python 3.11
- Distributed launch via `mlx.launch` with a two-host JACCL hostfile
- OpenAI-compatible HTTP server endpoint

### Reproduction

1. Start the distributed server with the JACCL backend and wait for
   `/v1/models` to return HTTP 200.
2. Submit one normal completion request; it returns HTTP 200.
3. Submit two otherwise valid completion requests concurrently (for example,
   two simultaneous `POST /v1/chat/completions` requests).
4. Repeat with both requests using the same generation mode. The failure is
   not specific to mixing reasoning/thinking modes.

### Expected behavior

The server should serialize or safely schedule concurrent requests, or return
an explicit overload response without terminating either distributed rank.

### Actual behavior

Both requests fail with an empty HTTP response. The distributed server logs
include messages equivalent to:

```
_pickle.UnpicklingError: invalid load key
[jaccl] Recv failed with error code -12
rank 0 exited with code 255
```

The peer rank exits as part of the same failure. After the crash, the server
endpoint is unavailable until restarted.

### Additional observations

- A single request succeeds immediately before the concurrent pair.
- The failure reproduces with `thinking=false` + `thinking=false`,
  `thinking=true` + `thinking=true`, and mixed modes.
- The failure occurs in both supervised and manually launched server modes.
- Serializing admission at the HTTP boundary avoids the concurrent crash, but
  that is only a workaround and removes parallel request handling.

Could the JACCL transport or distributed server path support concurrent
request admission? If concurrency is currently unsupported, an explicit
server-side queue/lock or a documented limitation would prevent a process-wide
crash.

nachogala:
Additional reproduction result: serialization of HTTP admission does not eliminate the distributed failure.

[...]

PR #1748, opened 2026-08-17
Fix distributed server crash ([jaccl] Recv failed -12) on per-request model switching

stepnoy:
## Problem

Running `mlx_lm.server` distributed over two nodes (`mlx.launch --backend jaccl`, tensor parallel), any request whose `model` field does not exactly match the loaded model kills the whole cluster:

```
libc++abi: terminating due to uncaught exception of type std::invalid_argument: [jaccl] Recv failed with error code -12
[WARN] Node with rank 0 exited with code 255
```

This is trivially triggered by well-behaved clients: `/v1/models` advertises `Path(cli_args.model).resolve()`, which is **rank-local** (e.g. `/Volumes/storage A/...` on node A vs `/Volumes/storage B/...` on node B), so a client that echoes the advertised id back sends a path that exists on one rank only. Any unknown model string (`"model": "x"`) triggers it too.

Related: #1737 — the `UnpicklingError` there looks like a sibling symptom of the same lockstep break.

## Root cause

A mismatching `model` field triggers `ModelProvider.load` → `_load` → `sharded_load` on every rank. The requested path may exist on one rank only: that rank fails fast and returns to its idle heartbeat `all_sum`s, while the other proceeds into load/generation collectives. The mismatched collectives pair up, receive work requests pile up, and jaccl aborts in `post_recv` (`jaccl/rdma.h`): `ibv_post_recv` returns **ENOMEM (12)**, surfaced as `Recv failed with error code -12`. Rank 0 can even log a 200 before the transport dies, so from the client side the response just disappears.

## Fix (two commits)

1. **Guard** — in distributed mode, once a model is loaded, a request for a different model raises a `ValueError` instead of reloading. The raise is symmetric on every rank (the request is broadcast before load), so the collective lockstep survives and the client receives a clean explanatory error.
2. **Rank-uniform model id** — `/v1/models` now advertises the CLI string itself (canonical, identical on all ranks), and the HTTP layer (rank 0, before the request is broadcast) normalizes aliases of the served model — the CLI string or any path resolving to it — to `default_model`.

Non-distributed serving is untouched, including dynamic model switching.

## Testing

2× Mac Studio M3 Ultra (Thunderbolt RDMA link, JACCL), Qwen3.8-27B 8-bit, mlx 0.32.0 / mlx-lm 0.31.3 (patch rebased onto current `main`):

- 8 sequential chat completions at temp=0 — stable (previously the cluster died on the first or second one)
- 4× at temp=0.7 — stable
- both former killers (`model` = the resolved rank-0 path, `model` = `"x"`) — clean error response, cluster stays alive, the next request succeeds
- clients echoing the `/v1/models` id (Open WebUI) now work; tool calls and reasoning unaffected; concurrent requests from two clients served without incident
- zero `-12` occurrences in the server log

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -373,6 +373,21 @@ class ModelProvider:
 
         model_key = (model_path, adapter_path, draft_model_path)
         if self.model_key != model_key:
+            if self.is_distributed and self.model_key is not None:
+                # Dynamic model switching is rank-divergent in distributed
+                # mode: the requested path may exist on one rank only, so one
+                # rank fails fast while the other enters sharded_load /
+                # generation collectives. Their all_sums then pair with the
+                # failed rank's idle heartbeats, receive work requests pile up
+                # and jaccl dies with `Recv failed with error code -12`
+                # (ENOMEM from ibv_post_recv). Raising here is symmetric on
+                # every rank (the request is broadcast before load), so the
+                # lockstep survives and the client gets a clean error.
+                raise ValueError(
+                    "Distributed server cannot switch models per request; "
+                    f"serving {self.model_key[0]!r}. Request that exact id "
+                    "or 'default_model'."
+                )
             self._load(*model_key)
 
         return self.model, self.tokenizer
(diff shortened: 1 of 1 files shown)


zcbenz:

PR #1752, opened 2026-08-18
Declare MLX norm convention in converted models

### LINKED ISSUE #1445: Record RMSNorm (1+weight) storage state in converted metadata so runtimes don't have to guess
### Summary

Several architectures (Gemma, Qwen3.5, Qwen3-Next, …) store RMSNorm weights as the
**deviation from 1** and apply `(1 + weight)` at load. The catch: a *converted* bundle
can legitimately be in **either** storage state — norms stored **raw** (deviation,
mean ≈ 0) or **already shifted** (mean ≈ 1) — and **nothing in the converted files
records which**. So every runtime has to *guess* the state from weight statistics, and
that guess has repeatedly shipped as a silent correctness bug.

This is a request to have mlx-lm's converter **record the norm storage state
explicitly** so downstream runtimes can stop guessing.

### Why guessing fails (two concrete, shipped examples)

1. **Gemma `(1 + weight)` misapplication** — the unit-offset convention is a documented
   Gemma architecture feature, and applying/omitting the `+1` incorrectly has surfaced as
   a real bug before (e.g. ml-explore/mlx-examples#1057). Whether a given converted bundle
   already has the offset baked in is not recorded anywhere.

2. **Qwen3.5 intermittent degenerate loads** — in the Swift runtime (vmlx-swift), the
   heuristic decided whether to apply `+1` by inspecting the **first** norm weight it
   happened to iterate. Swift randomizes `Dictionary` order per process, and a few norm
   weights in a raw bundle legitimately have mean > 0.5, so **~1 in 13 loads of the same
   bundle silently loaded a degenerate model** (flat logits, garbage output) — no error,
   no log. Fix + full write-up: osaurus-ai/vmlx-swift#102. The corrected heuristic is an
   order-independent majority vote, which is robust — **but it is still a heuristic
   substituting for information the converter could simply have stated.**

Both bugs have the same root: the storage state is derived from the weights instead of
declared, and any single- or few-sample derivation is fragile because the signal is noisy
and the same architecture ships in *both* states.

### The underlying gap

- safetensors' `__metadata__` is a free-form `string → string` header and mandates **no**
  semantic keys.
- `config.json` has **no** field for norm storage state.

So there is no declared source of truth, and each runtime re-derives it independently
(and sometimes wrongly).

### Proposal

Have the mlx-lm converter emit a small, explicit declaration of the norm storage state
— in the safetensors `__metadata__` header and/or `config.json` — and have the readers
**honor it when present**, falling back to a heuristic only for legacy bundles that
predate it. Concretely, something like one of:

```jsonc
// config.json  (or safetensors __metadata__)
"rmsnorm_weight_offset": "unit"   // norms stored as deviation-from-1, apply (1 + w)
// vs
"rmsnorm_weight_offset": "zero"   // norms stored already-shifted, apply as-is
```

The exact key/name is yours to bless — the ask is that **one canonical, documented key
exists** and that the converter populates it, so “declare it, don’t re-derive it” becomes
possible across the ecosystem (mlx-lm, mlx-vlm, mlx-swift/vmlx, and any third-party
converter). vmlx already reads such a key (`norm_convention`) the moment a bundle
provides one; today none do.

### Offer

Happy to send the converter-side PR (populate the key on convert/quantize) if there's
agreement on the key name and semantics.

sfeng01-tech:
Fixes #1445

## Summary

- declare `norm_convention: "mlx"` in converted `config.json` files and safetensors metadata
- add a regression test that verifies both declarations

## Why

Converted MLX weights are already in MLX runtime form. Declaring that storage convention lets downstream runtimes avoid guessing whether an architecture-specific `1 + weight` RMSNorm adjustment is still required.

## Validation

```text
python -m unittest tests.test_utils.TestUtils.test_make_shards \
  tests.test_utils.TestUtils.test_save_declares_mlx_norm_convention
```

Result: 2 tests passed.

## Limitations

- Existing converted bundles are unchanged. They need to be reconverted or have equivalent metadata backfilled to carry this declaration.
- Downstream runtimes must honor the `norm_convention` field to benefit from it.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -57,6 +57,7 @@ MODEL_REMAPPING = {
 }
 
 MAX_FILE_SIZE_GB = 5
+MLX_NORM_CONVENTION = "mlx"
 
 
 def _parse_size(x):
@@ -797,7 +798,15 @@ def save_model(
         shard_name = shard_file_format.format(i + 1, shards_count)
         shard_path = save_path / shard_name
 
-        mx.save_safetensors(str(shard_path), shard, metadata={"format": "mlx"})
+        mx.save_safetensors(
+            str(shard_path),
+            shard,
+            metadata={
+                "format": "mlx",
+                # Norm weights are already in MLX runtime form.
+                "norm_convention": MLX_NORM_CONVENTION,
+            },
+        )
 
         for weight_name in shard.keys():
             index_data["weight_map"][weight_name] = shard_name
@@ -981,6 +990,9 @@ def save(
         src_path = hf_repo_to_path(hf_repo)
     else:
         hf_repo = None
+
+    # Converted weights are already in MLX runtime form.
+    config["norm_convention"] = MLX_NORM_CONVENTION
 
     dst_path = Path(dst_path)
     save_model(dst_path, model, donate_model=donate_model)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -2,9 +2,11 @@
 
 import json
 import os
+import struct
 import tempfile
 import unittest
 from pathlib import Path
+from unittest import mock
 
 import mlx.core as mx
 import mlx.nn as nn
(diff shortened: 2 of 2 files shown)


zcbenz:

PR #1755, opened 2026-08-18
Fix RWKV7 legacy checkpoint loading

sfeng01-tech:
## Summary

Fix loading legacy RWKV-7 FLA checkpoints that store time-mix parameters in a
single `model.layers.*.attn.x_x` tensor.

`mlx_lm.models.rwkv7` expects the current six-parameter layout:
`x_r`, `x_w`, `x_k`, `x_v`, `x_a`, and `x_g`. The loader now splits the legacy
`[6, hidden_size]` tensor into these parameters and removes the legacy key
before strict weight loading.

## Validation

- Added a regression test for legacy `attn.x_x` migration.
- Added a batched cached-generation test asserting chunked generation matches a
  full forward pass.
- End-to-end validation with `RWKV/RWKV7-Goose-World3-1.5B-HF` on a
  three-token prompt loaded successfully and produced `(1, 3, 50304)` bfloat16
  logits.
- Compared those logits against an independent PyTorch reference implementation
  of the checkpoint equations: greedy argmax matched at every prompt position;
  mean absolute error was `0.0461` and maximum absolute error was `0.3125`.
  The remaining drift is consistent with bfloat16/fused-kernel accumulation.

```text
python -m unittest \
  tests.test_models.TestModels.test_rwkv7_cached_generation_matches_full_prompt \
  tests.test_models.TestModels.test_rwkv7_sanitize_migrates_legacy_time_mix_weights \
  tests.test_models.TestModels.test_all_models
```

All tests pass.

### DIFF
--- a/mlx_lm/models/rwkv7.py
+++ b/mlx_lm/models/rwkv7.py
@@ -436,7 +436,14 @@ class Model(nn.Module):
         return self.model.layers
 
     def sanitize(self, weights):
-        for k, v in weights.items():
+        for k, v in list(weights.items()):
+            if k.endswith(".attn.x_x"):
+                prefix = k.removesuffix("x_x")
+                for name, value in zip(
+                    ("x_r", "x_w", "x_k", "x_v", "x_a", "x_g"), v
+                ):
+                    weights[f"{prefix}{name}"] = value[None, None, :]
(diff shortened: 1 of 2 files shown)


zcbenz:

PR #1760, opened 2026-08-19
Add DeepSeek-V4 DSML chat template + tolerate truncated invoke close tag (stacked on #1337)

nh13:
**Stacked on #1337** (@snagnever's `deepseek_dsml` parser) — that should land first; the commits unique to this PR are the two on top. Also depends on DeepSeek-V4 model support (#1189).

Adds the encode side plus a robustness fix #1337's parser needs in practice.

**1. `chat_templates/deepseek_dsml.py` — DSML chat template.** DeepSeek-V4 ships no Jinja template (only the reference Python `encoding_dsv4.py`), so there's nothing for `chat_template_type` to load and tools are never rendered into the prompt. This is a Python `apply_chat_template` reusing the existing `deepseek_v32` DSML helpers; only the V4-specific tool wording and `tool_calls` block tag live here. Byte-exact with the reference encoder for single-turn system+tools+user prompts (chat and thinking modes) and assistant tool-call replay. It maps the server's `enable_thinking` kwarg onto the reference's `thinking_mode`, and normalizes a top-level `tools=` kwarg onto a system message (synthesizing an empty one if absent) so tools aren't silently dropped when no system message is present. Enable with `chat_template_type: "deepseek_dsml"`.

**2. Lenient invoke close tag.** The 4/5/6-bit MLX quants systematically truncate `</｜DSML｜invoke>` to `</｜DSML｜inv>`; the strict `_INVOKE` close misses those calls (they come back as content). Widen it to `</｜DSML｜inv[^>]*>`.

**Tests:** byte-exact chat-template test (skips when the model's `encoding/` dir isn't cached) + a close-tag regression test. No model download required.

**Marker note:** #1501 (merged) replaced the token-based state machine with a text-based one, so the trailing-`>` marker workaround this branch inherits is no longer needed on `main` — the markers can return to full form once this rebases forward past #1189.

### DIFF
--- a/mlx_lm/chat_templates/deepseek_dsml.py
+++ b/mlx_lm/chat_templates/deepseek_dsml.py
@@ -81,6 +81,18 @@ tool_calls_template = "<{dsml_token}tool_calls>\n{tool_calls}\n</{dsml_token}too
 tool_output_template = "<tool_result>{content}</tool_result>"
 
 
+def _tool_call_with_str_args(tc: Dict[str, Any]) -> Dict[str, Any]:
+    """``encode_arguments_to_dsml`` ``json.loads()``es ``tc["arguments"]``, i.e.
+    it expects the OpenAI JSON *string* form. Some clients (e.g. agent loops
+    replaying their own prior calls) send ``arguments`` already parsed as a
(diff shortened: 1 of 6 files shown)


zcbenz:

PR #1762, opened 2026-08-19
Support 'sqrtsoftplus' moe routing activation for future Laguna Releases

tzachicohen:
This patch introduce support for future Laguna releases. 
It enables an additional MOE routing score activation function - sqrtsoftplus. It is chosen based on 'config.json' definition.
For backwards compatibility, if no definition is found in 'config.json', sigmoid will be chosen. 

Testing done:

- Linting tests.
- Unit tests. 
- Manual tests with un-released versions of Laguna.


zcbenz (approved):
Thanks!

PR #1764, opened 2026-08-19
Reject dynamic quantization targets at the BPW floor

### LINKED ISSUE #1758: dynamic_quant silently returns uniform quantization when target-bpw is at the floor of the default bit settings
I hit this while building matched-size quantized models for a comparison study. Three differently calibrated dynamic_quant runs at the same 4.5 target came out identical to each other, and all but identical to plain uniform 4-bit. I assumed I was doing something wrong, looks to be the arithmetic.

With the defaults (`--low-bits 4 --high-bits 5`, group size 64), a 4-bit layer already costs 4.5 bpw once you count each group's 16-bit scale and bias. So 4.5 is the exact floor of what those settings can express. Ask for it and the threshold search has nothing to decide: everything lands at low bits. The full calibration pass still runs first, and the output looks like success. There's no warning, and the help text for `--target-bpw` doesn't mention a valid range.

## What I expected

A mixed-precision allocation at 4.5 bpw, or failing that, the tool telling me these settings can't produce one.

## What I received

1 layer promoted out of ~200, after the full calibration pass, with a normal success message. Same on Llama-3.2-3B: 1 of 197 layers, with stock, matched-stock, and custom sensitivity files all producing identical models.

As a control, the same command with `--target-bpw 4.6` promotes 38 layers, so allocation works fine one notch above the floor. The cliff is exactly where the arithmetic says.

## Reproduce (defaults, default model)

```bash
python -m mlx_lm.quant.dynamic_quant --model Qwen/Qwen3-0.6B-base \  --mlx-path ./out_45 --target-bpw 4.5 --seed 123
python - <<'PY'
import json
q = json.load(open("out_45/config.json"))["quantization"]
per_layer = {k: v for k, v in q.items() if isinstance(v, dict) and "bits" in v}
print("layers promoted to high bits:", sum(1 for v in per_layer.values() if v["bits"] == 5))
PY
```

I get 1 layer promoted (out of ~200). Same on Llama-3.2-3B: 1 of 197 layers, with stock, matched-stock, and custom sensitivity files all producing identical models.

As a control, the same command with `--target-bpw 4.6` promotes 38 layers, so allocation works fine one notch above the floor. The cliff is exactly where the arithmetic says.

## Suggested fix

The floor is computable from the arguments alone, before the expensive calibration pass:

```python
# each group carries a 16-bit scale and 16-bit bias => 32/group_size bpw overhead
floor_bpw = args.low_bits + 32 / args.low_group_size
if args.target_bpw <= floor_bpw:
    raise ValueError(
        f"--target-bpw {args.target_bpw} is at or below the minimum "
        f"({floor_bpw:.4g}) expressible with --low-bits {args.low_bits} at "
        f"group size {args.low_group_size}; the result would be uniform "
        f"{args.low_bits}-bit quantization. Lower --low-bits or raise the target."
    )
```

A warning instead of an error also works if a uniform result is acceptable output. A post-search warning when zero layers get promoted would catch the general case, and a sentence in the `--target-bpw` help about the valid range would help too.

## Workaround

Set `--low-bits` below the default (e.g. `--low-bits 3`) so the floor drops and the allocation has room to decide, or keep targets comfortably above `low_bits + 32/group_size`. Either way you have to derive the floor yourself, which is what this issue is about.

Environment: mlx_lm 0.31.3 (current PyPI release), mlx 0.32.0 and 0.32.1, macOS 26.5.2, M4 Max. I checked main's dynamic_quant.py and it only differs from 0.31.3 by the trust-remote-code change from #1385, so this is current.

original4422:
## Summary

- reject `--target-bpw` values at or below the low-precision layout floor before distributed initialization, model loading, or calibration
- explain the computed lower bound in CLI help and the dynamic quantization guide
- cover the default and a custom low-bit/group-size configuration with a regression test

Fixes #1758.

## Root cause

Each quantized group stores a 16-bit scale and a 16-bit bias in addition to the packed weights. The low-precision floor is therefore:

```text
low_bits + 32 / low_group_size
```

With the defaults this is exactly 4.5 BPW. A target at or below that value leaves no room for a mixed-precision allocation, but the CLI previously completed model loading and calibration before silently producing an effectively uniform low-bit result.

The new check runs immediately after argument parsing and suggests lowering `--low-bits` or raising the target.

## Validation

- `python -m unittest tests.test_dynamic_quant -v`
- `pre-commit run --files mlx_lm/quant/dynamic_quant.py tests/test_dynamic_quant.py mlx_lm/LEARNED_QUANTS.md`
- CI-equivalent local suite on Apple Silicon with MLX 0.32.1 and the official test data: 213 tests passed, 1 skipped
- `mlx.launch --python .venv/bin/python -n 2 tests/model_parallel_tests.py`: passed on both ranks

### DIFF
--- a/mlx_lm/LEARNED_QUANTS.md
+++ b/mlx_lm/LEARNED_QUANTS.md
@@ -93,7 +93,8 @@ Some important options are:
 
 - `--target-bpw`: The target bits-per-weight. For a given set of quantization
   parameters only certain ranges are possible. For example, with the default
-  parameters a BPW in the range `[4.5, 5.5]` is achievable.
+  parameters the target must be greater than 4.5 BPW, since each 4-bit group
+  also stores a 16-bit scale and bias. Targets up to 5.5 BPW are achievable.
 - `--sensitivities`: A path to a precomputed sensitivities file.
 - `--low-bits`: The number of bits to use for the less sensitive layers.
 - `--high-bits`: The number of bits to use for the more sensitive layers.
--- a/mlx_lm/quant/dynamic_quant.py
+++ b/mlx_lm/quant/dynamic_quant.py
@@ -160,7 +160,13 @@ def main():
         help="Path to a pre-computed sensitivity JSON file.",
     )
     parser.add_argument(
-        "--target-bpw", type=float, default=5.0, help="Target bits per weight."
+        "--target-bpw",
+        type=float,
+        default=5.0,
+        help=(
+            "Target bits per weight. Must be greater than low-bits + "
+            "32 / low-group-size (4.5 with the defaults)."
+        ),
     )
     parser.add_argument("--low-bits", type=int, default=4)
     parser.add_argument("--low-group-size", type=int, default=64)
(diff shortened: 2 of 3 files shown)


zcbenz:

PR #1765, opened 2026-08-19
docs(dynamic_quant): document nondeterministic sensitivity computation

### LINKED ISSUE #1759: dynamic_quant produces different models across runs with identical arguments and a fixed --seed
While building replicate arms for a quantization comparison I noticed that running dynamic_quant twice with the same arguments and the same `--seed` gives back two different models. Different bit allocations, different weights, different measured bpw.

## Reproduce

```bash
python -m mlx_lm.quant.dynamic_quant --model Qwen/Qwen3-0.6B-base --mlx-path ./run_a --target-bpw 5.0 --seed 123
python -m mlx_lm.quant.dynamic_quant --model Qwen/Qwen3-0.6B-base --mlx-path ./run_b --target-bpw 5.0 --seed 123
shasum -a 256 run_a/model.safetensors run_b/model.safetensors
```

## What I expected

The same command with the same `--seed` producing the same model, every time.

## What I received

On my machine (M4 Max, macOS 26.5.2, mlx 0.32.0, mlx_lm 0.31.3, the current PyPI release), five identical invocations gave five distinct models: promoted-layer counts of 132, 134, 135, 136 and 138, reported bpw from 4.992 to 5.013, five different weight hashes. Another same-seed pair under mlx 0.32.1 also differed (135 vs 134 layers promoted). A same-seed pair on Qwen2.5-1.5B-Instruct drifted further: 121 vs 132 layers promoted, 4.986 vs 5.062 bpw. On Llama-3.2-3B at low=3/high=5 I've also seen three runs at a 4.5006 target come out at 4.4745, 4.5352 and 4.5195. Three different model families, all with the same behavior.

I checked whether it was my environment: plain uniform quantization (`mlx_lm.convert -q --q-bits 4 --q-group-size 64`) run twice produces byte-identical outputs on the same machine, so the nondeterminism is specific to the calibrated path.

I also tried to isolate where it enters. With the same seed, the drawn calibration data is byte-identical between runs, so the seed is doing its job on sampling. But the sensitivities computed over that identical data differ on all 197 layers of Qwen3-0.6B (median relative difference about 3%, max 45%). Going one level deeper: a single gradient computation on one identical batch, run twice back-to-back in the same process, already returns slightly different values. So the nondeterminism is in the kernel-level reductions rather than anything in this repo's accumulation loop, and it compounds from there: the sensitivity scores wobble around 3%, and the binary-searched threshold lands somewhere different each time. 

For what it's worth, the chunk permutation in load_data is the only random operation in the path, and main() seeds it explicitly right before each use - so I don't think any of this variation is intended as far as I can tell.

## Suggested fix

Two options, smaller one first:

1. A sentence in the `--seed` help saying it fixes the calibration sample draw but not the output. That alone turns a surprise into documented behavior.

2. A real deterministic mode, though since single-gradient computations already differ, that likely needs deterministic reductions in mlx core rather than anything here. Anyone comparing quantization methods would use it, since right now the run-to-run spread has to be treated as a variance component before a between-method difference means anything.

## Workaround

For comparisons: build replicate runs and treat the run-to-run spread as a variance component, so a between-method difference has to clear the tool's own noise. If exact size matching matters, skip the threshold search entirely and spend a fixed bit budget over the sensitivity ranking (rank layers, promote in order until the budget is spent). That sidesteps both the nondeterministic threshold and the size miss.

axiom-of-choice:
Fixes #1759

Body:
What

The --seed flag fixes the calibration sample draw, but gradient reductions in Metal are nondeterministic, so sensitivity scores and the resulting bit allocation can vary between runs even with the same seed. Three model families exhibit this (Qwen3-0.6B, Qwen2.5-1.5B, Llama-3.2-3B per the issue).

This PR documents the behavior in two places:

1. --seed help text — explains what the seed controls and doesn't control, and suggests replicate runs for method comparisons.
2. estimate_sensitivities docstring — notes the source of nondeterminism (Metal gradient reductions) and the observed variance (~3% median, ~45% max).

Why documentation and not a code fix

The nondeterminism enters at the kernel level: a single gradient computation on one identical batch, run twice in the same process, already returns slightly different values. This is the same class of nondeterminism as CUDA's non-deterministic atomicAdd. Fixing it requires deterministic reductions in mlx core, not in this repo.

The issue's suggestion #1 (document the behavior) turns a surprise into documented behavior. The suggested workaround (replicate runs, treat spread as variance) is included in the help text.

Testing

No behavioral change — only documentation. python -m mlx_lm.quant.dynamic_quant --help renders the new seed description correctly.

Fixes #1759.

### DIFF
--- a/mlx_lm/quant/dynamic_quant.py
+++ b/mlx_lm/quant/dynamic_quant.py
@@ -46,6 +46,15 @@ def estimate_sensitivities(
     gradient_accum_dtype: mx.Dtype = mx.float32,
     gradient_checkpoint: bool = False,
 ):
+    """Estimate per-layer sensitivity to quantization.
+
+    Sensitivity = alignment between the gradient and (low_q - high_q) weight
+    difference, averaged over calibration batches.  The gradient computation
+    uses nondeterministic Metal reductions, so repeated calls with identical
+    inputs can produce slightly different values (median ~3%, max ~45%
+    relative difference observed).  The resulting bit allocation is therefore
+    not bit-reproducible across runs even with the same seed.
+    """
     def qdq(w, bits, group_size):
         w, s, b = mx.quantize(w, bits=bits, group_size=group_size)
         return mx.dequantize(w, scales=s, biases=b, bits=bits, group_size=group_size)
@@ -152,7 +161,19 @@ def main():
     parser.add_argument(
         "--mlx-path", default="mlx_model", help="Path to save the model"
     )
-    parser.add_argument("--seed", type=int, default=123)
+    parser.add_argument(
+        "--seed",
+        type=int,
+        default=123,
+        help=(
+            "Seed for the calibration sample draw. Fixes which data are"
+            " used but does not guarantee identical output: gradient"
+            " reductions in Metal are nondeterministic, so sensitivity"
+            " scores and the resulting bit allocation can vary between"
+            " runs. For method comparisons, build replicate runs and"
+            " treat the spread as a variance component."
+        ),
+    )
     parser.add_argument(
         "--sensitivities",
         type=str,


zcbenz:

Issue #1699, opened 2026-08-09
gemma3n broken with batch KV caches: cache.state unpack crash + RoPE offset mutated in place across update_and_fetch


Cyb3rb1ade:
## Summary

Running a gemma3n text model through the batched generation path (`BatchKVCache` / `BatchRotatingKVCache`, i.e. `BatchGenerator` and any server built on it) fails in two independent ways:

1. **Crash** — `ValueError: too many values to unpack (expected 2, got 4)` in `Gemma3nAttention.__call__` at `keys, values = cache.state`, because batch caches return a 4-tuple `(k, v, offset, left_padding)` while plain `KVCache`/`RotatingKVCache` return 2-tuples.
2. **Degenerate output** — after working around (1), generation produces repetition loops (e.g. `HierHierHierHier...`) because the RoPE offset used for the queries is silently mutated by the cache update (details below). The same model + prompt generates correctly with plain caches.

## Environment

- mlx-lm 0.31.3 (latest release), mlx 0.32.0
- macOS, Apple Silicon
- Model: any gemma3n text conversion, e.g. `mlx-community/Huihui-gemma-3n-E4B-it-abliterated-lm-8bit`
- Code in `main` (`mlx_lm/models/gemma3n.py`) is identical as of 2026-08-09.

## Reproduction

```python
from mlx_lm import load
from mlx_lm.generate import generate
from mlx_lm.models.cache import (
    BatchKVCache, BatchRotatingKVCache, KVCache, RotatingKVCache, CacheList,
)

model, tok = load("mlx-community/Huihui-gemma-3n-E4B-it-abliterated-lm-8bit")
prompt = tok.apply_chat_template(
    [{"role": "user", "content": "What is the capital of France?"}],
    tokenize=False, add_generation_prompt=True)

def to_batch(c):
    if type(c) is KVCache:
        return BatchKVCache([0])
    if isinstance(c, RotatingKVCache):
        return BatchRotatingKVCache(c.max_size, [0])
    if isinstance(c, CacheList):
        return CacheList(*(to_batch(x) for x in c.caches))
    return c

# Plain caches: works -> "The capital of France is Paris."
print(generate(model, tok, prompt=prompt, max_tokens=48,
               verbose=False, prompt_cache=model.make_cache()))

# Batch caches (what _make_cache() produces for the batched engine):
batch = [to_batch(c) for c in model.make_cache()]
print(generate(model, tok, prompt=prompt, max_tokens=48,
[...]

Issue #1693, opened 2026-08-09
Generation failures are reported as HTTP 404 and malformed config is ignored


Thump604:
### Problem

The server maps arbitrary generation exceptions to HTTP 404, which incorrectly presents runtime/input failures as missing resources. Separately, malformed optional generation configuration can be silently ignored.

### Expected behavior

Missing resources should remain 404, invalid request/configuration data should produce a client error, and unexpected generation failures should surface as server errors without being mislabeled.

### Proposed fix

Draft PR #1685 preserves meaningful HTTP exceptions, reports malformed optional configuration, and maps unexpected generation failures accurately, with focused server regressions.

PR #1766, opened 2026-08-19
Fix 18 server/config/packaging/model issues (#1081, #1653, #1670, #1691, #1693, #1695, #1148, #1133, #1248, #229, #1649, #1732, #1692, #1563, #1699, #118, #1183, #1627/#1236/#1604)

amanyagami:
## Summary

A batch of small, independent fixes for open issues that don't depend on
model numerics/GPU access — CLI/config parsing, error handling, HTTP
status mapping, auth, packaging/typing gaps, cache handling, and tool-call
parsing. Each item below is a self-contained change; happy to split into
separate PRs if that's preferred for review.

## Fixes

- **#1081** — `_BaseCache` now has a default `trim()` that raises
  `NotImplementedError` instead of `AttributeError`, so
  `is_trimmable() == True` without a `trim()` override fails loudly at
  the point of the bug instead of several frames away.
- **#1653** — `mlx_lm.chat` now imports `readline` (guarded for
  platforms without it), which makes `corridor_input`'s `\x01`/`\x02`
  markers actually get consumed instead of leaking to the terminal, and
  adds persistent chat history (Up/Down) across sessions.
- **#1670** — A `lora_parameters` config missing `scale` now raises a
  specific, actionable `ValueError` instead of a bare `KeyError`,
  including an alpha→scale conversion hint for configs written for
  PEFT's alpha/rank convention.
- **#1691** — Importing OLMo without `ai2-olmo` installed now raises a
  catchable `ImportError` instead of calling `sys.exit(1)`, which
  otherwise kills the whole host process (server, notebook, etc.).
- **#1693** — `mlx_lm.server` generation failures now map to the
  correct HTTP status: `FileNotFoundError` → 404, `ValueError` /
  `TypeError` / `KeyError` → 400, anything else → 500 (previously
  everything was reported as 404). `chat_template_kwargs` is now also
  type-validated.
- **#1695** — Adds an optional `--api-key` bearer-token flag. Binding
  to a non-loopback host now requires `--api-key` at startup, and the
  wildcard CORS default is no longer used for an authenticated
  deployment.
- **#1148** — On a failure partway through generation, the server now
  terminates SSE streams with an error event + `[DONE]` and
  force-closes non-streaming connections, instead of a client
  potentially waiting on a connection that will never get more bytes.
- **#1133** — A loaded model's basename is registered as an alias, so
  a short model name in a request resolves to the already-loaded model
  instead of falling through to a HuggingFace Hub lookup.
- **#1248 / #229** — Fixes `ModelProvider.load()`'s adapter
  resolution, which had two bugs: it looked up `_adapter_map` using the
  already-rewritten `model_path` (so `--adapter-path` was silently
  dropped for the default model), and it let a stale `default_model`
  map entry silently override an explicit per-request `adapters` value.
- **#1649** — `_download()` now tries
  `snapshot_download(local_files_only=True)` first and only falls back
  to the network on a cache miss, avoiding a Hub API round-trip on
  every single `load()` call for an already-cached model (~300ms → <1ms
  on a warm cache, per the measurements in the issue).
- **#1732** — Adds `@overload` signatures to `load()` so
  `model, tokenizer = load(...)` type-checks cleanly under
  `mypy --strict` / pyrefly instead of requiring a 3-tuple unpack.
- **#1692** — Raises `python_requires` to `>=3.10` to match
  `transformers>=5.7.0`'s actual floor, and declares `huggingface-hub`,
  `regex`, and `tqdm` as direct `install_requires` since several
  modules (including some console-script entry points like
  `mlx_lm.lora`, `mlx_lm.share`, the `quant/*` scripts) import them
  directly rather than only transitively.
- **#1563** — `qwen3.py`/`qwen3_moe.py`: transformers ≥ 5 nests rope
  config under a `rope_parameters` dict instead of top-level
  `rope_theta`/`rope_scaling` (confirmed against actual
  `transformers==5.15.1` — `Qwen3Config().to_dict()` has no top-level
  `rope_theta` at all anymore). Both `ModelArgs` had `rope_theta` as a
  required field with no fallback, so any newly-published Qwen3(-MoE)
  config in the new shape crashed with `TypeError`. Added
  `__post_init__` lifting, matching the pattern already used by
  `qwen3_5.py`.
- **#1699** — `gemma3n.py`: `Gemma3nAttention` had two bugs on the
  batched-cache path (`BatchKVCache`/`BatchRotatingKVCache`): (1)
  `keys, values = cache.state` crashes on a batch cache's 4-tuple
  state; (2) the queries' RoPE offset was bound *before*
  `update_and_fetch` but applied *after* — batch caches keep `offset`
  as an `mx.array` mutated in place by the update, so the offset
  silently became the post-update value, shifting every query position
  by the prompt length and producing repetition-loop garbage output.
  Fixed by slicing `cache.state[:2]` and moving the queries' RoPE
  application before `update_and_fetch`. Added two regression tests
  using a real (tiny) gemma3n model — one asserting batch-cache and
  plain-cache logits match bit-for-bit (verified to fail with both the
  crash and, after a partial fix, silently-wrong logits before the
  full fix).
- **#118** — `cache.py`/`gemma3_text.py`: `make_prompt_cache`'s generic
  `max_kv_size` fallback only applies when a model has no custom
  `make_cache()` — every sliding-window model (gemma3, gemma3n, gemma4,
  cohere2, ...) defines its own with no parameters, so `--max-kv-size`
  was silently ignored for all of them, and Gemma 3's `RotatingKVCache`
  became permanently untrimmable past its hardcoded `sliding_window`
  (1024) regardless of what was requested. `make_prompt_cache` now
  forwards `max_kv_size` to `model.make_cache()` when its signature
  accepts it (checked via `inspect.signature`, so the ~40 other models
  are completely unaffected); `gemma3_text.Model.make_cache` now caps
  its sliding layers at `min(sliding_window, max_kv_size)`.
- **#1183** — Adds `context_length` to each entry in the `/v1/models`
  response (from `config.json`'s `max_position_embeddings`, falling
  back to a nested `text_config` for Qwen3.5/3.6-style configs), so
  clients can size `max_tokens` without hardcoding per-model defaults.
- **#1627 / #1236 / #1604** — `tool_parsers/qwen3_coder.py`:
  `_convert_param_value` had three related crashes/silent-drops. A
  float-formatted integer arg (e.g. `"140.0"` for an `integer` schema
  field, which the model emits in practice) raised `ValueError` from a
  bare `int(param_value)` — combined with the server silently dropping
  failed tool calls, this produced an infinite generate-parse-fail-retry
  loop against a fully prompt-cached, deterministic request (observed
  ~390 iterations in the reporter's case). Separately, `ast.literal_eval`
  on an object/array/untyped parameter crashed with
  `SyntaxError`/`ValueError` — unhandled, propagating out of the HTTP
  request handler — on any value that isn't valid JSON *and* isn't a
  valid Python literal (an ISO 8601 timestamp, a large free-text/code
  argument, truncated JSON). Both paths (and the float branch, latent
  but the same shape) now fall back to the raw string on failure
  instead of raising.

## Not included

A handful of issues in this sweep turned out to need no code change —
either already fixed on `main`, or don't reproduce. Each was verified
against real artifacts (live model downloads, the actual installed
`transformers` version, actual tokenizer files) rather than assumed:

- **#1128**, **#1500** — already fixed on `main`.
- **#939** (RoPE mutation) — fixed by #1003; the existing
  `test_su_scaled_rope_no_mutation`/`test_yarn_rope_no_mutation` tests
  already cover it.
- **#983** (`BatchRotatingKVCache.merge` shape mismatch) — fixed by #999.
- **#1041** (Devstral tekken tokenizer emits literal `Ġ`) — downloaded
  the actual tokenizer.json from the named model; the detokenizer
  correctly converts byte-level BPE tokens to spaces when driven the
  way `generate.py`/`server.py` actually drive it. The reported repro
  only reproduces if the caller re-fetches `TokenizerWrapper.detokenizer`
  (a property that constructs a fresh instance every access) per
  token instead of holding one instance — not something mlx-lm's own
  code does.
- **#1257** (XTC sampling crash on gemma-4-e4b) — downloaded and ran
  the full named model with the issue's exact XTC request; generates
  correctly. Fixed by #1575 (per-row XTC threshold for batched
  logits), landed after this issue was filed.
- **#1126** (Gemma 4 tool-call regex on unbalanced braces in strings)
  — fixed by #1150; verified against both of the issue's exact failing
  inputs.
- **#1131**'s reported log string doesn't exist anywhere in this repo
  — it belongs to an unrelated third-party wrapper.
- **#972** is unrelated bot-generated content.

## Testing

Ran the real test suite (Linux x86_64 + `mlx[cpu]`, real model
downloads from the Hub — not mocked). **Full suite: 0 failures.**

- `tests/test_server.py`: 35/35 passed (includes a live end-to-end
  check of the new `context_length` field against a real downloaded
  model, and the `--api-key` auth checks)
- `tests/test_utils.py`, `test_tuner_utils.py`, `test_prompt_cache.py`,
  `test_tokenizers.py`, `test_chat.py`: 43 passed, 6 subtests passed
- `tests/test_models.py`: 91/91 passed, 3 skipped (Metal-only kernel
  tests — see below)
- `tests/test_gguf.py`, `test_generate.py`, `test_finetune.py`: 42/42
  passed
- `tests/test_tool_parsing.py`, `test_sample_utils.py`: all passing,
  including new regression tests for the qwen3_coder parser fixes

Along the way this surfaced and fixed several things beyond the issue
list itself:

1. **A real bug in this PR's own `--api-key` change**: `_check_auth`
   crashed with `AttributeError` on any `cli_args` object lacking that
   attribute — exactly what `test_server.py`'s test double does, and
   what any library caller that builds `ModelProvider` by hand would
   hit too. Fixed with `getattr(..., default=None)`.
2. **Three pre-existing test failures**, confirmed to reproduce
   identically on unmodified `main` (a clean worktree at the same base
   commit) before being touched, fixed rather than left alone:
   - `test_bitnet`: `BitLinear.execute_matmul_kernel` unconditionally
     called a Metal-only custom kernel with no fallback, so any
     BitNet-style model hard-crashes off Metal. Added a plain-ops
     reimplementation of the packed-ternary matmul, verified against a
     hand-computed example.
   - `test_gated_delta` / `test_gated_delta_masked`: call
     `gated_delta_kernel`, a genuinely Metal-only custom kernel with
     no portable equivalent (returns `None` off Metal, by design — the
     ops fallback is covered separately by
     `test_gated_delta_precision`, which passes on CPU). Added
     `@unittest.skipUnless(mx.metal.is_available(), ...)`.
   - `test_cache_to_quantized`: `rtol=4e-2` over ~152k logits was tight
     enough that the CPU backend's int8-quantized-cache accumulation
     order pushes 2 outlier values (max observed ~4.97e-2) just over
     the line, with the predicted token unaffected — quantization
     rounding noise, not a correctness bug. Bumped to `rtol=6e-2` with
     the reasoning documented in a comment.

Also spot-verified the trickier fixes with standalone logic/type tests
(mocked HTTP handlers, `mypy --strict` for the `load()` overloads, a
readline history round-trip, a hand-derived numeric check for the
BitLinear unpacking) and, where a real download was feasible, against
actual named models/tokenizers/configs rather than synthetic
approximations.


zcbenz:
Please submit as separate PRs.

PR #1767, opened 2026-08-21
cache.py: default trim() raises NotImplementedError instead of AttributeError (#1081)

amanyagami:
Any subclass that overrides is_trimmable() to return True must also
override trim(); previously the base class had no trim() at all, so
that invariant being broken failed several frames away with a bare
AttributeError instead of failing loudly at the point of the bug.

---

Split out of #1766 per review feedback ("Please submit as separate PRs"). Full test suite (CPU backend, real model downloads, no mocks) passes for this change; see commit message for specifics.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -146,6 +146,23 @@ class _BaseCache:
     def is_trimmable(self):
         return False
 
+    def trim(self, n):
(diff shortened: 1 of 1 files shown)


zcbenz:
At the moment we don't take minor changes like this.

PR #181, opened 2025-05-16
Pixtral text support, pipe input_embeddings through llama arch

mattjcly:
Pipes Pixtral model (i.e., https://huggingface.co/mlx-community/pixtral-12b-4bit) to underlying llama model arch

Demo:
```
/Users/matt/Workspace/forks/mlx-lm/.venv/bin/python -m mlx_lm.generate --model /Users/matt/.cache/lm-studio/models/mlx-community/pixtral-12b-4bit --prompt <s>[INST]Compute the square root of 32[/INST] --max-tokens 500 --temp 0
==========
To compute the square root of 32, you can use the following steps:

1. **Prime Factorization**:
   - Find the prime factors of 32.
   - \( 32 = 2 \times 2 \times 2 \times 2 \times 2 \)
   - This can be written as \( 32 = 2^5 \).

2. **Square Root Calculation**:
   - The square root of a number is a value that, when multiplied by itself, gives the original number.
   - For \( 32 = 2^5 \), the square root is \( 2^{5/2} \).

3. **Simplify the Expression**:
   - Simplify \( 2^{5/2} \) by breaking it down:
     - \( 2^{5/2} = 2^{(1+1)/2} = 2^1 \times 2^{1/2} = 2 \times \sqrt{2} \).

Therefore, the square root of 32 is:
\[ \sqrt{32} = 4\sqrt{2} \]

So, the square root of 32 is \( 4\sqrt{2} \).
==========
Prompt: 11 tokens, 14.503 tokens-per-sec
Generation: 260 tokens, 19.168 tokens-per-sec
Peak memory: 7.004 GB
```
`self.text_config["num_attention_heads"]` section needed to resolve:
```
  File "/Users/matt/Workspace/forks/mlx-lm/mlx_lm/models/pixtral.py", line 31, in __init__
    self.language_model = llama.Model(llama.ModelArgs.from_dict(args.text_config))
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matt/Workspace/forks/mlx-lm/mlx_lm/models/base.py", line 17, in from_dict
    return cls(
           ^^^^
TypeError: ModelArgs.__init__() missing 1 required positional argument: 'num_attention_heads'
```

PR #1768, opened 2026-08-21
Add M-RoPE support so vision embeddings keep their grid positions

lpalbou:
## What this fixes

`initialize_rope` currently returns a plain `nn.RoPE` for Qwen3-VL / Qwen3.5
models. For text-only input that is exactly correct — M-RoPE's three axes all
equal the sequence index there, so it degenerates to 1-D RoPE and the fused
kernel is the right thing to use.

It stops being correct the moment image embeddings arrive through
`input_embeddings`. Image patches then get flattened 1-D indices instead of
their `(t, h, w)` grid coordinates, and the spatial layout is silently lost —
no error, just quietly worse answers.

That matters because `input_embeddings` (#179, #181) is exactly the seam
external integrators use to attach a vision encoder to an mlx-lm text model.
I hit this building vision support on top of mlx-lm: `generate_step` happily
accepts my image embeddings, but there is no way to tell it where those
patches actually sit in the image.

There is also a smaller bug alongside it: the existing `rope_type == "mrope"`
branch is effectively dead for this family, because Qwen3.5 configs report
their type as `"default"` and carry `mrope_section` alongside it. Keying off
the type alone never reaches that branch.

## What it does

- Adds an `MRoPE` rotary in `rope_utils.py`. With `position_ids=None` it
  delegates to `nn.RoPE`, so the text path keeps the fused kernel.
- Selects M-RoPE on the presence of `mrope_section` rather than on the rope
  type, which is what actually identifies these configs.
- Threads an optional `position_ids` through `qwen3_5` and `generate_step`,
  mirroring how `input_embeddings` is already plumbed. Generated tokens
  continue from one past the largest prompt position, so all three axes
  advance together once the prompt is consumed — which is correct, since
  generated tokens are ordinary text.
- Adds `does_model_support_position_ids`, mirroring
  `does_model_support_input_embeddings`. Passing `position_ids` to a model
  that does not accept them raises rather than silently ignoring them.

Linear-attention (Gated DeltaNet) layers carry no rotary embedding, so they
ignore positions entirely.

## Why I need it

I maintain a library that gives Apple-Silicon users vision on MLX checkpoints by
keeping `mlx-lm` as the decoder and using `mlx-vlm` purely as a vision encoder,
handing the merged embeddings over through `input_embeddings`. That works today
across nine local checkpoints. But for the Qwen families I have to attach a
`rope_1d_substituted` marker to every response, because I know the image patches
are being placed at flattened 1-D positions and I would rather say so than hide
it. This PR is what would let me drop that marker.

## An isolated A/B, since the aggregate number hides what this actually does

Same checkpoint (`mlx-community/Qwen3.8-27B-4bit`), same image (the word ZEPPELIN
rendered in bold), same merged embeddings computed once by the vision encoder and
then decoded three ways. The ONLY variable is how positions are assigned:

| decode path | answer |
|---|---|
| the reference VLM runtime (true M-RoPE) | `ZEPPELIN` |
| mlx-lm today (1-D RoPE substituted) | **`ZIPPELIN`** |
| mlx-lm with this PR (`position_ids` supplied) | **`ZEPPELIN`** |

One character, and it is the kind of character error that matters: fine glyph
discrimination is exactly what 2-D positional information supports. Reproduced on
`mlx-community/Ornith-1.0-9B-4bit` too, where all three agree — so the gap is not
universal, which is why the aggregate below is 13/14 vs 11/14 rather than
something dramatic.

## Measured effect

`mlx-community/Qwen3.8-27B-4bit`, 14 spatial questions over 3 synthetic images
(quadrant layout, object counting, left-to-right ordering). Identical weights
and identical merged embeddings in every run — the only thing that changes is
the position encoding. Vision embeddings and reference answers come from
mlx-vlm; scoring is against ground truth, not against agreement.

| lane | correct |
|---|---|
| mlx-lm before this PR (1-D RoPE) | 11 / 14 |
| mlx-lm with this PR (`position_ids` supplied) | **13 / 14** |
| mlx-vlm reference (native M-RoPE) | 13 / 14 |

The two questions that flip are the position-sensitive ones — "what colour is
the shape directly above the green one" (was `RED`, ground truth `BLUE`) and
"how many blue squares are there" (was `4`, ground truth `3`). The remaining
failure is shared with the reference and is just my 48-token generation cap
truncating a list.

Before this PR I also saw a 320x220 rectangle reported as a square, which is
the kind of aspect-ratio judgment that needs 2-D positions to survive.

## No regression on the text path

- `MRoPE` with `position_ids=None` calls the same fused kernel as before.
- End-to-end check on the same 27B model, greedy decode, 3 prompts x 40 tokens:
  **token-identical** to `main`.
- At the rotary itself, the explicit cos/sin path agrees with the fused kernel
  to `2.4e-7` when the three axes are sequential.
- `tests/test_models.py`, `test_generate.py`, `test_prompt_cache.py`: same 9
  pre-existing failures before and after, +2 passing from this PR.

## Tests

`test_mrope` covers selection on `mrope_section`, the `position_ids=None`
fallback being bit-identical to `nn.RoPE`, sequential positions matching 1-D
RoPE, grid positions actually changing the encoding, and partial-rotary
passthrough.

`test_qwen3_5_position_ids` covers the end-to-end threading through
`generate_step` and the rejection path for models without the parameter.

## Scope and caveats

- This wires up `qwen3_5` (which also covers `qwen3_5_moe` and the Ornith
  checkpoints, since they share `model_type`).
- **Correction to an earlier version of this description:** I said the other
  M-RoPE families would need "the same one-line threading". That understates it.
  `qwen3_5` routes through `qwen3_next`'s attention, so the threading was
  contained. `qwen3_vl` delegates to `qwen3.Model`, so covering it means
  threading positions through the base Qwen3 text model — a wider blast radius
  that I did not want to fold into this PR unreviewed. Happy to do it as a
  follow-up, or here if you prefer.
- The benchmark is 14 questions on one model. It is enough to show the
  direction and that the two lanes converge, not enough to quote a rate.
- The explicit cos/sin path is slower than the fused kernel, which is why it
  only runs when `position_ids` is actually supplied.
- Video (`t > 0` across frames) follows the same code path but I have not
  tested it.

One more guard, added after review of my own change: the M-RoPE branch is
restricted to the unscaled rope types. A config pairing an `mrope_section` with
`linear`/`yarn`/`llama3`/`longrope` scaling still needs that scaling applied, and
`MRoPE` does not implement it — so those configs fall through to the existing
behaviour rather than silently losing their scale. Covered in the test.

Happy to adjust the API shape — in particular whether `position_ids` belongs
on `generate_step` or should be carried some other way.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -383,6 +383,16 @@ def generate_step(
                 "position_ids must have shape (3, L) or (3, 1, L), got "
                 f"{position_ids.shape}."
             )
+        # A length mismatch would silently misalign the per-chunk slicing
+        # below, which is the exact failure mode explicit positions exist to
+        # remove. Check it against the same sequence input_embeddings is
+        # checked against.
+        seq_len = len(input_embeddings) if input_embeddings is not None else len(prompt)
+        if position_ids.shape[-1] != seq_len:
+            raise ValueError(
+                f"position_ids sequence length ({position_ids.shape[-1]}) must "
+                f"match the sequence length of the prompt ({seq_len})."
+            )
         next_position = int(position_ids.max()) + 1
 
     tokens = None
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -31,7 +31,11 @@ from .models.cache import (
 )
 from .sample_utils import make_sampler
 from .tokenizer_utils import TokenizerWrapper
-from .utils import does_model_support_input_embeddings, load
+from .utils import (
+    does_model_support_input_embeddings,
+    does_model_support_position_ids,
+    load,
+)
 
 DEFAULT_PROMPT = "hello"
 DEFAULT_MAX_TOKENS = 100
@@ -310,6 +314,7 @@ def generate_step(
     quantized_kv_start: int = 0,
     prompt_progress_callback: Optional[Callable[[int, int], None]] = None,
     input_embeddings: Optional[mx.array] = None,
+    position_ids: Optional[mx.array] = None,
 ) -> Generator[Tuple[mx.array, mx.array], None, None]:
     """
     A generator producing token ids based on the given prompt from the model.
@@ -336,6 +341,12 @@ def generate_step(
            when ``kv_bits`` is non-None. Default: ``0``.
         prompt_progress_callback (Callable[[int, int], None]): A call-back which takes the
            prompt tokens processed so far and the total number of prompt tokens.
+        position_ids (mx.array, optional): Explicit positions of shape
+          ``(3, L)`` or ``(3, 1, L)`` for M-RoPE models. Required only when
+          the prompt contains image or video embeddings, whose positions are
+          grid coordinates rather than sequence indices. Generated tokens
+          continue from one past the largest prompt position.
+          Default: ``None``.
         input_embeddings (mx.array, optional): Input embeddings to use instead of or in
           conjunction with prompt tokens. Default: ``None``.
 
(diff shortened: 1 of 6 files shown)


zcbenz (commented):
Actually I think we should add vision support to mlx-lm and this seems to be a good start, but it will take some time for us to check this PR leaves us room to add vision support in future.

PR #1769, opened 2026-08-21
Fix cached model buffers on server swaps

### LINKED ISSUE #1757: mlx_lm.server never clears MLX's buffer pool on model swap, retaining every model it has loaded
## Summary

`mlx_lm.server` never releases MLX's buffer pool, so a server that is asked for more than one model retains the weights of every model it has loaded — while reporting only one model live. `generate.py` and `evaluate.py` call `mx.clear_cache()` (6 and 2 sites respectively); `server.py`, the one long-lived process where retention compounds, calls it zero times.

This is not a leak — the memory is a cache and `mx.clear_cache()` reclaims all of it. The problem is that nothing ever calls it, and MLX's default cache limit is effectively all of RAM.

## Where

`ModelProvider._load()` drops the previous model's references and immediately allocates the next one:

```python
# Remove the old model if it exists.
self.model_key = None
self.model = None
self.tokenizer = None
self.draft_model = None

# Load the model and tokenizer
...
```

Those four assignments return the old weights to MLX's Metal buffer pool, not to the OS. The next `load()` allocates fresh buffers because the shapes differ, and the pool keeps the old set indefinitely.

## Reproduction

```python
import gc
import mlx.core as mx
from mlx_lm import load

GB = 2**30
snap = lambda l: print(f"{l:24s} active={mx.get_active_memory()/GB:5.2f} cache={mx.get_cache_memory()/GB:5.2f}")

m, _ = load("mlx-community/Llama-3.2-3B-Instruct-4bit"); mx.eval(m.parameters())
snap("A loaded (1.7GB)")
del m; gc.collect()
snap("A deleted")

m, _ = load("mlx-community/Qwen2.5-Coder-14B-Instruct-4bit"); mx.eval(m.parameters())
snap("B loaded (7.7GB)")
del m; gc.collect()
snap("B deleted")

mx.clear_cache()
snap("after clear_cache()")
```

```
A loaded (1.7GB)         active= 1.68 cache= 0.00
A deleted                active= 0.00 cache= 1.68
B loaded (7.7GB)         active= 7.74 cache= 1.68
B deleted                active= 0.00 cache= 9.42   <-- both models retained
after clear_cache()      active= 0.00 cache= 0.00   <-- RSS 4.82GB -> 0.53GB
```

The pool converges on the union of size classes rather than growing without bound, so two models cost you both models forever. Serving them through `mlx_lm.server` reproduces the same numbers via HTTP: request model A, then model B, and `/v1/models` reports one model while the process holds two.

## Why the default makes this bite

```
>>> mx.set_cache_limit(2**60)   # returns the previous value
97925254348                     # 91.2 GB on a 96 GiB machine
```

The cache limit defaults to essentially all of physical memory, so MLX will not self-limit before the machine is in trouble. Reasonable on a dedicated inference box; on a workstation also running containers and a GUI there is no margin.

## Real-world impact

On a 96 GiB M3 Ultra, a config drift had one client requesting `Qwen3.6-35B-A3B-4bit` while the daemon was started with `Qwen3.6-35B-A3B-8bit`. `mlx_lm.server` serves whatever model a request names rather than rejecting it, so this produced 25 swaps across 36 loads over two days. The same process, serving the same model, measured:

- 37.6 GB resident on day 1 (one 35 GB model — correct)
- 85.1 GB resident on day 2 (still one model live; ~48 GB of pooled dead buffers)

A 60 GB model then started alongside it. Total demand reached 167 GB, macOS compressed it into 52 GB of RAM with 0.2 GB free for two and a half hours, and the kernel panicked on the WindowServer userspace watchdog. Nothing in the server's own reporting indicated it was holding 48 GB it would never use again.

## Suggested fix

Call `mx.clear_cache()` in `_load()` after dropping the old references and before loading the new model — the one moment where nothing is live and clearing cannot cost a reuse:

```python
 self.draft_model = None
+mx.clear_cache()
```

Optionally also expose a startup `--cache-limit` that calls `mx.set_cache_limit()`, so operators can bound the pool without a wrapper.

Note this is distinct from #1546, which addresses pooled memory from *completed requests* via an idle timer and a `/v1/cache/evict` endpoint. That PR does not touch `_load()`, and its idle eviction is opt-in and defaults to disabled, so model-swap retention persists with it applied.

## Versions

- mlx-lm 0.31.3 (current release) and `main` as of 2026-08-18 — neither contains any `clear_cache` in `server.py`
- mlx 0.32.0, mlx-metal 0.32.0
- macOS 26.5.2, M3 Ultra (Mac15,14), 96 GiB

tandede:
## Summary

- clear MLX cached buffers after releasing the current model and before allocating its replacement
- leave the initial model load unchanged so startup does not perform an unnecessary cache clear
- add regression coverage for both the replacement ordering and the initial-load path

Fixes #1757

## Testing

- `python -m pytest -q tests/test_server.py` (27 passed)
- `pre-commit run --files mlx_lm/server.py tests/test_server.py`
- manual MLX allocation check: a 64 MiB old model allocation dropped from 67,108,872 active bytes to 8 active bytes with 0 cached bytes before the replacement loader ran

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -312,10 +312,14 @@ class ModelProvider:
             )
 
         # Remove the old model if it exists.
+        had_loaded_model = self.model is not None or self.draft_model is not None
         self.model_key = None
         self.model = None
         self.tokenizer = None
         self.draft_model = None
+        if had_loaded_model:
+            # Return buffers from the previous model before allocating its replacement.
+            mx.clear_cache()
 
         # Load the model and tokenizer
         if self.is_distributed:
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -5,6 +5,8 @@ import io
 import json
 import threading
 import unittest
+from types import SimpleNamespace
+from unittest.mock import patch
 
 import mlx.core as mx
 import requests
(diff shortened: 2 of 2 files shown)


zcbenz:
Automatic PR is banned in this repo.

PR #1770, opened 2026-08-21
Add AI usage policy

zcbenz:
This is basically the same with https://github.com/ml-explore/mlx/pull/4331 in MLX repo, with an additional pull request template on adding models to help us verify the correctness of implementation.


nastya236 (approved):
Amazing!

PR #1777, opened 2026-08-24
Include prefilled prompt tokens in logits-processor history

### LINKED ISSUE #1773: mlx_lm.generate.generate_step() omits prompt-prefill tokens from logits-processor history.
## Summary

`generate_step()` does not expose the full prompt history to logits processors.

When a multi-token prompt is prefetched, the prefetched tokens are added to the model KV cache but not to the `tokens` value passed to logits processors. As a result, the first processor call sees only the final prompt token instead of the complete prompt.

## Environment

- mlx-lm: 0.31.3
- mlx: 0.32.0
- OS: macOS 15.7.3 on Apple Silicon
- Python: 3.10.18

## Minimal reproduction

The following uses a recorder processor that only records the token history it receives:

```python
import mlx.core as mx

prompt_ids = mx.array([2, 5, 7, 9])
first_history = []

def record_history(tokens, logits):
    if not first_history:
        first_history.append(tokens.tolist())
    return logits

list(
    generate_step(
        prompt_ids,
        model,  # which is any already-loaded compatible model.
        max_tokens=1,
        logits_processors=[record_history],
    )
)

print(first_history[0])
```

## Actual result

```text
[9]
```

The first processor call receives only the final prompt token.

## Expected result

```text
[2, 5, 7, 9]
```

The processor should receive the full semantic token history, with each prompt token represented exactly once.

## Why this matters

Stateful logits processors such as repetition, presence, and frequency penalties use the supplied token history as their context window.

Because prompt-prefill tokens are missing from that history, these processors cannot apply their configured context window to earlier prompt text. They only see the final prompt token and tokens generated during the current completion.

This is especially problematic for multi-turn generation, where important conversation context may be in the prefetched part of the prompt.

## Implementation detail

In `generate_step()`, the prefill loop calls `_model_call()` directly. The local `tokens` value is updated only inside `_step()`, which is called for the remaining final prompt token and subsequent generated tokens.

This makes the model cache history and the processor-visible token history inconsistent.

gyanu2507:
Fixes #1773

generate_step prefills all but the last prompt token through _model_call and never puts those ids on `tokens`. _step then concatenates only the leftover token, so the first logits-processor call gets `[last_prompt_token]` instead of the whole prompt.

That's the history repetition/presence/frequency penalties use. Anything in the prefetched prefix is invisible to them.

Same concat _step already does, just during prefill. Dummy-model test covers the four-token repro from the issue.

- [x] I understand it is strictly prohibited to use AI to write PR description


tests/test_generate_step.py
```diff
@@ -0,0 +1,49 @@
+# Copyright © 2024 Apple Inc.
```
simeetnayan81: Hey, I want to understand, why do we need a new file here for the tests? 
I already see test.test_generate.test_batch_generate_with_logits_processors which has tests written for the module. Previously the test only used a single token and that is why prefill never ran previously.

Better: add a new method on TestGenerate, keep the dummy model local so it stays  under a unit test:
Maybe something like this
```python
def test_generate_step_processor_sees_prefilled_prompt(self):
    # DummyLM + [2, 5, 7, 9], prefill_step_size=2
    # first processor history == [2, 5, 7, 9]
```

Or 

A second TestCase in the same file (no setUpClass) is also fine if you want dummy tests not sitting on the real-model class. 


Also maybe, instead of using dummy model if we can use already existing Qwen in the class for this test as well, it'd be more consistent.

simeetnayan81 (commented):
Thanks for raising the PR. The code looks good to me but I have added a comment for the tests. Thanks. 

cc: @nastya236, @zcbenz

tests/test_generate_step.py
gyanu2507: Good call on both counts — done in ab58b24. `tests/test_generate_step.py` is gone; the test is now a method on `TestGenerate` using the Qwen model from `setUpClass`, so the `DummyLM` stub went away with it.

You're right about why the existing coverage missed this: `test_batch_generate_with_logits_processors` encodes `"hello"`, and with `total_prompt_tokens - prompt_processed_tokens > 1` never true, the prefill loop doesn't run at all. The new test passes `prefill_step_size=2` over a 4-token prompt so the loop runs twice and leaves one token for `_step`.

Confirmed it's load-bearing: with the `generate.py` hunk reverted the first processor call sees `[9]` instead of `[2, 5, 7, 9]`, which is the repro from #1773. `tests/test_generate.py` is 27 tests, all passing.

gyanu2507:
@simeetnayan81 moved the coverage onto `TestGenerate` with the Qwen model and deleted `tests/test_generate_step.py` in ab58b24. Please take another look.

simeetnayan81 (approved):
Hey, thanks for working on the feedback. LGTM.  Thanks.

zcbenz (approved):
Thanks!

PR #1778, opened 2026-08-24
Make "state" of cache return full state

### LINKED ISSUE #1250: BatchRotatingKVCache: rotated=False reloads as True from saved prompt cache

### LINKED ISSUE #1384: Gemma 3n: mlx_lm.server returns garbage output (shared-KV layers incompatible with batched cache)

zcbenz:
Make padded kv cache return raw keys/values/offset in `state`, so they would still be padded after restoration, which is required for optimized sdpa kernels to kick in.

A few more refactoring to make this happen:
* Make `load/save_prompt_cache` recognize scalars in cache state and do automatic serialization/deserialization, so it is possible to return scalars in `state` and get rid of `meta_state`.
* Add a `keys_and_values()` method to return sliced keys/values, instead of relying on internal state of `state` property.

As a side effect also fixes https://github.com/ml-explore/mlx-lm/issues/1250, fixes #1384.

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -52,12 +52,29 @@ def save_prompt_cache(
         metadata (Dict[str, str]): Optional metadata to save along with model
             state.
     """
-    metadata = {} if metadata is None else metadata
-    cache_data = [c.state for c in cache]
-    cache_info = [c.meta_state for c in cache]
-    cache_data = dict(tree_flatten(cache_data))
+    scalars = [""]  # must be non-empty otherwise tree_flatten drops it
+
+    # Convert scalars to mx.array so they can be serialized
+    def map_data(item):
+        i, data = item
+        if isinstance(data, mx.array):
+            return item
+        elif isinstance(data, (int, float, bool)):
+            scalars.append((i, "scalar"))
+            return (i, mx.array(data))
+        elif isinstance(data, str):
+            scalars.append((i, "string"))
+            return (i, mx.array([c for c in data.encode("utf-8")]))
+        elif data is None:
+            scalars.append((i, "none"))
+            return (i, mx.array([]))
+        else:
+            raise ValueError(f"Model state is not serializable: {data}")
+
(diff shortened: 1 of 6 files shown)


zcbenz:
@michalk8 Can you do a review of this PR?

mlx_lm/models/cache.py
```diff
@@ -73,13 +91,20 @@ def load_prompt_cache(file_name, return_metadata=False):
...
-        globals()[c].from_state(state, meta_state)
-        for c, state, meta_state in zip(classes, arrays, info)
-    ]
+    metadata, classes, scalars = cache_metadata
+    for pos, type in scalars[1:]:
+        array = arrays[pos]
+        if type == "scalar":
+            arrays[pos] = array.item()
+        elif type == "string":
+            arrays[pos] = "".join(chr(i) for i in array.tolist())
```
michalk8, line 103: `bytes(array.tolist()).decode("utf-8")`, this breaks on, e.g.:

```python
x = 'café'
print("".join(chr(i) for i in x.encode('utf-8')))  # cafÃ©
```

michalk8 (changes requested):
Just the one comment for string serialisation, otherwise LGTM!

mlx_lm/models/cache.py
zcbenz: I changed to use `[ord(char) for char in data]` which should work universally.

michalk8 (approved):
LGTM from my side!

PR #1779, opened 2026-08-24
Remove dead code in generate.py

zcbenz:
The `_make_cache` is unused and `_make_new_cache` is the same with `make_prompt_cache`.


nastya236 (approved):

nastya236:
THanks, it was a lot of dead code..

PR #1781, opened 2026-08-24
Proposal: Add --prefill-step-size also to mlx_lm.generate and mlx_lm.chat for speed/memory usage trade-off

davidparkercoder:
## Summary

`prefill_step_size` is the largest lever on peak memory during long-prompt prefill, but it is only reachable from the command line on two of the four entry points. #943 added `--prefill-step-size` to `mlx_lm.server` and
`mlx_lm.benchmark`. `mlx_lm.generate` and `mlx_lm.chat` still hardcode the 2048 default, so a user who runs either of those has no way to set it short of writing Python.

This adds the same flag to both, with the same name, type and default.

## Why it matters

Measured on an M5 Max 128 GB, mlx-lm 0.31.3, macOS 26.6.1, `mlx-community/Qwen3.8-27B-4bit`, at a full 128K context:

| prefill step size | peak GB | generation tok/s | prefill tok/s |
|---|---|---|---|
| 2048 (default) | 39.78 | 21.31 | 431.5 |
| 1024 | 32.26 | 21.37 | 438.1 |
| 512 | 28.54 | 21.40 | 438.9 |
| 256 | 26.52 | 21.29 | 418.5 |

Going from 2048 to 512 drops peak memory by 11.2 GB, or 28%, at a 128K context.

## Changes

- `mlx_lm/generate.py`: add `DEFAULT_PREFILL_STEP_SIZE`, add the
  `--prefill-step-size` argument, and pass it through to `generate`.
- `mlx_lm/chat.py`: same, passed through to `stream_generate`.
- `tests/test_generate.py`: parser tests for the default and an override.
- `tests/test_chat.py`: extend the existing all-args test and add a default test.
- `README.md`: document the flag under "Long Prompts and Generations".

No behaviour changes when the flag is not passed. Both entry points keep the existing 2048 default, and the plumbing is a keyword argument that `generate_step` already accepts.

## Verification

```
python -m unittest test_chat test_generate.TestGenerateArgs
Ran 7 tests in 0.014s
OK
```

Confirmed end to end that the value reaches `generate_step`: with `--prefill-step-size 8` it receives 8, and with the flag absent it receives 2048.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,11 +200,18 @@ Face Space.
 
 - A rotating fixed-size key-value cache.
 - Prompt caching
+- A configurable prefill step size
 
 To use the rotating key-value cache pass the argument `--max-kv-size n` where
 `n` can be any integer. Smaller values like `512` will use very little RAM but
 result in worse quality. Larger values like `4096` or higher will use more RAM
 but have better quality.
+
+Long prompts are read in steps, and the step size sets how many prompt tokens
+are processed at once. Smaller steps use less peak memory while the prompt is
+read, at some cost to prompt processing speed. To change it pass
+`--prefill-step-size n`. The default is `2048`, and `mlx_lm.generate`,
+`mlx_lm.chat`, `mlx_lm.server` and `mlx_lm.benchmark` all accept it.
 
 Caching prompts can substantially speedup reusing the same long context with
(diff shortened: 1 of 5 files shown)


RohanGautam:
Lgtm! Useful improvement.

nastya236 (approved):
Looks great, thank you!

PR #1787, opened 2026-08-26
Stop gradients through Qwen3 MoE expert-routing indices

### LINKED ISSUE #1786: [BUG] Qwen3 MoE LoRA backward fails through discrete expert-routing indices
## Summary

LoRA fine-tuning of `Qwen3-30B-A3B-Instruct-2507` in MLX 4-bit format fails deterministically on the first backward pass:

```text
ValueError: [gather_axis] Cannot calculate VJP with respect to indices.
```

Inference and validation forward passes succeed. The failure also occurs with `grad_checkpoint: false`, so gradient checkpointing is not the cause.

## Environment

- macOS 26.6.2, arm64
- Apple M3 Ultra, 256 GiB unified memory
- Python 3.12.14
- MLX 0.32.2
- MLX-LM 0.31.3
- model: `lmstudio-community/Qwen3-30B-A3B-Instruct-2507-MLX-4bit`
- fine-tuning type: LoRA
- trainable layers: 1
- batch size: 1
- maximum sequence length in the minimal reproduction: 128

The relevant Qwen3 MoE code is unchanged on current `main` at `74e7cf931e84ef7c2f63e875adf414e20decc1c5`.

## Root cause

`Qwen3MoeSparseMoeBlock.__call__` computes integer expert indices from differentiable router scores:

```python
inds = mx.argpartition(gates, kth=-k, axis=-1)[..., -k:]
scores = mx.take_along_axis(gates, inds, axis=-1)
y = self.switch_mlp(x, inds)
```

The indices remain on the automatic-differentiation path. During backward propagation, MLX consequently requests a VJP with respect to the discrete indices and raises.

This is the Qwen3 MoE routing site. It is distinct from the DSA sparse-attention indexer discussed in #1449 and #1452, although the underlying non-differentiable-index problem is analogous to ml-explore/mlx#3787.

## Proposed fix

Detach the discrete expert selection immediately after `argpartition`:

```python
inds = mx.argpartition(gates, kth=-k, axis=-1)[..., -k:]
inds = mx.stop_gradient(inds)
```

The selected scores and model activations remain differentiable. The change has identity forward semantics and prevents autograd from requesting a gradient for an integer selection.

## Validation

With the change applied to current `main`:

- a one-step real-model LoRA run succeeds with gradient checkpointing enabled;
- a twenty-step run succeeds with checkpointing enabled and 1,736-token inputs;
- validation runs at iterations 1, 5, 10, 15 and 20 complete;
- adapters save at iterations 10 and 20 and the final adapter reloads for generation;
- tiny-model regression tests pass both with and without MLX-LM gradient checkpointing;
- `tests/test_finetune.py` and the two new tests pass together: 18 passed;
- Black and isort pre-commit hooks pass on both changed files.

The longer run uses disposable repeated text to exercise the training path. Its falling loss is not a model-quality result.

nicejazzburger:
## What

Stops gradients through the integer top-k expert indices produced in `Qwen3MoeSparseMoeBlock`.

## Why

The indices derive from differentiable router scores and are consumed by `take_along_axis` and `SwitchGLU`. Without an explicit barrier, LoRA backward propagation asks MLX for a VJP with respect to the indices and fails with:

```text
ValueError: [gather_axis] Cannot calculate VJP with respect to indices.
```

Integer expert selections are non-differentiable. `mx.stop_gradient(inds)` preserves the forward computation while preventing the invalid gradient request. Router scores and model activations remain differentiable.

## Tests

Adds `tests/test_qwen3_moe_gradients.py`, which constructs a small one-layer Qwen3 MoE model and verifies finite backward propagation:

- without gradient checkpointing;
- with MLX-LM gradient checkpointing.

Validation completed on current `main`:

```text
black....................................................................Passed
isort....................................................................Passed
..................                                                       [100%]
18 passed
```

The change also completed one-step and twenty-step checkpointed LoRA smoke runs on `Qwen3-30B-A3B-Instruct-2507-MLX-4bit`, including adapter save and reload. The longer run used 1,736-token disposable inputs.

Fixes #1786

Related: ml-explore/mlx#3787, #1449, #1452


zcbenz (commented):
Can you move the test to `tests/test_tuner_trainer.py`?

zcbenz (approved):

PR #1795, opened 2026-08-27
Stop gradients through MoE routing indices in granitemoe, granitemoehybrid, lfm2_moe

adityak74:
## What

Wraps the integer top-k expert-routing indices in `mx.stop_gradient` in three MoE model families: `granitemoe`, `granitemoehybrid`, and `lfm2_moe`.

## Why

LoRA fine-tuning (`mlx_lm.lora --train`) on any of these families currently crashes in the backward pass:

```text
ValueError: [gather_axis] Cannot calculate VJP with respect to indices. Use stop_gradient on indices to stop gradients from being computed.
```

The indices derive from differentiable router logits and are consumed by `take_along_axis` and the switch MLP, so without an explicit barrier MLX is asked for a VJP w.r.t. integer indices. Most sibling MoE models in the repo (`olmoe`, `mixtral`, `qwen2_moe`, `phimoe`, `glm4_moe`, `bailing_moe`, `ernie4_5_moe`, `exaone_moe`, `afmoe`, ...) already guard their routing indices with `mx.stop_gradient`; these three were missing it.

This is the same fix as #1787 (`qwen3_moe`), applied to the remaining affected families. A survey of `mlx_lm/models/` shows these three are the last index-selecting MoE models without the guard.

## Test

Adds `tests/test_moe_routing_gradients.py`: a tiny-config backward pass through each of the three models (no weight downloads). Fails 3/3 before the fix, passes 3/3 after. Also verified end-to-end that the same one-line pattern unblocks real `mlx_lm.lora --train` runs (found while fine-tuning Qwen3-30B-A3B / surveying MoE trainability on an M5 Pro).

https://claude.ai/code/session_01GZWrsNwempeEyaCaQiZbWH


zcbenz (commented):
Can you move the tests to tests/test_tuner_trainer.py like https://github.com/ml-explore/mlx-lm/pull/1787/changes?

adityak74:
Done — moved the three backward-pass tests into `tests/test_tuner_trainer.py` following the #1787 structure (same `_assert_backward` helper shape), and removed the separate test file. Full `test_tuner_trainer.py` passes locally (5 passed).

zcbenz (approved):

PR #1796, opened 2026-08-28
fix: handle single-quoted values with commas in pythonic tool parser

### LINKED ISSUE #1785: mlx_lm "pythonic" tool-call parser truncates arguments containing single-quoted strings with embedded commas
### Environment

- `mlx-lm` version: 0.31.3
- `mlx` version: 0.31.2
- macOS 26.5.2, Apple Silicon (arm64)
- Model: `LiquidAI/LFM2.5-2.6B-MLX` (4bit quant), served via `mlx_lm.server`

### Summary

The `pythonic` tool-call parser (`mlx_lm/tool_parsers/pythonic.py`) truncates string
arguments at the first comma when the argument value is **single-quoted** — which is
exactly the format `LiquidAI/LFM2.5-2.6B-MLX`'s own chat template emits (see
`format_arg_value` in the model's `chat_template.jinja`, which always wraps string
arguments in single quotes and escapes embedded ones). Any comma appearing inside the
string content (not just literal list-separator commas) breaks the parse.

### Root cause

`_tool_args_regex` only special-cases **double**-quoted values:

```python
_tool_args_regex = re.compile(r'(\w+)=(?:"([^"]*)"|([^,]+))(?:,\s*|$)', re.DOTALL)
```

Anything not wrapped in `"..."` falls into the `([^,]+)` branch, which is a naive
"everything up to the next comma" match — including commas that are part of the quoted
string's *content*, not argument separators.

### Reproduction

```python
text = (
    "<|tool_call_start|>[write(filePath='/tmp/hello.py', "
    "content='# Hello World in Python\\nprint(\"Hello, World!\")')]"
    "<|tool_call_end|>"
)
parse_tool_call(text)
```

**Actual:**

```python
{'name': 'write', 'arguments': {'filePath': '/tmp/hello.py',
 'content': '\'# Hello World in Python\\nprint("Hello'}}
```

`content` is silently truncated right before `, World!")'` — the comma inside the
Python `print(...)` string being written ends the match.

**Expected:**

```python
{'name': 'write', 'arguments': {'filePath': '/tmp/hello.py',
 'content': '# Hello World in Python\nprint("Hello, World!")'}}
```

Run `repro.py` in this directory to see both outputs side by side (no GPU/Metal needed).

### Impact

Any tool call whose string argument contains a comma (extremely common for `content`/code
arguments, like a `write` or `edit` file tool) gets silently truncated with no error —
the response comes back well-formed but with corrupted data, so it's easy to miss until
the written file/output looks wrong downstream.

### Related work already in the tracker

This exact bug (single-quoted argument values with embedded commas) doesn't appear to be
filed anywhere yet — searched issues/PRs for `pythonic`, `tool_call comma`, `single quote
tool`, `ast.literal_eval`, `content argument cut off`, nothing matches. But there's a lot
of adjacent, currently-unmerged work worth being aware of before touching this file:

- [#864](https://github.com/ml-explore/mlx-lm/pull/864) (merged) — originally added the
  `pythonic` parser's use for LFM2, with auto-detection keyed on `<|tool_list_start|>`.
- [#1260](https://github.com/ml-explore/mlx-lm/pull/1260),
  [#1333](https://github.com/ml-explore/mlx-lm/pull/1333),
  [#1436](https://github.com/ml-explore/mlx-lm/pull/1436) (all open) — three independent,
  competing PRs fixing the same auto-detection gap: LFM2.5 renamed its markers to
  `<|tool_call_start|>`/`<|tool_call_end|>`, so `_infer_tool_parser`'s
  `<|tool_list_start|>` check never fires and the pythonic parser never gets selected for
  LFM2.5 models. (We hit this too — see our workaround below.)
- [#1707](https://github.com/ml-explore/mlx-lm/pull/1707) (open) — fixes `pythonic.py`'s
  handling of multi-call blocks (`[f1(...), f2(...)]`) with a quote-aware call-boundary
  regex, but leaves `_tool_args_regex` (the actual bug here) completely untouched — verified
  by diffing the PR. Does not fix this issue.
- [#1246](https://github.com/ml-explore/mlx-lm/pull/1246) (open) — proposes a dedicated
  `lfm2.py` parser that parses the whole `[...]` block with `ast` in one pass (also handles
  dotted names like `grocery.orderIngredients` and multi-call lists, which the shared
  `pythonic` parser doesn't). As a side effect this would also fix the truncation bug here,
  since it never falls back to comma-splitting — but it's a separate file gated behind its
  own auto-detection rule, not a patch to the shared parser.

None of the above are merged, so this bug is live on `main` as of this writing (verified by
fetching `mlx_lm/tool_parsers/pythonic.py` directly from `main` — byte-identical to the
buggy 0.31.3 release). Given the amount of overlapping work already in flight, whoever
triages this should probably decide whether to patch `pythonic.py` in isolation (this
issue's suggested fix) or fold it into #1246's more complete rewrite, rather than have a
third parallel patch to reconcile.

### Suggested fix

Replace the regex-based key/value splitting with real Python-literal parsing via `ast`,
which correctly handles quoting style (single/double), escapes, and embedded commas/parens
since it uses Python's actual grammar instead of approximating it. Full diff in `fix.patch`;
the changed function:

```python
_tool_call_regex = re.compile(r"\[(\w+)\((.*?)\)\]", re.DOTALL)


def parse_tool_call(text: str, tools: Any | None = None):
    match = _tool_call_regex.search(text)
    if not match:
        raise ValueError("No function provided.")

    func_name = match.group(1)
    args_str = match.group(2)

    arguments = {}
    if args_str:
        try:
            call_node = ast.parse(f"f({args_str})", mode="eval").body
        except (SyntaxError, RecursionError, MemoryError):
            # Pathologically nested/long argument text can blow the parser's
            # recursion depth or memory (e.g. long unary/binary operator
            # chains) rather than raising a plain SyntaxError - treat those
            # the same as unparseable input instead of crashing the caller.
            call_node = None

        if isinstance(call_node, ast.Call):
            for kw in call_node.keywords:
                if kw.arg is None:
                    continue
                try:
                    arguments[kw.arg] = ast.literal_eval(kw.value)
                except (ValueError, SyntaxError):
                    arguments[kw.arg] = ast.unparse(kw.value)

    return dict(name=func_name, arguments=arguments)
```

Verified against the failing case above, the parser's own existing docstring example
(`function_name(arg1="value1", arg2=2)`), and a battery of adversarial nesting inputs
(see "Self-review" below) — see `repro.py`.

### Self-review — what this fix does *not* cover

Before proposing this, I stress-tested it adversarially rather than assuming the `ast`
rewrite was a strict improvement. Two real findings came out of that, plus two known,
unchanged limitations:

1. **Fixed during review: uncaught `RecursionError`/`MemoryError` on pathological input.**
   The first version of this fix only caught `SyntaxError` around `ast.parse`. Testing
   deeply-nested/long argument expressions found that CPython's parser doesn't uniformly
   degrade to `SyntaxError` — e.g. `ast.parse("f(x=" + "1+" * 200_000 + "1)")` raises an
   *uncaught* `RecursionError`, and long unary chains (`-` / `not`) raise `MemoryError`
   instead. Either would crash the request-handling thread on a sufficiently
   long/adversarial tool-call argument. Fixed by widening the `except` clause to
   `(SyntaxError, RecursionError, MemoryError)`. `repro.py` includes these cases as
   regression tests.

2. **Not fixed: the outer call-boundary regex is still non-quote-aware.**
   `_tool_call_regex = re.compile(r"\[(\w+)\((.*?)\)\]")` is unchanged — it's non-greedy up
   to the *first* literal `)]` in the text, including one embedded inside a quoted string
   argument (very plausible for a code-writing tool, e.g. `content='items[0])] # foo'`).
   When that happens, `args_str` gets truncated mid-string, which is no longer valid Python,
   so `ast.parse` raises `SyntaxError` and **all** arguments are dropped — worse than the
   original regex-based parser, which degrades to a partial/mangled value but at least keeps
   the other keys. This is the same *class* of bug as the one this issue reports, just one
   layer up (call-boundary matching instead of argument-splitting), and it's pre-existing —
   present identically in the unmodified `pythonic.py` today, not something this patch
   introduces. A complete fix needs a quote-aware call-boundary matcher, which is roughly
   what #1707 already builds (for a different reason — multi-call splitting). This patch
   deliberately doesn't attempt that here to keep the diff minimal and reviewable; flagging
   it so whoever picks this up doesn't consider the job done after just the `_tool_args_regex`
   fix.

3. **Not fixed: dotted function names** (e.g. `grocery.orderIngredients(...)`) — the outer
   regex's `\w+` for the function name doesn't allow `.`. Pre-existing, unrelated to this
   bug, already called out explicitly in #1246's description.

4. **Not fixed: multi-call blocks** (`[f1(...), f2(...)]`) — `_tool_call_regex.search()`
   only ever matches the first call. Pre-existing; that's what #1707 is for.

Net effect: this patch is a minimal, narrowly-scoped, verified fix for the exact reported
symptom (comma inside a single/double-quoted string argument) plus the recursion/memory
hardening found while testing it — not a claim that the `pythonic` parser is fully robust
afterward. Given items 2–4 above and the overlapping PRs already in flight, a maintainer
may reasonably prefer folding this into #1246 instead of merging it standalone.

Happy to open a PR with this change if useful.

axiom-of-choice:
## Summary

The pythonic tool-call parser's regex only special-cases double-quoted values. Single-quoted values (which LFM2.5's chat template emits) fall into the `([^,]+)` branch, which truncates at the first embedded comma.

**Before:** `[write(filePath='/tmp/hello.py', content='# Hello, world!')]` parses `content` as `# Hello` (truncated at the comma).

**After:** The regex adds a `'([^']*)'` branch that handles single-quoted values as a third alternative, and the value-selection logic picks from the three groups (double-quoted, single-quoted, unquoted).

## Changes

- `mlx_lm/tool_parsers/pythonic.py`: added `'([^']*)'` to `_tool_args_regex`, updated group indexing
- `tests/test_tool_parsing.py`: added `test_pythonic_single_quoted_args_with_commas` with two cases (single-quoted with embedded comma + double-quoted regression)

## Tests

```
8 passed in 4.74s
```

Closes #1785.

### DIFF
--- a/mlx_lm/tool_parsers/pythonic.py
+++ b/mlx_lm/tool_parsers/pythonic.py
@@ -14,7 +14,9 @@ Parses assistant responses containing tool calls in formats like:
 
 
 _tool_call_regex = re.compile(r"\[(\w+)\((.*?)\)\]", re.DOTALL)
-_tool_args_regex = re.compile(r"""(\w+)=(?:"([^"]*)"|'([^']*)'|([^,]+))(?:,\s*|$)""", re.DOTALL)
+_tool_args_regex = re.compile(
+    r"""(\w+)=(?:"([^"]*)"|'([^']*)'|([^,]+))(?:,\s*|$)""", re.DOTALL
+)
 
 
 def parse_tool_call(text: str, tools: Any | None = None):
--- a/mlx_lm/tool_parsers/pythonic.py
+++ b/mlx_lm/tool_parsers/pythonic.py
@@ -14,7 +14,7 @@ Parses assistant responses containing tool calls in formats like:
 
 
 _tool_call_regex = re.compile(r"\[(\w+)\((.*?)\)\]", re.DOTALL)
-_tool_args_regex = re.compile(r'(\w+)=(?:"([^"]*)"|([^,]+))(?:,\s*|$)', re.DOTALL)
+_tool_args_regex = re.compile(r"""(\w+)=(?:"([^"]*)"|'([^']*)'|([^,]+))(?:,\s*|$)""", re.DOTALL)
 
 
 def parse_tool_call(text: str, tools: Any | None = None):
@@ -30,8 +30,8 @@ def parse_tool_call(text: str, tools: Any | None = None):
         matches = _tool_args_regex.findall(args_str)
         for pair in matches:
             key = pair[0].strip()
-            # pair[1] is quoted value, pair[2] is unquoted value
-            value = pair[1] if pair[1] else pair[2].strip()
+            # pair[1] is double-quoted, pair[2] is single-quoted, pair[3] is unquoted
+            value = pair[1] if pair[1] else (pair[2] if pair[2] else pair[3].strip())
 
             # Try to parse the value using ast.literal_eval
             try:
--- a/mlx_lm/tool_parsers/pythonic.py
+++ b/mlx_lm/tool_parsers/pythonic.py
@@ -1,7 +1,7 @@
 # Copyright © 2026 Apple Inc.
 
 import ast
-from typing import Any, Dict, List
+from typing import Any
 
 import regex as re
 
@@ -42,7 +42,7 @@ def parse_tool_call(text: str, tools: Any | None = None):
 
             arguments[key] = value
 
-    return dict(name=func_name, arguments=arguments)
+    return {"name": func_name, "arguments": arguments}
 
 
 tool_call_start = "<|tool_call_start|>"
--- a/tests/test_tool_parsing.py
+++ b/tests/test_tool_parsing.py
@@ -154,6 +154,22 @@ class TestToolParsing(unittest.TestCase):
                 }
                 self.assertEqual(tool_call, expected)
 
+    def test_pythonic_single_quoted_args_with_commas(self):
+        # LFM2.5 emits single-quoted strings; embedded commas must not truncate
+        test_case = (
+            "[write(filePath='/tmp/hello.py', "
+            "content='# Hello, world!')]"
+        )
+        tool_call = pythonic.parse_tool_call(test_case, None)
+        self.assertEqual(tool_call["name"], "write")
+        self.assertEqual(tool_call["arguments"]["filePath"], "/tmp/hello.py")
+        self.assertEqual(tool_call["arguments"]["content"], "# Hello, world!")
+
+        # Double-quoted still works
+        test_case = '[search(query="hello, world")]'
+        tool_call = pythonic.parse_tool_call(test_case, None)
+        self.assertEqual(tool_call["arguments"]["query"], "hello, world")
+
     def test_qwen3_coder_single_quoted_params(self):
         tools = [
             {
--- a/tests/test_tool_parsing.py
+++ b/tests/test_tool_parsing.py
@@ -155,10 +155,7 @@ class TestToolParsing(unittest.TestCase):
 
     def test_pythonic_single_quoted_args_with_commas(self):
         # LFM2.5 emits single-quoted strings; embedded commas must not truncate
-        test_case = (
-            "[write(filePath='/tmp/hello.py', "
-            "content='# Hello, world!')]"
-        )
+        test_case = "[write(filePath='/tmp/hello.py', " "content='# Hello, world!')]"
         tool_call = pythonic.parse_tool_call(test_case, None)
         self.assertEqual(tool_call["name"], "write")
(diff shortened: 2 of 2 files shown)


zcbenz (changes requested):
Can you fix lint error?

axiom-of-choice:
> Can you fix lint error?

I think it should pass now. Thanks for raising

zcbenz (approved):

PR #1799, opened 2026-08-28
fix: apply 1/sqrt(head_v_dim) readout scale in gated_delta_update

### LINKED ISSUE #1798: gated_delta_update: missing 1/sqrt(head_v_dim) readout scale — incoherent output under real weights (qwen4_exp, possibly qwen3_next)
# mlx-lm: gated_delta_update missing 1/sqrt(head_v_dim) readout scale

**Repo:** ml-explore/mlx-lm — gated_delta.py confirmed on main.
**Dupe check:** No existing issue. Closest is #1792 (dt_bias per-head indexing) — different bug.

---

## Bug

`gated_delta_update` (both the Metal kernel path and the ops fallback) does not scale its output
by `1/sqrt(head_v_dim)` before returning. The llama.cpp reference for `GATED_DELTA_NET` applies
this unconditionally at `ggml/src/ggml-cpu/ops.cpp:10824`:

    attn_data[j] = sum * scale;  // scale = 1.0f / sqrtf((float)S_v), S_v = head_v_dim

For qwen4_exp (head_v_dim = 128): scale ≈ 0.0884.

`qwen3_next.py` in mlx-lm calls `gated_delta_update` and passes raw output directly to the gated
norm with no post-scale — same omission. Whether qwen3_next is affected in practice depends on
whether the HF transformers reference applies this scale (I have not verified that; see Open
question below).

## Evidence

From-scratch MLX port of Qwen3.8-Flash-Next (qwen4_exp architecture), instrumented A/B against
llama.cpp with --eval-callback on the Q4_K_XL GGUF (coherent at greedy temp=0).

Layer 0 (linear/DeltaNet layer), first three elements of block output:

| | values |
|---|---|
| Port, no readout scale | [-0.273, 0.231, -0.252] |
| llama.cpp reference    | [-0.0558, 0.1065, -0.1232] |
| Port, with 1/sqrt(128) | [-0.0684, 0.1035, -0.1289] |

Unscaled output: ~11x too large (raw delta ÷ 0.088), several elements sign-flipped vs reference.
The gated RMS-norm partially compensates; layer-0 net divergence is ~5x. Error compounds over 48
layers.

Symptom under real weights at every quant level (4-bit, 6-bit, 8-bit experts): incoherent
generation, top-1 for "The capital of France is" was `,` at p=0.33. After adding
`out = out * (1.0 / head_v_dim**0.5)` post-`gated_delta_update`, layer-0 matched within noise,
`Paris` became top-1 at p=0.43, and greedy chat completions were clean.

All other subcomponents (PLE path, q/k normalisation, conv1d, state threading) were cleared
bit-for-bit by the same A/B. The readout scale was the only structural divergence.

## Minimal fix (caller side)

```python
out, state = gated_delta_update(q, k, v, a, b, A_log, dt_bias, state, mask)
out = out * (1.0 / head_v_dim ** 0.5)  # ggml GATED_DELTA_NET ops.cpp:10824
```

Applying it inside `gated_delta_update` would break callers that already absorb the scale into
weights — safer to apply at the model layer.

## Regression test (catches the bug)

The bug was invisible to toy-weight tests because the gated norm masked small-scale differences.
A test that reconstructs the pipeline manually and asserts output matches the *scaled* variant
but not the *raw* variant catches it reliably. Landed in our port as
`test_delta_readout_has_inv_sqrt_head_v_dim_scale`.

## Open question for maintainers

Does `modeling_qwen3_next.py` on HF apply `1/sqrt(head_v_dim)` after the delta readout? If yes,
the shipped qwen3_next path in mlx-lm is also producing ~11x-scaled delta outputs (with the gated
norm absorbing some error), and coherent-looking results come from partial compensation. If no,
qwen3_next weights were trained without this scale and only qwen4_exp-family models need it
applied on the mlx side.

Happy to send a PR for qwen4_exp. Can include qwen3_next if the HF reference confirms it applies
the scale there too.

axiom-of-choice:
## What

`gated_delta_update` omits the `1/sqrt(head_v_dim)` readout scale that the ggml reference applies unconditionally (`ops.cpp:10824`). For `head_v_dim=128` the output is ~11x too large, compounding over layers into incoherent generation.

## Fix

Scale output by `v.shape[-1] ** -0.5` inside `gated_delta_update`, after both kernel and ops paths. All four callers (qwen3_next, qwen3_5, kimi_linear, bailing_moe_v3) omit this scale, so applying it centrally is safe.

## Test

- New `test_gated_delta_readout_scale` asserts the scale is applied and is not identity.
- Updated `test_gated_delta_lower_bound` to account for the scale.
- All 5 `gated_delta` tests pass.

Fixes #1798

### DIFF
--- a/mlx_lm/models/gated_delta.py
+++ b/mlx_lm/models/gated_delta.py
@@ -610,5 +610,12 @@ def gated_delta_update(
         state = mx.zeros((B, Hv, Dv, Dk), dtype=mx.float32)
 
     if not use_kernel or mx.default_device() != mx.gpu or not mx.metal.is_available():
-        return gated_delta_ops(q, k, v, g, beta, state, mask)
-    return gated_delta_kernel(q, k, v, g, beta, state, mask)
+        y, state = gated_delta_ops(q, k, v, g, beta, state, mask)
+    else:
+        y, state = gated_delta_kernel(q, k, v, g, beta, state, mask)
+
+    # Readout scale: ggml GATED_DELTA_NET applies 1/sqrt(head_v_dim)
+    # unconditionally (ops.cpp:10824). Without it the raw delta output is
+    # ~1/head_v_dim^0.5 too large, compounding over layers.
+    y = y * (v.shape[-1] ** -0.5)
+    return y, state
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -3457,7 +3457,8 @@ class TestModels(unittest.TestCase):
             lower_bound=lower_bound,
         )
 
-        self.assertTrue(mx.allclose(output, expected))
+        # gated_delta_update applies 1/sqrt(head_v_dim) readout scale
+        self.assertTrue(mx.allclose(output, expected * (v.shape[-1] ** -0.5)))
         self.assertTrue(mx.allclose(state, expected_state))
 
     def test_gated_delta_precision(self):
@@ -3547,6 +3548,36 @@ class TestModels(unittest.TestCase):
                 y = y[:, s:e]
                 self.assertTrue(mx.allclose(y, y_gt, rtol=1e-4, atol=1e-4))
                 self.assertTrue(mx.allclose(st, st_gt, rtol=1e-4, atol=1e-3))
+
+    def test_gated_delta_readout_scale(self):
+        """gated_delta_update must scale output by 1/sqrt(head_v_dim)."""
+        mx.random.seed(0)
+        B, T = 1, 2
+        Hk, Hv, Dk, Dv = 2, 4, 16, 16
+
+        q = mx.random.normal(shape=(B, T, Hk, Dk))
+        k = mx.random.normal(shape=(B, T, Hk, Dk))
+        v = mx.random.normal(shape=(B, T, Hv, Dv))
+        a = mx.random.normal(shape=(B, T, Hv))
+        b = mx.random.normal(shape=(B, T, Hv))
+        A_log = mx.zeros((Hv,))
+        dt_bias = mx.zeros((Hv,))
+
+        # Raw (unscaled) output from the ops path
+        beta = mx.sigmoid(b)
+        g = mx.exp(-mx.exp(A_log.astype(mx.float32)) * nn.softplus(a + dt_bias))
+        state = mx.zeros((B, Hv, Dv, Dk), dtype=mx.float32)
+        y_raw, _ = gated_delta_ops(q, k, v, g, beta, state)
+
+        # Scaled output from gated_delta_update
+        y_scaled, _ = gated_delta_update(
+            q, k, v, a, b, A_log, dt_bias, use_kernel=False
+        )
+
+        scale = Dv**-0.5
+        self.assertTrue(mx.allclose(y_scaled, y_raw * scale, rtol=1e-5, atol=1e-6))
+        # Verify the scale is not identity
+        self.assertFalse(mx.allclose(y_scaled, y_raw, rtol=1e-2, atol=1e-3))
 
 
 if __name__ == "__main__":


zcbenz:
@wyanzhao Could you please take a look at this?

wyanzhao:
Sure, I will take a look into this.

wyanzhao:
Thanks for taking this on, and for the clear write-up. I went through the same three implementations before responding, and I'd like to suggest holding off on this one as written, because as far as I can tell the factor is already applied. Every in-tree caller folds the required `Dk**-0.5` into **q** before calling `gated_delta_update`, so multiplying the raw recurrence output by `Dv**-0.5` here applies it a second time at the shared-helper boundary. Details below so you can check my reasoning.

There are now six call sites across five model files:

- [`qwen3_next.py:283-287`](https://github.com/ml-explore/mlx-lm/blob/e93eaf7f491a756eafcfef2e607ff7f5210fe51a/mlx_lm/models/qwen3_next.py#L283-L287)
- [`qwen3_5.py:180-184`](https://github.com/ml-explore/mlx-lm/blob/e93eaf7f491a756eafcfef2e607ff7f5210fe51a/mlx_lm/models/qwen3_5.py#L180-L184)
- [`kimi_linear.py:352-364`](https://github.com/ml-explore/mlx-lm/blob/e93eaf7f491a756eafcfef2e607ff7f5210fe51a/mlx_lm/models/kimi_linear.py#L352-L364)
- [`bailing_moe_v3.py:79-82, 372-376`](https://github.com/ml-explore/mlx-lm/blob/e93eaf7f491a756eafcfef2e607ff7f5210fe51a/mlx_lm/models/bailing_moe_v3.py#L79-L82)
- [`kimi_k3.py:364-372`](https://github.com/ml-explore/mlx-lm/blob/e93eaf7f491a756eafcfef2e607ff7f5210fe51a/mlx_lm/models/kimi_k3.py#L364-L372) and [`450-460`](https://github.com/ml-explore/mlx-lm/blob/e93eaf7f491a756eafcfef2e607ff7f5210fe51a/mlx_lm/models/kimi_k3.py#L450-L460)

They all use the same algebraic pattern:

```python
inv_scale = k.shape[-1] ** -0.5
q = (inv_scale**2) * mx.fast.rms_norm(q, None, eps)
k = inv_scale * mx.fast.rms_norm(k, None, eps)
```

Because `rms_norm(x, eps_rms) = sqrt(D) * x / sqrt(sum(x**2) + D * eps_rms)`, this pattern already gives q the extra `Dk**-0.5` factor relative to k. It is exactly `q = l2norm(q) * Dk**-0.5`, `k = l2norm(k)` in the Kimi/Bailing callers that use `eps_rms = eps_l2 / Dk`. The two Qwen callers pass `eps_rms=1e-6` directly, so they use a slightly different epsilon convention, but the `Dk**-0.5` scale is still already present.

That matches the HF reference. Both [`torch_chunk_gated_delta_rule`](https://github.com/huggingface/transformers/blob/ac3244569528944b9d5773cafea525cd8a8b63de/src/transformers/models/qwen3_next/modeling_qwen3_next.py#L419-L425) and [`torch_recurrent_gated_delta_rule`](https://github.com/huggingface/transformers/blob/ac3244569528944b9d5773cafea525cd8a8b63de/src/transformers/models/qwen3_next/modeling_qwen3_next.py#L539-L545) always scale **q** by `query.shape[-1] ** -0.5`, i.e. `1/sqrt(Dk)`. The recurrent state does not depend on q and the readout is linear in q, so this factor can be moved from q to the readout, but it must be applied only once.

The ggml line cited by the PR is that same single factor in a different location. llama.cpp's two non-fused builders scale q by `1/sqrt(S_k)` ([chunk path](https://github.com/ggml-org/llama.cpp/blob/b356fa2624643b6d5753162ae43efff8cdd4d8cb/src/models/delta-net-base.cpp#L25-L47), [recurrent path](https://github.com/ggml-org/llama.cpp/blob/b356fa2624643b6d5753162ae43efff8cdd4d8cb/src/models/delta-net-base.cpp#L299-L321)). Its fused builder skips the q-side scale and the op scales the readout by `1/sqrt(S_v)` ([builder](https://github.com/ggml-org/llama.cpp/blob/b356fa2624643b6d5753162ae43efff8cdd4d8cb/src/models/delta-net-base.cpp#L381-L402), [CPU op](https://github.com/ggml-org/llama.cpp/blob/b356fa2624643b6d5753162ae43efff8cdd4d8cb/ggml/src/ggml-cpu/ops.cpp#L10965-L11031)). All three llama.cpp builders assert `S_k == S_v`, so its fused implementation cannot distinguish those constants. It is not evidence for adding an independent `Dv**-0.5` factor to mlx-lm, which supports `Dk != Dv`.

I also compared `gated_delta_update` against a NumPy port of HF's recurrent rule in fp32, using the q/k preprocessing from the corresponding mlx-lm callers. For the Kimi-style case, the reference applies the same recurrence elementwise to vector decay. The check covers the Metal kernel and ops paths, `Dk != Dv`, and scalar and vector decay:

| case | main `e93eaf7` rel. max diff (kernel / ops) | PR `690a242` fitted output/reference (kernel / ops) |
|---|---:|---:|
| Qwen3-Next defaults: Hk=16, Hv=32, Dk=Dv=128 | `1.347e-6 / 1.312e-6` | `0.088388 / 0.088388` |
| `qwen3_5.py` ModelArgs defaults: Hk=16, Hv=64, Dk=192, Dv=128 | `9.426e-7 / 9.899e-7` | `0.088388 / 0.088388` |
| Kimi-style vector decay, Dk=Dv=128 | `1.667e-7 / 1.848e-7` | `0.088388 / 0.088388` |

For all three PR cases, the relative max error is `0.9116`. The bounded conclusion here is about the raw `gated_delta_update` output: this PR adds another `Dv**-0.5` multiplier, which is `0.088388` for `Dv=128`. I am not claiming an 11.3x end-to-end logit or generation change; every current model applies a downstream RMS normalization that can partly absorb a uniform scale.

Two additional points:

- The existing [`test_gated_delta_lower_bound`](https://github.com/ml-explore/mlx-lm/blob/e93eaf7f491a756eafcfef2e607ff7f5210fe51a/tests/test_models.py#L3539-L3567) already detects the helper-contract change. Running this PR's implementation with that unmodified test fails at `assertTrue(mx.allclose(output, expected))`. The PR changes the expectation to `expected * Dv**-0.5` and adds a test that only checks the newly introduced multiplier against the raw ops path; neither change compares the new behavior with an external reference.
- The branch currently conflicts with main because #1626 added the `k.shape[-1]` kernel guard in the same hunk. The PR description also counts four callers, while current main has six call sites in five models.

Regarding #1798: the reporter's from-scratch Qwen3.8-Flash-Next port is not available in mlx-lm, so its exact preprocessing cannot be inspected here. The reported approximately 11x discrepancy is consistent with that out-of-tree port missing the q-side `Dk**-0.5` factor, but it does not establish that as fact. If the port omitted that preprocessing, the correction belongs at its model call site, as in the five in-tree models; it does not justify adding an unconditional second scale to the shared helper. The issue itself also cautions that applying the scale inside `gated_delta_update` can break callers that already absorb it.

I would be glad to review a follow-up that documents the helper contract (q is expected pre-normalized with the `Dk**-0.5` factor) and/or adds an HF-reference comparison test so this cannot be reintroduced. The comparison script is included below.

<details>
<summary>Comparison script (run as <code>python check_scale_revised.py &lt;mlx-lm checkout&gt;</code>)</summary>

```python
#!/usr/bin/env python3
"""Compare mlx-lm gated_delta_update with an HF-derived recurrence."""

import hashlib
import importlib.util
import subprocess
import sys
from pathlib import Path

import mlx.core as mx
import numpy as np


if len(sys.argv) != 2:
    raise SystemExit("usage: python check_scale_revised.py <mlx-lm checkout>")

root = Path(sys.argv[1]).resolve()
source = root / "mlx_lm/models/gated_delta.py"
spec = importlib.util.spec_from_file_location("gd", source)
if spec is None or spec.loader is None:
    raise RuntimeError(f"cannot load {source}")
gd = importlib.util.module_from_spec(spec)
spec.loader.exec_module(gd)


def git_revision(path: Path) -> str:
    result = subprocess.run(
        ["git", "-C", str(path), "rev-parse", "HEAD"],
        check=True,
        capture_output=True,
        text=True,
    )
    return result.stdout.strip()


print("checkout:", root)
print("revision:", git_revision(root))
print("gated_delta.py sha256:", hashlib.sha256(source.read_bytes()).hexdigest())
print("mlx:", mx.__version__, "| device:", mx.default_device(), "| metal:", mx.metal.is_available())


def softplus(x: np.ndarray) -> np.ndarray:
    return np.log1p(np.exp(-np.abs(x))) + np.maximum(x, 0)


def sigmoid(x: np.ndarray) -> np.ndarray:
    return 1 / (1 + np.exp(-x))


def l2norm(x: np.ndarray, eps: float = 1e-6) -> np.ndarray:
    return x / np.sqrt((x * x).sum(-1, keepdims=True) + eps)


def hf_reference(
    q: np.ndarray,
    k: np.ndarray,
    v: np.ndarray,
    g_log: np.ndarray,
    beta: np.ndarray,
) -> np.ndarray:
    """NumPy port of HF torch_recurrent_gated_delta_rule.

    The vector-decay case is its elementwise extension used by MLX KDA callers.
    """
    q = l2norm(q) * (q.shape[-1] ** -0.5)
    k = l2norm(k)
    batch, steps, h_k, d_k = k.shape
    h_v, d_v = v.shape[-2:]
    repeat_factor = h_v // h_k
    q = np.repeat(q, repeat_factor, axis=2)
    k = np.repeat(k, repeat_factor, axis=2)

    state = np.zeros((batch, h_v, d_k, d_v))
    output = np.zeros((batch, steps, h_v, d_v))
    for t in range(steps):
        decay = np.exp(g_log[:, t])
        if decay.ndim == 3:  # vector decay [B,Hv,Dk]
            state *= decay[..., None]
        else:  # scalar decay [B,Hv]
            state *= decay[..., None, None]

        k_t = k[:, t]
        q_t = q[:, t]
        v_t = v[:, t]
        beta_t = beta[:, t][..., None]
        memory = (state * k_t[..., None]).sum(-2)
        delta = (v_t - memory) * beta_t
        state += k_t[..., None] * delta[..., None, :]
        output[:, t] = (state * q_t[..., None]).sum(-2)
    return output


def relative_max(actual: np.ndarray, reference: np.ndarray) -> float:
    return float(np.abs(actual - reference).max() / np.abs(reference).max())


def fitted_scale(actual: np.ndarray, reference: np.ndarray) -> float:
    actual_flat = actual.reshape(-1)
    reference_flat = reference.reshape(-1)
    denominator = np.vdot(reference_flat, reference_flat).real
    return float(np.vdot(reference_flat, actual_flat).real / denominator)


def run_case(
    name: str,
    *,
    batch: int,
    steps: int,
    h_k: int,
    h_v: int,
    d_k: int,
    d_v: int,
    rms_eps: float,
    vector_gate: bool,
    seed: int = 0,
) -> None:
    rng = np.random.default_rng(seed)
    q = rng.standard_normal((batch, steps, h_k, d_k))
    k = rng.standard_normal((batch, steps, h_k, d_k))
    v = rng.standard_normal((batch, steps, h_v, d_v))

    gate_shape = (batch, steps, h_v, d_k) if vector_gate else (batch, steps, h_v)
    parameter_shape = (h_v, d_k) if vector_gate else (h_v,)
    a = rng.standard_normal(gate_shape)
    b = rng.standard_normal((batch, steps, h_v))
    A_log = rng.standard_normal(parameter_shape) * 0.5
    dt_bias = rng.standard_normal(parameter_shape) * 0.5
    g_log = -np.exp(A_log) * softplus(a + dt_bias)
    beta = sigmoid(b)
    reference = hf_reference(q, k, v, g_log, beta)

    f32 = lambda x: mx.array(x, dtype=mx.float32)
    q_mlx, k_mlx, v_mlx = f32(q), f32(k), f32(v)
    inv_scale = d_k**-0.5
    q_mlx = (inv_scale**2) * mx.fast.rms_norm(q_mlx, None, rms_eps)
    k_mlx = inv_scale * mx.fast.rms_norm(k_mlx, None, rms_eps)

    gate_kind = "vector" if vector_gate else "scalar"
    print(
        f"\n[{name}] B={batch} T={steps} Hk={h_k} Hv={h_v} "
        f"Dk={d_k} Dv={d_v} gate={gate_kind} rms_eps={rms_eps:g}"
    )
    for use_kernel in (True, False):
        output, _ = gd.gated_delta_update(
            q_mlx,
            k_mlx,
            v_mlx,
            f32(a),
            f32(b),
            f32(A_log),
            f32(dt_bias),
            use_kernel=use_kernel,
        )
        output = np.array(output.astype(mx.float32), dtype=np.float64)
        path = "kernel" if use_kernel else "ops"
        print(
            f"  {path:6s} rel.max.diff={relative_max(output, reference):.3e} "
            f"fitted(output/reference)={fitted_scale(output, reference):.6f}"
        )


run_case(
    "Qwen3-Next defaults",
    batch=1,
    steps=8,
    h_k=16,
    h_v=32,
    d_k=128,
    d_v=128,
    rms_eps=1e-6,
    vector_gate=False,
)
run_case(
    "qwen3_5.py ModelArgs defaults",
    batch=1,
    steps=8,
    h_k=16,
    h_v=64,
    d_k=192,
    d_v=128,
    rms_eps=1e-6,
    vector_gate=False,
)
run_case(
    "Kimi-style vector gate",
    batch=1,
    steps=8,
    h_k=8,
    h_v=8,
    d_k=128,
    d_v=128,
    rms_eps=1e-6 / 128,
    vector_gate=True,
)
```

</details>

axiom-of-choice:
Thanks,  the measurement settles it. 

The six in-tree callers already fold `Dk**-0.5` into q (via the `inv_scale`/`inv_scale**2` pattern), so applying `Dv**-0.5` at the helper boundary double-scales the output (fitted `0.088388` = `1/sqrt(128)`), which would break every caller rather than fix #1798.

I'll close this PR and take the follow-up instead: document the helper contract (q expected pre-normalized with the `Dk**-0.5` factor) and add an HF-reference comparison test so the scale can't be reintroduced.

axiom-of-choice:
Superseded by #1823, which documents the helper contract (callers fold `Dk**-0.5` into q) and pins it with an HF-reference test instead of adding a second scale.

axiom-of-choice:
Superseded by #1823.

PR #1800, opened 2026-08-28
Fix MLA quantized-KV crash in deepseek_v3 and glm4_moe_lite

katlun-lgtm:
File "mlx_lm/models/deepseek_v3.py", line 147, in __call__
  AttributeError: 'tuple' object has no attribute 'swapaxes'

  main @ 96830a6:  FAILED (errors=5)
    ValueError: [broadcast_shapes] (2,2,4,4,64) and (2,2,4,1,4,64) cannot be broadcast
    ValueError: [reshape] Cannot reshape array of size 524288 into shape (2,64,1,64)
  this branch:     Ran 84 tests — OK (skipped=1)
  pre-commit:      black Passed, isort Passed

  Moonlight-16B-A3B-Instruct-4bit, M3 Max, wikitext-2 ppl @ 1k tokens
  fp16 6.64 · kv8 6.65 · kv4 6.80 · kv3 7.62   (--kv-bits 8 greedy output identical to fp16)

  The MLA path treats the cache output as an array, but a quantized cache returns a three-element tuple.
  
  The previous rank check applied the mask to the wrong head axis when multiple query heads share a KV head.

 
- AI usage disclosure: 
  - Claude Code wrote the patch (deepseek_v3.py, glm4_moe_lite.py, base.py, the two tests)
  - it ran the test runs and the ppl measurements on my M3 Max; every number in the PR is its tool output
  - it merged current main into the branch and pushed the branch to my fork
  - i reviewed the diff, understand it, and own it
  - the PR text is my writing

### DIFF
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -90,11 +90,14 @@ def quantized_scaled_dot_product_attention(
             q_indices = mx.arange(kL - qL, kL)
             k_indices = mx.arange(kL)
             mask = q_indices[:, None] >= k_indices[None]
-        if n_repeats > 1 and mask.ndim >= 3:
-            if mask.shape[-3] == 1:
-                mask = mx.expand_dims(mask, -3)
-            else:
+        if n_repeats > 1 and 3 <= mask.ndim < scores.ndim:
+            if mask.shape[-3] == n_q_heads:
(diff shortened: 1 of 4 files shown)


zcbenz (changes requested):
We don't do manual `quantized_matmul` and `dequantize` in model implementations.

PR #1774, opened 2026-08-22
Add tensor parallel sharding for Qwen3 MoE

azamamirza:
Add tensor parallel sharding for Qwen3 MoE

## Problem

`qwen3_moe` has no `Model.shard()`, so Qwen3 MoE checkpoints (e.g.
Qwen3-30B-A3B, Qwen3-235B-A22B) cannot be tensor-parallel sharded across
multiple machines the way `glm4_moe`, `deepseek_v3`, and other MoE models
already can.

## Change

Follows the established `glm4_moe` pattern exactly:

- `Qwen3MoeSparseMoeBlock` gains a `sharding_group`; when set, the block
  applies `sum_gradients` on the input and `mx.distributed.all_sum` on the
  output. The router gate stays replicated. The expert-weighted sum gets
  `.astype(y.dtype)` for dtype parity with `glm4_moe`.
- `Model.shard(group)`: attention `q/k/v_proj` sharded all-to-sharded,
  `o_proj` sharded-to-all, with `n_heads` / `n_kv_heads` divided by the
  group size; dense-MLP layers (`mlp_only_layers` / `decoder_sparse_step`)
  sharded via `shard_linear`; MoE layers sharded in place on
  `switch_mlp.{gate,down,up}_proj` via `shard_inplace`.
- Adds a tiny `qwen3_moe` config (including a dense `mlp_only_layers=[0]`
  layer so both MLP branches are exercised) to
  `tests/model_parallel_tests.py::test_shard`.

## Verification

- `tests/model_parallel_tests.py`: 1 passed, 4 subtests passed (the new
  qwen3_moe subtest shards with a size-1 group and matches the unsharded
  output).
- `tests/test_models.py -k qwen3`: 3 passed.
- Hardware verification of the same change (originally applied on top of
  the 0.31.3 release): bit-exact vs unsharded at group size 1 (float and
  4-bit); max abs logit diff ~3e-8 at group size 2 on a local ring (float
  and 4-bit); sustained 2-node run of mlx-community/Qwen3-235B-A22B-4bit
  sharded across two Apple Silicon machines over RDMA (Thunderbolt 5),
  producing coherent generations.
- `black` / `isort --profile=black` clean.

PR #1801, opened 2026-08-29
Add pipeline parallelism for Qwen3 MoE

azamamirza:
## Problem

`qwen3_moe` has no pipeline support, so Qwen3 MoE checkpoints cannot be
split layer-wise across machines the way `glm4_moe`, `deepseek_v3`, and
other MoE models already can via `PipelineMixin`. (Companion to the
tensor-parallel `shard()` added in #1774; this adds the pipeline axis.)

## Change

Follows the established `glm4_moe` / `deepseek_v3` pattern exactly:

- `Qwen3MoeModel` mixes in `PipelineMixin` and its `__call__` gains the
  standard pipeline plumbing: `recv_like` from the previous rank, run
  `pipeline_layers`, `send` to the next rank with the `mx.depends` tie on
  the last cache, and the final `all_gather` broadcast. With
  `pipeline_size == 1` (the default) every branch is skipped and behavior
  is unchanged.
- `Model.layers` returns `self.model.pipeline_layers` (as in `glm4_moe`).
- Adds `tests/model_parallel_tests.py::test_pipeline`, a structural test
  in the style of the existing `test_shard`: a tiny `qwen3_moe` config is
  pipelined on a size-1 group and must match the un-pipelined output.

## Verification

Rebased onto current `main` (which already includes #1774 and #1787):

- `tests/model_parallel_tests.py`: 2 passed, 5 subtests passed.
- `tests/test_models.py -k qwen3`: 3 passed.
- Multi-node hardware verification (2 Apple Silicon MacBooks over an RDMA
  JACCL ring, Thunderbolt 5), `Qwen3-235B-A22B` 4-bit, greedy, identical
  prompt: 2-node **pipeline 28.41 tok/s decode, TTFT 7.23 s** vs 2-node
  tensor-parallel sharding 24.56 tok/s / TTFT 11.91 s (+16% decode,
  -39% TTFT); generated text checked coherent and consistent between the
  two modes.
- `black` / `isort --profile=black` clean.

### DIFF
--- a/mlx_lm/models/qwen3_moe.py
+++ b/mlx_lm/models/qwen3_moe.py
@@ -9,6 +9,7 @@ from mlx.nn.layers.distributed import shard_inplace, shard_linear, sum_gradients
 
 from .activations import swiglu
 from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .pipeline import PipelineMixin
 from .switch_layers import SwitchGLU
 
 
@@ -181,7 +182,7 @@ class Qwen3MoeDecoderLayer(nn.Module):
         return out
 
 
-class Qwen3MoeModel(nn.Module):
+class Qwen3MoeModel(PipelineMixin, nn.Module):
     def __init__(self, args: ModelArgs):
         super().__init__()
         self.args = args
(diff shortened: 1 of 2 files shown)


nastya236:
Thank you for your contribution, will test it as soon as possible.

nastya236 (approved):
Looks good to me! Tested on 4 m5 ultra mesh.

PR #1804, opened 2026-08-29
Feat/server kv cache quant partial

soobrosa:
# Problem

`mlx-lm`'s server has no way to enable KV cache quantization. The library supports it (`kv_bits`, `kv_group_size`, `quantized_kv_start`), but only through the Python API, so anyone running `mlx_lm.server` was stuck with an unquantized cache that grows linearly with context length. `max_kv_size` wasn't exposed either.

# Solution

Adds `--kv-bits`, `--kv-group-size`, `--quantized-kv-start`, and `--max-kv-size` to the server CLI and threads them through to generate.

Not every cache type supports quantization. Without a check, the server would start normally, quantize whatever layers happen to support it, and silently leave the rest alone. So model load now inspects the cache: if any layer can't be quantized with the requested settings, it logs which layers and how many, and the server exits with a non-zero code.

Noted that the model loads in a background thread. Raising there just kills the thread, and the main server keeps accepting connections that block forever waiting on a model that never arrives. The main thread now joins on the first load and exits if it failed.

Tradeoff: with `--kv-bits` set, requests are processed one at a time. The batched path doesn't support quantized caches yet. This is documented in the flag's help text.

- AI usage disclosure: 
I use AI as a helper but I do my judgment and apply as a human actor.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -292,7 +292,12 @@ def maybe_quantize_kv_cache(prompt_cache, quantized_kv_start, kv_group_size, kv_
         return
     for e, c in enumerate(prompt_cache):
         if hasattr(c, "to_quantized") and c.offset >= quantized_kv_start:
-            prompt_cache[e] = c.to_quantized(group_size=kv_group_size, bits=kv_bits)
+            try:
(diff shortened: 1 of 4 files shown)


michalk8:
Closing this in favor of #1832, as this PR contains some extra stuff I'd consider adding in a future PR (like failing early at the start-up, etc.)

PR #1810, opened 2026-08-31
fix: plumb evaluate batch-size to generate_until, accept max_gen_toks

### LINKED ISSUE #1808: mlx_lm.evaluate: Metal Insufficient Memory on generative tasks on 16 GB Macs (--batch-size not applied to generation, max_gen_toks ignored)
### Environment

- MacBook Air M5 (base), 16 GB unified memory, macOS 26.6.1
- mlx 0.32.2 and mlx 0.31.2 (both reproduce), mlx-lm 0.31.3, lm_eval 0.4.12
- Model: `mlx-community/Qwen3-4B-Instruct-2507-4bit` (3.35 GB)

### What happens

```
mlx_lm.evaluate --model mlx-community/Qwen3-4B-Instruct-2507-4bit \
  --task mmlu_pro --limit 2 --batch-size 1
```

crashes during generation:

```
RuntimeError: [METAL] Command buffer execution failed: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
```

System memory is >70% free at launch, and `--batch-size` 1 vs 2 vs default makes no difference. On the same machine, `mlx_lm.generate` with this model and LoRA training on a 9B model both work fine, so generative eval tasks look effectively unusable on 16 GB Macs.

### Why (from reading evaluate.py / generate.py)

Two things compound:

1. `EvalHarnessLM.generate_until` calls `batch_generate` without any batch size, so `BatchGenerator`'s default `completion_batch_size=32` applies and every request decodes concurrently (mmlu_pro at `--limit 2` is 28 requests, all at once). The `--batch-size` flag is only used in the loglikelihood scoring loop, which is why it has no effect here.
2. `generate_until` reads `opt.get("max_gen_tokens", DEFAULT_MAX_TOKENS)`, but lm-eval passes the key `max_gen_toks`, so per-task generation caps are discarded and every request gets `DEFAULT_MAX_TOKENS = 8192`.

28 concurrent sequences with ~2k-token prompts and up to 8k generation each means the KV caches grow toward tens of GB on a 4B model, a genuine OOM on 16 GB.

### Confirmation

Monkeypatching the call with `completion_batch_size=4, prefill_batch_size=2` makes the exact same command complete normally (28/28 samples, exact_match 0.679, in line with published full-set numbers for this quant).

A fix could be to plumb `--batch-size` through to `batch_generate` as `completion_batch_size` and to accept `max_gen_toks` alongside `max_gen_tokens`.

axiom-of-choice:
Closes #1808.

## Bugs
Two bugs in `mlx_lm.evaluate` that make generative eval tasks unusable
on 16 GB Macs:

1. `generate_until` calls `batch_generate` without batch-size args, so
   `BatchGenerator`'s default `completion_batch_size=32` applies and
   every request decodes concurrently. The `--batch-size` flag was only
   used in the loglikelihood scoring loop.
2. `max_tokens` reads `opt.get("max_gen_tokens", ...)` but lm-eval
   passes the key `max_gen_toks`, so per-task generation caps are
   discarded and every request gets `DEFAULT_MAX_TOKENS = 8192`.

## What is the fix

- Pass `completion_batch_size=self._batch_size` and
  `prefill_batch_size=max(self._batch_size // 2, 1)` to `batch_generate`
- Accept `max_gen_toks` as a fallback key alongside `max_gen_tokens`

## Checks

- `pre-commit run --all-files` (black + isort) — clean

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -343,7 +343,8 @@ class MLXLM(LM):
 
         # TODO consider multi-token, per-prompt stop conditions
         max_tokens = [
-            self._max_tokens or opt.get("max_gen_tokens", DEFAULT_MAX_TOKENS)
+            self._max_tokens
+            or opt.get("max_gen_tokens", opt.get("max_gen_toks", DEFAULT_MAX_TOKENS))
             for opt in options
         ]
 
@@ -354,6 +355,8 @@ class MLXLM(LM):
             max_tokens=max_tokens,
             verbose=True,
             sampler=self._sampler,
+            completion_batch_size=self._batch_size,
+            prefill_batch_size=max(self._batch_size // 2, 1),
         ).texts
 
         for e, (text, opt) in enumerate(zip(completions, options)):


mlx_lm/evaluate.py
```diff
@@ -343,7 +343,8 @@ def generate_until(self, requests) -> list[str]:
 
         # TODO consider multi-token, per-prompt stop conditions
         max_tokens = [
-            self._max_tokens or opt.get("max_gen_tokens", DEFAULT_MAX_TOKENS)
+            self._max_tokens
+            or opt.get("max_gen_tokens", opt.get("max_gen_toks", DEFAULT_MAX_TOKENS))
```
michalk8, line 347: I think we can drop `max_gen_tokens` and use `max_gen_toks`.

mlx_lm/evaluate.py
```diff
@@ -354,6 +355,8 @@ def generate_until(self, requests) -> list[str]:
             max_tokens=max_tokens,
             verbose=True,
             sampler=self._sampler,
+            completion_batch_size=self._batch_size,
+            prefill_batch_size=max(self._batch_size // 2, 1),
```
michalk8, line 359: Why halve this? Would just use the `self._batch_size`.

michalk8 (changes requested):
Can you please also add one test?

michalk8:
Actually just saw #1812, closing this in favor of that PR!

PR #1812, opened 2026-09-01
Fix/mlx_lm.evaluate ignoring max_gen_toks and --batch-size during generation

### LINKED ISSUE #1808: mlx_lm.evaluate: Metal Insufficient Memory on generative tasks on 16 GB Macs (--batch-size not applied to generation, max_gen_toks ignored)

dasashreeya:
Fixes #1808

At `evaluate.py:346`, the code reads `max_gen_tokens`, while `lm-eval` emits `max_gen_toks`, so it falls back to 8192 tokens instead of applying `mmlu_pro`’s 2048-token limit. At `evaluate.py:350`, `--batch-size` is not passed to `batch_generate`, leaving `BatchGenerator`’s 8/32 defaults in effect; the flag is only read in the loglikelihood loop at `evaluate.py:303`.

```bash
mlx_lm.evaluate --model mlx-community/Qwen3-4B-Instruct-2507-4bit \
  --task mmlu_pro --limit 2 --batch-size 1
```

On main, this fails with `[METAL] Command buffer execution failed: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)`. Fixing the token key alone still crashes. With both fixes, peak memory is 3.44 GiB at batch size 1, 10.13 GiB at 4, and 16.97 GiB at the new default of 8.

Prefill memory is 5.41 / 8.21 / 9.49 GiB at batch sizes 4 / 8 / 16. Passing the old default of 16 into prefill would use 1.28 GiB more than 8, so the default moves to 8.

On 70 samples at batch size 1, the `mmlu_pro` exact-match score changes from 0.6143 to 0.5857. Ten completions exceed 2048 tokens, with a maximum of 5140, so the existing `BENCHMARKS.md` result produced under the 8192-token budget may decrease slightly when regenerated; `n=70` is small.

Three tests were added; two fail on main, while the `--max-tokens` precedence test already passes.

Per CONTRIBUTING: I directed the investigation, code, tests, and measurements using Claude Code; the runs were on my M4 Pro. I reviewed the diff, and I take responsibility for every line.

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -342,8 +342,9 @@ class MLXLM(LM):
         ]
 
         # TODO consider multi-token, per-prompt stop conditions
+        # lm-eval names the per-task generation cap "max_gen_toks".
         max_tokens = [
-            self._max_tokens or opt.get("max_gen_tokens", DEFAULT_MAX_TOKENS)
+            self._max_tokens or opt.get("max_gen_toks", DEFAULT_MAX_TOKENS)
             for opt in options
         ]
 
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -355,6 +355,8 @@ class MLXLM(LM):
             max_tokens=max_tokens,
             verbose=True,
             sampler=self._sampler,
+            prefill_batch_size=self._batch_size,
+            completion_batch_size=self._batch_size,
         ).texts
 
         for e, (text, opt) in enumerate(zip(completions, options)):
@@ -406,7 +408,13 @@ def main():
     parser.add_argument(
         "--output-dir", default=".", help="Output directory for result files."
     )
-    parser.add_argument("--batch-size", type=int, default=16, help="Batch size")
+    parser.add_argument(
+        "--batch-size",
+        type=int,
+        default=8,
+        help="Maximum number of sequences processed at once, both when "
+        "scoring and when generating.",
+    )
     parser.add_argument("--num-shots", type=int, default=None, help="Number of shots")
     parser.add_argument(
         "--max-tokens",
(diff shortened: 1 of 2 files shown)


michalk8 (approved):
LGTM, thanks!

PR #1815, opened 2026-09-01
fix: gate wired_limit on the active device

### LINKED ISSUE #1813: wired_limit() raises KeyError on Apple Silicon when the default device is CPU
## Summary

I'm trying to immigrate [XHToken/Spark-X2.5-4B](https://huggingface.co/XHToken/Spark-X2.5-4B) into mlx_lm.

`wired_limit()` in `mlx_lm/generate.py` gates on `mx.metal.is_available()`, but then queries a key that only exists in the **GPU** device info. On Apple Silicon, Metal is available even when the default device has been set to CPU, so any generation with `mx.set_default_device(mx.cpu)` dies with:

```
KeyError: 'max_recommended_working_set_size'
```

The guard should test the device actually in use, not whether Metal exists on the machine.

## Reproduction

Environment: macOS 15 (Darwin 25.5.0), Apple M5 Pro, Python 3.11.15, `mlx` 0.32.2, `mlx-lm` 0.31.3.

No model needed — a 4×4 linear layer reproduces it:

```python
import mlx.core as mx
import mlx.nn as nn
from mlx_lm.generate import wired_limit

m = nn.Linear(4, 4)

print(mx.metal.is_available())   # True
print(mx.default_device())       # Device(gpu, 0)

with wired_limit(m):             # fine
    pass

mx.set_default_device(mx.cpu)
print(mx.default_device())       # Device(cpu, 0)

with wired_limit(m):             # KeyError: 'max_recommended_working_set_size'
    pass
```

Through the normal generation path:

```python
import mlx.core as mx
mx.set_default_device(mx.cpu)

model, tokenizer = load("mlx-community/Llama-3.2-3B-Instruct-4bit")
generate(model, tokenizer, prompt="hi", max_tokens=8)
```

```
  File ".../mlx_lm/generate.py", line 705, in stream_generate
    with wired_limit(model, [generation_stream]):
  File ".../mlx_lm/generate.py", line 238, in wired_limit
    max_rec_size = mx.device_info()["max_recommended_working_set_size"]
KeyError: 'max_recommended_working_set_size'
```

## Cause

`mx.device_info()` returns different keys per device:

```python
mx.set_default_device(mx.gpu); sorted(mx.device_info())
# ['architecture', 'device_name', 'max_buffer_length',
#  'max_recommended_working_set_size', 'memory_size', 'resource_limit']

mx.set_default_device(mx.cpu); sorted(mx.device_info())
# ['architecture', 'device_name']
```

`wired_limit()` branches on machine capability rather than the active device:

```python
if not mx.metal.is_available():
    yield          # taken on Linux/CUDA, where metal is absent
else:
    ...
    max_rec_size = mx.device_info()["max_recommended_working_set_size"]   # CPU -> KeyError
```

On a CUDA host `mx.metal.is_available()` is `False`, so the early return hides the problem. On Apple Silicon with CPU selected, it takes the `else` branch and queries a GPU-only key.

## CPU generation works once this guard is bypassed

To confirm the guard is the only blocker, I monkey-patched `wired_limit` to a no-op context manager and changed nothing else, then ran a full generation on CPU with an 8-bit quantized model:

```python
import contextlib, sys, mlx.core as mx
mx.set_default_device(mx.cpu)
import mlx_lm.generate
G = sys.modules["mlx_lm.generate"]

@contextlib.contextmanager
def _noop(*a, **k):
    yield
G.wired_limit = _noop
# ... load + generate as usual
```

```
device: Device(cpu, 0)
==========
2226
==========
Prompt: 28 tokens, 0.061 tokens-per-sec
Generation: 5 tokens, 0.070 tokens-per-sec
Peak memory: 4.413 GB
```

Correct output, so nothing else in the generation path objects to running on CPU. (Throughput is of course terrible — dequantizing an affine-quantized model on CPU is not an optimized path. The value here is debugging and cross-checking GPU results, not speed.)

## Suggested fix

Gate on the active device:

```python
if not mx.metal.is_available() or mx.default_device() != mx.gpu:
    try:
        yield
    finally:
        pass
else:
    ...
```

Or defensively, `mx.device_info().get("max_recommended_working_set_size")` and skip the wiring when it's absent — that also covers future backends whose device info lacks the key.

## Why this may be worth fixing despite #235

I'm aware of #235 ("How to use CPU for inference?") and that CPU inference isn't a priority. Two reasons this still seems worth a small patch:

1. **It's a crash, not a missing feature.** `mx.set_default_device(mx.cpu)` is a documented MLX call, and everything downstream of `wired_limit` works fine on CPU — I ran a full 4B model end-to-end on CPU by bypassing only this function. The three-line guard is the only thing standing in the way.
2. **CPU is the reference path for debugging Metal numerics.** I hit this while cross-checking a fused-SDPA divergence between backends (Metal prefill differs from CPU by ~5e-4 at every head_dim). CPU execution is how you establish ground truth for that class of bug.

For context on how this surfaces in the wild: [XHToken/Spark-MLX-LLM](https://github.com/XHToken/Spark-MLX-LLM) documents `--device cpu` as its primary Quick Start command. That command cannot run on Apple Silicon with mlx-lm 0.31.3 — it was validated on CUDA, where the early return hides the issue.

Happy to send a PR.

axiom-of-choice:
`wired_limit()` and the other `set_wired_limit()` call sites gate on `mx.metal.is_available()`, but query `device_info()["max_recommended_working_set_size"]`, a GPU-only key. On Apple Silicon with `mx.set_default_device(mx.cpu)`, Metal is available but the key is missing, so generation dies with `KeyError`.

Gate on the active device instead, so the wired-limit path is skipped on CPU. Fixes #1813.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -1594,7 +1594,7 @@ class BatchGenerator:
         self._gen_tokens_counter = 0
         self._steps_counter = 0
 
-        if mx.metal.is_available():
+        if mx.metal.is_available() and mx.default_device() == mx.gpu:
             self._old_wired_limit = mx.set_wired_limit(
                 mx.device_info()["max_recommended_working_set_size"]
             )
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -229,7 +229,7 @@ def wired_limit(model: nn.Module, streams: Optional[List[mx.Stream]] = None):
     async eval could be running pass in the streams to synchronize with prior
     to exiting the context manager.
     """
-    if not mx.metal.is_available():
+    if not mx.metal.is_available() or mx.default_device() != mx.gpu:
         try:
             yield
         finally:
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -26,7 +26,11 @@ from .models.cache import (
 )
 from .sample_utils import make_sampler
 from .tokenizer_utils import TokenizerWrapper
-from .utils import does_model_support_input_embeddings, load
+from .utils import (
+    does_model_support_input_embeddings,
+    load,
+    maybe_set_recommended_wired_limit,
+)
 
 DEFAULT_PROMPT = "hello"
 DEFAULT_MAX_TOKENS = 100
@@ -229,35 +233,33 @@ def wired_limit(model: nn.Module, streams: Optional[List[mx.Stream]] = None):
     async eval could be running pass in the streams to synchronize with prior
     to exiting the context manager.
     """
-    if not mx.metal.is_available() or mx.default_device() != mx.gpu:
-        try:
-            yield
-        finally:
-            pass
-    else:
-        model_bytes = tree_reduce(
-            lambda acc, x: acc + x.nbytes if isinstance(x, mx.array) else acc, model, 0
+    old_limit = maybe_set_recommended_wired_limit()
+    if old_limit is None:
+        yield
+        return
+
+    model_bytes = tree_reduce(
+        lambda acc, x: acc + x.nbytes if isinstance(x, mx.array) else acc, model, 0
+    )
+    max_rec_size = mx.device_info()["max_recommended_working_set_size"]
+    if model_bytes > 0.9 * max_rec_size:
+        model_mb = model_bytes // 2**20
+        max_rec_mb = max_rec_size // 2**20
+        print(
+            f"[WARNING] Generating with a model that requires {model_mb} MB "
+            f"which is close to the maximum recommended size of {max_rec_mb} "
+            "MB. This can be slow. See the documentation for possible work-arounds: "
+            "https://github.com/ml-explore/mlx-lm/tree/main#large-models"
         )
-        max_rec_size = mx.device_info()["max_recommended_working_set_size"]
-        if model_bytes > 0.9 * max_rec_size:
-            model_mb = model_bytes // 2**20
-            max_rec_mb = max_rec_size // 2**20
-            print(
-                f"[WARNING] Generating with a model that requires {model_mb} MB "
-                f"which is close to the maximum recommended size of {max_rec_mb} "
-                "MB. This can be slow. See the documentation for possible work-arounds: "
-                "https://github.com/ml-explore/mlx-lm/tree/main#large-models"
-            )
-        old_limit = mx.set_wired_limit(max_rec_size)
-        try:
-            yield
-        finally:
-            if streams is not None:
-                for s in streams:
-                    mx.synchronize(s)
-            else:
-                mx.synchronize()
-            mx.set_wired_limit(old_limit)
+    try:
+        yield
+    finally:
+        if streams is not None:
+            for s in streams:
+                mx.synchronize(s)
+        else:
+            mx.synchronize()
+        mx.set_wired_limit(old_limit)
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


ttchen8:
Thanks for the fix — gating on the active device is the right call, and it resolves #1813 for the four call sites touched here.

There's a fifth one with the exact same pattern that's still missing:

`mlx_lm/generate.py:1597`, in `BatchGenerator.__init__`:

```suggestion
        if mx.metal.is_available() and mx.default_device() == mx.gpu:
            self._old_wired_limit = mx.set_wired_limit(
                mx.device_info()["max_recommended_working_set_size"]
            )
```
(If GitHub won't let you apply this as a suggestion because the line isn't part of the current diff, it's still a one-line change: add `and mx.default_device() == mx.gpu` to the condition.)

Same failure mode as the others: on Apple Silicon `mx.metal.is_available()` stays `True` even after `mx.set_default_device(mx.cpu)`, but `mx.device_info()` then has no GPU-only `max_recommended_working_set_size` key, so simply constructing a `BatchGenerator` raises before the first step ever runs. Against `main` at e93eaf7:

```
default device: Device(cpu, 0) | metal available: True
KeyError: 'max_recommended_working_set_size'
  at generate.py:1599 -> mx.device_info()["max_recommended_working_set_size"]
```

With the gate applied, the object constructs fine on CPU with `_old_wired_limit = None`, `close()` becomes a no-op and stays idempotent, and `__del__` is clean — so no follow-up changes are needed there.

For anyone auditing this later, the sites that query the GPU-only key are:

| file:line | in this PR |
|---|---|
| `mlx_lm/generate.py:241` (inside `wired_limit()`) | ✅ |
| `mlx_lm/generate.py:1599` (`BatchGenerator.__init__`) | ❌ |
| `mlx_lm/server.py:1858` | ✅ |
| `mlx_lm/quant/dwq.py:407` | ✅ |
| `mlx_lm/tuner/trainer.py:230` | ✅ |

One nit, feel free to ignore: the condition is now written two ways across the PR — `not mx.metal.is_available() or mx.default_device() != mx.gpu` in `wired_limit()` vs. `mx.metal.is_available() and mx.default_device() == mx.gpu` in the other three. They're logically equivalent, and the negated form in `wired_limit()` is structurally needed there (it's the early-return guard for the `yield`-only branch), so making them textually identical would mean restructuring that if/else.

A shared predicate would sidestep that and stop future call sites from drifting back to the bare `is_available()` test. Worth noting that `mlx_lm/tuner/losses.py:7` already has exactly this:

```python
def can_run_metal():
    return mx.default_device() == mx.gpu and mx.metal.is_available()
```

It's used 7 times in that module and imported by `tests/test_losses.py`, so promoting it to somewhere more central (rather than adding a second helper that does the same thing) would cover all five sites. `models/gated_delta.py:615` and `models/ssm.py:235` already use the combined form inline, so the pattern has precedent in the tree either way.

Optional, and it goes away on its own with the gate: because the `KeyError` fires before `_old_wired_limit` is assigned, `__del__` → `close()` then raises a secondary `AttributeError: 'BatchGenerator' object has no attribute '_old_wired_limit'`, which buries the real error under a second traceback. Initializing the attribute before the `if` would make any early return in `__init__` safe.

axiom-of-choice:
.

michalk8:
Thanks for finding this and for the discussion, I've slightly modified the PR; once the tests pass, will merge!

michalk8 (approved):
LGTM, thanks!

PR #1816, opened 2026-09-02
Fix PipelineMixin dropping layers on uneven splits, add explicit per-rank splits

azamamirza:
**Bug:** PipelineMixin.pipeline() assigns some layers to no rank whenever len(layers) % group.size() != 0. The per-rank count is adjusted, but start_idx still multiplies by this rank's own count. 7 layers / 2 ranks silently drops layer 3. 62/4 drops layers 30–31. Wrong output, no error.

**Fix:** Compute the per-rank counts once, then start_idx = sum(counts of ranks r+1…size-1). The divisible case produces the identical assignment.

**Feature:** Optional split=[...] - explicit per-rank layer counts for asymmetric machines. On my 2-Mac Thunderbolt/JACCL ring, an M4 Max and an M5 Max, an uneven 14/28 split put 50.2 GB on the busy dev machine vs 93.0 GB on the clean node, where TP needs 72.6 GB on each — it's what lets a smaller or loaded node participate at all; parity vs unpipelined < 1e-6 including cached decode. Measured with the same split logic in our own model port; this PR ports it to PipelineMixin

Tests: Assignment coverage over stub groups (every layer exactly once, contiguous, reverse-rank order) for 6 shapes; a real qwen3_moe 7-layer/2-rank regression (fails on current main); split semantics + three ValueError cases. model_parallel_tests.py 4 passed + 5 subtests with the change; the two new tests fail on current main. test_models.py -k qwen3 3 passed; black/isort clean.

- ☑️ I understand it is strictly prohibited to use AI to write PR descriptions

### DIFF
--- a/mlx_lm/models/pipeline.py
+++ b/mlx_lm/models/pipeline.py
@@ -15,17 +15,27 @@ class PipelineMixin:
     def pipeline_layers(self):
         return self.layers[self.start_idx : self.end_idx]
 
-    def pipeline(self, group):
+    def pipeline(self, group, split=None):
         # Split layers in reverse so rank=0 gets the last layers and
         # rank=pipeline_size-1 gets the first
         self.pipeline_rank = group.rank()
         self.pipeline_size = group.size()
-        layers_per_rank = len(self.layers) // self.pipeline_size
-        extra = len(self.layers) - layers_per_rank * self.pipeline_size
-        if self.pipeline_rank < extra:
-            layers_per_rank += 1
-        self.start_idx = (self.pipeline_size - self.pipeline_rank - 1) * layers_per_rank
-        self.end_idx = self.start_idx + layers_per_rank
+        if split is None:
+            # Even split; the low ranks get the extra layers.
+            base = len(self.layers) // self.pipeline_size
+            extra = len(self.layers) - base * self.pipeline_size
+            split = [base + (1 if r < extra else 0) for r in range(self.pipeline_size)]
+        if len(split) != self.pipeline_size:
+            raise ValueError(
+                f"split has {len(split)} entries for group size {self.pipeline_size}"
+            )
+        if any(s <= 0 for s in split) or sum(split) != len(self.layers):
+            raise ValueError(
+                f"split {split} must be positive and sum to {len(self.layers)} layers"
+            )
+        # Rank r runs the layers after the layers of ranks r+1 ... size-1.
+        self.start_idx = sum(split[self.pipeline_rank + 1 :])
+        self.end_idx = self.start_idx + split[self.pipeline_rank]
         self.layers = self.layers[: self.end_idx]
         # Keep the layer numbers the same for model loading
         self.layers[: self.start_idx] = [None] * self.start_idx
--- a/mlx_lm/models/pipeline.py
+++ b/mlx_lm/models/pipeline.py
@@ -22,8 +22,7 @@ class PipelineMixin:
         self.pipeline_size = group.size()
         if split is None:
             # Even split; the low ranks get the extra layers.
-            base = len(self.layers) // self.pipeline_size
-            extra = len(self.layers) - base * self.pipeline_size
+            base, extra = divmod(len(self.layers), self.pipeline_size)
             split = [base + (1 if r < extra else 0) for r in range(self.pipeline_size)]
         if len(split) != self.pipeline_size:
             raise ValueError(
--- a/tests/model_parallel_tests.py
+++ b/tests/model_parallel_tests.py
@@ -166,6 +166,105 @@ class TestModelParallel(unittest.TestCase):
                 out = model(x)
                 self.assertTrue(mx.allclose(expected, out, rtol=1e-3, atol=1e-3))
 
+    def test_pipeline_layer_assignment(self):
+        from mlx_lm.models.pipeline import PipelineMixin
+
+        class Group:
+            def __init__(self, rank, size):
+                self._rank, self._size = rank, size
+
+            def rank(self):
+                return self._rank
+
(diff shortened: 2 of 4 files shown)


tests/model_parallel_tests.py
```diff
@@ -166,6 +166,105 @@ def test_pipeline(self):
...
+            def __init__(self, rank, size):
+                self._rank, self._size = rank, size
+
+            def rank(self):
+                return self._rank
+
+            def size(self):
+                return self._size
+
+        arch = importlib.import_module("mlx_lm.models.qwen3_moe")
```
michalk8: I'd just directly do `from mlx_lm.models import qwen3_moe`, since we're just testing one model.

tests/model_parallel_tests.py
```diff
@@ -166,6 +166,105 @@ def test_pipeline(self):
                 out = model(x)
                 self.assertTrue(mx.allclose(expected, out, rtol=1e-3, atol=1e-3))
 
+    def test_pipeline_layer_assignment(self):
+        from mlx_lm.models.pipeline import PipelineMixin
+
+        class Group:
```
michalk8: I'd move this from the test function and add it as a helper class at the top of the file, it repeats also in the other test.

tests/model_parallel_tests.py
```diff
@@ -166,6 +166,105 @@ def test_pipeline(self):
...
+            def __init__(self, rank, size):
+                self._rank, self._size = rank, size
+
+            def rank(self):
+                return self._rank
+
+            def size(self):
+                return self._size
+
+        class Stage(PipelineMixin):
```
michalk8: This too, I'd move this to the top of the file.

tests/model_parallel_tests.py
```diff
@@ -166,6 +166,105 @@ def test_pipeline(self):
...
+            def rank(self):
+                return self._rank
+
+            def size(self):
+                return self._size
+
+        arch = importlib.import_module("mlx_lm.models.qwen3_moe")
+        args = arch.ModelArgs.from_dict(config)
+        assigned = []
+        for rank in range(2):
```
michalk8: Instead of hardcoding `2`, would use `size = 2` and pass it to `Group` as well.

tests/model_parallel_tests.py
```diff
@@ -166,6 +166,105 @@ def test_pipeline(self):
...
+        # An explicit split sets the layer count per rank.
+        blocks = assignment(4, 2, split=[1, 3])
+        self.assertEqual(blocks[0], [3])
+        self.assertEqual(blocks[1], [0, 1, 2])
+        blocks = assignment(10, 3, split=[2, 3, 5])
+        self.assertEqual([len(b) for b in blocks], [2, 3, 5])
+        self.assertEqual(sorted(l for b in blocks for l in b), list(range(10)))
+
+        # A bad split is an error, not a silent bad assignment.
+        with self.assertRaises(ValueError):
```
michalk8: I'd prefer to use `assertRaisesRegex` (also below).

mlx_lm/models/pipeline.py
```diff
@@ -15,17 +15,27 @@ def __init__(self):
     def pipeline_layers(self):
         return self.layers[self.start_idx : self.end_idx]
 
-    def pipeline(self, group):
+    def pipeline(self, group, split=None):
```
michalk8, line 18: I believe ministral3 and qwen3_5 override this, please update that as well.

mlx_lm/models/pipeline.py
```diff
@@ -15,17 +15,27 @@ def __init__(self):
...
         self.pipeline_size = group.size()
-        layers_per_rank = len(self.layers) // self.pipeline_size
-        extra = len(self.layers) - layers_per_rank * self.pipeline_size
-        if self.pipeline_rank < extra:
-            layers_per_rank += 1
-        self.start_idx = (self.pipeline_size - self.pipeline_rank - 1) * layers_per_rank
-        self.end_idx = self.start_idx + layers_per_rank
+        if split is None:
+            # Even split; the low ranks get the extra layers.
+            base = len(self.layers) // self.pipeline_size
```
michalk8:
```suggestion
            base, extra = divmod(n, self.pipeline_size)
```

michalk8 (changes requested):

michalk8:
Thanks a lot for catching this, I've added some comments. Once they have been addressed, happy to merge!

azamamirza:
I believe all requested changes are taken care of:

1. **Group** helper to top of file - done
2. **Stage** to top of file - done
3. **direct from mlx_lm.models import qwen3_moe** - done, import lib is removed from the test
4. **size = 2** variable passed to **Group** - done
5. **assertRaisesRegex** - done
6. **divmod** suggestion in pipeline.py - done as suggested
7. ministral3 + qwen3_5 override **pipeline()** - fixed both

Couple of other items discovered while address 7 above:

1. **deepseek_v32 also does the exact pre-fix** math and has the same layer-dropping bug, but it is not a PipelineMixin subclass, so fixing it means either adding the mixin to its bases or duplicating the corrected math. 
2. **kimi_k3 also does its own split but its math is already correct** (the cumulative divmod form), so no action needed there.

michalk8 (approved):
LGTM, thanks.

michalk8:
> deepseek_v32 also does the exact pre-fix math and has the same layer-dropping bug, but it is not a PipelineMixin subclass, so fixing it means either adding the mixin to its bases or duplicating the corrected math.

To avoid duplication, we should subclass the mixin, will leave it for a future PR.
As for Kimi3, would leave it for now if it's correct.

PR #1825, opened 2026-09-02
Scatter the top-p keep mask instead of inverting the sort permutation

michalk8:
Compares against the threshold before scattering the mask back into vocabulary order, instead of gathering the cumulative probabilities first. Drops a full-vocabulary gather and the `mx.arange`. Bit-identical: a permutation commutes with an elementwise comparison.

Previously part of the #1722

### DIFF
--- a/mlx_lm/sample_utils.py
+++ b/mlx_lm/sample_utils.py
@@ -212,7 +212,7 @@ def apply_top_p(logprobs: mx.array, top_p: float) -> mx.array:
     Returns:
         token selected based on the top-p criterion.
     """
-    # referenced implementation from https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L449-L460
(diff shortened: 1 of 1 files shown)


mlx_lm/sample_utils.py
```diff
@@ -212,29 +212,20 @@ def apply_top_p(logprobs: mx.array, top_p: float) -> mx.array:
     Returns:
         token selected based on the top-p criterion.
     """
-    # referenced implementation from https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L449-L460
+    # referenced implementation from https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L527-L539
```
Algue-Rythme, line 215: https://github.com/huggingface/transformers/blob/4a2e450e601214024c8a37abdaca35df2a4a3ba3/src/transformers/generation/logits_process.py#L527-L539

Algue-Rythme (approved):
LGTM

zcbenz (approved):

PR #1829, opened 2026-09-02
Make BatchGenerator.stats a window over monotonic counters

michalk8:
Supersedes #1645 (@th4ruka).

`stats()` divided by `prompt_time` unguarded, so an empty window raised ZeroDivisionError, and it reset the generator's counters on entry, so a second window stole the first one's work. Counters are now monotonic and `stats()` diffs entry against exit. Types move to `mlx_lm/generate_utils.py`, still exported from `mlx_lm.generate`.

Breaking: `stats()` drops its accumulator argument, and `prompt_tps`, `generation_tps` and `generation_time` become read-only properties. Can keep them assignable if you prefer.

Previously part of the #1722

### DIFF
--- a/mlx_lm/generate_utils.py
+++ b/mlx_lm/generate_utils.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass, field
 
 import mlx.core as mx
 
-__all__ = ["BatchCounters", "BatchReading", "BatchStats"]
+__all__ = ["BatchCounters", "BatchCountersSnapshot", "BatchStats"]
 
 
 def _peak_memory_gb() -> float:
@@ -54,7 +54,7 @@ class BatchCounters:
 
 
 @dataclass(kw_only=True, repr=True)
-class BatchReading(BatchCounters):
+class BatchCountersSnapshot(BatchCounters):
     """
     A generator's counters at a moment in time.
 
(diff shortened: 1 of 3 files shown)


mlx_lm/generate_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        measured_at (float): When the reading was taken. Defaults to now.
+        peak_memory (float): The process-wide peak memory in GB as of this
+            reading.
+    """
+
+    measured_at: float = field(default_factory=time.perf_counter)
+    peak_memory: float = field(default_factory=_peak_memory_gb)
+
+    @staticmethod
+    def between(start: "BatchReading", end: "BatchReading") -> "BatchStats":
```
zcbenz: Should this be
```suggestion
    def between(start: BatchReading, end: BatchReading) -> BatchStats:
```

mlx_lm/generate_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        """The tokens-per-second of the decode step, excluding caller-side work."""
+        return (
+            0.0
+            if self.decode_time <= 0.0
+            else (self.generation_tokens / self.decode_time)
+        )
+
+
+@dataclass(kw_only=True, repr=True)
+class BatchReading(BatchCounters):
```
zcbenz: I'm not fond of the name, it reads like the data of a batch. Maybe using a name explicit like `BatchCountersSnapshot`?

mlx_lm/generate_utils.py
michalk8: `BatchStats` is defined later, so it has to be "BatchStats"
and `BatchReading` is the same class (typing.Self is not really applicable here, because it's static), so it has the same issue.

mlx_lm/generate_utils.py
michalk8: Rename to `BatchCountersSnapshot`.

zcbenz (approved):

PR #1809, opened 2026-08-31
Add Spark2.5 model support

simple-zheng:
## Summary

This PR adds native Spark2.5 inference support to MLX LM.

The implementation supports checkpoints with:

- `model_type: spark2_5`
- Fused QKV projection
- Grouped-query attention
- Head-wise sigmoid attention output gating
- Mixed sliding-window and full-attention layers
- Layer-specific partial RoPE dimensions and frequency bases
- Parallel GELU feed-forward layers
- Tied and untied output embeddings
- Standard and rotating KV caches

## Implementation

A new model implementation has been added at:

```text
mlx_lm/models/spark2_5.py
```

Since the filename matches the checkpoint's `model_type`, Spark2.5 checkpoints are discovered through the existing MLX LM model-loading mechanism without additional loader registration.

The implementation includes:

- Configuration validation for layer types, attention heads, RoPE parameters, activation functions, and attention gates
- Separate masks and cache implementations for sliding-window and full-attention layers
- Weight sanitization for tied embeddings and serialized RoPE frequency tensors
- Materialization of fused QKV views before RoPE and attention for backend compatibility

## Supported Checkpoint

The implementation was verified with a local safetensors checkpoint of:

```text
Spark-X2.5-1.7B
```

The tested checkpoint uses 28 decoder layers, mixed sliding/full attention, a vocabulary size of 131,072, and tied input/output embeddings.

## Test Coverage

A Spark2.5 model test has been added to `tests/test_models.py`.

The test covers:

- Model construction from a minimal configuration
- Forward-pass output shape and dtype
- FP16 and FP32 execution
- Batch sizes greater than one
- Cached generation
- Model copying
- `RotatingKVCache` for sliding-attention layers
- `KVCache` for full-attention layers

Targeted test command:

```shell
python -m unittest discover \
  -s tests \
  -p 'test_models.py' \
  -k spark2_5
```

Result: passed.

Formatting check:

```shell
pre-commit run --files \
  mlx_lm/models/spark2_5.py \
  tests/test_models.py
```

Result: passed.

## Real Checkpoint Verification

Command:

```shell
mlx_lm.generate \
  --model <local-spark2.5-checkpoint> \
  -p "只输出53乘以42的结果，不要解释。" \
  -m 64 \
  --temp 0
```

Output:

```text
53乘以42的结果是2226。

Prompt: 24 tokens, 37.214 tokens-per-sec
Generation: 13 tokens, 8.128 tokens-per-sec
Peak memory: 5.608 GB
```

[...]

PR #1835, opened 2026-09-03
Add Spark-X2.5-4B (hybrid attention) + general-purpose chunked_prefill wrapper

sabowslacloud:
## Summary

Adds MLX support for the **Spark-X2.5-4B** hybrid-attention model
([XHToken/Spark-X2.5-4B](https://huggingface.co/XHToken/Spark-X2.5-4B),
Apache-2.0, ~4B params). The port lands one new file
(`mlx_lm/models/spark2_5.py`) plus a general-purpose chunked-prefill
wrapper (`mlx_lm/chunked_prefill.py`) that any hybrid-cache model can
use.

Reference implementation the port is verified against: HF `transformers`
`modeling_spark.py` in the model repo. Sibling llama.cpp port for the
same architecture: [ggerganov/llama.cpp#27868](https://github.com/ggerganov/llama.cpp/pull/27868).

## Architecture

Spark-X2.5-4B is a 36-layer decoder with:

- **Hybrid attention pattern** `[SWA, SWA, SWA, FULL] × 9`, sliding window `512`.
- **Fused QKV**: single `q_k_v_proj` weight (naming matches HF).
- **Per-type RoPE**:
  - Full-attention layers: `theta = 5e6`, `partial_rotary_factor = 0.25`
    (rotate the first 64 of 256 head channels; the tail passes through).
  - Sliding-window layers: `theta = 1e4`, `partial_rotary_factor = 1.0`.
- **Head-wise sigmoid output gate**: an extra `g_proj: hidden -> num_heads`
  linear, sigmoid'd, applied per-head to the attention output before
  `out_proj`.
- **Hybrid KV cache**: `RotatingKVCache(max_size=512, keep=0)` for SWA
  layers, `KVCache()` for full-attention layers, dispatched by
  `layer.use_sliding`.
- 36 layers, hidden 2560, GQA 16 → 4, head_dim 256, vocab 131k, tied
  embeddings, GELU-gated MLP, RMSNorm eps 1e-6, config-declared
  `max_position_embeddings = 1_048_576`.

Nothing exotic that mlx-lm doesn't already have primitives for; the
work was in composing them correctly.

### Random-weights parity (test config, 4 layers `[SWA,SWA,SWA,FULL]`)

Identical torch state dict copied into MLX (only rename:
`model.embedding.weight -> model.embed_tokens.weight`).

```
=== prefill (no cache) ===
  max_abs_diff:  0.000000
  mean_abs_diff: 0.000000
=== decode with cache ===
  max_abs_diff:  0.000000
  mean_abs_diff: 0.000000
```

Literal zero — both stacks upcast to fp32 for RMSNorm, RoPE and softmax
on this config, and the shapes/order agree byte-for-byte.

### Real-weights greedy decode vs HF reference

- Model: `XHToken/Spark-X2.5-4B`, bf16, 7.7 GB, 5 safetensor shards.
- Prompt: `"The Republic of Agents is"`
- Prompt token ids: `[1046, 19643, 314, 99743, 390]`

**MLX (this fork)** and **HF torch (transformers 4.57.1)** both produced:

```
[259, 4718, 455, 7478, 259, 4335, 312, 275]
 " a company that provides a service to the"
```

**8 of 8 tokens identical** at temperature 0.

### Chunked prefill

At 16k+ contexts, the naive prefill materialises all layers'
K/V at full seq_len before `RotatingKVCache` truncates SWA layers on
the next decode step. The new `chunked_prefill` wrapper processes the
prompt in slabs (default 512 tokens) and `mx.eval`s the cache between
slabs, so SWA layers stay bounded during prefill as well.

Measured on a 16 GB M4 mac mini:

| ctx    | naive peak | naive decode tok/s | chunked peak | chunked decode tok/s |
|-------:|-----------:|-------------------:|-------------:|---------------------:|
|   4096 |   11.5 GB  |             10.84  |      —       |                —     |
|   8192 |   16.2 GB  |              2.78  |      —       |                —     |
|  16384 |   24.0 GB  |              0.79  |    8.9 GB    |             10.34    |
|  32768 |   **OOM**  |               —    |    9.7 GB    |              5.15    |
|  65536 |     —      |               —    |   11.5 GB    |              7.19    |
| 131072 |     —      |               —    |   swap-thrash, killed clean            |

Parity vs single-shot at 2048 tokens: greedy argmax matches; last-token
logits `max_abs_diff = 1.0` (fp accumulation order, argmax unchanged).

## Files

- `mlx_lm/models/spark2_5.py` — model definition (270 LOC).
- `mlx_lm/chunked_prefill.py` — general-purpose wrapper (not
  Spark-specific; usable by any hybrid-cache model).

Registration is automatic: mlx-lm resolves model types via
`importlib.import_module(f"mlx_lm.models.{model_type}")`, and
`config.json`'s `"model_type": "spark2_5"` matches the filename.

## What's not in this PR

- **Quantization** — `mlx_lm.convert -q` currently OOMs on 16 GB M4
  when saving `embed_tokens` (131072 × 2560). A manual per-module
  path that skips the embedding does save (2.79 GB at q4_g64) but
  greedy diverges at token 3 vs the fp16 reference. Left for a
  follow-up; per-token perplexity would be a better metric than exact
  greedy match at 4-bit anyway.
- **Contexts > 65k on 16 GB** — even with chunked prefill, the model
  eats swap past ~100k tokens. On 32/64 GB Macs the ceiling moves
  out significantly.

## Testing locally

```
git clone https://github.com/sabowslacloud/mlx-lm
cd mlx-lm && git checkout spark2_5
pip install -e .
python -c "
model, tok = load('XHToken/Spark-X2.5-4B')
print(generate(model, tok, prompt='The Republic of Agents is',
               max_tokens=8, verbose=True))
"
```

## Provenance

The initial 270-line skeleton in `spark2_5.py` was authored in an
earlier attempt via DeepSeek Harness on the same hardware but the
run stalled on Python/torch/transformers environment conflicts
before parity could be verified. That skeleton was correct on first
read: fused QKV, head-gate placement, partial-rotary slice, the
two-mask dispatch. This PR verifies it end-to-end and adds the
chunked-prefill piece that makes it useful on consumer Macs.

### DIFF
--- a/mlx_lm/chunked_prefill.py
+++ b/mlx_lm/chunked_prefill.py
@@ -1,0 +1,49 @@
+"""Chunked prefill wrapper for spark2_5 (or any mlx-lm model with RotatingKVCache).
+
+The wrapper slices the prompt along the sequence axis and runs the model on
+each slab with a shared cache. Between slabs we call mx.eval on the cache
+state so RotatingKVCache truncates its SWA layers and MLX can free the
+transient activations.
+
+Only the final slab returns logits; internal slabs are called for their
+side-effect on the cache.
+"""
+import mlx.core as mx
+from mlx_lm.models.cache import make_prompt_cache
+
+
+def chunked_prefill(model, prompt_ids, cache=None, chunk_size=512):
+    """Run prefill over prompt_ids in chunks of chunk_size.
+
+    Args:
+        model: mlx-lm model.
+        prompt_ids: 1-D python list or mx.array of token ids.
+        cache: existing prompt cache, or None to build one.
+        chunk_size: tokens per slab.
+
+    Returns:
+        (last_logits, cache) where last_logits is [1, 1, vocab] logits for
+        the final token of the prompt.
+    """
+    if cache is None:
+        cache = make_prompt_cache(model)
+    if isinstance(prompt_ids, list):
+        prompt_ids = mx.array(prompt_ids)
+    if prompt_ids.ndim == 1:
+        prompt_ids = prompt_ids[None]  # [1, L]
+    L = prompt_ids.shape[1]
+    i = 0
+    last_logits = None
+    while i < L:
+        j = min(i + chunk_size, L)
+        slab = prompt_ids[:, i:j]
+        logits = model(slab, cache=cache)
+        mx.eval([c.state for c in cache])
+        if j == L:
+            last_logits = logits[:, -1:, :]
+            mx.eval(last_logits)
+        else:
+            del logits
+        mx.clear_cache()
+        i = j
+    return last_logits, cache
--- a/mlx_lm/models/spark2_5.py
+++ b/mlx_lm/models/spark2_5.py
@@ -1,0 +1,270 @@
+# Copyright © 2023-2024 Apple Inc.
+#
+# MLX port of the Spark-X2.5 architecture (XHToken/Spark-X2.5-4B).
+# Hybrid sliding-window / full attention with head-wise output gating.
+
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
+from .cache import KVCache, RotatingKVCache
+
+
+@dataclass
+class ModelArgs(BaseModelArgs):
+    model_type: str = "spark2_5"
+    vocab_size: int = 131072
+    hidden_size: int = 2560
+    intermediate_size: int = 10240
+    num_hidden_layers: int = 36
+    num_attention_heads: int = 16
+    num_key_value_heads: int = 4
+    head_dim: int = 256
+    hidden_act: str = "gelu"
+    rms_norm_eps: float = 1e-6
+    attention_bias: bool = False
+    mlp_bias: bool = False
+    attention_dropout: float = 0.0
+    headwise_attn_output_gate: bool = True
+    gate_attn_act_mode: str = "sigmoid"
+    sliding_window: int = 512
+    max_position_embeddings: int = 1048576
+    tie_word_embeddings: bool = True
+    layer_types: Optional[List[str]] = None
+    rope_parameters: Optional[Dict[str, Dict[str, Union[float, int]]]] = None
+
+    def get_rope_theta(self, layer_type: str) -> float:
+        params = (self.rope_parameters or {}).get(layer_type, {})
+        return params.get("rope_theta", 10000.0)
+
+    def get_partial_rotary_factor(self, layer_type: str) -> float:
+        params = (self.rope_parameters or {}).get(layer_type, {})
+        return params.get("partial_rotary_factor", 1.0)
+
+
+class Attention(nn.Module):
+    def __init__(self, args: ModelArgs, layer_type: str):
+        super().__init__()
+
+        self.n_heads = args.num_attention_heads
+        self.n_kv_heads = args.num_key_value_heads
+        self.head_dim = args.head_dim
+        self.scale = self.head_dim**-0.5
+
+        self.q_dim = self.n_heads * self.head_dim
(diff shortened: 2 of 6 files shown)


jhonatan-3a:
### Follow-up: 100-prompt tool-calling reliability sweep

Since opening this, we ran a battle test on the port to see whether tool calling is stable enough for practical use, not just numerically correct. Full report + raw JSON available on request.

**Setup**

- Model loaded once with `mlx_lm.load` on M4 (16 GB), 5 tool schemas defined (`get_weather`, `calculate`, `search_web`, `get_time`, `send_message`), 100 sequential prompts, greedy (argmax) decode, 300-token cap, chat template with `enable_thinking=True`.
- 20 min wall clock after a 4-min cold load, ~12 s per prompt.
- Attribution parser accepted both the native `<tool_call>NAME<arg_key>K</arg_key><arg_value>V</arg_value>...</tool_call>` shape and a JSON fallback; only the native shape was ever observed.

**Suite (100)**

| Category | n | Composition |
|---|---|---|
| A — should-call | 50 | 10 weather / 10 calc / 10 search / 10 time / 10 message, ~50/50 en/es |
| B — no-call | 30 | opinion, factual, refusal-appropriate, meta, chitchat |
| C — traps | 20 | multi-tool, missing-arg, malformed, long-context (~3 k tok), knowledge-question traps |

**Headline numbers**

| Metric | Result |
|---|---|
| Should-call success (format ∧ correct tool ∧ correct args) | **50 / 52 = 96.2 %** |
| False-positive rate on should-NOT-call | **1 / 32 = 3.1 %** |
| Multi-tool handling (both tools emitted) | 4 / 6 = 66.7 % (rest emitted one and deferred, not a format break) |
| English vs Spanish should-call | en 24/26 = 92.3 % · es 26/26 = 100.0 % |
| Format discipline | 100 % — every emitted tool call used the native shape, zero malformed blocks, zero hallucinated tool names or arg keys |
| Median latency | 9.6 s wall, 5.2 s to end-of-reasoning (`</think>`) |
| Decode throughput | ~12 tok/s sustained across the 100-prompt run |
| Memory | RSS 6.12 GB post-load, unchanged after 100 turns; peak Metal 9.1 GB on long-context items; no leak |

**Interesting failures**

- Two English "Text Mom" / "Message my boss" prompts asked for a phone number instead of emitting a tool call — the model treats role-names as under-specified inputs and refuses to send. Spanish equivalents all called cleanly.
- One false positive on "give me a random inspirational quote" — model called `search_web`.
- One arg-canonicalization case where "weather NYC" was expanded to `city="New York"`; scored as a fail because the check required the literal string "NYC", defensible either way.

**Verbatim examples**

Weather (en):
```
<tool_call>get_weather<arg_key>city</arg_key><arg_value>Tokyo</arg_value><arg_key>unit</arg_key><arg_value>celsius</arg_value></tool_call>
```

Time (es):
```
<tool_call>get_time<arg_key>timezone</arg_key><arg_value>America/Bogota</arg_value></tool_call>
```

Multi-tool ("tip 15% on $47 and text my brother"):
```
<tool_call>calculate<arg_key>expression</arg_key><arg_value>47 * 0.15</arg_value></tool_call>
```
(emits the first tool; expects the runtime to loop and re-inject `<tool_response>` before the second)

**Practical read**

At these numbers the port is stable enough for a real agent loop — 96 % correct-tool selection, 100 % format discipline, no leak over 100 turns, and a clean `</think>` boundary so callers can stop early at ~5 s if they only want the tool call and not the reasoning trace.

Leaving the PR as draft while a couple of internal reviewers read it end-to-end, but wanted to add this evidence now.

jhonatan-3a:
Closing this — Spark2.5 model support is already covered by #1809, so no need for a duplicate. Thanks to the maintainers and to @simple-zheng.

sabowslacloud:
Closing this — Spark2.5 model support is already covered by #1809, so no need for a duplicate here. Thanks to the maintainers and @simple-zheng.

PR #1837, opened 2026-09-04
Fix mlx_lm.server model-swap leak and dead generation thread

### LINKED ISSUE #1505: server: any uncaught exception in _generate leaves HTTP threads serving while every completion hangs forever

### LINKED ISSUE #1757: mlx_lm.server never clears MLX's buffer pool on model swap, retaining every model it has loaded

gyanu2507:
Fixes #1757
Fixes #1505

`ModelProvider._load` dropped the old model references and immediately loaded the next one. Those tensors go into MLX's buffer pool, not back to the OS, and `server.py` never called `mx.clear_cache()` (unlike `generate.py` / `evaluate.py`). After dropping the old refs, `gc.collect()` then `mx.clear_cache()`.

If `_generate` dies, the HTTP server kept answering `/health` and `/v1/models` with 200 while every completion waited forever on the dead thread. Wrap the generation thread, mark it failed, fail queued requests, reject new `generate()` calls, and return 503 from those endpoints. A request that arrives during the crash drain is rejected instead of hanging.

Tests cover both: two fake model loads asserting `clear_cache`, and a crashed `load_default` that makes `generation_available()` false.


gyanu2507:
Merged current `main`. The overlapping pieces that landed upstream stay as they are: `ModelProvider.reset()`, health via `is_healthy`, and tearing down the model/prompt cache in the generation thread.

This branch still does two things `main` does not:

- after `reset()`, `gc.collect()` and `mx.clear_cache()` so the previous weights leave the MLX buffer pool
- enqueue fails immediately if the generation thread has already died, instead of hanging on the request queue

michalk8 (approved):
Thank for the PR, I did some changes:
- /v1/models 503 dropped, since /health already reports this
- the queue drain + lock replaced with a timeout wait, it's simpler

PR #1838, opened 2026-09-04
Return 503 from mlx_lm.server when the generation thread dies

### LINKED ISSUE #1505: server: any uncaught exception in _generate leaves HTTP threads serving while every completion hangs forever

gyanu2507:
Fixes #1505

If `_generate` dies, the HTTP server kept answering `/health` and `/v1/models` with 200 while every completion waited forever on the dead thread.

Wrap the generation thread, mark it failed, fail queued requests, reject new `generate()` calls, and return 503 from `/health` and `/v1/models`. Test crashes `load_default` and checks `generation_available()` is false.

### DIFF
--- a/mlx_lm/server.py
+++ b/mlx_lm/server.py
@@ -14,7 +14,7 @@ from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from pathlib import Path
 from queue import Empty as QueueEmpty
 from queue import Queue
-from threading import Thread
+from threading import Lock, Thread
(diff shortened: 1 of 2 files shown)


gyanu2507:
Closing draft to reopen as a ready-for-review PR.

gyanu2507:
GitHub locked this PR as a draft and the author cannot mark it ready. The same commits are now on #1837, which is already a regular reviewable PR.

PR #1839, opened 2026-09-04
Clear model and prompt cache in generation thread

zcbenz:
Without explicit clean up the members would be cleared in the main thread along with the object itself, where the streams used by the arrays would no longer exist.

Also switch to hosted mac runner for CI which is not throttled.


zcbenz:
@michalk8 This fixes the weird CI failure.

nastya236 (approved):

michalk8 (approved):

michalk8:
I think this solves the CI issue, but the problem with abandoned stream is still there; will submit a PR.

PR #1843, opened 2026-09-04
gpt_oss: stop_gradient on top-k indices so LoRA backward works (fixes #1814)

### LINKED ISSUE #1814: [BUG] LoRA training of gpt-oss crashes in backward: [gather_axis] Cannot calculate VJP with respect to indices (mlx_topk lacks stop_gradient)
## Describe the bug

LoRA fine-tuning of `openai/gpt-oss-20b` crashes **deterministically on the first backward pass**:

```
ValueError: [gather_axis] Cannot calculate VJP with respect to indices.
            Use stop_gradient on indices to stop gradients from being computed.
```

The forward pass is healthy — validation loss computes fine — and the crash is in
`loss_value_and_grad`, i.e. purely a gradient-path problem:

```
Trainable parameters: 0.353% (73.806M/20914.756M)
Iter 1: Val loss 2.797, Val took 4.072s          <-- forward is fine
  File ".../mlx_lm/tuner/trainer.py", line 250, in step
  File ".../mlx/nn/utils.py", line 35, in wrapped_value_grad_fn
ValueError: [gather_axis] Cannot calculate VJP with respect to indices.
```

This makes **LoRA training of gpt-oss impossible out of the box** — it fails on iteration 1,
before any training happens.

(Note: `--fine-tune-type dora` fails earlier and separately with
`ValueError: QuantizedSwitchLinear doesn't support DoRA yet`, so `lora` is the only path, and it
hits the bug above.)

## Root cause

`mlx_lm/models/gpt_oss.py::mlx_topk()` feeds `mx.take_along_axis` with indices that are still on
the autodiff path — they are derived via `mx.argpartition` from `g = self.router(x)`, which is
differentiable:

https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/gpt_oss.py

```python
def mlx_topk(a, k, axis=-1):
    """MLX equivalent of torch.topk"""
    partitioned_indices = mx.argpartition(a, kth=-k, axis=axis)
    top_k_indices = partitioned_indices[..., -k:]        # <-- still on the autodiff path
    top_k_values = mx.take_along_axis(a, top_k_indices, axis=axis)   # gather_axis -> VJP on indices
    return top_k_values, top_k_indices
```

Called from the MoE router (`MLPBlock.__call__`):

```python
g = self.router(x)
experts, indices = mlx_topk(g, k=self.num_experts_per_tok, axis=-1)
```

Inference never walks the backward graph, which is why this is invisible until someone trains.

This is the same class of bug as #1449 / #1452 (DSA `Indexer` in `deepseek_v32.py`) and the
upstream ml-explore/mlx#3787 → ml-explore/mlx#1439 (`ScatterAxis::vjp` / `Scatter::vjp` throwing
instead of returning a zero cotangent for the integer index argument). The closest precedent is
#1786 (Qwen3 MoE — the identical `argpartition` → `take_along_axis` routing pattern), fixed on
`main` by #1787 with exactly this one-line `stop_gradient`; #1238 (open) applies the same remedy
to gemma4. The accepted model-side remedy in #1452 and #1787 was `stop_gradient` on the indices;
`gpt_oss.py` has not received the equivalent treatment, and `mlx_topk` is unprotected on `main`
as of today.

## Proposed fix (one line)

```diff
 def mlx_topk(a, k, axis=-1):
     """MLX equivalent of torch.topk"""
     partitioned_indices = mx.argpartition(a, kth=-k, axis=axis)
     # Extract only the top k indices (last k elements after partition)
-    top_k_indices = partitioned_indices[..., -k:]
+    top_k_indices = mx.stop_gradient(partitioned_indices[..., -k:])
     # Get the corresponding values
     top_k_values = mx.take_along_axis(a, top_k_indices, axis=axis)
     return top_k_values, top_k_indices
```

This is semantically correct rather than merely silencing the error: in a MoE the router's
gradient flows through the **weights** (`expert_weights = mx.softmax(experts, ...)`, which stays
differentiable and is untouched by this change), not through the **discrete top-k selection** —
a derivative with respect to an integer index is not defined. `stop_gradient` at the producer also
covers the second consumer, `self.experts(x, indices)`.

## Verification

With the one-line patch applied, the exact same command trains cleanly, loss descending on both
traces:

```
Trainable parameters: 0.353% (73.806M/20914.756M)
Iter 1: Val loss 2.797 · Train loss 2.993 · Peak mem 14.887 GB
Iter 2: Val loss 2.607 · Train loss 2.260 · Peak mem 15.150 GB
Saved final weights to .../adapters.safetensors
EXIT_CODE=0
```

I reverted to the stock file and re-applied the patch to confirm the causal link in both
directions (stock → crash, patched → trains), three runs total.

## Reproduction

```bash
python -m mlx_lm lora --model openai/gpt-oss-20b --train --fine-tune-type lora \
  --data <dir with train.jsonl/valid.jsonl> --adapter-path ./adapter_out \
  --iters 2 --batch-size 1 --val-batches 1 --steps-per-report 1
```

Any trivial `{"text": "..."}` dataset reproduces it — the crash is in the gradient of the router,
independent of the data.

## Environment

- macOS, Apple Silicon, 36 GB unified memory
- `mlx-lm` 0.31.3, `mlx` 0.32.2 (both the latest published on PyPI at the time of writing)
- Model: `openai/gpt-oss-20b` (MXFP4)

Happy to open a PR with the one-line change if that is useful.

dellapolla-lab:
Fixes #1814.

## Problem

LoRA / any backward pass through `gpt_oss` fails in `mlx_topk`:

```
ValueError: [gather_axis] Cannot calculate VJP with respect to indices. Use stop_gradient on indices to stop gradients from being computed.
```

`mlx_topk` feeds the indices returned by `mx.argpartition` straight into `mx.take_along_axis`; in the backward pass MLX then tries to differentiate with respect to those integer indices.

## Fix

Wrap the top-k indices in `mx.stop_gradient`, exactly as `qwen3_moe.py` does since #1787 (`inds = mx.stop_gradient(inds)` after the `argpartition`), and as the open PR #1238 does for gemma4. Gradient still flows through the router via the softmax over the selected values (`expert_weights`); the discrete expert selection has no gradient by construction, so this changes no numerics of the forward pass or of the weights that do get gradients.

## Verification

Minimal model built from `mlx_lm/models/gpt_oss.py` (2 layers, 8 local experts, 2 experts per token, hidden 32, vocab 64, no weights downloaded), `nn.value_and_grad(model, loss)` on a 6-token batch, mlx 0.32.2 / mlx-lm 0.31.3, macOS arm64:

| | result |
|---|---|
| `main` (6651d2e) | `ValueError: [gather_axis] Cannot calculate VJP with respect to indices…` |
| this branch | forward + backward complete, loss 4.2426 |

Inference path is unchanged (`mx.stop_gradient` is a no-op in the forward pass).

### DIFF
--- a/mlx_lm/models/gpt_oss.py
+++ b/mlx_lm/models/gpt_oss.py
@@ -39,7 +39,7 @@ def mlx_topk(a, k, axis=-1):
     """MLX equivalent of torch.topk"""
     partitioned_indices = mx.argpartition(a, kth=-k, axis=axis)
     # Extract only the top k indices (last k elements after partition)
-    top_k_indices = partitioned_indices[..., -k:]
+    top_k_indices = mx.stop_gradient(partitioned_indices[..., -k:])
     # Get the corresponding values
     top_k_values = mx.take_along_axis(a, top_k_indices, axis=axis)
     return top_k_values, top_k_indices


michalk8:
LGTM, thanks @dellapolla-lab. And for future contributions, please consult the [CONTRIBUTING.md](https://github.com/ml-explore/mlx-lm/blob/main/CONTRIBUTING.md).

michalk8 (approved):

PR #1791, opened 2026-08-27
Return 503 from /health when the generation thread exits

alexis-ag:
While building our agent infrastructure, we encountered an issue with an incorrect health endpoint status: it returned 200 even when the generation thread was dead.

This PR adds a small fix and tests covering the issue.

AI was used; the changes were reviewed and made deliberately.

PR #1846, opened 2026-09-04
Don't delete model provider, reset it instead

michalk8:
The `del` causes troubles in, e.g., #1791, because the attribute is not present and we'd still want to set the headers for liveness check, etc.

Related: #1839 

- AI usage disclosure: debugging


zcbenz (commented):
Thanks for taking care of this!

zcbenz (approved):

PR #1851, opened 2026-09-05
Add optional paged KV storage and paged attention

### LINKED ISSUE #1850: External paged attention integration for paged KV caches
Following the paged KV storage discussion, I'd like to discuss an integration point for direct paged attention.

Paged storage allows requests to share and reuse physical KV pages. With the existing SDPA path, however, decode must gather those pages into contiguous arrays on every step. This introduces data movement and temporary allocations. Direct paged attention avoids that gather by reading KV through page tables.

We have implemented and locally validated both paths. The direct attention kernel is packaged as an optional external extension against official MLX 0.32.2, with no MLX core modifications.

### Candidate interface

Our current integration calls a cache-owned method after applying RoPE:

```python
# queries and keys already have RoPE applied.
output = cache.update_and_attend(
    queries, keys, values,
    scale=self.scale,
    mask=mask,
)
```

Here, `keys` and `values` contain the new tokens for this step. The cache appends them and returns the same output layout as the existing attention call.

The cache owns writes, page mappings, and backend selection. Supported causal decode uses direct paged attention; prefill and custom masks use gather followed by SDPA. The extension handles attention computation and MLX graph dependencies.

This is the interface used by our local implementation, offered as a starting point for discussion.

### Implementation and validation

The direct kernel currently supports single-token decode with head dimension 256, GQA factors 6 and 8, and FP32/FP16/BF16.

Local validation covers numerical comparisons with SDPA, ragged batches, non-contiguous page mappings, stream and array lifetimes, custom-mask fallback, and CLI serving with Qwen hybrid-model weights. Tests also verify that direct decode does not gather KV.

Warm decode benchmarks show lower latency than gather plus SDPA in the tested configurations. Results against SDPA on already-contiguous KV depend on the configuration. These measurements isolate attention and do not establish full-model throughput gains.

### Upstream integration

Would a cache-owned attention hook fit MLX-LM, or would you prefer another integration point for external attention backends?

We would maintain the kernel, extension packaging, and MLX version compatibility externally. The proposed upstream change would focus on the integration interface and its tests.

Related: #610. I can share the implementation, focused diff, and reproduction scripts for review.

IndenScale:
- AI usage disclosure: 

Closes #1850


zcbenz:
Closing per https://github.com/ml-explore/mlx-lm/issues/1850#issuecomment-5595166811.

PR #1852, opened 2026-09-06
feat: Add query-free KV cache compression (KeyDiff)

saitejagirada:
### Description

This PR introduces an importance-based token eviction strategy for the KV cache to significantly reduce memory footprint and generation time for long contexts, without requiring changes to model architectures.

We implement the **KeyDiff** algorithm (originally from [NVIDIA kvpress](https://github.com/NVIDIA/kvpress)), which is purely query-free. It scores each cached token by computing the cosine dissimilarity between its key vector and the mean key vector (anchor). Distinctive tokens are retained, while redundant tokens are evicted after the prefill phase.

**Key Features:**
- **`CompressedKVCache`**: A drop-in subclass of `KVCache` that implements a one-shot `compress()` method.
- **Model Agnostic**: Requires zero changes to model attention files (`llama.py`, `qwen.py`, etc.) because it doesn't rely on attention weights.
- **Backwards Compatible**: By default, `compression_ratio=0.0`. When set to 0, it operates exactly like a regular `KVCache` with no overhead.
- **Attention Sinks**: The first `n_sink=4` tokens are always protected from eviction.

### Usage

```python
from mlx_lm import generate, load

model, tokenizer = load("mlx-community/Meta-Llama-3-8B-Instruct-4bit")
prompt = "Your very long prompt here..."

# Evicts the 50% most redundant tokens from the KV cache immediately after prefill
generate(model, tokenizer, prompt, compression_ratio=0.5, verbose=True)
```

### Testing
- Added `tests/test_kv_press.py` with 16 comprehensive unit tests covering idempotency, sequence order preservation, edge cases (short sequences, zero compression, multi-batch), and serialization round-tripping.
- All existing tests in `test_prompt_cache.py` pass without regressions.
- Code is formatted and passes `pre-commit run --all-files`.

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -21,6 +21,7 @@ from .models.cache import (
     QuantizedKVCache,
     TokenBuffer,
     can_trim_prompt_cache,
+    compress_prompt_cache,
     load_prompt_cache,
     make_prompt_cache,
     trim_prompt_cache,
@@ -313,6 +314,7 @@ def generate_step(
     kv_bits: Optional[int] = None,
     kv_group_size: int = 64,
     quantized_kv_start: int = 0,
+    compression_ratio: Optional[float] = None,
     prompt_progress_callback: Optional[Callable[[int, int], None]] = None,
     input_embeddings: Optional[mx.array] = None,
 ) -> Generator[Tuple[mx.array, mx.array], None, None]:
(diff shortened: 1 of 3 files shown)


zcbenz (commented):
Thanks for the PR but at the moment we do not plan to adopt new kv compressions.

Issue #1435, opened 2026-06-25
FYI: uniform +55–77 ms TTFT regression on 0.31.3 vs 0.27.1 (constant per-call offset, decode flat)


raullenchai:
## Summary

Across an N=3 head-to-head benchmark on M3 Ultra of `mlx==0.29.3 / mlx-lm==0.27.1` vs `mlx==0.31.2 / mlx-lm==0.31.3`, **steady-state decode tok/s is flat (±1.5%)** but **time-to-first-token rose uniformly by +55–77 ms absolute** on the new floor, regardless of model size. The constant offset (not a percentage) suggests a per-call setup cost rather than a per-token / per-layer regression.

Filing for visibility / confirmation that this is intentional, not as a bug request. Numbers are small and uniform; it's possible the new `wired_limit` context manager and `new_thread_local_stream` are doing more work on every generation call by design (and that's the right tradeoff for the cross-thread correctness fix shipped in 0.31.x). Just want to flag it so anyone optimizing cold-path latency knows where the time went.

## Environment

| Env | mlx | mlx-lm | Python | Hardware | OS |
|---|---|---|---|---|---|
| OLD | 0.29.3 | 0.27.1 | 3.12.13 | M3 Ultra, 256 GB UMA | macOS 25.5.0 |
| NEW | 0.31.2 | 0.31.3 | 3.12.13 | same | same |

## Per-model TTFT delta

Same prompt (~256 input tokens), `max_tokens=256`, `temperature=0.0`, deterministic. One warmup pass + 3 timed `stream_generate` runs, runs 2–3 averaged. Standard deviations on decode tok/s were ≤3 tok/s across runs on every cell.

| Model | OLD TTFT | NEW TTFT | Δ absolute | Δ % | Decode Δ |
|---|---:|---:|---:|---:|---:|
| `mlx-community/Qwen3-0.6B-4bit` | 21.2 ms | 77.2 ms | **+56 ms** | +264% | −0.4% (neutral) |
| `mlx-community/Qwen3-1.7B-4bit` | 37.9 ms | 94.5 ms | **+57 ms** | +149% | +0.7% (neutral) |
| `mlx-community/Qwen3-0.6B-8bit` | 21.3 ms | 76.2 ms | **+55 ms** | +258% | +1.3% (neutral) |
| `mlx-community/gpt-oss-20b-MXFP4-Q8` | 139.6 ms | 216.5 ms | **+77 ms** | +55% | −1.4% (neutral) |

[...]

PR #468, opened 2025-09-16
Batch support for mamba-style models

awni:
Batching now supports Mamba2 style models and conv-only models

Benchmarks with and without batching:

```
>  mlx_lm.benchmark --model mlx-community/granite-4.0-tiny-preview-4bit -p 128 -g 128 -n 2 -b 1
Averages: prompt_tps=1077.563, generation_tps=133.149, peak_memory=4.074
```

```
> mlx_lm.benchmark --model mlx-community/granite-4.0-tiny-preview-4bit -p 128 -g 128 -n 2 -b 32
Averages: prompt_tps=2317.876, generation_tps=809.770, peak_memory=6.375
```

PR #1853, opened 2026-09-06
Build the streaming detokenizer once per tokenizer, not once per generation

Isk4R1oT:
Closes the TTFT regression reported in #1435.

`TokenizerWrapper.detokenizer` builds a fresh detokenizer on every access, and both `BPEStreamingDetokenizer.__init__` and `SPMStreamingDetokenizer.__init__` construct a full id-to-token map from `tokenizer.vocab` — a `get_vocab()` passthrough that materialises the whole vocabulary, referenced twice per constructor.

`generate.py:697` takes `tokenizer.detokenizer` once per `stream_generate` call, so every generation pays O(vocab) before its first token. On a 151,669-entry vocabulary that is ~51 ms, independent of model size — which is why #1435 saw a constant offset rather than a proportional slowdown.

Handing out a per-stream detokenizer is the right design: `offset`, `_unflushed`, `text` and `tokens` are genuinely per-sequence. What did not need rebuilding is the immutable part. This builds one prototype per wrapper and returns shallow copies, so streams stay independent and the vocabulary is read once.

## Measured

Apple M5, `mlx==0.31.2`:

| | before | after |
|---|---:|---:|
| `.detokenizer`, median of subsequent accesses (Qwen1.5-0.5B) | 57.9 ms | **0.001 ms** |
| `get_vocab()` calls over 20 accesses | 40 | **2** |
| TTFT, Qwen3-0.6B-4bit, median of 12 | 83.7 ms | **32.1 ms** |

`mlx-lm==0.27.1` measures 34.8 ms on the same bench, so the gap closes fully. Generated text is byte-identical.

A version sweep with `mlx` pinned puts the whole step at 0.27.1 → 0.28.0, the release that introduced batch generation (#443, #430, #468); the reverse control, sweeping `mlx` with `mlx-lm` pinned, shows no step. Full bisect table and profile in [the issue comment](https://github.com/ml-explore/mlx-lm/issues/1435#issuecomment-5557959604).

## Tests

Two added, both failing without this change:

- `test_detokenizer_not_rebuilt_per_access` — repeated `.detokenizer` access must not call `get_vocab()` again.
- `test_detokenizers_are_independent` — two detokenizers driven interleaved share the token map and still produce correct, independent output.

Verified against `BPEStreamingDetokenizer` (Qwen1.5, Llama-3.2), `SPMStreamingDetokenizer` (Mistral-7B-Instruct-v0.3) and `NaiveStreamingDetokenizer`. `black` and `isort --profile=black` are clean.

## Pre-existing failure, unrelated to this PR

`test_tokenizers` fails on `Llama-3.2-1B-Instruct-4bit` and `Falcon3-7B-Instruct-4bit` with `'a,b' != 'a ,b'` on `transformers==5.16.1`, which now warns that `clean_up_tokenization_spaces=True` is destructive for BPE tokenizers. I confirmed this reproduces identically on unmodified `v0.31.3`, so it is not introduced here — but it does mean the file is not green on current `transformers`, and it may be worth a separate issue.

## Two things noticed while working on this, not changed here

- `tests/test_tokenizers.py` sets `tokenizer._detokenizer = NaiveStreamingDetokenizer(tokenizer)` and then calls `check_tokenizer`, which reads `tokenizer.detokenizer` — the property, which ignores `_detokenizer`. That subtest currently re-tests the BPE path rather than the naive one.
- `TokenizerWrapper.__getattr__` returns `self._detokenizer` for `attr == "detokenizer"`. Since the property shadows `__getattr__`, that branch is only reachable when the property itself raises `AttributeError`, in which case the original error is replaced by `'TokenizerWrapper' object has no attribute '_detokenizer'`. It cost me some time; happy to open a separate PR if you would like it removed.

### DIFF
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -1,3 +1,4 @@
+import copy
 import importlib
 import json
 import warnings
@@ -451,8 +452,22 @@ class TokenizerWrapper:
     def detokenizer(self):
         """
         Get a stateful streaming detokenizer.
+
+        The detokenizer's per-stream state is exactly what ``reset()``
+        installs; everything else it holds (the id-to-token map, the byte
+        decoder) is derived from the tokenizer and immutable. Building one
+        prototype per wrapper and handing out shallow copies keeps every
+        stream independent while materialising the vocabulary only once --
+        rebuilding it per call costs O(vocab) on the first token of every
+        generation.
         """
-        return self._detokenizer_class(self)
+        prototype = self.__dict__.get("_detokenizer_prototype")
+        if prototype is None:
+            prototype = self._detokenizer_class(self)
+            self.__dict__["_detokenizer_prototype"] = prototype
+        detokenizer = copy.copy(prototype)
+        detokenizer.reset()
+        return detokenizer
 
     def __getattr__(self, attr):
         if attr == "detokenizer":
--- a/tests/test_tokenizers.py
+++ b/tests/test_tokenizers.py
@@ -72,6 +72,63 @@ class TestTokenizers(unittest.TestCase):
         tokenizer._detokenizer = NaiveStreamingDetokenizer(tokenizer)
         self.check_tokenizer(tokenizer)
 
+    def test_detokenizer_not_rebuilt_per_access(self):
+        """Accessing .detokenizer must not re-read the vocabulary.
+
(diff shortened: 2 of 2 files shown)


zcbenz (changes requested):
Thanks for the PR but the code is too hacky, I'm closing this in favor of #1873.

PR #1855, opened 2026-09-07
Fix padding-token inflation in batched perplexity evals

LxYuan0420:
This PR is:
- to fix the padding-token log-probability being included in the score of shorter documents.
- to add a regression test that batches a short and a long document together and verifies that each result matches scoring the same document individually.

TLDR:
mlx_lm.evaluate reports different perplexity for the same model on the same data depending on --batch-size, because every document that isn't the longest in its batch gets one padding-token log-probability added to its score

####  How reproduce (change to different batch size):
```bash
$ mlx_lm.evaluate --model mlx-community/SmolLM2-135M-Instruct --tasks wikitext --batch-size 1 --no-apply-chat-template

$ mlx_lm.evaluate --model mlx-community/SmolLM2-135M-Instruct --tasks wikitext --batch-size 8 --no-apply-chat-template

$ mlx_lm.evaluate --model mlx-community/SmolLM2-135M-Instruct --tasks wikitext --batch-size 16 --no-apply-chat-template
```

#### Before-screenshot:
Screenshot 2026-09-07 at 10 35 55 AM

#### After-screenshot:

Screenshot 2026-09-07 at 10 27 16 AM

#### Evidence table:

| Version | Batch size | Word perplexity | Byte perplexity | Bits/byte |
|---|---:|---:|---:|---:|
| Before | 1 | 26.8173617726 | 1.8497836726 | 0.8873565612 |
| Before | 8 | 26.8773976779 | 1.8505573757 | 0.8879598664 |
| Before | 16 | 26.8815974797 | 1.8506114473 | 0.8880020201 |
| After | 1 | 26.8173617726 | 1.8497836726 | 0.8873565612 |
| After | 8 | 26.8168858929 | 1.8497775342 | 0.8873517736 |
| After | 16 | 26.8168858929 | 1.8497775342 | 0.8873517736 |

before the fix, bs=8 and bs=16 inflated Wikitext perplexity relative to bs=1; after the fix, batched scoring converges and no longer adds the padding-token score. However, after the fix, bs=1 is still off from bs=8/16 by about 0.00048 word perplexity, which is tiny compared with the previous ~0.06 drift.

### DIFF
--- a/mlx_lm/evaluate.py
+++ b/mlx_lm/evaluate.py
@@ -303,7 +303,7 @@ class MLXLM(LM):
         for i in tqdm(range(0, len(inputs), self._batch_size)):
             batch = inputs[i : i + self._batch_size]
             scores, lengths, _ = self._score_fn(batch)
-            mask = mx.arange(scores.shape[-1]) < lengths[:, None]
+            mask = mx.arange(scores.shape[-1]) < lengths[:, None] - 1
             all_scores.extend((mask * scores).sum(axis=-1).tolist())
 
         return all_scores
--- a/tests/test_evaluate.py
+++ b/tests/test_evaluate.py
@@ -71,21 +71,22 @@ class TestMLXLM(unittest.TestCase):
                 logits = mx.arange(11, dtype=mx.float32)
                 return mx.broadcast_to(logits, (*inputs.shape, logits.shape[0]))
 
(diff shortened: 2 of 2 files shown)


michalk8:
Thanks a lot for catching this! I've modified the test to use mocks instead, once they pass, will merge this.

michalk8 (approved):

PR #1856, opened 2026-09-07
Fix gradients through routing indices

dogukanveziroglu:
I noticed a few related issues after [mlx#3820](https://github.com/ml-explore/mlx/pull/3820). This fixes the backpropagation errors by adding `mx.stop_gradient` to selection indices while preserving gradients through scores and model outputs.

I used ai for the tests.


zcbenz (changes requested):
The test is over-complicated for such small change, a simple test like the one in https://github.com/ml-explore/mlx-lm/pull/1787 would be preferred but it is not really meaningful adding the same test for all the models. Since this is a trivial change, I think we should just ignore the tests.

zcbenz (approved):

PR #1857, opened 2026-09-07
Fix config parsing for Nemotron-H checkpoints saved by transformers

brunopistone:
Loading or converting a Nemotron-H checkpoint written by a recent transformers release fails in three separate places:

- `num_hidden_layers` is a property of `layers_block_type` in transformers, so it never appears in the saved config and ModelArgs raises "missing 1 required positional argument: 'num_hidden_layers'". Derive the layer count from the block list instead, accept `layer_types` as an alias for `layers_block_type`, map both the current and the legacy block-type spellings, and raise a listed-values error for unknown ones.

- transformers tags non-finite floats, e.g. `{"__float__": "Infinity"}`, so that config.json stays valid JSON for every parser. load_config passed the tag through untouched, so time_step_limit reached mx.clip as a dict. Decode the tags on load and re-encode them on save, which keeps a converted config readable by both libraries.

- A config missing a required key surfaced as a bare `__init__` `TypeError` naming only the first missing argument and no model type. Report the model type and every missing key instead.

Each new test fails on main with the corresponding error.

- AI usage disclosure: 
The patch was co-written with Claude Code under my directions. Manual reviews and live tests were performed.

### DIFF
--- a/mlx_lm/models/base.py
+++ b/mlx_lm/models/base.py
@@ -12,13 +12,19 @@ from mlx.utils import tree_map
 class BaseModelArgs:
     @classmethod
     def from_dict(cls, params):
-        return cls(
-            **{
-                k: v
-                for k, v in params.items()
-                if k in inspect.signature(cls).parameters
-            }
-        )
+        accepted = inspect.signature(cls).parameters
+        kwargs = {k: v for k, v in params.items() if k in accepted}
+        missing = [
+            name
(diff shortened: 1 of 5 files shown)


michalk8 (approved):
I've reverted the change in `ModelArgs.__init__`, as it's out-of-scope of this PR. Otherwise, LGTM!

PR #1317, opened 2026-05-28
Add MiniCPM5 XML tool-call parser

scrappylabsai:
## Summary

Adds a tool-call parser for [openbmb/MiniCPM5-1B](https://huggingface.co/openbmb/MiniCPM5-1B) (and its `-MLX`, `-GGUF`, `-SFT` variants).

The model emits tool calls in XML — not JSON — so generic parsers don't pick it up, and `mlx_lm.server` currently warns:

```
WARNING - Received tools but model does not support tool calling.
```

The MiniCPM5 wire format:

```xml
<function name="get_weather">
  <param name="city">Tokyo</param>
  <param name="date">2024-06-27</param>
</function>
```

Multi-line / special-character values may be wrapped in `<![CDATA[...]]>`. Attributes can be single- or double-quoted.

## Implementation

- `mlx_lm/tool_parsers/minicpm5.py` — stateless `parse_tool_call(text, tools)` returning a single `{"name", "arguments"}` dict; raises `ValueError` for the `ToolCallFormatter` to log + drop.
- Schema-aware: rejects unknown functions, unknown params, duplicate params, missing required params, and `<param>` tags without a `name=` attribute.
- Type coercion: strings pass through; other types try strict JSON, then `ast.literal_eval`, then fall back to raw.
- Auto-detected by `<function name=` + `<param name=` in the chat template.

Logic ported from [SGLang's `MiniCPM5Detector`](https://github.com/sgl-project/sglang/pull/25600).

## Tests

`tests/test_tool_parsing.py`:
- Added MiniCPM5 cases to the existing multi-parser sweep (numeric args + string args).
- New `test_minicpm5` covering: CDATA multi-line params, non-string typed params (array, boolean), body-only segment form, missing-required validation, unknown-param validation, duplicate-param validation, missing-name-attr validation, single-quoted attributes.

```
Ran 6 tests in 0.001s
OK
```

## Smoke test

Live tested against `openbmb/MiniCPM5-1B-MLX` running under `mlx_lm.server`:

Single call:
```json
{
  "finish_reason": "tool_calls",
  "tool_calls": [{
    "function": {
      "name": "get_weather",
      "arguments": "{\"city\": \"Tokyo\", \"date\": \"2024-06-27\"}"
    },
    "type": "function",
    "id": "..."
  }]
}
```

Two consecutive `<function>...</function>` blocks both extracted cleanly via the state machine (one segment per pair).

## Test plan

[...]

PR #1859, opened 2026-09-08
Add MiniCPM5 tool calling support

arvind1705:
OpenBMB has released new model [openbmb/MiniCPM5-2B-MLX](https://huggingface.co/openbmb/MiniCPM5-2B-MLX).
Currently tool calling is broken. When used with mlx-lm below logs are there 

`WARNING - Received tools but model does not support tool calling.`

So based on PR reference from VLLM - https://github.com/vllm-project/vllm/pull/43175

I've added below -

- **New tool parser** `mlx_lm/tool_parsers/minicpm5.py`: Parses MiniCPM5's XML-style `<function name="..."><param name="...">...</param></function>` blocks using `xml.etree.ElementTree` with a regex fallback. Respects tool schema for type-aware deserialization and handles CDATA blocks.

- **Auto-detection** in `_infer_tool_parser()`: Added `elif "<function name=" in chat_template: return "minicpm5"` so the parser is automatically registered when loading MiniCPM5 models.

- **Tests** in `tests/test_tool_parsing.py`: Covers numeric/string params, multiple calls, CDATA, and empty arguments.

- AI usage disclosure: I've used AI to generate few parts code and tests. I've validated against the behaviour of MiniCPM5 locally and also checked vllm implementation also


arvind1705:
Same change in https://github.com/ml-explore/mlx-lm/pull/1317.
Closing this one.

PR #1861, opened 2026-09-08
Update DeepSeek-V3.2 to use PipelineMixin

twallgren:

The issue this is addressing:
deepseek_v32.py had its own independent pipeline implementation, duplicating logic that already existed in the shared PipelineMixin base class but not benefiting from bug fixes.

The solution:
Converted DeepseekV32Model to inherit PipelineMixin like every other pipeline-capable model so it benefits from the shared, tested pipeline logic (including uneven-split support) instead of maintaining another copy of similar code.

Note: an earlier version of this PR also fixed a layer-drop bug directly in pipeline.py's core split calculation, but that was independently fixed and merged upstream in https://github.com/ml-explore/mlx-lm/pull/1816 first so that part was dropped from this PR to avoid duplicating it.

I needed these fixes to get DeepSeek-V3.2-family models running correctly in pipeline mode on my 3-node cluster.

### DIFF
--- a/mlx_lm/models/deepseek_v32.py
+++ b/mlx_lm/models/deepseek_v32.py
@@ -12,6 +12,7 @@ from .activations import swiglu
 from .base import BaseModelArgs, create_attention_mask, scaled_dot_product_attention
 from .cache import CacheList, KVCache
 from .mla import MultiLinear
+from .pipeline import PipelineMixin
 from .rope_utils import initialize_rope
 from .switch_layers import SwitchGLU
 
@@ -412,7 +413,7 @@ class DeepseekV32DecoderLayer(nn.Module):
         return h + r
 
 
-class DeepseekV32Model(nn.Module):
+class DeepseekV32Model(PipelineMixin, nn.Module):
     def __init__(self, config: ModelArgs):
         super().__init__()
         self.vocab_size = config.vocab_size
@@ -421,28 +422,7 @@ class DeepseekV32Model(nn.Module):
             DeepseekV32DecoderLayer(config, idx)
             for idx in range(config.num_hidden_layers)
         ]
-        self.start_idx = 0
-        self.end_idx = len(self.layers)
-        self.num_layers = self.end_idx
-
         self.norm = nn.RMSNorm(config.hidden_size, eps=config.rms_norm_eps)
-        self.pipeline_rank = 0
-        self.pipeline_size = 1
-
-    def pipeline(self, group):
-        # Split layers in reverse so rank=0 gets the last layers and
-        # rank=pipeline_size-1 gets the first
-        self.pipeline_rank = group.rank()
-        self.pipeline_size = group.size()
-        layers_per_rank = len(self.layers) // self.pipeline_size
-        extra = len(self.layers) - layers_per_rank * self.pipeline_size
-        if self.pipeline_rank < extra:
-            layers_per_rank += 1
-        self.start_idx = (self.pipeline_size - self.pipeline_rank - 1) * layers_per_rank
-        self.end_idx = self.start_idx + layers_per_rank
-        self.layers = self.layers[: self.end_idx]
-        self.layers[: self.start_idx] = [None] * self.start_idx
-        self.num_layers = len(self.layers) - self.start_idx
 
     def __call__(
         self,
@@ -455,23 +435,22 @@ class DeepseekV32Model(nn.Module):
         pipeline_size = self.pipeline_size
 
         if cache is None:
-            cache = [None] * self.num_layers
+            cache = [None] * len(self.pipeline_layers)
         mask = create_attention_mask(
-            h, cache[0][0] if cache[0] else None, return_array=True
+            h, cache[0][0] if cache and cache[0] else None, return_array=True
         )
 
         # Receive from the previous process in the pipeline
-
         if pipeline_rank < pipeline_size - 1:
             h = mx.distributed.recv_like(h, (pipeline_rank + 1))
 
-        for i in range(self.num_layers):
-            h = self.layers[self.start_idx + i](h, mask, cache[i])
+        for layer, c in zip(self.pipeline_layers, cache):
+            h = layer(h, mask, c)
 
         # Send to the next process in the pipeline
         if pipeline_rank != 0:
             h = mx.distributed.send(h, (pipeline_rank - 1) % pipeline_size)
-            if cache[-1] is not None:
+            if cache and cache[-1] is not None:
                 cache[-1][0].keys = mx.depends(cache[-1][0].keys, h)
 
         # Broadcast h while keeping it in the graph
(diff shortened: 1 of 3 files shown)


azamamirza:
Ran into the deepseek_v32 side of this yesterday while working on #1816: it has the same silently-dropped-layers bug that PR fixed in the mixin (7 layers on 2 ranks ran layer 3 on no rank), and @michalk8 suggested exactly this subclassing approach as a follow-up in the #1816 discussion. You got there first, and this version is cleaner than what I had staged: dropping the override entirely also picks up the new split= argument for free.

Verified locally on Apple Silicon (M5 Max):

1. Applied my regression tests from the #1816 follow-up work to this branch unmodified: all pass, including a 7-layer / 2-rank uneven assignment check on the real DeepseekV32Model (fails on current main, passes here) and a tiny-config forward-equality test through test_pipeline with a size-1 group.
2. tests/model_parallel_tests.py: 4 passed + 8 subtests. tests/test_models.py -k "deepseek or qwen3": 6 passed.
3. The sanitize() change reads correct to me but I did not execute a pipeline-sharded checkpoint load to verify that half.
4. We hit the original issue testing on a 2 Mac ring of uneven hardware (M4 Max and M5 Max machines with differing specs).

This PR has no committed tests; the diff below adds the deepseek_v32 coverage and applies cleanly to this branch if you want to fold it in.

<details>
<summary>regression tests (applies to this branch)</summary>

```diff
diff --git a/tests/model_parallel_tests.py b/tests/model_parallel_tests.py
index 1d9bb89..af73c83 100644
--- a/tests/model_parallel_tests.py
+++ b/tests/model_parallel_tests.py
@@ -6,7 +6,7 @@ import unittest
 import mlx.core as mx
 
 import mlx_lm
-from mlx_lm.models import qwen3_moe
+from mlx_lm.models import deepseek_v32, qwen3_moe
 from mlx_lm.models.pipeline import PipelineMixin
 
 
@@ -170,6 +170,29 @@ class TestModelParallel(unittest.TestCase):
                 "max_position_embeddings": 256,
                 "tie_word_embeddings": False,
             },
+            {
+                "model_type": "deepseek_v32",
+                "vocab_size": 128,
+                "hidden_size": 64,
+                "index_head_dim": 16,
+                "index_n_heads": 2,
+                "index_topk": 4,
+                "intermediate_size": 128,
+                "moe_intermediate_size": 32,
+                "num_hidden_layers": 4,
+                "num_attention_heads": 4,
+                "num_key_value_heads": 4,
+                "n_shared_experts": 1,
+                "n_routed_experts": 4,
+                "kv_lora_rank": 8,
+                "q_lora_rank": 8,
+                "qk_rope_head_dim": 8,
+                "v_head_dim": 16,
+                "qk_nope_head_dim": 16,
+                "num_experts_per_tok": 2,
+                "first_k_dense_replace": 1,
+                "max_position_embeddings": 256,
+            },
         ]
         mx.random.seed(0)
         for config in test_configs:
@@ -247,15 +270,40 @@ class TestModelParallel(unittest.TestCase):
         }
 
         size = 2
-        args = qwen3_moe.ModelArgs.from_dict(config)
-        assigned = []
-        for rank in range(size):
-            model = qwen3_moe.Model(args)
-            model.model.pipeline(Group(rank, size))
-            assigned.extend(
-                i for i, l in enumerate(model.model.layers) if l is not None
-            )
-        self.assertEqual(sorted(assigned), list(range(7)))
+        dsv32_config = {
+            "model_type": "deepseek_v32",
+            "vocab_size": 128,
+            "hidden_size": 64,
+            "index_head_dim": 16,
+            "index_n_heads": 2,
+            "index_topk": 4,
+            "intermediate_size": 128,
+            "moe_intermediate_size": 32,
+            "num_hidden_layers": 7,
+            "num_attention_heads": 4,
+            "num_key_value_heads": 4,
+            "n_shared_experts": 1,
+            "n_routed_experts": 4,
+            "kv_lora_rank": 8,
+            "q_lora_rank": 8,
+            "qk_rope_head_dim": 8,
+            "v_head_dim": 16,
+            "qk_nope_head_dim": 16,
+            "num_experts_per_tok": 2,
+            "first_k_dense_replace": 1,
+            "max_position_embeddings": 256,
+        }
+        for arch, arch_config in ((qwen3_moe, config), (deepseek_v32, dsv32_config)):
+            with self.subTest(model_type=arch_config["model_type"]):
+                args = arch.ModelArgs.from_dict(arch_config)
+                assigned = []
+                for rank in range(size):
+                    model = arch.Model(args)
+                    model.model.pipeline(Group(rank, size))
+                    assigned.extend(
+                        i for i, l in enumerate(model.model.layers) if l is not None
+                    )
+                self.assertEqual(sorted(assigned), list(range(7)))
 
 
 if __name__ == "__main__":
```

</details>

twallgren:
Thanks @azamamirza for the review, feedback, testing, and regression tests! I applied your diff and confirmed that the tests pass. I'll remove the draft status.
Since you're working on the uneven split topic, you might be interested in https://github.com/ml-explore/mlx-lm/pull/1862 and https://github.com/ml-explore/mlx-lm/pull/1863 as well.

mlx_lm/models/deepseek_v32.py
```diff
@@ -455,23 +435,22 @@ def __call__(
         pipeline_size = self.pipeline_size
 
         if cache is None:
-            cache = [None] * self.num_layers
+            cache = [None] * len(self.pipeline_layers)
         mask = create_attention_mask(
-            h, cache[0][0] if cache[0] else None, return_array=True
+            h, cache[0][0] if cache and cache[0] else None, return_array=True
```
michalk8: I don't think these changes are needed, can you please revert?

tests/model_parallel_tests.py
```diff
@@ -246,15 +269,40 @@ def test_pipeline_uneven_model(self):
...
-        args = qwen3_moe.ModelArgs.from_dict(config)
-        assigned = []
-        for rank in range(size):
-            model = qwen3_moe.Model(args)
-            model.model.pipeline(Group(rank, size))
-            assigned.extend(
-                i for i, l in enumerate(model.model.layers) if l is not None
-            )
-        self.assertEqual(sorted(assigned), list(range(7)))
+        dsv32_config = {
```
michalk8: Don't think this is needed, as both Qwen3 and DeepSeek V3.2 use the pipeline mixin, would remove not to clutter the tests.

michalk8 (changes requested):
Can you please split this PR into two, one for DeepSeek to use PipelineMixin (happy to merge) and the Qwen3 MoE in a separate one (I'm not sure about this one yet, if it's indeed a bug, it might be present in other models).

tests/model_parallel_tests.py
twallgren: Agreed! I've remove it.

mlx_lm/models/deepseek_v32.py
twallgren: Agreed! Done!

twallgren:
Thanks for the review and feedback, @michalk8! I've split the PR into two as you requested. The new PR is #1875.

michalk8 (approved):
LGTM, thanks!

PR #1863, opened 2026-09-08
Support uneven tensor-parallel sharding in shard() (GQA/MLA-aware)

twallgren:

The issue this is addressing:
Tensor parallelism required the model's attention head count to divide evenly across the ranks.

The solution:
Adds support to more model files for uneven sharding so ranks can get different amounts of work and adds a small shared helper in pipeline.py used by all of them to compute the per-rank split sizes.

This depends on https://github.com/ml-explore/mlx/pull/4465 so it should not be merged until that PR is merged.

I needed this change so I could use tensor parallelism with a wider variety of models on my 3 node cluster. I've been running it successfully for a while and testing it with a variety of models with great results.


azamamirza:
Read through the sharding approach: distributing whole KV-head groups so n_heads_local stays an exact multiple of n_kv_heads_local on every rank is the right invariant for GQA under uneven splits, and the zero-head rank check failing loudly matches how the pipeline side now behaves. I have not run this branch since it's gated on ml-explore/mlx#4465.

An offer for when 4465 lands: I run a 4-Mac Thunderbolt/JACCL ring with genuinely uneven hardware (an M4 Max and an M5 Max with different memory headroom), which is the exact deployment this PR exists for. Happy to validate real uneven tensor-parallel inference on it, including memory-per-rank and throughput against the even-split baseline and against uneven pipeline splits from #1816. 

That's data a single-machine test can't produce. Ping me here when it's testable.

PR #1866, opened 2026-09-09
Fix NameError in benchmark script

LxYuan0420:
This PR is:
- to name error issue in the benchmark script (introduced from lint cleanup PR #1860 )
- to rename the loop variable to `_response` and the return to `return _response`

note: In Python, `_response` is still a normal variable. The leading underscore tells lint tools: "this variable is intentionally unused in the loop body” since we return it after the loop, it still carries the last streamed response.

Before:
Screenshot 2026-09-09 at 10 49 36 PM

After:
Screenshot 2026-09-09 at 10 49 41 PM

### DIFF
--- a/mlx_lm/benchmark.py
+++ b/mlx_lm/benchmark.py
@@ -125,7 +125,7 @@ def main():
     prompt = prompts[0]
 
     def single_bench():
-        for _ in stream_generate(
+        for _response in stream_generate(
             model,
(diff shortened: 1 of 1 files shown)


michalk8 (approved):
Thanks for catching this, LGTM!

zcbenz (approved):

PR #1868, opened 2026-09-09
fix: prevent negative trim() from corrupting cache offset

saitejagirada:
### Description

This PR fixes a bug in the cache `trim()` methods where passing a negative `n` value artificially *increases* the cache `offset`, corrupting internal state and leading to reads of uninitialized memory.

When `n < 0`, `min(self.offset, n)` evaluates to `n` (since any negative value is less than a non-negative offset). Subtracting a negative `n` from `self.offset` then increases it:

```python
# offset=10, n=-5
n = min(10, -5)   # n = -5
self.offset -= -5  # offset becomes 15!
```

This PR adds `max(0, n)` to clamp negative values in the `trim()` methods of `KVCache`, `ConcatenateKVCache`, `QuantizedKVCache`, and `RotatingKVCache`.

### Verification

Tested end-to-end on `mlx-community/Qwen1.5-0.5B-Chat-4bit` (Apple Silicon, macOS).

**Test script:**
```python
from mlx_lm import load
from mlx_lm.models.cache import make_prompt_cache, trim_prompt_cache
import mlx.core as mx

model, tokenizer = load("mlx-community/Qwen1.5-0.5B-Chat-4bit")
tokens = mx.array(tokenizer.encode("Explain what machine learning is in simple terms."))
logits = model(tokens[None], cache=cache)
mx.eval(logits, *[c.state for c in cache])

print(f"Before: offset={cache[0].offset}")
trim_prompt_cache(cache, -5)
print(f"After:  offset={cache[0].offset}")
```

**Before fix (`main` branch):**
```
Cache offset before trim: 10
Cache offset after trim(-5): 15       ← offset INCREASED
trim() returned: -5                   ← returned a negative trim count
⚠️  OFFSET INCREASED from 10 to 15!
```

**After fix (`fix/negative-trim` branch):**
```
Cache offset before trim: 10
Cache offset after trim(-5): 10       ← offset unchanged
trim() returned: 0                    ← safely clamped to 0
✅ Offset unchanged at 10 — negative trim safely clamped
```

### DIFF
--- a/mlx_lm/models/cache.py
+++ b/mlx_lm/models/cache.py
@@ -231,7 +231,7 @@ class ConcatenateKVCache(_BaseCache):
         return True
 
     def trim(self, n):
-        n = min(self.offset, n)
+        n = min(self.offset, max(0, n))
         self.offset -= n
         return n
 
@@ -321,7 +321,7 @@ class QuantizedKVCache(_BaseCache):
         return True
 
     def trim(self, n):
-        n = min(self.offset, n)
+        n = min(self.offset, max(0, n))
         self.offset -= n
         return n
 
@@ -389,7 +389,7 @@ class KVCache(_BaseCache):
         return True
 
     def trim(self, n):
-        n = min(self.offset, n)
+        n = min(self.offset, max(0, n))
         self.offset -= n
         return n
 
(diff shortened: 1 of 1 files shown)


zcbenz (changes requested):
Thanks for the PR but I don't think we would need to guard against every crafted case the does not happen in reality.

PR #1870, opened 2026-09-10
Chunkwise-parallel gated delta rule for training

adityak74:
## Problem

The gated delta layers fall back to `gated_delta_ops` whenever the Metal kernel is unavailable, and that includes **every training step** — `use_kernel=not self.training`, because the kernel has no VJP. `gated_delta_ops` is a sequential loop over timesteps, so autograd retains one state per token.

Measured on one layer (Hk=16, Hv=32, Dk=Dv=128), forward and backward, M4 Pro:

| seq | peak memory | time |
|---:|---:|---:|
| 512 | 4.29 GB | 0.36 s |
| 1024 | 10.18 GB | 1.16 s |
| 2048 | 31.44 GB | 4.20 s |
| 4096 | 108.33 GB | 39.02 s |

About 8 MB per token per layer, growing linearly, and superlinear in time. Five model families route through this during training: `qwen3_5`, `qwen3_next`, `kimi_k3`, `kimi_linear`, `bailing_moe_v3`.

## Change

`gated_delta_ops` becomes a thin dispatcher: scalar gating without a mask takes a new `gated_delta_chunkwise`, everything else goes to `gated_delta_sequential`, which is the existing loop under a name that says what it is. The existing kernel tests now compare against `gated_delta_sequential` so they keep testing what they tested before.

`gated_delta_chunkwise` is the chunkwise parallel form of the gated delta rule ([arXiv:2406.06484](https://arxiv.org/abs/2406.06484), [arXiv:2412.06464](https://arxiv.org/abs/2412.06464)). Splitting the sequence into chunks of 64 and solving one unit lower-triangular system per chunk keeps `O(T / chunk)` states instead of `O(T)` and replaces the per-step loop with matmuls.

Same setup, with the new path:

| seq | peak memory | time | vs sequential |
|---:|---:|---:|---|
| 512 | 0.33 GB | 0.04 s | 13x mem, 10x speed |
| 1024 | 0.97 GB | 0.09 s | 10x mem, 12x speed |
| 2048 | 1.74 GB | 0.23 s | 18x mem, 19x speed |
| 4096 | 4.73 GB | 0.60 s | **23x mem, 65x speed** |

Inference is untouched: that path uses the Metal kernel and never reaches this code.

## Scope

The dispatch prefers the chunkwise path only for scalar gating without a mask. Vectorized gating (`g.ndim == 4`) and masked sequences continue to use the reference loop, so the change cannot alter results for cases it does not cover.

## Two implementation notes

`mx.linalg.tri_inv` is CPU-only and has no VJP, so it cannot appear in a training graph. The unit-triangular inverse is instead built from matmuls by 2x2 block recursion in log2(n) levels, which autodiff handles natively and which parallelizes across the batch — for the batched 64x64 case here that also measured faster than the LAPACK path (5.9 ms vs 7.4 ms for 1024 matrices, before counting the device round trip).

The textbook derivation divides by the cumulative decay `G_j`, which underflows and sends the intermediate writes to infinity on strongly decaying heads. This implementation carries bounded ratios `G_j / G_i` (at most 1 for `i <= j`) instead. The tests cover that regime explicitly.

## Testing

`tests/test_gated_delta.py` adds a `TestGatedDeltaChunkwise` case covering:

- agreement with `gated_delta_ops` at mild (0.2) and strong (4.0) decay
- sequence lengths that are and are not a multiple of the chunk size
- grouped-query heads (`Hv // Hk == 4`)
- gradients through all five inputs

Outputs match to ~1e-6 relative in fp32 and gradients to ~1e-4. The existing kernel tests pass unchanged (11 passed, 32 subtests).

Verified to merge cleanly into current `main`, and the merged tree passes the suite (training dispatch agrees with the kernel to 1.5e-6).

### DIFF
--- a/mlx_lm/models/gated_delta.py
+++ b/mlx_lm/models/gated_delta.py
@@ -548,6 +548,29 @@ def gated_delta_ops(
     mask: Optional[mx.array] = None,
 ) -> Tuple[mx.array, mx.array]:
     """
+    Ops-based gated delta rule, used whenever the Metal kernel is unavailable.
+
+    That includes every training step, since the kernel has no VJP. Scalar
+    gating without a mask takes the chunkwise-parallel path, which is
+    equivalent to the sequential recurrence but retains O(T / chunk) states
+    for autograd instead of one per timestep; everything else falls back to
+    :func:`gated_delta_sequential`.
+    """
+    if mask is None and g.ndim == 3:
+        return gated_delta_chunkwise(q, k, v, g, beta, state)
+    return gated_delta_sequential(q, k, v, g, beta, state, mask)
+
+
+def gated_delta_sequential(
+    q: mx.array,
+    k: mx.array,
+    v: mx.array,
+    g: mx.array,
+    beta: mx.array,
+    state: Optional[mx.array] = None,
+    mask: Optional[mx.array] = None,
+) -> Tuple[mx.array, mx.array]:
+    """
     Ops-based reference implementation for prompt prefill (sequential loop).
     Supports both scalar and vectorized gating.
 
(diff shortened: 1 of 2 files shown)


Goekdeniz-Guelmez:
this is already implemented in [MLX-LM-LoRA](https://github.com/Goekdeniz-Guelmez/mlx-lm-lora) which already provides chunked and efficient long context training for all linear models within MLX-LM. you can also manually change the chunk size via:

```python
from mlx_lm_lora.trainer.sft_trainer import SFTTrainingArgs
args = SFTTrainingArgs(..., recurrence_chunk_size=512, ...)
```

works with the other algorithms like DPO, ORPO etc. too.

adityak74:
Thanks for the pointer — I hadn't seen `recurrent_patch.py`. Having read it, I think the two are complementary rather than duplicates, and it's worth being precise about why.

`enable_memory_safe_recurrences` keeps `_gated_delta_step_ops`, the sequential per-token loop, and wraps each block of `chunk_size` steps in `mx.checkpoint`. That is activation checkpointing of the existing recurrence: memory falls to O(T / chunk) retained states, the arithmetic is unchanged, and the backward pass recomputes the token loop inside each block. Time stays superlinear in T, as in the table above (4.20 s at 2,048 → 39.02 s at 4,096 for one layer), and checkpointing adds a forward recompute on top.

This PR changes the arithmetic. `gated_delta_chunkwise` is the chunkwise-parallel form of the gated delta rule (arXiv:2406.06484 §4, arXiv:2412.06464): within a chunk the recurrence is solved as one unit-lower-triangular system and the intra- and inter-chunk contributions are matmuls, so there is no per-token loop in either direction. Same outputs to bf16 tolerance (the tests compare against `gated_delta_sequential`), O(T / chunk) states without recomputation, and the 135× at 4,096 in the table is from removing the loop, not from checkpointing it.

The two compose: your checkpoint wrapper could sit around `gated_delta_chunkwise` blocks just as it sits around the step loop now, which would give the memory of both and the speed of this one. Happy to add a benchmark against `recurrent_patch` at 2,048 and 4,096 if the maintainers want it for the record.

adityak74:
As offered above, here is the benchmark against `mlx_lm_lora.recurrent_patch`, same one-layer setup as the table in the description (Hk=16, Hv=32, Dk=Dv=128, batch 1, bf16 inputs, float32 state), forward + backward w.r.t. q, k, v, M4 Pro 64 GB, chunk 64 for both chunked paths. The checkpointed path is `recurrent_patch.checkpointed_gated_delta` reproduced verbatim around mlx-lm's `_gated_delta_step_ops`; script is `scripts/bench_gdn_paths.py` in the repo linked in the description.

| T | path | peak memory | fwd+bwd |
|---:|---|---:|---:|
| 1,024 | sequential (`gated_delta_ops`, current training path) | 7.09 GB | 1.88 s |
| 1,024 | checkpointed sequential (mlx-lm-lora) | 2.89 GB | 1.34 s |
| 1,024 | chunkwise (this PR) | 1.08 GB | 0.05 s |
| 2,048 | sequential | 21.20 GB | 6.94 s |
| 2,048 | checkpointed sequential | 5.19 GB | 3.35 s |
| 2,048 | chunkwise | 1.97 GB | 0.19 s |
| 4,096 | sequential | (108 GB, table above) | (39 s) |
| 4,096 | checkpointed sequential | 9.78 GB | 6.77 s |
| 4,096 | chunkwise | 4.95 GB | 0.56 s |

So the two do different things, as expected from the code: checkpointing the step loop brings memory down 2–4× and time down at most 2× (the recompute is offset by less allocator pressure), while the chunkwise form is 2× smaller than that and 12–18× faster at 2,048–4,096 because the per-token loop is gone in both directions. They compose if anyone wants both.

PR #1873, opened 2026-09-10
Fix detokenizer getting re-created for every generation

### LINKED ISSUE #1435: FYI: uniform +55–77 ms TTFT regression on 0.31.3 vs 0.27.1 (constant per-call offset, decode flat)

zcbenz:
It used to be cached per tokenizer but #430 changed the behavior, not exactly sure of the reason but it does not seem reasonable creating a new object every time when the property is accessed.

Close https://github.com/ml-explore/mlx-lm/issues/1435

### DIFF
--- a/mlx_lm/generate.py
+++ b/mlx_lm/generate.py
@@ -702,6 +702,7 @@ def stream_generate(
         prompt = mx.array(prompt)
 
     detokenizer = tokenizer.detokenizer
+    detokenizer.reset()
 
     kwargs["max_tokens"] = max_tokens
 
--- a/mlx_lm/tokenizer_utils.py
+++ b/mlx_lm/tokenizer_utils.py
@@ -503,12 +503,12 @@ class TokenizerWrapper:
         """
         Get a stateful streaming detokenizer.
         """
-        return self._detokenizer_class(self)
+        if not hasattr(self, "_detokenizer"):
+            self._detokenizer = self._detokenizer_class(self)
+        return self._detokenizer
 
     def __getattr__(self, attr):
-        if attr == "detokenizer":
-            return self._detokenizer
-        elif attr == "eos_token_ids":
+        if attr == "eos_token_ids":
             return self._eos_token_ids
         elif attr.startswith("_"):
             return self.__getattribute__(attr)
--- a/tests/test_tokenizers.py
+++ b/tests/test_tokenizers.py
@@ -67,7 +67,9 @@ class TestTokenizers(unittest.TestCase):
 
         # Try one with a naive detokenizer
         tokenizer = load_tokenizer("mlx-community/Llama-3.2-1B-Instruct-4bit")
-        tokenizer._detokenizer = NaiveStreamingDetokenizer(tokenizer)
+        detokenizer = NaiveStreamingDetokenizer(tokenizer)
+        tokenizer._detokenizer = detokenizer
+        self.assertEqual(tokenizer.detokenizer, detokenizer)
         self.check_tokenizer(tokenizer)
 
     def test_special_tokens(self):


michalk8 (changes requested):
For `generate.py`, this works, but for server; to quickly reproduce (roughly simulates what the server does, I observed this as well launching the server and passing two requests):

```python
from mlx_lm.utils import load_tokenizer
tok = load_tokenizer('mlx-community/Qwen1.5-0.5B-Chat-4bit')
A = tok.encode('over 40 years', add_special_tokens=False)
B = tok.encode('disengage, allowing now', add_special_tokens=False)
def run(da, db):
    a = b = ''
    for x, y in zip(A, B):
        da.add_token(x); a += da.last_segment
        db.add_token(y); b += db.last_segment
    return a, b
D = tok._detokenizer_class
print('separate:', run(D(tok), D(tok)))
s = D(tok)
print('shared  :', run(s, s))
```

I was actually cleaning up the detokenizer, and think I have some solution there. If you prefer, I'll fix it there, otherwise the server.py needs to be fixed here.

zcbenz:
Closing in favor of https://github.com/ml-explore/mlx-lm/pull/1879.

PR #1875, opened 2026-09-10
Fix qwen3_moe sanitize() function's layer-0 MoE assumption

twallgren:

The issue this is addressing:
qwen3_moe.py's sanitize() function decides whether to stack MoE weights by checking for a fixed key on layer 0. If layer 0 isn't a MoE layer then that key won't exist and the gate will end up skipping the stacking for every layer.
Seems like all of the Qwen3 MoE configs that are out right now have a MoE layer at layer 0 so they're not affected by this bug, but that could change so we should be prepared for that situation.

The solution:
Changed qwen3_moe.py's sanitize check from a fixed layer-0 check to presence-based detection across whichever local MoE layers actually exist in the checkpoint.

### DIFF
--- a/mlx_lm/models/qwen3_moe.py
+++ b/mlx_lm/models/qwen3_moe.py
@@ -260,9 +260,18 @@ class Model(nn.Module):
     def sanitize(self, weights):
         if self.args.tie_word_embeddings:
             weights.pop("lm_head.weight", None)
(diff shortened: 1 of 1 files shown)


michalk8 (approved):
Thanks a lot, I've added your fix to rest of the models.

PR #1877, opened 2026-09-11
Reject unsupported compressed-tensors formats instead of 4-bit affine

### LINKED ISSUE #1865: compressed-tensors: float-quantized (FP8) checkpoints silently load as 4-bit affine
### Summary

`mlx_lm` treats every `compressed-tensors` checkpoint whose `format` it doesn't explicitly recognise as **4-bit affine**, including `float-quantized` (FP8). The stored tensors are `F8_E4M3`, so their bytes get reinterpreted as packed 4-bit integers rather than rejected.

### Where

`mlx_lm/utils.py`, in the `compressed-tensors` branch. On `main`:

```python
elif quant_method == "compressed-tensors":
    if quantization_config.get("format") == "nvfp4-pack-quantized":
        quantization = {"group_size": 16, "bits": 4, "mode": "nvfp4"}
    elif quantization_config.get("format") == "mxfp4-pack-quantized":
        quantization = {"group_size": 32, "bits": 4, "mode": "mxfp4"}
    else:
        quantization = {"group_size": 32, "bits": 4, "mode": "affine"}   # <-- fp8 lands here
```

On the current PyPI release (0.31.3, `utils.py:380`) there is no format check at all — every `compressed-tensors` model becomes 4-bit affine.

`compressed-tensors` is a *container* format that carries int4, int8 and fp8 alike. The `else` branch assumes one of them.

### Reproduction

Any `llm-compressor` `FP8_DYNAMIC` export. A public one:
`orcarouter/Gemma-4-26B-A4B-it-Uncensored-FP8` (gated, auto-approve).

Its `config.json`:

```json
"quantization_config": {
  "quant_method": "compressed-tensors",
  "format": "float-quantized",
  "config_groups": {"group_0": {
    "targets": ["Linear", "Gemma4TextExperts"],
    "weights":           {"num_bits": 8, "type": "float", "symmetric": true, "dynamic": false},
    "input_activations": {"num_bits": 8, "type": "float", "strategy": "token", "dynamic": true}
  }}
}
```

`recipe.yaml` confirms `scheme: FP8_DYNAMIC`. Reading the safetensors headers shows the dtypes are exactly what the config says:

```
dtype BF16     count=12473   e.g. model.embed_vision.embedding_projection.weight
dtype F8_E4M3  count=1132    e.g. model.language_model.layers.0.experts.0.down_proj.weight
```

with a paired per-output-channel scale:

```
weight : ...down_proj.weight        F8_E4M3  [2816, 704]
scale  : ...down_proj.weight_scale  BF16     [2816, 1]
```

`num_bits: 8` / `type: "float"` is unambiguous in the config, but nothing in the branch reads it.

### Expected

Either dequantize/support FP8, or raise an explicit "unsupported compressed-tensors format" error. Silently selecting a different bit width and quantization mode is the worst of the three outcomes: it produces a model that loads and generates rather than one that fails.

### Suggested fix

Read `num_bits` and `type` out of `config_groups` instead of inferring everything from `format`, and make the fallback an error rather than `affine`:

```python
elif quant_method == "compressed-tensors":
    fmt = quantization_config.get("format")
    if fmt == "nvfp4-pack-quantized":
        quantization = {"group_size": 16, "bits": 4, "mode": "nvfp4"}
    elif fmt == "mxfp4-pack-quantized":
        quantization = {"group_size": 32, "bits": 4, "mode": "mxfp4"}
    elif fmt == "pack-quantized":
        quantization = {"group_size": 32, "bits": 4, "mode": "affine"}
    else:
        raise ValueError(
            f"unsupported compressed-tensors format {fmt!r}; "
            "dequantize to bf16 before converting"
        )
```

`float-quantized` FP8 is recoverable without much work — weights are `F8_E4M3` with a static per-output-channel `<name>.weight_scale`, so `w.to(float32) * scale` reconstructs bf16, and the activation scales are dynamic with nothing stored. Happy to open a PR for either the guard alone or the guard plus an FP8 dequantization path, whichever you'd prefer.

### Environment

- `mlx-lm` 0.31.3 (latest on PyPI) and current `main`
- macOS 26.3, Apple M1 Max, 64 GB

gyanu2507:
Fixes #1865.

`compressed-tensors` is a container. The old `else` branch treated every unrecognized format as 4-bit affine, including `float-quantized` FP8. Those checkpoints store F8_E4M3 weights, so they loaded and generated garbage instead of failing.

Known packed formats still map the same way. Anything else, including FP8, raises with a message to dequantize first.

Tested with `tests.test_utils` coverage for `pack-quantized`, `nvfp4-pack-quantized`, and `float-quantized`.

### DIFF
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -303,6 +303,27 @@ def hf_repo_to_path(hf_repo):
     )
 
 
+def compressed_tensors_quantization(quantization_config: dict) -> dict:
+    """Map a compressed-tensors config to an MLX quantization dict.
+
+    ``compressed-tensors`` is a container format. Only packed integer formats that
+    MLX already implements are accepted. Unknown formats such as ``float-quantized``
+    (FP8) used to fall through to 4-bit affine and silently misinterpret F8_E4M3
+    weights.
+    """
+    fmt = quantization_config.get("format")
+    if fmt == "nvfp4-pack-quantized":
+        return {"group_size": 16, "bits": 4, "mode": "nvfp4"}
+    if fmt == "mxfp4-pack-quantized":
+        return {"group_size": 32, "bits": 4, "mode": "mxfp4"}
+    if fmt == "pack-quantized":
+        return {"group_size": 32, "bits": 4, "mode": "affine"}
+    raise ValueError(
+        f"unsupported compressed-tensors format {fmt!r}; "
+        "dequantize to bf16 before converting"
+    )
+
+
 def load_config(model_path: Path) -> dict:
     with open(model_path / "config.json", "r") as f:
         config = json.load(f)
@@ -434,12 +455,7 @@ def load_model(
             config["quantization_config"] = quantization
             _quantize(quantization)
         elif quant_method == "compressed-tensors":
-            if quantization_config.get("format") == "nvfp4-pack-quantized":
-                quantization = {"group_size": 16, "bits": 4, "mode": "nvfp4"}
-            elif quantization_config.get("format") == "mxfp4-pack-quantized":
-                quantization = {"group_size": 32, "bits": 4, "mode": "mxfp4"}
-            else:
-                quantization = {"group_size": 32, "bits": 4, "mode": "affine"}
+            quantization = compressed_tensors_quantization(quantization_config)
             config["quantization"] = quantization
             config["quantization_config"] = quantization
             _quantize(quantization)
--- a/mlx_lm/utils.py
+++ b/mlx_lm/utils.py
@@ -303,7 +303,7 @@ def hf_repo_to_path(hf_repo):
     )
 
 
-def compressed_tensors_quantization(quantization_config: dict) -> dict:
+def _compressed_tensors_quantization(quantization_config: dict) -> dict:
     """Map a compressed-tensors config to an MLX quantization dict.
 
     ``compressed-tensors`` is a container format. Only packed integer formats that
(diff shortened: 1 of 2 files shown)


mlx_lm/utils.py
```diff
@@ -303,6 +303,27 @@ def hf_repo_to_path(hf_repo):
     )
 
 
+def compressed_tensors_quantization(quantization_config: dict) -> dict:
```
michalk8, line 306: I'd rename to `_compressed_tensors_quantization` to mark it as private.

tests/test_utils.py
```diff
@@ -288,6 +288,20 @@ def test_normal_model_unaffected_by_default(self):
         self.assertIsInstance(model, nn.Module)
         self.assertEqual(loaded_config["model_type"], "llama")
 
+    def test_compressed_tensors_known_formats(self):
+        self.assertEqual(
+            utils.compressed_tensors_quantization({"format": "nvfp4-pack-quantized"}),
```
michalk8, line 293: Can you also please add `mxfp4-pack-quantized` to the test case?

michalk8 (changes requested):
Two minor requests, then I'll be happy to merge.

gyanu2507:
Renamed to `_compressed_tensors_quantization`, and `mxfp4-pack-quantized` is in `test_compressed_tensors_known_formats` next to nvfp4 and pack-quantized.

michalk8 (approved):

PR #1880, opened 2026-09-11
Fix GPTQ error propagation indices across groups

### LINKED ISSUE #1878: GPTQ error propagation mixes global and block-local column indices
`gptq_quantize` has two issues, the error buffer gets indexed by the whole-matrix column instead of the position in the group, so after the first group the write silently misses; and the in-group update runs to `k+j` instead of stopping at `j`, so it spills into the next group. Fixing one without the other doesn't help.

Basically, the current behavior doesn't throw any errors, but it just results in values that are slightly off of what the GPTQ paper says they should be, I don't know if this has a huge impact on the quality of the outputs of a given model, but it definitely does cause different quantized weights.

Proposed correction:

```diff
- W[..., k : k + j] -= e @ Hinv[k : k + 1, k : k + j]
- err[..., k : k + 1] = e
+ W[..., k : j] -= e @ Hinv[k : k + 1, k : j]
+ err[..., k - i : k - i + 1] = e
```

### Reproduction (this code was generated by AI)

Save the following as `repro.py` and run `python repro.py` in an environment containing MLX and mlx-lm. It selects the CPU and downloads no weights. It checks the original source expressions before creating an in-memory corrected function.

<details>
<summary>Complete reproduction script</summary>

```python
"""CPU-only, no weights/downloads: demonstrate GPTQ block indexing errors.

Run with the mlx-lm reference environment's Python. This checks the installed
implementation and constructs an in-memory version with both indices fixed.
"""
import argparse
import importlib.util
import inspect
import json
from importlib.metadata import version

import mlx.core as mx

mx.set_default_device(mx.cpu)

import mlx.nn as nn
from mlx_lm.quant import gptq

parser = argparse.ArgumentParser(description=__doc__)
parser.add_argument('--upstream-file', help='Optional pinned upstream gptq.py to execute instead of the installed module')
args = parser.parse_args()
if args.upstream_file:
    spec = importlib.util.spec_from_file_location('gptq_repro_upstream', args.upstream_file)
    assert spec and spec.loader
    gptq = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(gptq)

source = inspect.getsource(gptq.gptq_quantize)
bad_window = 'W[..., k : k + j] -= e @ Hinv[k : k + 1, k : k + j]'
bad_index = 'err[..., k : k + 1] = e'
assert bad_window in source and bad_index in source, 'Installed GPTQ differs; review the source before using this reproduction.'
fixed_source = source.replace(bad_window, 'W[..., k : j] -= e @ Hinv[k : k + 1, k : j]').replace(bad_index, 'err[..., k - i : k - i + 1] = e')
namespace = dict(vars(gptq))
exec(compile(fixed_source, '<gptq with both block indices fixed>', 'exec'), namespace)

# A group-local buffer cannot store a global column index in later groups.
i, k, group_size = 32, 32, 32
err = mx.zeros((1, group_size))
err[..., k:k + 1] = mx.ones((1, 1))
mx.eval(err)
assert mx.sum(err).item() == 0
err[..., k - i:k - i + 1] = mx.ones((1, 1))
assert mx.sum(err).item() == 1

# In the first group, k=1 already propagates into the next group. The later
# err @ Hinv block update then visits that column again.
k, j = 1, 32
assert list(range(k, k + j))[-1] == j
assert list(range(k, j))[-1] == j - 1

# Execute the actual quantizer, with and without the two-line correction.
class Model(nn.Module):
    def __init__(self, weight):
        super().__init__()
        self.proj = nn.Linear(weight.shape[1], weight.shape[0], bias=False)
        self.proj.weight = weight

    def __call__(self, x):
        return self.proj(x)

mx.random.seed(7)
weight = mx.random.normal((4, 128))
calibration = mx.random.normal((256, 128))
# Correlated columns exercise compensation across group boundaries.
calibration = calibration + 0.3 * mx.sum(calibration, axis=-1, keepdims=True)
mx.eval(weight, calibration)
kwargs = dict(data=calibration, bits=4, group_size=32,
              fallback_bits=4, fallback_group_size=32, batch_size=32)
original, _ = gptq.gptq_quantize(Model(mx.array(weight)), **kwargs)
corrected, _ = namespace['gptq_quantize'](Model(mx.array(weight)), **kwargs)
mx.eval(original.parameters(), corrected.parameters())
different_words = mx.sum(original.proj.weight != corrected.proj.weight).item()
assert different_words > 0, 'This fixture did not expose a packed-weight difference.'
print(json.dumps({'mlx': version('mlx'), 'mlx_lm': version('mlx-lm'),
                  'device': 'cpu', 'later_group_global_write_sum': 0,
                  'later_group_local_write_sum': 1,
                  'packed_weight_words_changed': different_words}, indent=2))
```

</details>

Output (MLX 0.32.2, mlx-lm 0.31.3):

```json
{"mlx":"0.32.2","mlx_lm":"0.31.3","device":"cpu","later_group_global_write_sum":0,"later_group_local_write_sum":1,"packed_weight_words_changed":23}
```

joshuarossi:
Fixes #1878

I was trying to implement GPTQ in my project (mlx-bun) using mlx-lm as a reference, and the numbers were not matching due to this bug. gptq_quantize mixes up the local group index and the global position index, this fixes that. The test in tests/test_gptq.py fails on main, and passes after the fix.

- AI usage disclosure: 
I used AI to find the bug, and then fixed it, and ran the code. The patch and test were drafted using ai, and then ran and reviewed by me.

### DIFF
--- a/mlx_lm/quant/gptq.py
+++ b/mlx_lm/quant/gptq.py
@@ -123,8 +123,10 @@ def gptq_quantize(
 
                 e = gptq_error(w, d, scales, biases)
 
-                W[..., k : k + j] -= e @ Hinv[k : k + 1, k : k + j]
-                err[..., k : k + 1] = e
+                # Update the remaining columns of this group now; the
+                # columns after the group are updated once per group below.
+                W[..., k:j] -= e @ Hinv[k : k + 1, k:j]
+                err[..., k - i : k - i + 1] = e
                 mx.eval(err, W)
 
             W[..., j:] -= err @ Hinv[i:j, j:]
--- a/tests/test_gptq.py
+++ b/tests/test_gptq.py
@@ -1,0 +1,102 @@
+# Copyright © 2026 Apple Inc.
+
+import unittest
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from mlx_lm.quant.gptq import gptq_quantize
+
+
+class OneLinear(nn.Module):
+    def __init__(self, weight):
+        super().__init__()
+        self.proj = nn.Linear(weight.shape[1], weight.shape[0], bias=False)
+        self.proj.weight = weight
+
+    def __call__(self, x):
+        return self.proj(x)
+
+
+def reference_gptq(weight, data, bits, group_size):
+    """Column-by-column GPTQ without the lazy per-group update.
+
+    Every column's error is propagated to all remaining columns immediately,
+    which is the form in the paper (Algorithm 1 without the block batching).
+    The Hessian and its inverse are formed the same way as in gptq_quantize.
+    """
+    H = data.T @ data
+    with mx.stream(mx.cpu):
+        damp = 1e-2 * mx.mean(mx.diag(H))
+        diag = mx.arange(H.shape[0])
+        H[diag, diag] += damp
+        H = mx.linalg.cholesky(H)
+        H = mx.linalg.cholesky_inv(H)
+        Hinv = mx.linalg.cholesky(H, upper=True)
+    mx.eval(Hinv)
+
+    n_bins = 2**bits - 1
+    W = weight.astype(mx.float32)
+    all_scales = []
+    all_biases = []
+    for i in range(0, W.shape[-1], group_size):
+        j = i + group_size
+        _, scales, biases = mx.quantize(W[..., i:j], bits=bits, group_size=group_size)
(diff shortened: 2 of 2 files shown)


michalk8 (approved):
Thanks a lot for spotting this, LGTM!

PR #1881, opened 2026-09-12
Recover qwen3_coder parameter names when the model drops the '>'

CryptoRhinoGH:
Qwen3 coder (and related models with the same parser) dropping closing tags for parameter names. 
_parse_xml_function_call located parameter names with match_text.index(">"), which misses a dropped '>', scanning past it to the next one in the buffer breaking the parameter name parsing, and raises ValueError when none is present. 
Reusing the same tolerant pattern which is already used for the function name, but added to a module-level _name_regex, utilizing a similar fix from #1811. Alternatively there have been similar fixes for parameter values, but never parameter names on a module-level. 

- AI usage disclosure: Utilized claude opus 5 to fix this as i came across this bug while working with Qwen3.6-35B-A3B-4bit, used it to check the issue across by benchmarking multiple tool call loops to verify issue before fixes.


michalk8 (approved):
I've added one extra test case, otherwise LGTM, thanks!

Issue #1894, opened 2026-09-16
Can we add support for deepseek v4.1?


Teejer:
I would really like to try it out on my 512gb mac. How hard is it to add?

PR #1895, opened 2026-09-16
Add DeepSeek-V4.1 model

Goekdeniz-Guelmez:
Addresses #1894


Goekdeniz-Guelmez:
Text backbone implementation of DeepSeek-V4.1-Flash. The implementation was checked against the pending
Transformers reference using a random init 172,720-parameter checkpoint.

Not tested against the real weights.

PR #1902, opened 2026-09-19
Fix lm-eval continuation scoring after partial context truncation

sqliangx:
### Summary

This PR fixes a scoring bug in the lm-eval adapter's `MLXLM.loglikelihood()` method. Currently, `MLXLM.loglikelihood()` uses the truncated prefix length to slice untruncated full sequences. As a result, it incorrectly includes context tokens in continuation scoring after partial context truncation. 

For example, with a character tokenizer, context `abc`, continuation `de`, and `max_tokens=3`, the context becomes `bc`. The code should score `de`, but it currently scores `cde` instead.

The PR fixes the bug by saving the original continuation boundary before truncating the prefix. This keeps the scoring focused on the continuation. The existing truncation policy and fully truncated `(-inf, False)` behavior stay the same.

### Testing

This PR adds a test covering no truncation and two partial-truncation cases. Both partial-truncation cases fail before the fix and pass afterward.

- AI usage disclosure: bug finding


michalk8 (approved):
Thanks for spotting this!

PR #1912, opened 2026-09-22
Keep the most likely token when top_p is small enough to mask everything

NishchayMahor:
`apply_top_p` keeps tokens whose ascending cumulative probability exceeds `1 - top_p`. Once `top_p` is small enough that `1 - top_p` rounds to `1.0` in the logprob dtype, nothing satisfies that — the cumulative sum tops out just below 1.0 — so **every token is masked**.

Sampling then draws from all `-inf` and returns arbitrary tokens, silently:

```python
from mlx_lm.sample_utils import make_sampler, apply_top_p
# logprobs: 4096-vocab, argmax is token 3801
apply_top_p(logprobs, 1e-8)        # 0 tokens survive
make_sampler(temp=0.8, top_p=1e-8)(logprobs[None])
# -> 2225, 1846, 3514, 3109, 1033 ...  (uniform noise, not 3801)
```

Asking for the most aggressive nucleus filter gives noise instead of near-greedy output — the opposite of what was requested, with no error raised.

## Why it matters at realistic values

The threshold is precision dependent, and the low-precision dtypes that on-device inference actually uses fail well inside the range a user might type:

| logprob dtype | masks everything at |
|---|---|
| float32 | `top_p ≤ 1e-7` |
| float16 | `top_p ≤ 1e-4` |
| bfloat16 | `top_p ≤ 1e-3` |

`make_sampler` admits any `0 < top_p < 1.0`, so all of these reach the filter. For float32 over a 4096 vocab the cumulative sum reaches `0.9999994`, while `1 - 1e-8` rounds to exactly `1.0`.

## Fix

Force the last entry of the ascending keep mask — the most likely token — so at least one always survives. That matches what the filter is asking for: with a tiny `top_p`, the correct nucleus is the single most likely token.

Values in the normal range are untouched. Over a 4096 vocab, `top_p=0.5` keeps 30 tokens and `top_p=0.9` keeps 312, before and after.

## Tests

Added `test_apply_top_p_keeps_the_top_token_at_tiny_thresholds`, covering float32 / float16 / bfloat16 at `top_p` of 1e-8, 1e-4 and 1e-3, asserting both that something survives and that it is the argmax. It fails on current `main`:

```
FAILED tests/test_sample_utils.py::TestSampleUtils::test_apply_top_p_keeps_the_top_token_at_tiny_thresholds
1 failed, 9 passed
```

With the fix, `tests/test_sample_utils.py` and `tests/test_generate.py` are 49 passed. `pre-commit` (black, isort, ruff) passes on both files.


michalk8 (commented):
(withdrawn)

michalk8 (approved):
Thanks, I kept your tests but instead of forcing at least one token to be present, I've decided to use a different implementation.
If you want to have a flag which forces some min amount of tokens always present (like in HF), feel free to open another PR!

